PR #133, opened 2024-08-02
Remove broken labels key in bug_report issue template

bdbch:
The issue template currently is broken because labels expects a string. This should fix it.


mauro-balades (approved):

mauro-balades:
Thanks!

PR #192, opened 2024-08-09
`alpha.yml` use zstd compression for appimage, drop libfuse2 dependency.

Samueru-sama:
Now on my system the appimage takes 1.7 seconds to start vs 2.5 seconds with the older appimagetool.

image

And the appimage no longer needs `libfuse2` as the static runtime is being used by default now.


mauro-balades:
Thanks!

PR #208, opened 2024-08-10
`alpha.yml` add zsync update to appimage

Samueru-sama:
This change would enable the appimage to be updated with [AppImageUpdate](https://github.com/AppImageCommunity/AppImageUpdate) (delta updates). 

I also did a small change, originally appimagetool was being told to make a `zen.AppImage` and once it got moved to `dist` it was renamed with that `${{ matrix.generic == true && 'generic' || 'specific' }}` that determines the type. 

I instead just told appimagetool to make the appimage have that name directly. 

Something I'm a bit worried is that the same `${{ matrix.generic == true && 'generic' || 'specific' }}` also goes to the zsync file, hopefully it goes right. I tried to run the workflow on my fork to check that but I really don't know how it works 😅

### DIFF
--- a/.github/workflows/alpha.yml
+++ b/.github/workflows/alpha.yml
@@ -288,10 +288,11 @@ jobs:
           ls -al
           find .
           ls -al "$APPDIR"
-          ARCH=x86_64 ./appimagetool-x86_64.AppImage --comp zstd --mksquashfs-opt -Xcompression-level --mksquashfs-opt 10 "$APPDIR" zen.AppImage
+          ARCH=x86_64 ./appimagetool-x86_64.AppImage --comp zstd --mksquashfs-opt -Xcompression-level --mksquashfs-opt 10 \
+          -u "gh-releases-zsync|$GITHUB_REPOSITORY_OWNER|desktop|latest|zen-${{ matrix.generic == true && 'generic' || 'specific' }}.AppImage.zsync" \
+          "$APPDIR" zen-${{ matrix.generic == true && 'generic' || 'specific' }}.AppImage
           mkdir dist
-          mv zen.AppImage* dist/.
-          mv ./dist/zen.AppImage ./dist/zen-${{ matrix.generic == true && 'generic' || 'specific' }}.AppImage
+          mv zen*AppImage* dist/.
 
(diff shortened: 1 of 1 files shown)


mauro-balades (approved):

mauro-balades:
I'll make a test run see if it works

mauro-balades:
Also, please let me know if you are on the discord server so I can hive you the contributor role!

Samueru-sama:
> Also, please let me know if you are on the discord server so I can hive you the contributor role!

Thanks for the offer but I'm not in the discord, I didn't even know it existed lol.

Samueru-sama:
Looks like the appimages build went thru, that's great!

Btw if in the future the appimage name changes to include the version of zen, for example `zen-v123456-specific.AppImage` in that case where the version would be in the zsync there has to be a wilcard like this: `...|desktop|latest|zen-*${{ matrix.generic == true && 'generic' || 'specific' }}.AppImage.zsync...` for appimageupdate to keep working.

Samueru-sama:
Uh oh.

@mauro-balades While the appimage build went thru, and the zsync files were generated and moved to dist, they not available in the latest release?

mauro-balades:
Yeah, you didn't add the files to the upload list.

What does zsync do anyways?

Samueru-sama:
> Yeah, you didn't add the files to the upload list.
> 
> What does zsync do anyways?

I just noticed that the problem is likely here: `path: ./dist/zen-${{ matrix.generic == true && 'generic' || 'specific' }}.AppImage`

I think adding a wildcard at the end of AppImage will make it also upload the zsync file. 

>What does zsync do anyways?

It allows [delta updates](https://github.com/AppImageCommunity/AppImageUpdate) in the appimage, that way the whole file doesn't have to downloaded again and only the diff.

mauro-balades:
But Firefox already has built-in updates, why would you need updates in the appimage?

Samueru-sama:
> But Firefox already has built-in updates, why would you need updates in the appimage?

That doesn't work here since the appimage is a read-only squashfs image.

As far as I know the auto-update thing firefox has is with their standalone binary.

PR #213, opened 2024-08-10
`alpha.yml` include zsync file in release

Samueru-sama:
Hopefully this is all that's needed.

### DIFF
--- a/.github/workflows/alpha.yml
+++ b/.github/workflows/alpha.yml
@@ -298,7 +298,7 @@ jobs:
         uses: actions/upload-artifact@v4
         with:
           name: zen-${{ matrix.generic == true && 'generic' || 'specific' }}.AppImage
-          path: ./dist/zen-${{ matrix.generic == true && 'generic' || 'specific' }}.AppImage
+          path: ./dist/zen-${{ matrix.generic == true && 'generic' || 'specific' }}.AppImage*
 
   release:
     if: ${{ github.event.inputs.create_release == 'true' }}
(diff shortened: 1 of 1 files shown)


.github/workflows/alpha.yml
```diff
@@ -298,7 +298,7 @@ jobs:
         uses: actions/upload-artifact@v4
         with:
           name: zen-${{ matrix.generic == true && 'generic' || 'specific' }}.AppImage
-          path: ./dist/zen-${{ matrix.generic == true && 'generic' || 'specific' }}.AppImage
+          path: ./dist/zen-${{ matrix.generic == true && 'generic' || 'specific' }}.AppImage*
 
   release:
```
mauro-balades, line 303: Also needed to be added here (release workflow). And I don't think you can wildcard when uploading so please add the full path.

mauro-balades (changes requested):

.github/workflows/alpha.yml
Samueru-sama: Sorry but when you say here you mean the changes I did in line 301 or you mean the list of files to be unloaded in line 358?

Normally I just tell Upload artifact the path to dist and it picks all the files in there without having to give the name of the files.

I'm also more confused when you said the full path, because the list of files to upload doesn't have full paths 😅

mauro-balades: You just uploaded the artifact but as you can see, below you have a list of files to upload to the release

Samueru-sama: > You just uploaded the artifact but as you can see, below you have a list of files to upload to the release

Done I think

mauro-balades:
Looks good

Samueru-sama:
It works!

image

mauro-balades:
nice!

Issue #217, opened 2024-08-11
Add more bytecode caching strategy


gunir:
### What feature would you like to see?

By default Firefox-based browsers create bytecode cache every 5 times of loading the same page

dom.script_loader.bytecode_cache.strategy=0 (default)
```
# Choose one strategy to use to decide when the bytecode should be encoded and saved. The following strategies are available right now:
# * -2 : (reader mode) The bytecode cache would be read, but it would never be saved.
# * -1 : (eager mode) The bytecode would be saved as soon as the script is seen for the first time, independently of the size or last access time.
# * 0 : (default) The bytecode would be saved in order to minimize the page-load time. 
```
-1 is too often, -2 is never, 0 is too ineffective, so we need more values

### What platform are you seeing the problem on?

Linux

### Additional context

By modifying this file: https://hg.mozilla.org/mozilla-central/file/tip/dom/script/ScriptLoader.cpp#l2479

and replace that block of code with:
```
  switch (strategy) {
    case -2: {
      // Reader mode, keep requesting alternate data but no longer save it.
      LOG(("ScriptLoadRequest (%p): Bytecode-cache: Encoding disabled.",
           aRequest));
      aRequest->MarkSkippedBytecodeEncoding();
      return;
    }
    case -1: {
      // Eager mode, skip heuristics!
      hasSourceLengthMin = false;
      hasFetchCountMin = false;
      break;
    }
    case 1: {
      // 1!
      hasSourceLengthMin = true;
      hasFetchCountMin = true;
      sourceLengthMin = 1024;
      // If we were to optimize only for speed, without considering the impact
      // on memory, we should set this threshold to 2. (Bug 900784 comment 120)
      fetchCountMin = 1;
      break;
    }
    case 2: {
      // 2!
      hasSourceLengthMin = true;
      hasFetchCountMin = true;
      sourceLengthMin = 1024;
      // If we were to optimize only for speed, without considering the impact
      // on memory, we should set this threshold to 2. (Bug 900784 comment 120)
      fetchCountMin = 2;
      break;
    }
    case 3: {
      // 3!
      hasSourceLengthMin = true;
      hasFetchCountMin = true;
      sourceLengthMin = 1024;
      // If we were to optimize only for speed, without considering the impact
      // on memory, we should set this threshold to 2. (Bug 900784 comment 120)
      fetchCountMin = 3;
      break;
    }
    default:
    case 0: {
      hasSourceLengthMin = true;
      hasFetchCountMin = true;
[...]

PR #219, opened 2024-08-11
Add more level to dom.script_loader.bytecode_cache.strategy

gunir:
https://github.com/zen-browser/desktop/issues/217


mauro-balades:
Did you use surfer export?

gunir:
> Did you use surfer export?

No, I ended up using Github Diff

mauro-balades:
That won't work when importing later. I'll try to do it later with surfer

PR #220, opened 2024-08-11
Update zen-browser.js to enable Browser Toolbox by default

GunGunGun:
This is very helpful for people who want to tinker with Zen's UI.

Maybe for convenience we should create `chrome` folder inside Zen's profile by default, and `userChrome.css`+`userContent.css` like Floorp.

Press Ctrl+Shift+Alt+I to open Browser Toolbox and click Style Editor, filter `userChrome.css` to start modifying.


mauro-balades:
Thanks!

PR #239, opened 2024-08-13
feat: Add warning for changing the languages requested

Pdzly:


mauro-balades:
Is this really important to have? Because creating new patches means more work for me each time firefox updates

mr-cheffy (approved):
👍🏽

Pdzly:
@mauro-balades i would say it is nice to have but definitly not a must.

Its just a reminder for the user.

PR #277, opened 2024-08-16
Fix scrollbar visible issue when having a lot of tabs

gunir:
Issue: https://discord.com/channels/1088172780480114748/1094002792735191150/1273822650296172594
image

Result: https://discord.com/channels/1088172780480114748/1094002792735191150/1273852505431998596
image


mauro-balades:
I don't think we should remove it, maybe a change in the UI but not removing it

gunir:
> I don't think we should remove it, maybe a change in the UI but not removing it

You can make it smaller by changing this, and we should because by default Windows scrollbar is ugly, I will submit a patch later:

image

PR #311, opened 2024-08-18
Changed marketplace to store

n7itro:
More consistent naming as mentioned in visual issue 9 in issue 304


mauro-balades:
Thanks! You don't need to create a new fork each time btw

n7itro:
> Thanks! You don't need to create a new fork each time btw

I know, there was just something wrong with the previous one so instead of fixing it I created a new fork (still learning version control 😝)

PR #316, opened 2024-08-18
Removed unused "awesome bar"

n7itro:
The "Awesome Bar Settings" don't appear to do anything, so can they be removed to avoid any confusion? If they are for a future feature they can be added back later.
I also changed "Look And Feel" to "Look and Feel"


mauro-balades:
Thanks!

PR #380, opened 2024-08-21
Closing the last tab should not cause the browser to close, this is bad default inherited from Firefox

gunir:
Closing the last tab should not cause the browser to close, this is bad default inherited from Firefox

### DIFF
--- a/src/browser/app/profile/zen-browser.js
+++ b/src/browser/app/profile/zen-browser.js
@@ -162,3 +162,6 @@ pref("sidebar.verticalTabs", false, locked);
 // Better Windows theming
 pref("widget.non-native-theme.scrollbar.style", 2);
 pref("widget.non-native-theme.use-theme-accent", true);
+
+// Closing the last tab should not cause the browser to close, this is bad default inherited from Firefox
+pref("browser.tabs.closeWindowWithLastTab", false);


DrLarck:
Hi !

You can disable this behavior by following these steps:
1. Go to `about:config`
2. Search for `browser.tabs.closeWindowWithLastTab` then set its value to `false` and you'll be fine!
C.f [this reddit post](https://www.reddit.com/r/firefox/comments/zwrwae/comment/j1wgxc3/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button)

Perhaps let the user decide if they want Zen Browser to behave like this during the onboarding?

Anthonyy232:
A setting toggle would be nice

prathapperil (approved):

src/browser/app/profile/zen-browser.js
```diff
@@ -162,3 +162,6 @@ pref("sidebar.verticalTabs", false, locked);
 // Better Windows theming
 pref("widget.non-native-theme.scrollbar.style", 2);
 pref("widget.non-native-theme.use-theme-accent", true);
+
+// Closing the last tab should not cause the browser to close, this is bad default inherited from Firefox
```
nithin891233t, line 166: Make default screen with todo lists, tasks and calendar including desktop icons. Like a landing page after closing last tab

clay-risser:
Perhaps the default behavior should be opening a new tab when the last tab is closed. Until that tab state changes, the new tab should not have an X button because it would make no sense to close an empty tab that simply spawns a new tab.

ViniciusDev26 (approved):

fndov:
>     2. Search for `browser.tabs.closeWindowWithLastTab` then set its value to `false` and you'll be fine!
this should be a toggle if implemented, because then the browser will always stay open no matter how many times you press control w to close unless you use alt f4 or another keybinding to quit

The-Lost-Light:
> > ```
> > 2. Search for `browser.tabs.closeWindowWithLastTab` then set its value to `false` and you'll be fine!
> > ```
> 
> this should be a toggle if implemented, because then the browser will always stay open no matter how many times you press control w to close unless you use alt f4 or another keybinding to quit

firefox can use Ctrl-Q to quit by default

src/browser/app/profile/zen-browser.js
```diff
@@ -162,3 +162,6 @@ pref("sidebar.verticalTabs", false, locked);
...
+pref("browser.tabs.closeWindowWithLastTab", false);
```
rrojan, line 167: It would be best to let the user change this in the settings page imo. Let me know what you think.

Some people like me have gotten heavily used to closing windows by Ctrl+w-ing the last tab.

0cool-f: Firefox is the only (major) browser that behave like this by default.

rrojan: @0cool-f We could still have a toggle like this
image

SanjayBalaji: I have been using brave for quite some time now and I didn't know that this toggle existed. Sometimes, I have felt that the window should not close after closing the last tab and I never thought that there would be a toggle to switch this. But over time I got used to window closing after last tab, that now it feels slightly weird after using this toggle.

mauro-balades:
Why would you close it?

Issue #383, opened 2024-08-21
[Bug] Text on Websites blurry


AnTheMaker:
### What happened?

Hey there! First of all - great job with Zen! It looks and feels amazing, like a younger sibling of Firefox!
I'm using it on Linux (Zorin OS 17), and it seems like the text on pretty much all websites is a bit blurry when browsing. Sometimes it gets more sharp when scrolling, but it's overall super weird.
Would be awesome if you could look into this!
Keep up the great work!
~ An

### Reproducible?

- [X] I have checked that this issue cannot be reproduced on Mozilla Firefox.

### Version

1.0.0-a.26

### Severity impact

High

### What platform are you seeing the problem on?

Linux

camflan:
I have the same issue on macOS and none of the config changes have improved the situation. This (Zen) is only browser I have installed that has this behavior, including Firefox

NOCanoa:
@camflan can u create a new user profile and check if it also looks like that

n7itro:
Zen uses different font rendering preferences by default.
Can you try setting `gfx.font_rendering.cleartype_params.rendering_mode` to -1 or 2

camflan:
This is what it looks like on a brand new user profile.

<img width="3120" alt="Screenshot 2024-08-21 at 3 33 17 PM" src="https://github.com/user-attachments/assets/acfa8d7d-ad18-4aff-9f02-0b5533c96c84">
<img width="1257" alt="Screenshot 2024-08-21 at 3 33 11 PM" src="https://github.com/user-attachments/assets/efcefeca-36eb-401f-be44-047b2aa3cb48">
<img width="957" alt="Screenshot 2024-08-21 at 3 33 01 PM" src="https://github.com/user-attachments/assets/7464b4d1-000d-4a79-8b98-ee148531882f">
<img width="1071" alt="Screenshot 2024-08-21 at 3 32 58 PM" src="https://github.com/user-attachments/assets/57f3a55f-98a1-4499-8e98-2e9da2982fc6">


It's weird. You can see in the last 2 where I've moved my mouse over some areas of the screen and they've sharpened up. Specifically if you look around the Tabs > "Ctrl+Tab" text

camflan:
I've set these settings and I think it's improved. I'm still seeing random text that's fuzzy, but now most sites are rendered as expected,

[...]

PR #464, opened 2024-08-23
Disable Edge font style for Mac and Linux

gunir:
Disable Edge font style for Mac

https://github.com/zen-browser/desktop/issues/383#issuecomment-2305649865


win98se:
The OP of #383 is facing the same problem on Linux.
Also, the two lines are related to ClearType (and also DirectWrite), which most probably is Windows-only.

gunir (commented):
Exclude Linux too

PR #591, opened 2024-08-24
Add Snap Package

0xGingi:


mauro-balades:
Do I need to submit it anywhere?

0xGingi:
You would need to publish it to snapcraft

0xGingi:
I believe there's a github action to automatically update the snap package when you update the version

mauro-balades:
Why has it been closed?

win98se:
I don't even understand why is this PR being downvoted so bad, despite there are much controversies with Canonical, that owns Snapcraft and Ubuntu.

Arxari:
Because snap is bad and no one wants it to continue existing.

Arxari:
OK but by supporting snap it's just enabling the issue, every normal distro supports flatpak out of the box, all except Ubuntu. By supporting snap it hurts the Linux desktop because it furthers distribution divide which makes it harder for developers to distribute apps.

0xGingi:
I don't use snaps or flatpaks, but both are still good to have, especially with something like a browser, accessibility is kinda important to gain more users

Arxari:
They literally made the Ubuntu app store not work with flatpaks, so you have to install a separate store. Also I'm sorry for excluding Debian from the list, but I have yet to run into a user which daily drives it.

Fedora comes with it, Mint comes with it, OpenSUSE, ZorinOS, Manjaro, Pop OS, Elementary, Vanilla, Bazzite, Nobara...

Literally except Ubuntu, all the most desktop-used distros support it (without the app store bs) and ship with it.

The ones that support Snap out of those are; Zorin (prioritizes flatpak), Manjaro (prioritizes AUR).

0xGingi:
yeah, no, not saying it's scummy, just, for the user who uses Ubuntu/Snaps and just wants to install the browser, best thing is to have the snap

Arxari:
But you do realize that by supporting snap you are literally creating this entire problem of flatpak apps not supported even they should be?

mauro-balades:
Im locking it, but ill be considering about it

Issue #475, opened 2024-08-23
Missing `privacy.resistFingerprinting.letterboxing` Option


albu-ali:
### What happened?

**Description:**

The `privacy.resistFingerprinting.letterboxing` option, available in Firefox-based browsers like LibreWolf, is not present in the Zen browser. This option is crucial for users aiming to create a privacy-focused profile.

**Importance of the Option:**

- **Fingerprinting Protection:** This option helps prevent websites from fingerprinting users based on their screen size and resolution, minimizing their digital footprint.
- **Consistency with Firefox:** Firefox and other Firefox-based browsers offer this option, setting a standard for privacy features.

**Steps to Reproduce:**

1. Go to `about:config`.
2. Search for `privacy.resistFingerprinting.letterboxing`.
3. Observe that the option is not available.

**Additional Information:**

- The lack of this option may lead users to switch to other browsers with more comprehensive privacy features.
- This option exists in browsers like LibreWolf, and users who want to switch to your browser need this option to maintain their privacy settings and delete all other FF base browser like LibreWolf and some others.
- I love Zen browser and I want to be the best browser for all users, so plz consider add this option.  

### Reproducible?

- [ ] I have checked that this issue cannot be reproduced on Mozilla Firefox.

### Version

1.0.0-a.27 (64-bit)

### Severity impact

Medium

### What platform are you seeing the problem on?

Linux

gunir:
This is an hidden config, it's not exposed in about:config, it only exposes in some privacy browsers like Mullvad.

albu-ali:
> This is an hidden config, it's not exposed in about:config, it only exposes in some privacy browsers like Mullvad.

And LibreWolf and Tor browser 

I know it's hidden and that's is the problem because other people don't know that when they want to switch browser from like Mullvad or librewolf or any other FF base browser

PR #628, opened 2024-08-24
Exposes Letterboxing https://github.com/zen-browser/desktop/issues/475

gunir:
Exposes Letterboxing https://github.com/zen-browser/desktop/issues/475


mauro-balades:
Just to be sure, this doesnt enable it, right? it just exposes

gunir:
> Just to be sure, this doesnt enable it, right? it just exposes

Of course, it's false

qtchaos:
Don't see why this shouldn't be merged, if Zen is supposed to be as customizable as possible and also privacy focused, then simply exposing the config option doesn't do any harm.

mauro-balades:
Yeah, I was just making sure, haha

PR #694, opened 2024-08-25
chore: Adding description to .desktop file

upsidedownsweetfood:
I'm aware that this is kind of a really dumb thing to add, but it did bother me


mauro-balades:
Thanks!

PR #698, opened 2024-08-25
fix: use generic fullscreen detection pseudo-class

### LINKED ISSUE #614: [BUG] Entering fullscreen mode does not hide any browser UI elements
### What happened?

When I enter fullscreen mode, both native to firefox and tab bar remain visible indefinitely.

This occurs only on Zen Browser in Wayland KDE on Arch Linux. This can not be reproduced on stable, nor on nightly versions of Firefox.

This has no effect on the problem:
* Setting or unsetting `browser.fullscreen.autohide` in `about:config`
* Forcing start in XWayland mode by setting `GDK_BACKEND=x11`
* Using different methods to enter fullscreen (`F11` key, Toolbar button, KWin action)
* Enabling/Disabling native titlebar
* Clean re-install

Only way to hide elements is to activate Zen Compact mode manually after entering fullscreen.

### Reproducible?

### Version

Mozilla Zen 1.0.0-a.28

### Severity impact

Medium

Linux

### Relevant log output

```shell
No messages are being emitted to Ctrl-Shift-G console, except for tab height message.
```

ekorchmar:
Closes #614 

Replaces all instances of `inDOMFullscreen` pseudo-class with `inFullscreen` pseudo-class to ensure consistent behavior regardless of how window became (borderless) fullscreen.

Most notably, it allows toolbars to disappear whenever fullscreen state is initiated.

### DIFF
--- a/src/browser/base/content/zen-styles/zen-browser-container.css
+++ b/src/browser/base/content/zen-styles/zen-browser-container.css
@@ -1,5 +1,5 @@
 
-:root:not([inDOMFullscreen="true"]):not([chromehidden~="location"]):not([chromehidden~="toolbar"]) {
+:root:not([inFullscreen="true"]):not([chromehidden~="location"]):not([chromehidden~="toolbar"]) {
   & #tabbrowser-tabbox #tabbrowser-tabpanels .browserSidebarContainer {
     width: -moz-available;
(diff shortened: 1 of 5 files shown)


ekorchmar:
This should not be merged yet: it disables the toolbars in full-screen, but removes interactive behavior. Needs a fix before merging

ekorchmar:
Needs a better solution.

PR #739, opened 2024-08-26
fix collapsed tab shifting when audio playing

### LINKED ISSUE #608: chore(visual): favicon shifting when audio playing on small icon view
### What happened?

When having the sidebar in the small view (so that only the favicon is visible) and there is a tab playing audio, switching to it shifts the favicon to the left (by a few pixels) to create space for the mute button.

I'm assuming that this is not wanted behavior since the shift is so small that it doesn't actually make muting easier.

zen_example

Found the css in question which applies the margin but can't find it in this repo?

`📁tabs.css:247`
```css
/* Width/height & margins apply on tab icon stack children */
.tab-throbber,
.tab-icon-pending,
.tab-icon-image,
.tab-sharing-icon-overlay,
.tab-icon-overlay {
  height: 16px;
  width: 16px;

  &:not([pinned]) {
    margin-inline-end: 5.5px; <-- issue stems from this
  }
}
```

### Reproducible?

### Version

1.0.0-a.28

### Severity impact

Low

Windows

### Relevant log output

_No response_

n7itro:
Fixes #608 as tested in browser toolbox

### DIFF
--- a/src/browser/base/content/zen-styles/zen-tabs/vertical-tabs.css
+++ b/src/browser/base/content/zen-styles/zen-tabs/vertical-tabs.css
@@ -232,7 +232,7 @@
   }
   
   .tab-icon-overlay {
-    margin-inline-end: 0;
+    margin-inline-end: 0 !important;
     display: none; /* TODO: fix this */
   }
 


mauro-balades:
Thanks!

PR #926, opened 2024-08-26
Update preferences to fix branding, UI and in general be more secure

qtchaos:
Used [cachyos.cfg](https://github.com/CachyOS/CachyOS-Browser-Settings/blob/master/cachyos.cfg) to implement more preferences.

### DIFF
--- a/src/browser/app/profile/zen-browser.js
+++ b/src/browser/app/profile/zen-browser.js
@@ -30,7 +30,12 @@ pref('browser.translations.newSettingsUI.enable', true);
 pref("browser.urlbar.trimHttps", true);
 pref("browser.urlbar.untrimOnUserInteraction.featureGate", true);
 
-// Url bar
+// Search
+pref("browser.urlbar.suggest.searches", false);
+pref("browser.search.suggest.enabled", false);
+pref("browser.search.update", false);
+
+// URL bar
 pref('browser.urlbar.unitConversion.enabled', true);
 pref('browser.urlbar.trending.featureGate', false);
 pref('browser.urlbar.weather.featureGate', true);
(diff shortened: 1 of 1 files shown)


src/browser/app/profile/zen-browser.js
```diff
@@ -53,15 +58,63 @@ pref('browser.toolbars.bookmarks.visibility', 'never');
...
+pref("toolkit.crashreporter.infoURL", "");
+pref("security.protectionspopup.recordEventTelemetry", false);
+
+// Query stripping - ported from Brave (https://github.com/brave/brave-core/blob/master/components/query_filter/utils.cc#L23)
+pref("privacy.query_stripping.strip_list", "__hsfp __hssc __hstc __s _branch_match_id _branch_referrer _gl _hsenc _kx _openstat at_recipient_id at_recipient_list bbeml bsft_clkid bsft_uid dclid et_rid fb_action_ids fb_comment_id fbclid gbraid gclid guce_referrer guce_referrer_sig hsCtaTracking irclickid mc_eid ml_subscriber ml_subscriber_hash msclkid mtm_cid oft_c oft_ck oft_d oft_id oft_ids oft_k oft_lk oft_sk oly_anon_id oly_enc_id pk_cid rb_clickid s_cid ss_email_id twclid unicorn_click_id vero_conv vero_id vgo_ee wbraid wickedid yclid ymclid ysclid")
+
+// Disables fetching and updating safebrowsing lists from Google
+pref("browser.safebrowsing.malware.enabled", false);
+pref("browser.safebrowsing.phishing.enabled", false);
+pref("browser.safebrowsing.blockedURIs.enabled", false);
```
dewantawsif, line 82: And how is one supposed to enable this given they are not in the settings?

qtchaos: What you're saying is that theres no need to disable them because they cant be enabled? I dont get it.

dewantawsif: ...it's just a question

mauro-balades:
The settings are already included inside surfer's source code, thanks!

qtchaos:
> The settings are already included inside surfer's source code, thanks!

Every setting that I set I checked if it was already set, none of the major ones were included (I didn't see any). Even when I double checking and picking 5 at random, they weren't included?

qtchaos:
:trollface:

PR #968, opened 2024-08-27
fix: right-side tabs was incorrectly positioned on compact mode

elianiva:
"tabs on the right" didn't seem to affect compact mode, should be working now

https://github.com/user-attachments/assets/49e8da53-71de-441b-8f1b-236ed61c8af2


mauro-balades:
Thanks!

PR #1011, opened 2024-08-28
Enable https only by default.

bw8686:


mauro-balades:
So many commits for just a one-liner lol!

PR #1026, opened 2024-08-28
fix: properly hide and reveal toolbars in sizemode="fullscreen"

ekorchmar:
Behavior is taken from compact mode CSS. Works for left and right tabs positioning. May be broken for MacOS users, which will have to be fixed separately, but I have no feasible way to test it.

Note that upstream behavior of attaching JavaScript listeners to set and reset `top-margin` property is still there; but maybe so are may other "inherited" artifacts, and a clean-up is due sometime.

Closes issue #614


mithom:
does this also fix the scrollbar on fullscreen media like in youtube?

ekorchmar:
> does this also fix the scrollbar on fullscreen media like in youtube?

Not sure. Is this issue reported here on Github? Can't find anything

ekorchmar:
The issue is still open and this PR fixes it. Can I at least have a justification? I want to learn and contribute.

PR #1031, opened 2024-08-29
URL bar improvements

### LINKED ISSUE #721: URL bar pill doesn't change shape when searching
### What happened?

Some websites have additional emblems such as permissions after the HTTPS icon in the URL bar. When this occurs, the right side of the bar pill loses it's roundness so the two sections can merged into one block.

image

Unfortunately, when deleting the URL and beginning a search from a site where this occurs, the pill does not return to it's original shape.

image

### Reproducible?

### Version

1.0.0-a.29

### Severity impact

Low

Linux

### Relevant log output

_No response_

notsatvrn:
This pull request attempts to improve the look of the URL bar and search panel. There were a number of inconsistencies I noticed, and they should all be fixed now.

One thing I should note is that when the search menu is open the sizing and padding of some elements changes, and therefore the position. I think it looks better this way once it's open, but if the movement isn't ideal this could be reverted.

Fixes #721

### DIFF
--- a/src/browser/base/content/zen-styles/zen-urlbar.css
+++ b/src/browser/base/content/zen-styles/zen-urlbar.css
@@ -6,9 +6,9 @@
   --urlbarView-hover-background: var(--zen-colors-border);
   --urlbarView-highlight-background: var(--zen-colors-border);
   border: 1px solid var(--zen-colors-border);
+  border-radius: 999px;
   overflow: hidden;
   padding: 1px;
-  border-radius: 999px;
 }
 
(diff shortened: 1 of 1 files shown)


mithom:
i haven't been able to test these changes, but elements moving when focussing the urlbar doesn't sound good, even if it is small. This should be avoided at all cost imho. even in other browsers where the shape changes, the surroundings and the content do not move.

mauro-balades:
Thanks!

mauro-balades:
Thanks!

PR #1112, opened 2024-08-30
Fix right vertical tabs leftovers in fullscreen mode.

### LINKED ISSUE #1100: "Tabs on the right" keep some UI in fullscreen mode.
### What happened?

1. Enable "Tabs on the right"
2. Enter fullscreen mode e.g. go to YouTube video and press `f`.

The `.5rem` border on the left remains visible.

TL:DR I'll try to make a PR with a fix.

I've fixed it with following `userChome.css`. I think the border on left should be implemented exactly the same as in `zen-browser-container.css` e.g. margin instead of padding. I would move the margin style there. It would be easier to maintain and `!important` won't be needed. Another benefit would be better themes compatibility e.g. it will work with Disable Rounded Corners.

```
@media (-moz-bool-pref: "zen.tabs.vertical.right-side") and (not (-moz-bool-pref: "zen.view.compact")) {
  #navigator-toolbox {
    order: 6;
  }

  #zen-sidebar-splitter {
    order: 5;
  }

  @media (-moz-bool-pref: "zen.view.sidebar-expanded") {
    #navigator-toolbox:is([zen-user-hover="true"]:hover, :not([zen-user-hover="true"])) {
      padding-left: 0;
      padding-right: 2x;
    }
  }

  :root:not([inDOMFullscreen="true"]):not([chromehidden~="location"]):not([chromehidden~="toolbar"]) {
    & #tabbrowser-tabbox #tabbrowser-tabpanels .browserSidebarContainer {
      margin-left: var(--zen-element-separation) !important;
      margin-right: 0 !important;
    }
  }
}


#tabbrowser-tabbox {
  padding-left: 0 !important;
}
```

`padding-left: 0 !important;` should be just removed from `vertical-tabs.css`. It's needed only in `userChome.css` to override current `vertical-tabs.css`. 

### Reproducible?

### Version

1.0.0-a.33

### Severity impact

Low

Windows

### Relevant log output

_No response_

damianfijorek:
Fix #1100
- Use margin instead of padding as it's done for vertical tabs on the left.
- Move overrides close to the original values so it's easier to maintain the same selectors, get rid of `!important` and thus get better themes compatibility.

### DIFF
--- a/src/browser/base/content/zen-styles/zen-browser-container.css
+++ b/src/browser/base/content/zen-styles/zen-browser-container.css
@@ -2,14 +2,14 @@
 :root:not([inDOMFullscreen="true"]):not([chromehidden~="location"]):not([chromehidden~="toolbar"]) {
   & #tabbrowser-tabbox #tabbrowser-tabpanels .browserSidebarContainer {
     width: -moz-available;
-    margin: 0 var(--zen-element-separation) var(--zen-element-separation) 0;
+    margin: 0 var(--zen-element-separation) var(--zen-element-separation) 2px;
+    @media (-moz-bool-pref: "zen.tabs.vertical.right-side") {
+      margin: 0 2px var(--zen-element-separation) var(--zen-element-separation);
+    }
     box-shadow: 0 0 0 1px var(--zen-colors-border);
     clip-path: inset(-5px -5px -5px round var(--zen-webview-border-radius, var(--zen-border-radius)));
     border-radius: var(--zen-webview-border-radius, var(--zen-border-radius));
     transform: translate3d(0, 0, 0);
     overflow: hidden;
-
-    /* This fixes an issue with the left border */
-    margin-left: 2px;
   }
 }
--- a/src/browser/base/content/zen-styles/zen-compact-mode.css
+++ b/src/browser/base/content/zen-styles/zen-compact-mode.css
@@ -131,7 +131,13 @@
   #PersonalToolbar[collapsed="true"]{ display: none }
 
   :root:not([inDOMFullscreen="true"]) #tabbrowser-tabbox #tabbrowser-tabpanels .browserSidebarContainer {
-    margin-left: var(--zen-element-separation) !important;
+    @media (not (-moz-bool-pref: "zen.tabs.vertical.right-side")) {
+      margin-left: var(--zen-element-separation) !important;
+    }
+
+    @media (-moz-bool-pref: "zen.tabs.vertical.right-side") {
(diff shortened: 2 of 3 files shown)


damianfijorek:
I removed `padding-left: 0 !important;` because I handle it above with media queries.

I'm not sure why `padding: calc(var(--zen-tabbrowser-padding) * 1.5);` was added just to vertical tabs on the right. In the end it adds padding to top and bottom. Tabs on the left don't have that. If that is necessary I would write it explicitely e.g.:
```
      --zen-tabs-vertical-right-side-additional-padding: calc(var(--zen-tabbrowser-padding) * 1.5);
      padding-top: var(--zen-tabs-vertical-right-side-additional-padding);
      padding-bottom: var(--zen-tabs-vertical-right-side-additional-padding);
```

mauro-balades:
Could you send screenshots of the changes? I dont really know what's going on, sorry

damianfijorek:
5px left border remains in fullscreen mode e.g. YT video.

image

Same here. It's hard to keep up with the changes. 😅 I'll try to resolve conflicts after next release.

In general I think that it's better to collocate styles corresponding to vertical tabs on left and right instead of putting all overrides in one place and forcing them with `!important`. E.g. I have no idea why `padding: calc(var(--zen-tabbrowser-padding) * 1.5);` is there. I've not found anything similar for vertical tabs on the left in code nor style inspector.

damianfijorek:
Conflicts resolved.

mauro-balades:
You might want to check the recent updates because the changes you are doing are already implemented... just in another place on the CSS file

damianfijorek:
TL;DR I updated my branch yesterday and just checked recent updates. I still see issues that this commit tries to fix.

I see the effort in the code but I do not see it as an end user. Tabs on the right are always off. That's how I see them today. Big gap on the left no gap at the right and the referenced issues are not fixed.

image

Current approach seems to be to implement tabs on the left and then add overrides for tabs on the right. Often with `!important` and different, much simpler selectors. For example for tabs on the left **sometimes** there is `margin-left: 5px` but for the tabs on the right it's **always** `margin-left: 0` and `margin-right: 5px`.

I'm proposing, let's call it _if else_ approach. When something should be mirrored I used media queries and collocate those changes next to the rules for the tabs on the left. I hope this could make tabs on the right more stable. There is one place for selectors and when something changes for the tabs on the left it will be easier to find the place to change it for the tabs on the right - just below. 😉

mauro-balades:
This wont be needed anymore in the next release, as the UI changed quite a bit

PR #1131, opened 2024-08-31
chore: remove severity from bug report template

dw-0:
Determining criticality should be done by a maintainer, not by a user reporting a bug or issue. Giving users the ability to determine the criticality of their issue, in combination with using a bot for auto flagging, I see the following 2 problems arising:
- Non-technical users misjudging an issue. This can go both ways. Either they underestimate the problem, even though it is a serious problem, or they overestimate it, even though the problem may be minor or not serious at all.
- It is deliberately misused to artificially draw attention to a particular problem or mistake by knowingly choosing a higher severity for an issue than it actually is.

I created this PR because, in my opinion, I already see bug reports that apply to one of the two problems mentioned above. I am not going to point fingers right now, but filter for critical severity and decide on your own.

Having this labeling automatism will make it harder for maintainers, (future) contributors and people who just want to follow the current state of this project, as those labels do not reflect the assessment of severity by the maintainers of the project, instead the assessment is done by "random people on the internet". Maintainers would be required to re-label the issues if the previous label was wrong and contributors looking for work may focus on a specific severity and stumble across issues actually not fitting the label of severity they were looking for.

My suggestion would be (not included in this PR) to let the bot label all new issues with "Triage" (or something similar), so that those Issues then receive a correct lable by a maintainer that can actually assess the criticality.

This PR is obviously just a suggestion. So i am totally fine with this PR getting rejected in case the idea does not appeal.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -29,19 +29,6 @@ body:
     validations:
       required: true
   - type: dropdown
-    id: severity
-    attributes:
-      label: Severity impact
-      description: See label descriptions for more information.
-      options:
-        - Low
-        - Medium
-        - High
-        - Critical
-    validations:
-      required: true
-
-  - type: dropdown
     id: platform
     attributes:
       label: What platform are you seeing the problem on?


mauro-balades:
How could we prevent enhancements? Add a bit that checks for enhancement tags and close them automatically? Because enhancements/ questions should be I'm discussions

dw-0:
> Because enhancements/ questions should be I'm discussions

You can transfer a github issue into a new github discussion. 
https://docs.github.com/en/discussions/managing-discussions-for-your-community/moderating-discussions#converting-an-issue-to-a-discussion

So in case someone used the "Bug Report" button and actually created a feature request, i see two options apart from sifting the issue. In my opinion that step is mandatory, no matter what. Even in the current state.

Option 1: Transfer the github issue into a github discussion manually - i did a quick search trough available actions and it seems like there is no way of transferring an existing issue to a discussion with an existing github action.

Option 2: Flag the issue with a specific tag, maybe "Feature Request" or similar which then triggers a bot auto-responding to this issue, explaining to the user where feature requests should actually go and that this issue will be closed in favor of keeping the issues for actual issues only. The issue is then closed by the bot.

Issue #496, opened 2024-08-23
HDR playback does not work properly on macOS


miklosakos:
### What happened?

Loading a video in HDR on YouTube, like Jacob + Katie Schwarz's [COSTA RICA IN 4K 60fps HDR (ULTRA HD)](https://www.youtube.com/watch?v=LXb3EKWsInQ) will start playing back in HDR, however the actual video looks like HDR content being played back on an SDR display.
This bug does not happen on Mozilla Firefox running on macOS.

### Reproducible?

- [X] I have checked that this issue cannot be reproduced on Mozilla Firefox.

### Version

1.0.0-a.28 (64-bit)

### Severity impact

Medium

### What platform are you seeing the problem on?

macOS - aarch64

HarryHeres:
Same issue, will take a look into it and see if I can find the culprit. Confirming that HDR with normal Firefox works.

HarryHeres:
Can confirm, one of the git patches in src folder is causing this. Pin-pointing this down further later today...

mauro-balades:
I wonder what it could be

HarryHeres:
Can confirm that the issue remains with the clip-path and overflow:hidden attributes (needed for rounded borders) of the content window. 

A temporary solution would be to use https://zen-browser.app/themes/c6813222-6571-4ba6-8faf-58f3343324f6.

I will try to implement a more permanent solution while enabling for having the same visual appearance.

HarryHeres:
@mauro-balades I think this can get closed for now as we have a solution. I might get back to this later, but I think there are other issues with higher priority.

PR #1139, opened 2024-08-31
Fixing HDR on MacOS, adding Prettier

HarryHeres:
Fixing https://github.com/zen-browser/desktop/issues/496

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,2 @@
+tabwidth: 2
+printWidth: 100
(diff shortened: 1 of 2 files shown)


HarryHeres:
@mauro-balades There's one issue - when the video is not fully visible (e.g. after scrolling), it switches to SDR mode. After being fully visible, HDR mode comes back. This is due to the `overflow: hidden` property, however I have not found an appropriate solution to that.

mauro-balades:
I use clip-path because the website may overlap the border-radius

HarryHeres:
> I use clip-path because the website may overlap the border-radius

I tried to implement a lot of various little workarounds that would enable the clip-path attribute to stay, but I was, unfortunately, unsuccessful :/

mauro-balades:
You could use https://zen-browser.app/themes/c6813222-6571-4ba6-8faf-58f3343324f6

HarryHeres:
I will try to look for and implement a different solution then. Closing this for now.

PR #1161, opened 2024-09-01
Adding Homebrew 

HarryHeres:
This PR adds Homebrew Cask for the Zen Browser in order to be able to install the browser from command line. The "tap" is self hosted, pointing at this repository. 

The automatic update of the Cask version for new releases remains yet to be tested.

Also, macOS users installing via Homebrew will not be prompted that the application is damaged.


mauro-balades:
Hi, thanks a lot! 

Could the scripts moved to the `scripts/` folder? Also, how would updates work?

mauro-balades:
And the workflow, it should be separated (like where `flatpak` is, in order to avoid updating the repo if something failed on another build like windows)

mauro-balades (changes requested):
I've commented the changes I requested

HarryHeres:
> Hi, thanks a lot!
> 
> Could the scripts moved to the `scripts/` folder? Also, how would updates work?

As for the updates - on `brew install`, the browser version specified in the `version` attribute in `zen-browser.rb` will be installed. When a user executes `brew upgrade zen-browser`, the `livecheck` procedure will be performed. If new version is found in GitHub releases, it will be downloaded and installed accordingly. 

To ensure that the latest version is installed on `brew install` command, there's the update script included. 

PS: I'm no expert in Homebrew, but all of this was created by consulting their official docs :)

HarryHeres:
> Hi, thanks a lot!
> 
> Could the scripts moved to the `scripts/` folder? Also, how would updates work?

If you want, I can definitely move the script into the `scripts` folder, however I thought that it could be a part of the `.github` folder as it should be executed only by the pipeline(s)?

mauro-balades:
> > Hi, thanks a lot!
> > Could the scripts moved to the `scripts/` folder? Also, how would updates work?
> 
> If you want, I can definitely move the script into the `scripts` folder, however I thought that it could be a part of the `.github` folder as it should be executed only by the pipeline(s)?

In that case, you can store it inside `.github/workflows/src/` with the other scripts

~ mauro




Sep 1, 2024, 19:32 by ***@***.***:

HarryHeres:
@mauro-balades Updated based on your comments. Make sure to check out the workflow changes if they're correct :)

mauro-balades:
I think we can push directly instead of making a PR, no?

HarryHeres:
Sure, I also would prefer it that way but noticed that you create a PR with flatpaks so I figured I would leave it the same. But if you're okay with it, we can change it to pushing immediately to master.

mauro-balades:
I've already changed it, let me review if everything ok, and then we should be ready to go!

mauro-balades (approved):

PR #1206, opened 2024-09-03
Updating Homebrew GH actions

HarryHeres:


mauro-balades:
I've already fixed it, thanks!

PR #1212, opened 2024-09-04
Better and more performant rounded corners

gunir:
Better and more performant rounded corners

image

Works on every website, no performance penalty.


mauro-balades:
@gunir I would rather like this to be applied to the browser container because it wouldn't be applied to each vide, e.g. in split view... Could you please check that out? Thanks!

PR #1232, opened 2024-09-04
Update feature: fix inaccuracies with margins

rexxDigital:
Margins were off and variables are not used in certain places


mauro-balades:
Thanks!

Issue #1213, opened 2024-09-04
UI breaks in small window


bhaumik-tripathi:
### Captchas

- [X] I have read the instructions.
- [X] I have searched existing issues and avoided creating duplicates.
- [X] I am not filing an enhancement request.

### What happened?

Open the browser and escape from the full screen window to a floating window by clicking maximize/restore button next to close button or simple dragging after holding the flexible space in address bar. Return to normal mode if compact mode is activated. 
Now, on minimizing the size of floating window the address bar UI breaks with buttons beginning to overlap with each other. I use few theme store utilities too.
This doesn't happen with Firefox as it removes button from address bar when the window width is decreased and stops further minimizing after a certain width. Zen also stops further compression after a certain width but breaking the UI.
**Note:** I am using fedora 40 but not the flatpak version due to performance issues, I am using the unofficial community rpm package from [here](https://copr.fedorainfracloud.org/coprs/sneexy/zen-browser/) but packaging does not seems to effect the UI as I know, hence I filed the bug report.
[here](https://i.imgur.com/QpX6EDU.png)  is the broken address bar image.

### Reproducible?

- [X] I have checked that this issue cannot be reproduced on Mozilla Firefox.

### Version

1.0.0

### What platform are you seeing the problem on?

Linux

infiniwave:
Same issue on Windows.
![image](https://github.com/user-attachments/assets/080c9626-f941-469b-b320-06c024081f4b)

badjuice:
Maybe we should take note of what other browsers do in respect to this. This is how Microsoft Edge handles it.

On maximized window:
![image](https://github.com/user-attachments/assets/4e2b9248-53d9-48fe-bb1a-82f875b9baa9)

On resized window:
![image](https://github.com/user-attachments/assets/ff4dd43a-901f-4f8a-ba09-ba79b0dcc4f0)

Apparently it has a minimum width it allows the window to be resized to. And it hides most of the options from the toolbar.

bhaumik-tripathi:
> This doesn't happen with Firefox as it removes button from address bar when the window width is decreased and stops further minimizing after a certain width

This is apparently what firefox does too, even zen has a minimum width defined, and on observing you will find that it also removes buttons on decreasing width, but the minimum width and the icons present in the minimum width does not align properly, leading to the overlap/issue.

mauro-balades:
This has been fixed for the next release 👍🏽

PR #1237, opened 2024-09-04
Fix Responsivity of Toolbar

mohvn:
Adjusted the responsiveness of the Toolbar when resizing.

### Before:
before

### After:
dqwfwqd

Issue mentioned in https://github.com/zen-browser/desktop/issues/1213


mauro-balades:
Thanks a lot! Are the other changes just formatting? Because i cant really see *the* changes... But thanks!

mohvn:
> Thanks a lot! Are the other changes just formatting? Because i cant really see _the_ changes... But thanks!

yep, it might be my tsx code formatter

PR #1256, opened 2024-09-05
fix: Adjusted the responsiveness of the Toolbar when resizing (again).

mohvn:
Adjusted the responsiveness of the Toolbar when resizing again 'cause last ui update broke last responsiveness fix somehow

### Before:
before

### After:
after

Obs.: some extra changes in `zen-urlbar.css` was code formatting. 😉

### DIFF
--- a/src/browser/base/content/zen-styles/zen-urlbar.css
+++ b/src/browser/base/content/zen-styles/zen-urlbar.css
@@ -172,25 +172,52 @@ button.popup-notification-dropmarker {
   margin-top: 5px
 }
 
-@media (max-width: 550px) {
-  #urlbar-container {
-    width: calc(176px + 2 * (24px + 2 * var(--toolbarbutton-inner-padding)));
-  }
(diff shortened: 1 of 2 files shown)


mauro-balades:
Maybe the sizes need to be different depending if the legacy toolbar is enabled

mohvn:
> Maybe the sizes need to be different depending if the legacy toolbar is enabled

Thanks for the advice! I didn't know about these two options. I adjusted to both legacy and main toolbar modes.

### Legacy Mode:
legacy mode enabled

### Main Mode:
legacy mode disabled

mauro-balades:
Much better! I wonder if the values could simply be changed with a CSS variable to avoid duplication but ig it's fine now.

mohvn:
> Much better! I wonder if the values could simply be changed with a CSS variable to avoid duplication but ig it's fine now.

Ok, I'll think about a solution to improve that

PR #1260, opened 2024-09-05
Homebrew & docs update

HarryHeres:
Removing unnecessary stanza from the Cask and updating the docs with instructions on how to install the browser on MacOS.


mauro-balades:
👍🏽 Thanks!

PR #1265, opened 2024-09-05
Various fixes to Homebrew Cask

AlternateRT:
This PR makes various changes to `zen-browser.rb` so that it is conformant with the [style guidelines](https://github.com/Homebrew/homebrew-cask/blob/master/CONTRIBUTING.md#style-guide) outlined by the Homebrew Cask project.

It fixes the stanza order, adds URL verification to the `url` stanza, adds the missing `auto_updates` and `depends_on` stanza, makes the `postflight` stanza only clear the quarantine attribute, and expands the `zap` stanza so that it clear more files when uninstalling with `--zap`.

In addition, the version number on the cask has been updated, as previous GH Actions failed to do so.

### DIFF
--- a/.github/workflows/src/update-homebrew.sh
+++ b/.github/workflows/src/update-homebrew.sh
@@ -1,8 +1,8 @@
 
 latest_version=$1
-sed -i "s/version(.*)/version('${latest_version}')/" ./Casks/zen-browser.rb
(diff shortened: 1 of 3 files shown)


mauro-balades:
Thanks!

PR #1267, opened 2024-09-05
Fix vertical tabs on the right overflow.

damianfijorek:
The `vbox` overflows the parent `#navigator-toolbox`. It's fine for tabs on the left but when they are on the right the overflow is out of the window/screen. With `float: right` the `vbox` overflows to the left and is fully visible and has the same distance to the window edge.


mauro-balades:
I feel like we already fixed this... But thanks!

PR #1270, opened 2024-09-05
Fix in compact mode: Keep toolbar open during drag

burnt0rice:
I added the `:active` selector to the toolbar so that it doesn't disappear while dragging the window.


mauro-balades:
😁🤌

burnt0rice:
I know it's a little bit picky, hehe

PR #1292, opened 2024-09-06
Ensure tab becomes visible when selected in sidebar.

BrhmDev:
Ensure a tab is always visible when tab selected, new tab created. Especially useful when navigating tabs with key-binds.

D6yBtAAbj8


mauro-balades:
Thanks!

PR #1367, opened 2024-09-10
Remove Homebrew Cask files

AlternateRT:
Zen has been accepted into the Homebrew Cask repository (https://github.com/Homebrew/homebrew-cask/pull/184844), so it's no longer necessary to host the cask file within this repository (and tell people to use `brew tap ...`).

I've also updated the `README.md` instructions to reflect the fact that installation from [zen-browser.app/download](https://zen-browser.app/download) works, and included the system requirements for macOS.


HarryHeres:
Great PR! One question - as you removed the workflow, how will Homebrew react to a new release?

mauro-balades:
I dont like the idea of having 2 same packages in 2 different repos... I think it should stay here so it can be updated automatically

HarryHeres:
> I dont like the idea of having 2 same packages in 2 different repos... I think it should stay here so it can be updated automatically

If the main repo Cask can be updated automatically, we don't need to keep our own tap.

HarryHeres:
> @HarryHeres Well... casks in the main Homebrew repository are updated with the [`brew bump-cask-pr`](https://docs.brew.sh/Manpage#bump-cask-pr-options-cask) command - some popular/frequently bumped casks are autobumped by the Homebrew repository (e.g. [floorp](https://github.com/Homebrew/homebrew-cask/pull/184718)), but Zen is too new there for that at this moment.
> 
> In Zen's case, updating the cask would currently be done like so:
> 
> ```
> brew bump-cask-pr version=1.0.0-a.39 zen-browser
> ```
> 
> That automatically opens a PR there for a Homebrew maintainer to review. In theory, I suppose you could modify the workflow here to automatically open a PR there, but I don't know how to do that (hence why I opted to simply remove the workflow).
> 
> In any case, the cask contains the `auto_updates` stanza (see [optional stanza](https://docs.brew.sh/Cask-Cookbook#optional-stanzas)). Casks with this stanza are supposed to be upgraded with the embedded update functionality (see the [Homebrew FAQ](https://docs.brew.sh/FAQ#why-arent-some-apps-included-during-brew-upgrade)); for Zen that is `About Zen` - I had contemplated also removing the `brew upgrade --greedy` instructions because of this.

Ah, okay, I was wondering how that worked. Thank you for the overview. I think you can also remove `brew upgrade ...` section then since we will be mostly using the embedded auto-update functionality.

@mauro-balades After that is removed, this PR looks good to me.

AlternateRT:
I've removed the `brew upgrade ...` instructions in 1bbac89.

mauro-balades:
How do we update then?

HarryHeres:
> How do we update then?

Whenever we feel like there's a need to update the cask, we can upgrade directly via homebrew. For already installed, the Zen's updater will take care of the rest

mauro-balades:
Every release, I want it to be updated to the latest, but since the cask or however it's called is created by `not me`, how are going to be able to update the latest version automatically?

This is the same thing with flatpak, versions simply stayed behind until I was basically forced to make it a workflow haha

AlternateRT:
@mauro-balades I did a bit of research, and it turns out that Homebrew offers their own GitHub Actions for updating/bumping casks; I wrote a new workflow using those in 9f50343.

I'm no expert on writing workflows, but I think this should be able to automatically update the Homebrew cask on every release.

mauro-balades:
That's Great!

What do I need to set as API key?

AlternateRT:
The [documentation](https://github.com/Homebrew/actions/tree/master/bump-packages) has this to say:
> One should use the [Personal Access Token](https://github.com/settings/tokens/new?scopes=public_repo,workflow) for token input to this Action, not the default GITHUB_TOKEN, because brew bump creates a fork of the formula's tap repository (if needed) and then creates a pull request.

mauro-balades:
But you are the guy who created the brew, dont we need *your* token?

mauro-balades:
https://github.com/zen-browser/desktop/actions/runs/10984599633/job/30507781934#step:4:116 what is this?

PR #1435, opened 2024-09-14
Add in AppImage installation script and minor Flatpak install CMD

Muko-Tabi:
The AppImage installation script included downloading specific or generic, creating .desktop with icon, and uninstalling.
image

### DIFF
--- a/appimageinstall.sh
+++ b/appimageinstall.sh
@@ -1,12 +1,58 @@
 #!/bin/bash
 
+# Function to check if AVX2 is supported
+check_avx2_support() {
+    if grep -q avx2 /proc/cpuinfo; then
+        return 0  # AVX2 supported
+    else
+        return 1  # AVX2 not supported
+    fi
+}
+
+# Function to check if Zen Browser is installed
+check_installation_status() {
+    if [ -f ~/.local/share/AppImage/ZenBrowser.AppImage ]; then
+        return 0  # Zen Browser installed
+    else
+        return 1  # Zen Browser not installed
+    fi
+}
+
+# Function to check if zsync is installed
+check_zsync_installed() {
+    if command -v zsync &> /dev/null; then
+        return 0  # zsync is installed
+    else
+        return 1  # zsync is not installed
+    fi
+}
+
 # Kawaii ASCII Art for the script
 kawaii_art() {
-    echo "╔════════════════════════════════════════════════╗"
-    echo "║                                                ║"
-    echo "║   (ﾉ◕ヮ◕)ﾉ*:･ﾟ✧  Zen Browser Installer ✧ﾟ･:*   ║"
-    echo "║                                                ║"
-    echo "╚════════════════════════════════════════════════╝"
+    echo "╔════════════════════════════════════════════════════╗"
+    echo "║                                                    ║"
+    echo "║    (ﾉ◕ヮ◕)ﾉ*:･ﾟ✧  Zen Browser Installer ✧ﾟ･:*      ║"
+    echo "║                                                    ║"
+    
+    if check_avx2_support; then
+        echo "║    CPU: AVX2 Supported (Optimized Version)         ║"
+    else
+        echo "║    CPU: AVX2 Not Supported (Generic Version)       ║"
+    fi
+
+    if check_installation_status; then
+        echo "║    Status: Zen Browser Installed                   ║"
+    else
+        echo "║    Status: Zen Browser Not Installed               ║"
+    fi
+
+    if check_zsync_installed; then
+        echo "║    zsync: Installed (Needed for Updates)           ║"
+    else
+        echo "║    zsync: Not Installed (Needed for Updates)       ║"
+    fi
+
+    echo "╚════════════════════════════════════════════════════╝"
     echo
 }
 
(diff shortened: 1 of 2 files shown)


mauro-balades:
Love it!

mauro-balades:
Should we auto-deduce what version (generic or specific) we should choose?

Muko-Tabi:
Probably, it's mostly based on the CPU right? If so, what were the requirement again, let me try.

appimageinstall.sh
```diff
@@ -0,0 +1,174 @@
...
+    echo "| 2) Generic - Slow but compatible with older    |"
+    echo "|    devices                                     |"
+    echo "+------------------------------------------------+"
+    echo "| 3) Exit                                        |"
+    echo "+------------------------------------------------+"
+    read -p "Enter your choice (1/2/3): " version_choice
+
+    case $version_choice in
+        1)
+            url=$(curl -s https://api.github.com/repos/zen-browser/desktop/releases/latest | grep "browser_download_url.*zen-specific.AppImage\"" | cut -d '"' -f 4)
```
mauro-balades, line 116: We could just download from `https://github.com/zen-browser/desktop/releases/latest/download/zen-specific.AppImage`

appimageinstall.sh
```diff
@@ -0,0 +1,174 @@
...
+            exit 1
+            ;;
+    esac
+}
+
+main_menu() {
+    echo "(★^O^★) What would you like to do?"
+    echo "+------------------------------------------------+"
+    echo "| 1) Install                                     |"
+    echo "+------------------------------------------------+"
```
mauro-balades, line 140: What about an option to update? We have zsync appimages

mauro-balades (changes requested):

mauro-balades:
> Probably, it's mostly based on the CPU right? If so, what were the requirement again, let me try.

`grep avx2 /proc/cpuinfo`

appimageinstall.sh
Muko-Tabi: Oh yeah that works too, my habit of downloading from the JSON due to the version number from other AppImage. Let me change that.

appimageinstall.sh
Muko-Tabi: I could do that, but let's me test it out, never try to update with a zsync before.

Muko-Tabi:
Ok should be all there, wonder if I'm missing anything else
image

mauro-balades:
It looks great!

Muko-Tabi:
Sure!~

Samueru-sama:
@Muko-Tabi sorry if this isn't the right place but I've recently discovered the AppImage script and I have some questions/suggestions. 

Were you able to update the appimage using zsync directly? the zsync file is meant to be used with appimageupdatetool since the updateinfo isn't a fixed url that zsync can directly use. 

Also you don't have to do `appimage --extract` to extract the whole appimage, the runtime supports arguments to extract certain files, you can see this example in the [install script we have for zen browser at AM.](https://github.com/ivan-hc/AM/blob/main/programs/x86_64/zen-browser#L90-L108)

And updates with appimageupdatetool are handled this way: 

```
if command -v appimageupdatetool >/dev/null 2>&1; then
	cd "/opt/$APP" || exit 1
	appimageupdatetool -Or ./"$APP" && chmod a+x ./"$APP" && echo "$version" > ./version && exit 0
fi
```

that if appimageupdatetool is in PATH it gets used and if it for some reason fails during the update it will continue updating by comparing the URL of the files.

Also zsync file never has to be downloaded when using appimageupdatetool, the update info is already embedded in the appimage and you can check it with the `--appimage-updateinfo` flag

Muko-Tabi:
@Samueru-sama Hey there! For the zsync update part then yes it does work, but I wouldn't say it's optimized way since I haven't been digging around much in AppImage docs. It's mostly comparing the zsync on github and the current one, and replace the AppImage if zsync isn't matched.

The extract part is one I use from my other AppImage script. Sometime in AppImage, there's symlink of .desktop and .icon that I tried some way, but it doesn't work out, so I end up extracting the whole.

And I wouldn't mind if you do make changes to my script or if I have some time I will look into AppImage for updating it. I was mostly making this quick and for fun since there were problems with the Flatpak ver.

Samueru-sama:
> @Samueru-sama Hey there! For the zsync update part then yes it does work, but I wouldn't say it's optimized way since I haven't been digging around much in AppImage docs. It's mostly comparing the zsync on github and the current one, and replace the AppImage if zsync isn't matched.

Oh that makes sense, in that case zsync is only used for version comparison and no delta update is done.

> The extract part is one I use from my other AppImage script. Sometime in AppImage, there's symlink of .desktop and .icon that I tried some way, but it doesn't work out, so I end up extracting the whole.

We handle that in the `while [ "$COUNT" -lt 10 ]; do` in the script I linked where it extracts `.DirIcon` and if it is a symlink it reads it and then extracts what the symlink pointed to.

If you wonder why it gives 10 tries, that's just to prevent it from getting stuck in an infinite loop in case there is a symlink pointing to another symlink that poitns back to the original symlink in the AppImage lol

> And I wouldn't mind if you do make changes to my script or if I have some time I will look into AppImage for updating it. I was mostly making this quick and for fun since there were problems with the Flatpak ver.

No problem, I thought the zsync was directly being used for delta updates,

Issue #1528, opened 2024-09-18
Right-Clicking on Compact Bar Closes It


OwenD30:
### Captchas

- [X] I have read the instructions.
- [X] I have searched existing issues and avoided creating duplicates.
- [X] I am not filing an enhancement request.

### What happened?

When compact mode is enabled, when you right-click on a tab in the bar while it's visible, the correct menu shows up but the bar hides itself.

It would make more sense for the bar to stay visible while you are performing an action on one of the tabs, and only close once your cursor leaves the bar area.

### Reproducible?

- [X] I have checked that this issue cannot be reproduced on Mozilla Firefox.

### Version

1.0.1-a.2

### What platform are you seeing the problem on?

Linux

Issue #1208, opened 2024-09-03
Right-clicking on the vertical tab causes it to immediately collapse


FlawlessCasual17:
### Captchas

- [X] I have read the instructions.
- [X] I have searched existing issues and avoided creating duplicates.
- [X] I am not filing an enhancement request.

### What happened?

Whenever I right-click on the tab bar, it immediately closes.

![(zen)J181C4_09-03-2024](https://github.com/user-attachments/assets/882c22c7-f471-4706-9ffa-c4d8713fe4cd)

Note: This bug can be only be reproduced if the _**Expand Tabs on Hover (Wont work on compact mode)**_ option under the <ins>**How to expand tabs**</ins> section is enabled.

### Reproducible?

- [X] I have checked that this issue cannot be reproduced on Mozilla Firefox.

### Version

1.0.0-a.35 (64-bit)

### What platform are you seeing the problem on?

Windows

### Relevant log output

```shell
ZenThemeImporter: Initiating Zen theme importer ZenThemesImporter.mjs:38:13
ZenThemeImporter: Zen theme imported ZenThemesImporter.mjs:43:15
ZenWorkspaces: Initializing ZenWorkspaces... ZenWorkspaces.mjs:11:13
ZenThemeModifier: init browser layout ZenStartup.mjs:15:13
Zen CKS: Initializing shortcuts ZenKeyboardShortcuts.mjs:220:15
ZenWorkspaces: ZenWorkspaces initialized ZenWorkspaces.mjs:14:15
ZenThemeModifier: set tabs max-height to 803px ZenStartup.mjs:60:13
ZenThemeModifier: set tabs max-height to 802px 3 ZenStartup.mjs:60:13
ZenThemeMarketplaceParent(settings): Building themes list zen-settings.js:162:13
ZenThemeModifier: set tabs max-height to 802px ZenStartup.mjs:60:13
ZenThemeMarketplaceParent: Updating themes ZenThemeMarketplaceParent.sys.mjs:10:17
ZenThemeMarketplaceParent(settings): Building themes list zen-settings.js:162:13
ZenThemeModifier: set tabs max-height to 812px ZenStartup.mjs:60:13
ZenThemeMarketplaceParent: Uninstalling theme ZenThemeMarketplaceParent.sys.mjs:19:17
[...]

Issue #1024, opened 2024-08-28
Sidebar collapses when right clicking tab


NyxErinys:
### What happened?

With `Expand Tabs on Hover` set and side bar set to normal mode, when you go to right click on a tab the sidebar collapses. This happens across Windows, Linux, and MacOS.

This seems to be a side effect of how Firefox handles the right-click context menu compared to chromium-based browsers and Safari. Firefox continues to check hover state even when the right-click context menu is open so when you mouse in on the menu it just doesn't hover anything on the page, meanwhile the other two lock the hover state of the page underneath when the menu is open.

While this seems like a Firefox thing I thought I'd report it here because it affects something zen specific (the Sidebar).

Quick clip showing the main issue and then showing how Zen (and Firefox) handle hover and right click
[Screencast from 2024-08-28 16-14-33.webm](https://github.com/user-attachments/assets/dcf1101c-8849-4fe4-b27a-790a58875382)

### Reproducible?

- [ ] I have checked that this issue cannot be reproduced on Mozilla Firefox.

### Version

1.0.0.a.32-1

### Severity impact

Low

### What platform are you seeing the problem on?

Linux

mauro-balades:
This has been fixed for the next release, thanks!

PR #1532, opened 2024-09-18
Fix: keep sidebar visable on right click in compact mode, vertical tabs (expand on hover)

BrhmDev:
Keep sidebar visible on right click in compact mode, vertical tabs (expand on hover)

Also removed sidebar window-drag in compact mode and vertical tabs as I noticed it caused some weird behavior where the sidebar would close sometimes when you clicked on it.

Resolves: https://github.com/zen-browser/desktop/issues/1528, https://github.com/zen-browser/desktop/issues/1024, https://github.com/zen-browser/desktop/issues/1208

### DIFF
--- a/src/browser/app/profile/zen-browser.js
+++ b/src/browser/app/profile/zen-browser.js
@@ -100,12 +100,6 @@ pref('zen.keyboard.shortcuts.disable-firefox', false);
 pref('zen.tabs.dim-pending', true);
 pref('zen.themes.updated-value-observer', false);
 
-#ifdef XP_MACOSX
-pref('zen.themes.tabs.legacy-location', true);
-#else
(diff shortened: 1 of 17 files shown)


mauro-balades:
Woah, thanks!

grester:
Does this also address the same behaviour problem regarding the top bar in collapse mode?
Might I also add that when in "Customize toolbar" mode it also most of the type disappears and you have no idea where you're dragging the elements.

BrhmDev:
> Does this also address the same behaviour problem regarding the top bar in collapse mode? Might I also add that when in "Customize toolbar" mode, it also most of the times disappears and you have no idea where you're dragging the elements.

No, but could be refactored to a function that adds popup detection to an element. Then it could be used for those as well.

PR #1576, opened 2024-09-21
Add method to gZenUIManager to track popups on any element

BrhmDev:
Refactor https://github.com/zen-browser/desktop/pull/1532 so that we can track popups on any element.

Add popup tracking to top toolbar (haven't added css yet as it currently doesn't retract and compact mode is still being worked on)


mauro-balades:
Thanks!

PR #1597, opened 2024-09-22
Update better-fox.js - Remove sus tweaks

gunir:
Update better-fox.js - Remove sus tweaks


AvinashK13:
Also betterfox v130 (https://github.com/yokoffing/Betterfox/blob/130.0/user.js) is about to release.
zen code shows it is using old v128

Issue #941, opened 2024-08-26
MacOS window control hover does not keep top toolbar visable


JaeTLDR:
### What happened?

When Zen is in compact mode with Top Toolbar hidden, hovering the 'traffic light' controls blurs the Top Toolbar back to hidden making it difficult to minimize/zoom/close

### Reproducible?

- [ ] I have checked that this issue cannot be reproduced on Mozilla Firefox.

### Version

1.0.0-a.29

### Severity impact

Medium

### What platform are you seeing the problem on?

macOS - aarch64

PR #1726, opened 2024-09-25
[FIX] MacOS titlebar hover

HarryHeres:
When Compact mode was enabled and the top bar was shown with the hover event, when hovering over the native macOS window buttons, the top bar would start glitching. This PR removes native buttons and create custom ones while resembling the native buttons in terms of design.

Also closing #941


mauro-balades:
Isnt there a way to remove all the .patch files? It's a hustle to mantain when updating the firefox version....

HarryHeres:
> Isnt there a way to remove all the .patch files? It's a hustle to mantain when updating the firefox version....

Hmm, I'll try.

HarryHeres:
Also, I'll add the hover menu for the maximize button.

PR #1742, opened 2024-09-25
FIX: Remove Focus of the URL Bar at Startup

Matt-FTW:
When the browser launches, the urlbar is always focused, instead of the homepage. This behavior doesnt happen on Firefox, and it can be really annoying to always click out to unfocus the urlbar.

Closes:  #1169

### DIFF
--- a/src/browser/base/content/ZenStartup.mjs
+++ b/src/browser/base/content/ZenStartup.mjs
@@ -8,7 +8,6 @@
       window.SessionStore.promiseInitialized.then(async () => {
         this._changeSidebarLocation();
         this._zenInitBrowserLayout();
-        this._focusSearchBar();
       });
     },
 


Anthonyy232:
Can we make this a toggle? I think this behavior is extremely convenient and use it often.

mauro-balades:
It's not really a fix since this is a feature... I could make it a bit smarted like focusing only on "newtab"... But focusing the URL bar is really useful for some people

ScarsTRF:
For someone like me using something else like a self hosted [homepage](https://gethomepage.dev/) as their default loaded tab, it kind of makes it unusable in my case as a full keyboard user. For example I rely on my Keyboard actions to be focused to the page at first load rather than the search bar as I quickly navigate around my homelab with this page in mind. Having it as a toggle would be nice especially for those who find it daunting for them or even as an issue for those like me.

loustact:
I also use a custom hompeage that has its own search box. Whenever I open that homepage in Firefox, it always focuses on it. Then when I used Zen, it always focuses on the URL bar which is quite disruptive to what I always do. Since it makes sense that a need for this would be 50/50 depending on the person, this should be set as a toggle instead.

Soham52:
I too think that it's a bit disruptive, so adding an option to toggle this would be really good addition IMO

loustact:
I think this should be tracked as an issue so this will get noticed again.

PR #1747, opened 2024-09-26
Add Ctrl+Tab to cycle between recent tabs

gunir:
Add Ctrl+Tab to cycle between recent tabs


mauro-balades:
Just enable it yourself, I personally don't like It and many people don't either...

Issue #1718, opened 2024-09-25
Expanding tabs on hover makes the website layout "jump"


Sajito:
### Captchas

- [X] I have read the instructions.
- [X] I have searched existing issues and avoided creating duplicates.
- [X] I am not filing an enhancement request.

### What happened?

I have enabled "Expand tabs on hover". Since version 1.0.1-a.4 the page content visually jumps a bit, when hovering over the tab bar.

https://github.com/user-attachments/assets/3ab687a4-c885-4eb8-9763-8d17a3169524

### Reproducible?

- [X] I have checked that this issue cannot be reproduced on Mozilla Firefox.

### Version

1.0.1-a.5

### What platform are you seeing the problem on?

Linux

sjclayton:
@Sajito 

This isn't just since 1.0.1-a.4 - it's been doing that for a while...

Sajito:
It was working perfectly fine on multiple devices with 1.0.1-a.3. I know it existed before, but - at least from my pov - it's a regression since 1.0.1-a.4.

KarimElsayad247:
I think it's related to this same problem of "Search in page" tab making the content jump as well.

https://github.com/user-attachments/assets/89d63d51-664d-4cab-8d43-b9de44874888

yvvki:
This is a regressing issue from #957.

mateuszkozakiewicz:
The issue is not fixed in `1.0.1-a.6` even though changelog say it is
Additionally in `1.0.1-a.6` the sidebar itself shifts downwards when expanding there are some weird separators at the top

XInTheDark:
I can confirm, the website still shifts slightly upon expanding. And the separators have also appeared

https://github.com/user-attachments/assets/3c65ebd5-9437-4599-991b-8bf6c901daa9

Microwave-WYB:
This is a big problem... I can never just click on the tab I wanted to click. I must wait for the tab panel to expand, then find the tab again, move my cursor there and click.

https://github.com/user-attachments/assets/48d2123f-827d-42c4-ac47-f79829a78b7a

yvvki:
> This is a big problem... I can never just click on the tab I wanted to click. I must wait for the tab panel to expand, then find the tab again, move my cursor there and click.

@Microwave-WYB I think this is a separate issue of glitching tabs, the layout "jump" has since been fixed in 1.0.1-a.7.

Microwave-WYB:
Good call. I should open a separate issue.

[...]

PR #1757, opened 2024-09-27
Fix sidebar layout shift when expand on hover is on

### LINKED ISSUE #1581: Side bar with expand on hover become nightmare once started pinning tabs
### Captchas

- [ ] I am not filing an enhancement request.

### What happened?

If multiple tabs are pinned it arranges in grid on the top. But this makes the sidebar usage difficult as the tab arrangement in collapsed state differs from the expanded state. Provide the grid arrangement as optional.

https://github.com/user-attachments/assets/5fb13903-070e-4405-b4c8-962435d06e28

Also the Info message of empty Bookmarks bar overlays the sidebar. Whether the new tab placement is right? It should be placed to the end of tabs right(If the tab bar space get filled stick it above the list all tabs)

### Reproducible?

### Version

1.0.1-a.2 (64-bit)

Windows

### Relevant log output

_No response_

DanielSRS:
### Context:

When enabling **Expand Tabs On Hover** the layouts of collapsed sidebar and the expanded one are different in spacing and positioning of the elements making the hole UI jump and shifts every time the sidebar is expanded which is really uncomfortable to use.

The differences I noticed:
- Horizontal padding/margins are different, making the tabs shifts left and right
- Vertical paddings are different, making the content shift up and down
- Pined tabs changes to a grid layout when expanded, making the available space change, so all content jumps to newer positions, making especially hard to switch tabs.
- The workspaces + new page area also changes from a column to row layout, also making the content move position

This PR aims to adjust some of those things and make the experience a little better.

### What changed

To archive this, I updated some spacing/margins and changed some layouts to make the expanded sidebar looks the same as the collapsed sidebar, with same positioning of elements, changing only the available width between the two.

| Before the changes  | After changes |
| ------------- | ------------- |
| <video src="https://github.com/user-attachments/assets/110843c9-f4fd-491b-b6d4-7ec4ba40fb84"> | <video src="https://github.com/user-attachments/assets/923a9b00-650f-4e2a-949b-4c672afa9d78"> |

### Who is affected with this changes

I limited the changes to be applied only when the Expand Tabs On Hover setting is enabled, so all the others layout combinations are not expected to be impacted.

### Issues:

Fixes #1581
Related #1718

### Needs improvement:

I am not sure if this changes are aligned with the experience you intended to make, so if this were to be merged, there are two things to take notice:

- Sidebar animation on hover is disabled because it shifts contents to left before moving them to right, giving the same feeling of elements jumping
  - I tried adding a new one animating the width instead, but couldn't make it work

- Also zen-sidebar-icons-wrapper is still expanding vertically a few pixels and I couldn't figure out why. Is noticeable only when you have enough tabs to fill the vertical space

### DIFF
--- a/src/browser/base/content/zen-styles/zen-tabs/vertical-tabs.css
+++ b/src/browser/base/content/zen-styles/zen-tabs/vertical-tabs.css
@@ -122,6 +122,11 @@
     & > toolbarbutton:not(#zen-workspaces-button) {
       padding: 0 !important;
     }
+
+    @media (-moz-bool-pref: 'zen.view.sidebar-expanded.on-hover') and (not ((-moz-bool-pref: 'zen.view.compact') and (-moz-bool-pref: 'zen.view.compact.hide-tabbar'))) {
+      align-items: unset;
+    }
+
   }
 
   #newtab-button-container {
@@ -151,6 +156,13 @@
     }
   }
 
+  /**
+   * Match pinned tabs with other tabs spacing
+   */
+  #navigator-toolbox[zen-has-hover='true'] #vertical-pinned-tabs-container .tab-icon-image {
+    padding-right: 10px;
+  }
+
   /* Mark: toolbox as expanded */
   #navigator-toolbox:is(
     #navigator-toolbox[zen-user-hover='true'][zen-has-hover],
@@ -170,6 +182,15 @@
       margin-left: 2px;
       margin-right: 2px;
 
+      /**
+       * Do not change workspace button to row direction when expand on hover
+       * is on because it makes elements to jump to new positions. keep it with
+       * the same layout as when its collapsed
+       */
+      @media (-moz-bool-pref: 'zen.view.sidebar-expanded.on-hover') and (not ((-moz-bool-pref: 'zen.view.compact') and (-moz-bool-pref: 'zen.view.compact.hide-tabbar'))) {
+        flex-direction: column;
+      }
+
       & .zen-workspace-sidebar-icon[no-icon='true'] {
         display: none;
       }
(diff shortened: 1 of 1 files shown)


mauro-balades:
Thanks a lot!

Anthonyy232:
As mentioned, some animation would be nice before adding

DanielSRS:
Yeah, I should've kept the animation (removing the one line that unsets it, brings it back)

But I see there is a newer commit that replaces it with a width transition, which I think is better. I tried to do exactly that but didn't wok. Looking now, I probably forgot to test with an **_!important_**

mauro-balades:
I tried to add some a cool animation
[7HlTZPh.webm](https://github.com/user-attachments/assets/424bd1c7-fe7e-4d02-bdaf-3370dc51eda4)

Anthonyy232:
The out animation is good, but the lack of in animation looks a little unpolished. It should be the same as out animation but faster probably

DanielSRS:
I think I was able to make the transition added in 78f60b866e7a17fe4b74ef5be3099509f714634d work when expanding and collapsing the sidebar, but I ended up changing a lot of code. I will open a PR to discuss better

| Normal  | Slow |
| ------------- | ------------- |
| <video src="https://github.com/user-attachments/assets/f9caad78-25d6-4c3f-9459-a5000104a014"> | <video src="https://github.com/user-attachments/assets/f9e996e8-1776-4dd3-9c73-a869d696a0ba"> |

Anthonyy232:
Looks really good!

m3enzo:
Is this issue fixed in the latest release? Or will this be in a future build?

PR #1782, opened 2024-09-28
Fix web panel extensions, default web panel icons not visible

BrhmDev:
Fix web panel extensions by making the toolbarbutton the 'tab'.

Fixes after google fetch for favicons removed:
- Open site when adding to make sure icon is loaded.
- Store default site icons locally so they are loaded from the start.

Components PR: https://github.com/zen-browser/components/pull/31


mauro-balades:
Sorry but I dont think adding the default icons directly is a good option, specially if the default sites can change...

Issue #960, opened 2024-08-27
Brower cannot be started if built. [linux]


Autumnlight02:
### What happened?

I basically pulled it, ran the build and I continuesly get the following error:
![image](https://github.com/user-attachments/assets/a8a29619-24ec-4373-b1fb-f0ff04573fc0)

I ran the bootstrap command, the build command and then start

### Reproducible?

- [X] I have checked that this issue cannot be reproduced on Mozilla Firefox.

### Version

1.0.0

### Severity impact

High

### What platform are you seeing the problem on?

Linux

### Relevant log output

```shell
[autumnlight@autumnlight]npm run start

> zen-core@1.0.0 start
> cd engine && ./mach run --noprofile

 0:00.43 /var/home/autumnlight/Documents/github/personal/zen-browser/desktop/engine/obj-x86_64-pc-linux-gnu/dist/bin/zen -no-remote
JavaScript error: resource://gre/modules/ShortcutUtils.sys.mjs, line 142: TypeError: keyAttribute is null
console.info: "ZenThemeImporter: Initiating Zen theme importer"
console.info: "ZenThemeImporter: Zen theme imported"
console.info: "ZenWorkspaces: Initializing ZenWorkspaces..."
JavaScript error: chrome://browser/content/tabbrowser/tabbrowser.js, line 6064: TypeError: this.selectedBrowser is undefined
console.info: "ZenWorkspaces: ZenWorkspaces initialized"
JavaScript error: , line 0: uncaught exception: undefined
JavaScript error: , line 0: uncaught exception: undefined
JavaScript error: resource:///modules/sessionstore/SessionStore.sys.mjs, line 2749: TypeError: browser is undefined
console.info: "ZenThemeModifier: init browser layout"
console.info: "Zen CKS: Initializing shortcuts"
JavaScript error: chrome://browser/content/tabbrowser/tabbrowser.js, line 6064: TypeError: this.selectedBrowser is undefined
JavaScript error: chrome://browser/content/tabbrowser/tabbrowser.js, line 6071: TypeError: this.selectedTab is undefined
JavaScript error: resource:///modules/UrlbarInput.sys.mjs, line 3669: TypeError: this.window.gBrowser.selectedBrowser is undefined
JavaScript error: resource:///modules/ReportBrokenSite.sys.mjs, line 389: TypeError: tabbrowser.selectedBrowser is undefined
console.info: "ZenThemeModifier: set tabs max-height to" "577.6666870117188px"
[...]

PR #1795, opened 2024-09-29
Add necessary step for building to Readme

KarimElsayad247:
The browser would fails to launch properly without invoking this command, which I only discovered in a closed issue #960

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,7 +56,7 @@ winget install --id Zen-Team.Zen-Browser
 winget install --id Zen-Team.Zen-Browser.Optimized
 ```
 
-#### 
+####
 
 ### macOS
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -56,7 +56,7 @@ winget install --id Zen-Team.Zen-Browser
 winget install --id Zen-Team.Zen-Browser.Optimized
 ```
 
-#### 
+####
```
KarimElsayad247, line 59: Also this seems like a stray subheading

KarimElsayad247:
The alternative would've been to add it in the package.json, but adding an extra instruction seemed like a simper solution.

mauro-balades:
Thanks!

PR #1883, opened 2024-10-03
Temporary logo color change

Donnnno:
Hi!

We talked about this a bit on Discord, but wanted to give you the choice too.

This is a temporary logo change, till we have a new one! 

Using a colored background can result in looking less like a certain military logo, and look actually friendlier to use. 

image

image


AvinashK13:
I'm not aware of the discussion going on but the bg colour is horrendous!
Can we change the colour to something darker scheme pls?

Donnnno:
@AvinashK13 There are plans to make a new logo, and lots of submissions are already made on the Discord server.
However, that probably takes a while before it's here.

This color is actually part of one of the Zen base colors we have since the start

image

Anthonyy232:
It would probably be better to just wait.. multiple rebranding isn't exactly a good idea due to confusion

mauro-balades:
@Donnnno could we add it just for Zen Twilight branch?

Donnnno:
Done @mauro-balades 

I also made the gap a bit bigger in between the shapes.

mauro-balades:
Great, thanks!

PR #1892, opened 2024-10-03
Feature: Expand default workspace icons

kristijanribaric:
- Added additional default workspace icons to expand the available options.

- Improved the styling of the workspace icon selector to enhance the user experience when displaying more than two rows of icons. The container height is now limited to two rows, with a scrollbar introduced for any overflow, preventing the container from expanding beyond two rows.

https://github.com/user-attachments/assets/4ddc2b16-4bbd-447b-a385-df8e5084a019


mauro-balades:
I have a different idea on this topic and I would like to try it out, sorry!

PR #1895, opened 2024-10-04
Try to fix exec: zen: not found

Drsheppard01:
Same issue https://github.com/ungoogled-software/ungoogled-chromium-portablelinux/pull/121/files


Drsheppard01:
Not worked. Need to investigate this issue

PR #1933, opened 2024-10-06
fix performance

n7itro:
Fixes lag while scrolling as reported by users.

This isn't in Betterfox for a reason. Hardware acceleration always causes compatibility issues.


Mil0dV:
So not sure if this due to this, but: I have a lot lower cpu usage in google meets in Zen than FF, much appreciated. And now issues scrolling so far. Radeon 760M

PR #1943, opened 2024-10-07
Preinstall Brave Search and uBlockOrigin

GunGunGun:
Preinstall Brave Search and uBlockOrigin


GunGunGun:
Note: not finished yet, this PR requires more commits

mauro-balades:
Surfer does it for you. I don't know what you are doing with policies

sjclayton:
@GunGunGun 

Not sure this is necessary, if people want these, they have either already installed them / know how to do it on their own. By pre-installing you are relying on people being familiar with both, and knowing how to configure them. 

Also, it potentially opens Zen up to having to deal with additional issue reports arising from people not realizing that a problem they are facing could be due to the default configuration of either of these extensions.

CostcoFanboy:
Very against the concept of pre-installing any components from third parties.
If you want to create a "Start-Up" page with popular suggestions, I think that'd be okay.

mauro-balades:
As said before, this isn't how surfer manages extensions and search engines. Please checkout https://docs.gluon.dev/guides/includingaddons/

PR #1958, opened 2024-10-07
Installation of Zen on CachyOS

SoongVilda:
I’ve added both variants of `zen-browser-bin` and `zen-browser-avx2-bin`. CachyOS has Zen available in the repository.


mauro-balades:
Great!

PR #1983, opened 2024-10-09
Feature: Enable syncing of Workspace icons by default

kristijanribaric:
This PR adds a new preference (`services.sync.prefs.sync.zen.workspaces.icons`) to enable syncing of the Workspace icons across different devices. This allows users to maintain the same visual experience regardless of where they are using Zen.


mauro-balades:
Great!

PR #1994, opened 2024-10-10
Feature: Add option to change workspaces with side button and right-click

kristijanribaric:
This PR introduces a new preference (`zen.tabs.change-workspace-on-side-button-and-right-click`) that allows users to change workspaces using the mouse side button and right-click on the tab bar.

### Why not just mouse side buttons? 
Because Mozilla made it very hard to capture only mouse side buttons and this seems the best way to implement it.

### DIFF
--- a/src/browser/app/profile/zen-browser.js
+++ b/src/browser/app/profile/zen-browser.js
@@ -102,6 +102,7 @@ pref('zen.view.sidebar-collapsed.hide-mute-button', true);
 
 pref('zen.tabs.dim-pending', true);
 pref('zen.tabs.newtab-on-middle-click', true);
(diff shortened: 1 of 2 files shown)


sjclayton:
@kristijanribaric 

Ok, but which "side button" are you referring to here? If you are referring to mouse side buttons, then which one would it be, back or forward? 

If these are the buttons you're referring to, then there should be some pref to set which button this uses, not just enable or disable the option... as many users will have these buttons set to their default mouse behaviour of navigating back/forward or have them set to some other function (like I do). Overriding one of these buttons without specifying which could lead to users being confused by the resultant behaviour.

PR #2043, opened 2024-10-12
Fix change on hover setting not doing anything.

### LINKED ISSUE #1823: Change tabs on hover setting checkbox has no affect
### Captchas

### What happened?

The option `Change tabs on hover` in Look and Feel > Split View seems to be always getting enabled whether or not the box is checked.

### Reproducible?

### Version

1.0.1-a.6

Linux

### Relevant log output

_No response_

BrhmDev:
Change on hover didn't work because it changed a different pref than the one in zenViewSplitter.mjs, I chose to keep the version which is most consistent with other splitview prefs.

Closes: https://github.com/zen-browser/desktop/issues/1823

### DIFF
--- a/src/browser/components/preferences/zen-settings.js
+++ b/src/browser/components/preferences/zen-settings.js
@@ -1006,7 +1006,7 @@ Preferences.addAll([
     default: true,
   },
   {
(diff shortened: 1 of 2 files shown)


mr-cheffy:
Great, thanks!

PR #2050, opened 2024-10-12
Fix: Store pinned tab triggering principal in sessionstore

kristijanribaric:
Depends on: https://github.com/zen-browser/components/pull/50


mr-cheffy:
Thanks!

PR #2072, opened 2024-10-13
Disable upcoming annoying Search Terms feature

GunGunGun:
Disable upcoming annoying Search Terms feature


mr-cheffy:
Im not going to merge this as I quite like them. If people want it off, they can just turn it off in the preferences page. Thanks!

PR #2081, opened 2024-10-13
Fix: Open external links in workspace's default container

kristijanribaric:
Pass `fromExternal` to `ZenWorkspaces.getContextIdIfNeeded`

Depends on: https://github.com/zen-browser/components/pull/52


mr-cheffy:
Thanks a lot!

PR #2095, opened 2024-10-14
Feature: Force container tabs to a workspace where that container is set as default

kristijanribaric:
This PR introduces a new preference, "zen.workspaces.force-container-workspace", that allows users to control whether container tabs are automatically placed in a  workspace where that container is set as default. When enabled, this feature ensures that all tabs from a container are grouped together within a specific workspace.

When used alongside `Firefox Multi-Account Containers` extension, it enables always opening specific sites in a specific workspace.

The following files were modified:

- `src/browser/components/preferences/zen-settings.js`: Added the new preference definition.
- `src/browser/app/profile/zen-browser.js`: Set the default value for the new preference.
- `src/browser/components/preferences/zenTabsManagement.inc.xhtml`: Added a checkbox to control the new preference in the settings UI.

https://github.com/user-attachments/assets/4f616019-5d0e-4ac1-a04d-4d8a72c88090

https://github.com/user-attachments/assets/b648e49e-be8a-4c0e-8160-87d82c748b78

Depends on:
https://github.com/zen-browser/components/pull/53
https://github.com/zen-browser/l10n-packs/pull/74

### DIFF
--- a/src/browser/app/profile/zen-browser.js
+++ b/src/browser/app/profile/zen-browser.js
@@ -135,6 +135,7 @@ pref('zen.workspaces.enabled', true);
 pref('zen.workspaces.hide-default-container-indicator', true);
 pref('zen.workspaces.individual-pinned-tabs', true);
 pref('zen.workspaces.show-icon-strip', true);
+pref('zen.workspaces.force-container-workspace', true);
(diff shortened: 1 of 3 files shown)


LarveyOfficial:
🎉

PR #2113, opened 2024-10-15
Avoid misunderstanding for arch-based installation in README.md

VDuchauffour:
This PR replaces `CachyOS` installation instructions in favor of generic Arch-based distribution.
It also fixes the right repository for the pacakges (ie `AUR` instead of the official repository).
I specified `yay` as the AUR helper in the command.


mr-cheffy:
Thanks!

PR #2134, opened 2024-10-15
Fix wrong StartupWMClass in .desktop for Twilight AppImage

LeMoonStar:
When installing Zen Twilight through the AppImage installer script and executing it, we quickly notice that our desktop environment (GNOME, in my case) doesn't match the window to the correct `.desktop` file, causing no icon to appear.

This is because the Zen Twilight window has a different `WMClass` than the normal Zen installation: `zen-twilight` instead of `zen-alpha`. This PR is similar to #2004, as it does a simple find and replace on the `.desktop` file.

## Before this change

Zen Twilight is pinned, yet, the window appears with the default application icon next to the pinned Zen Twilight icon.
Screenshot from 2024-10-15 17-18-43

## After this change

The window is matches to the `.desktop` file and assigned the correct Icon
image

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -354,6 +354,7 @@ jobs:
 
           if [ "${{ inputs.update_branch }}" = "twilight" ]; then
             sed -i -e 's/Name=Zen Browser/Name=Zen Twilight/g' AppDir/zen.desktop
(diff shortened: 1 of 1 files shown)


mr-cheffy:
Great, thanks!

PR #2145, opened 2024-10-15
Fixing moving tab bar to right

### LINKED ISSUE #334: MacOS "Tabs on the right" option not functional
### What happened?

When opening the context menu in the Vertical Tab Bar and selecting "Tabs on the right" on my M3 Pro Macbook, nothing happens. However, when I perform the same action on my Windows 11 PC, it functions as expected. There is a check mark present on the option after selecting it on MacOS, but the Vertical Tab Bar remains on the left side.

### Reproducible?

### Version

1.0.0-a.23

### Severity impact

Low

macOS - aarch64

### Relevant log output

_No response_

harryheres:
Fixes #334


mr-cheffy:
Thanks a lot!

PR #2333, opened 2024-10-26
Fix an issue where workspace size is 0 on non-expanded toolbar

kwilcz:
When setting "Display workspaces as an icon strip" is disabled, the selected workspace is not correctly rendered on the screen.

## Before:
zen_xLFU6xiPhz

## After
zen_w3cnb5tiX8


mr-cheffy:
Thanks!

PR #2411, opened 2024-10-30
bugfix: Fix Temporary Containers Automatic mode | change firefox -> zen for default cookiestore id

### LINKED ISSUE #172: Temporary Containers Automatic Mode doesn't work on Zen Browser
### What happened?

# Temporary Containers Extension Issue

## Extension Details
- Name: Temporary Containers
- URL: https://addons.mozilla.org/en-US/firefox/addon/temporary-containers/

## Issue Description
The Temporary Containers extension is designed to automatically assign new tabs to temporary containers when the "Automatic Mode" setting is enabled. This feature works as expected in regular Firefox up to the current Developer Edition as of now, even if the extension is outdated. However, when using the Zen browser the "Automatic Mode" feature doesn't work.

## Steps to Reproduce
1. Install the Temporary Containers extension in Firefox.
2. Open the Temporary Containers extension settings.
3. Go to General and tick "Automatic Mode".
4. Open a new tab.

## Expected Result
When opening a new tab in step 4, the Temporary Containers extension should automatically assign the tab to a temporary container.

## Actual Result
When opening a new tab in step 4 in Zen Browser, the Temporary Containers extension does not automatically assign the tab to a temporary container.

##
- Would be nice if there's a workaround or the project can integrate the extension into it, since this is currently why I'm still using normal Firefox. Thanks in advance :D

### Reproducible?

### Version

Alpha build - 1.0.0-a.12

### Severity impact

Medium, Enhancement

Linux, Windows

### Relevant log output

_No response_

lbrooney:
Closes #172 

This PR fixes automatic mode for Temporary Containers by patching `firefox-default` to `zen-default`. I also included changes to the other store names to maintain consistency.

Based off brief testing this does not impact Workspaces. The default container of a workspace takes precedence over an automatic container generated say by Temporary Containers. 

I can't comment whether this fix breaks other container extensions. If they hardcoded checking cookieStoreId against `firefox-default`, then this fix would break them.

### Why temporary containers doesn't work
Inspecting the Temporary Containers extension and then opening a new tab, you will notice an uncaught error from `showOrHide` caused by L4943. 

Examining Temporary Containers source code, `showOrHide` is found in [`pageaction.ts`](https://github.com/stoically/temporary-containers/blob/main/src/background/pageaction.ts) with the line causing the throw at [L43](https://github.com/stoically/temporary-containers/blob/cbb39f7abbdf3c767bd1fbdc109394a0db169b05/src/background/pageaction.ts#L43). The correct behaviour is the `if else` on [L31](https://github.com/stoically/temporary-containers/blob/cbb39f7abbdf3c767bd1fbdc109394a0db169b05/src/background/pageaction.ts#L31) executing, not the else. 

In Zen that `if else` doesn't execute because Temporary Container assumes the default cookieStoreId is the [browser's name](https://github.com/stoically/temporary-containers/blob/cbb39f7abbdf3c767bd1fbdc109394a0db169b05/src/background/tmp.ts#L95)  (`zen`) appended with `-default`. However the default cookiestore id is set to [`firefox-default`](https://searchfox.org/mozilla-central/source/toolkit/components/extensions/parent/ext-toolkit.js#38). Thus the `else if` doesn't execute as `zen-default` does not equal `firefox-default` and the extension proceeds to the `else` causing the error.

More generally if the above error did not occur, Automatic mode would still not would as the necessary steps would not trigger [here](https://github.com/stoically/temporary-containers/blob/cbb39f7abbdf3c767bd1fbdc109394a0db169b05/src/background/tabs.ts#L177)(for on created automatic mode) or would return early [here](https://github.com/stoically/temporary-containers/blob/cbb39f7abbdf3c767bd1fbdc109394a0db169b05/src/background/request.ts#L250)(for on request Automatic Mode) as it would be looking for `zen-default` when the value is actually `firefox-default`.

### Breaks:
- [Simple Tab Groups](https://github.com/Drive4ik/simple-tab-groups) automatic temporary container functionality
  - checks against [`firefox-default`](https://github.com/Drive4ik/simple-tab-groups/blob/f56fb33f7e6d6764415297ec1fedac43a3e02cf0/addon/src/js/constants.js#L15)


mr-cheffy:
Id have to investigate this, as idk if it'll log out of all session for all users updating

lbrooney:
Based off a quick test, when using the tab without a container, I don't get logged out of accounts when I change from `firefox-default` to `zen-default` and vice versa. This was done in the dev environment. Also doesn't affect the session status in other containers. Excited to hear your results!

lbrooney:
Hi,
Just wondering if there were anything blocking this from being merged? If there's something I can do to help out or speed it along I'm willing to contribute.

Thanks for all your hard work!

lbrooney:
This PR would break [Simple Tab Groups](https://github.com/Drive4ik/simple-tab-groups) automatic temporary container functionality as it checks against [`firefox-default`](https://github.com/Drive4ik/simple-tab-groups/blob/f56fb33f7e6d6764415297ec1fedac43a3e02cf0/addon/src/js/constants.js#L15)

vraravam (approved):
LGTM!

avelino (approved):
LGTM
this worked to solve the problem reported here #172 _(I tested locally)_

vraravam:
can this be merged please @mr-cheff ?

JSGRD22:
Does this break the Firefox Multi-Account Containers extension? The extension is essential for properly syncing containers together with workspaces.

lbrooney:
I personally haven't observed any breaking behaviour, but my testing hasn't been thorough.

Looking at the code of Multi-Account Containers it doesn't look like it [breaks anything](https://github.com/search?q=repo%3Amozilla%2Fmulti-account-containers%20firefox-default&type=code). Every new "no container" tab is counted as a container tab, but that's just a fun stat counter. It may impact Container remote syncing used by MAC, but I don't use that feature and I wouldn't know what behaviour change to observe.

In regards to behaviours when interacting with the Temporary Containers extension, if a workspace has a default container set, that default container will be used upon opening a new tab instead of a temporary container(assuming Temporary Containers was in Automatic mode). 

For people who want to fix Temporary Containers Automatic Mode for Zen without this patch you can do the following:
1. Make sure to have Temporary Containters installed and that it's done its first install setup
1. Go to [about:debugging](about:debugging) 
1. In Debugging go to *This Zen*
1. Scroll until you see the *Temporary Containers* extension and click the **Inspect** button
1. A Developer Tools menu should pop out, click on **Storage** and within storage click the **Extension Storage** Tab and that should expand to display something with the name *moz-extension://<uuid>*, click it
1.  Some items should popup on the right, at the top should be **containerPrefix**, edit the item text from `zen` to `firefox`
1. Restart Zen and automatic mode of Temporary Containers should work.

Shape6253:
I know no maintainers like bumping, nonetheless, here's a bump

mr-cheffy:
Woudn't it be better if we fixed `getBrowserInfo` instead?

mr-cheffy:
https://github.com/zen-browser/desktop/commit/148e63e2265b24b2eff9462de45dd8b146cfda29 should be a more elegant solution that also fixes long term concerns we have with extension support.

PR #2421, opened 2024-10-30
POLLY for MacOS - Another try

GunGunGun:
POLLY for MacOS - Another try


mr-cheffy:
This was already enabled before by another commit, sorry!

PR #2506, opened 2024-11-01
Update better-fox.js

cristiancmoises:
I found this telemetry enabled. 
We can disable and lock or maybe only disable.


mr-cheffy:
What does this do?

JamiKettunen:
`toolkit.telemetry.hybridContent.enabled` doesn't even exist by default, even then it probably should be set to `false` assuming it exists at all in the modern firefox source

mr-cheffy:
Im sorry, but are these PRs done with AI? These configs flags dont even exist in firefox and the last one had functions that doesnt even exist.

PR #2509, opened 2024-11-01
Brutefox

cristiancmoises:
Performance enhance with some configurations for better security.  Have a nice day.

### DIFF
--- a/src/browser/app/profile/better-fox.js
+++ b/src/browser/app/profile/better-fox.js
@@ -21,7 +21,7 @@
  * SECTION: FASTFOX                                                         *
 ****************************************************************************/
 /** GENERAL ***/
-//pref("content.notify.interval", 100000); // (.10s); alt=500000 (.50s)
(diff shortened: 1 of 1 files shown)


Anthonyy232:
why
> pref("browser.download.animateNotifications", false);

also hard coding cache memory? doesn't really seem ideal if the name is as accurate as it is descriptive, especially 32mbs..
> pref("browser.cache.memory.capacity", 32768);

cristiancmoises:
> why
> 
> > pref("browser.download.animateNotifications", false);
No animations = better performance.
we can remove this.
> 
> also hard coding cache memory? doesn't really seem ideal if the name is as accurate as it is descriptive, especially 32mbs..
> 
> > pref("browser.cache.memory.capacity", 32768);
we can remove this too. 
I just add some features; the developers need choose the best to add in the official code.

mr-cheffy:
There are reasons why we didnt enable these configs by default, they are mostly anti-productivity and annoying. Sorry

Issue #2554, opened 2024-11-01
Hover-expand tab bar with translucent theme colors


KZDKM:
### Captchas

- [X] I have read the instructions.
- [X] I have searched existing issues and avoided creating duplicates.
- [X] I am not filing an enhancement request.

### What happened?

The newly added theme color option supports setting translucent theme colors, but when expand-on-hover is enabled for the vertical tab bar, the background expanded tab panel is also translucent, this greatly reduces the readability of tabs. 

The following userChrome snippet fixes the issue:
```css
#navigator-toolbox[zen-user-hover="true"][zen-has-hover="true"] {
  #TabsToolbar{
    /*ignores the alpha channel of the background and make panel opaque*/
    /*background: rgb(from var(--zen-main-browser-background) r g b / 1) !important;*/

    /*keep the panel translucent but add blur behind it*/
    backdrop-filter: blur(10px);
  }
}
```

Additionally, the opaque theme color could also be set to the background of `browserContainers` to avoid pages temporarily becoming completely transparent when loading.
```css
.browserContainer {
  background: rgb(from var(--zen-main-browser-background) r g b / 1) !important;
}
```

### Reproducible?

- [X] I have checked that this issue cannot be reproduced on Mozilla Firefox.

### Version

1.0.1

### What platform are you seeing the problem on?

Linux

PR #2555, opened 2024-11-01
Fix vertical tab expand on hover background for translucent theme color

KZDKM:
See #2554


mr-cheffy:
Hi! Sorry, but this has been fixed in an older commit as the compact mode/expand on hover functionality has changed a bit

PR #2576, opened 2024-11-02
Indicator for splitted tabs

GunGunGun:
Indicator for splitted tabs

image

### DIFF
--- a/src/browser/base/content/zen-styles/zen-decks.css
+++ b/src/browser/base/content/zen-styles/zen-decks.css
@@ -226,3 +226,9 @@
 #zenSplitViewModifier:not([has-enabled-realloc])[toast] #zenSplitViewModifierEnabledToast {
   display: revert;
 }
+
+/* Indicator for splitted tabs */
+.tabbrowser-tab.split:not(:hover) .tab-close-button {
+  display: unset!important;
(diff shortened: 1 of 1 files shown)


GunGunGun:
Linked to https://github.com/zen-browser/components/pull/72

Michael679089:
Nah, I would like it how Microsoft Edge does it:
1. There'll be a button on the top right that enables the split tabs.

image

Or maybe like:
1. You hover the tab to the other tab.
2. It will pop out an option after dropping the tab, which can be either split tabs or grouped tabs. 
3. But since it's only split tabs, it will automatically be split tabs.

Maybe the pop out option suggestion will come out after the group tabs update.

I just don't like this pull request, it looks like just visual clutter.

mr-cheffy:
This will be merged into special tab groups when they come out, thanks!

Spacellary:
I would ask for a potential enhancement on this, show split tab groups with some indicator such as a border to indicate they are grouped/a set:

zen_FKYm5ORBC4

...or make sure the "split tab indicator icon" is colored differently for each set of split tabs so we can differentiate them, this approach would be better and more flexible.

Is that colorization aspect for differentiation planned?

ajkuftic:
Another option could be a hierarchy/bracket view that shows an outline of tabs that shows them connected? 
split heirarchy v2

PR #2625, opened 2024-11-04
feat(ui): Add new "Pin Tab Globally" context menu option

kristijanribaric:
This PR introduces a new context menu option called "Pin Tab Globally" which allows users to pin tabs to the global pin list, making them accessible from any workspace.

The icon for this option is set to the same as the existing "Pin Tab" context menu option.

### DIFF
--- a/src/browser/themes/shared/zen-icons/icons.css
+++ b/src/browser/themes/shared/zen-icons/icons.css
@@ -893,7 +893,8 @@ menuitem[contexttype='fullscreen'][label*='Exit'] {
 #context_pinSelectedTabs,
 #context_unpinSelectedTabs,
 .customize-context-moveToPanel,
-#context_zen-replace-pinned-url-with-current {
+#context_zen-replace-pinned-url-with-current,
+#context_zen-pin-tab-global{
   --menu-image: url('pin.svg');
 }
 


Shockshwat:
I think you could use this icon for the global pin. 
pin-global
Edit: The SVG looks kind of distorted on my machine, Here is the PNG version. 
pin-global

kristijanribaric:
Nice one, thx!

kristijanribaric:
Is that an icon from source code or a custom one?

Shockshwat:
The icon from source code + added global icon at the bottom (custom)

mikedlr:
I wonder if there's a more general approach that could work better? I'd like to "pin tab mostly" but have certain workspaces where I really only want the things related to that workspace. Flexibility to have a pin on two workspaces but not all might be neat - or it might be too messy? 

Edit - and more importantly, maybe having a tab pinned on two workspaces but not many others. E.g. company email on "communications" generally + "my company" tab, but no on "other company" or "running club"

kristijanribaric:
Too messy.

This gives you ability to pin a tab:
a) to a specific workspaces
b) to all workspaces

You can have for example 3 tabs on every workspace that are pinned to that workspace and 3 global pins which would be shown across all workspaces

PR #2656, opened 2024-11-05
Fix: Parse user context ID correctly when pinning tabs

kristijanribaric:
This PR fixes an issue where the user context ID was not being parsed correctly when pinning tabs.

Previously, the user context ID was being stored as a string, which could lead to unexpected behavior. This commit ensures that the user context ID is parsed as an integer, which is the correct data type.

This change will ensure that pinned tabs are stored and retrieved correctly.


mr-cheffy (approved):

mr-cheffy:
Thanks!

PR #2665, opened 2024-11-05
Feature: Add new icons for essentials context items

kristijanribaric:
Basically the title

Oh and I updated context menu item showing logic.

Adding essentials context item:
image

Remove from essentials context item:
image

Depends on:
https://github.com/zen-browser/l10n-packs/pull/94


src/browser/base/zen-components/ZenPinnedTabManager.mjs
```diff
@@ -388,6 +391,9 @@
       for (let i = 0; i < tabs.length; i++) {
         const tab = tabs[i];
         tab.removeAttribute("zen-essential");
+        if(ZenWorkspaces.workspaceEnabled && ZenWorkspaces.getActiveWorkspaceFromCache.uuid) {
+          tab.setAttribute("zen-workspace-id", ZenWorkspaces.getActiveWorkspaceFromCache.uuid);
```
mr-cheffy, line 395: Essentials dont care about workspaces, no?

mr-cheffy (approved):

mr-cheffy:
@kristijanribaric Sorry, I meant to request changes... 😅

src/browser/base/zen-components/ZenPinnedTabManager.mjs
kristijanribaric: They do not. 
I modified `addToEssentials` to remove the workspace id so that essentials wouldn't be tied to a workspace in any way. When you are removing an essential using `removeEssentials`, that tab doesn't have a workspace id so it has to have one set - in this case the id of the current workspace is set

PR #2675, opened 2024-11-05
Fix browser background with macOS transparency on Twilight

greeeen-dev:
This PR fixes the browser background on macOS, allowing the background to be shown with transparency on macOS.

Instead of setting the whole window to be transparent, this approach sets the background for the whole window, then sets transparency for sidebar, splitter, navbar, and website view.
image

made in collaboration w/ mauro on zen discord


greeeen-dev:
closing because mauro requested it
Screenshot 2024-11-05 at 21 22 27

Issue #2636, opened 2024-11-04
Tab name weird padding


wiraki:
### Captchas

- [X] I have read the instructions.
- [X] I have searched existing issues and avoided creating duplicates.
- [X] I am not filing an enhancement request.

### What happened?

The text and icons of tab is not properly aligned. See in the screenshot how the text is almost touching the bottom border of the active tab surrounding box:

![image](https://github.com/user-attachments/assets/22a95a92-0707-4550-9985-9b0a56efa0ac)

Using the flatpak on linux.

I tried disabling all the zen mods, closing and re-opening the browser etc., but I couldn't figure out what is causing it.

Thanks for your efforts!!

### Reproducible?

- [X] I have checked that this issue cannot be reproduced on Mozilla Firefox.

### Version

1.0.1-a.17

### What platform are you seeing the problem on?

Linux

Richard-Woessner:
Hey @wiraki Im having trouble replicating this bug, can you send a screenshot of the entire browser?

wiraki:
Hey, thanks for the quick reaction. Here's a screenshot, took the opportunity to show the zen mods in case it's helpful. Although as mentioned, turning them all off did not solve the issue.

![image](https://github.com/user-attachments/assets/d45ca1c8-4ebd-43fc-8f38-7884f510d5d6)

Now I just tried loading another profile with default settings, and the issue is gone there. So it must be some setting/customization in my profile that is causing it. Do you have any idea where I should look? 

Edit: now with the screenshot

Richard-Woessner:
Okay I was trying to see why your tabs are so different from mine and I can see now you have a addon installed

https://zen-browser.app/mods/d8b79d4a-6cba-4495-9ff6-d6d30b0e94fe

Are you sure its turned off?

wiraki:
Yes, here is the same with mods off.

![image](https://github.com/user-attachments/assets/0a7d74e8-d1f9-4f52-904f-65b6f97e5d0d)

Richard-Woessner:
Okay so with the mod turned off it looks fine is that what you were asking?

wiraki:
> Okay so with the mod turned off it looks fine is that what you were asking?

No, it doesn't. In my last screenshot, you can see that the icon and tab name is not centered vertically in the box that highlights which tab is selected. 

However, I found what is causing it:

1. Right click in empty UI space
2. Select "Customize Toolbar"
3. At the bottom, set density to "Compact"

[...]

PR #2676, opened 2024-11-05
Removed ability to adjust density in CustomizeUI mode

Richard-Woessner:
Added patch to remove the Density button from the Customize UI settings. This is to avoid user confusion

Issue #2636

## After Update
image

## Before Update
image

### DIFF
--- a/src/browser/themes/shared/customizableui/customizeMode-css.patch
+++ b/src/browser/themes/shared/customizableui/customizeMode-css.patch
@@ -1,0 +1,20 @@
+diff --git a/browser/themes/shared/customizableui/customizeMode.css b/browser/themes/shared/customizableui/customizeMode.css
+index 622eff11b121abd354b48359951a1f8d6612380c..6d753dea4ffb4d84a57a8ab496130faa764a63c0 100644
+--- a/browser/themes/shared/customizableui/customizeMode.css
++++ b/browser/themes/shared/customizableui/customizeMode.css
+@@ -392,6 +392,15 @@ toolbarpaletteitem {
+   transform: scaleX(-1);
+ }
+ 
++/* 
++  Added for zen browser
++  This is a hack to hide the density button in the customize mode
++  In order to not confuse the user with the density options
++*/
++#customization-uidensity-button {
++  display: none !important;
++}
++
+ #customization-panelWrapper,
+ #customization-panelWrapper > .panel-arrowcontent,
+ #customization-panelHolder {


src/browser/components/customizableui/test/browser_customizemode_uidensity-js.patch
Richard-Woessner, line 1: This whole file is to test the Density button. I wasnt sure if there was a better way than commenting out the page, so let me know

Tellinq:
Doesn't the Touch density work fine? From my testing Touch density isn't indicating any alignment anomalies.

For me - I use Touch on my 2n1 devices, so this would be a negative change if Density is outright removed. Are you unable to hide Compact only?

MrPiggy105:
Compact works fine for me. The only confusion is perhaps confusing compact density with compact mode in Arc - we may want to rename it rather than remove it. I have a couple of friends who use compact density on Firefox and like it and I don't see why we should remove it from Zen.

Tellinq:
If you’re going to hide the Density button, I would only do this on macOS since the Touch density is available on Windows and Linux *which work fine*.

I strongly advise against this change for all operating systems. I get why you’d do this since you’re on macOS where only Compact and Normal are present, but that doesn’t remain true on Windows and Linux.

If you want to do this in a smart manner in removing it, remove the Compact option entirely, and THEN if the density option still shows, remove the Density option IF only one option is shown.

Richard-Woessner:
Maybe I'm wrong but the compact mode in the customizeUI options is different from zen's compact mode. The issue that this PR came from was from Linux but also caused an issue on my mac. I will test again and double check

Tellinq:
And you're right that Compact density is different from Zen's compact mode (as compact mode is hiding one or both toolbars unless hovering, while density intends to change the height and width of toolbar elements) - but that's not my point.

My point is that Touch also exists *and works fine*. If you're removing Density altogether, you're then removing Touch when that is completely unnecessary.

### Compact
image

### Normal
image

### Touch
image


Only issue I see with Compact is that the text alignment on tabs is not as ideal, while Normal and Touch are fine. This shouldn't need to warrant removing Density entirely - in my view what should be changed is just text alignment in Compact density. However if it's deemed unsupported then that's understandable. However, if you're going to remove anything related to the issue you're trying to solve, just remove Compact density as an available option, **not Density**.

If I'm being unclear... if you're going to remove anything... this is what you should remove, not what you are currently doing in your PR
image

----

For the record, not even the latest version of Firefox has Compact as available option (at least on Linux). This is simply all I'm asking out of this PR if ANYTHING regarding the issue you're trying to solve is to be removed

image

Richard-Woessner:
I see what you mean now. Sorry I'm at work so I couldn't dig as deep as I should. You're right I didn't see Touch because I'm on a mac. I will wait for how @mauro-balades wants to handle this. If he would rather, I fix the bug causing the padding issues or remove this feature because I would imagine that this would cause a lot more of a headache to make sure any change works well with all the Density options.

I will convert this into a draft and wait, but thanks for pointing this out

Richard-Woessner:
Okay I stand corrected this seems to be fixed now. Im going to close this but keep up a PR to remove toolbars button

PR #2701, opened 2024-11-06
Update tabbar to fade out tabs on both edges

thatcomputerguy0101:
This fixes the scroll edge detection of the tabbar by reverting it to use the elements list directly instead of filtering the tab list, which yields incorrect bounds due to Zen's changes to the tabbar. Additionally, the mask effect is now also applied to the top edge of the scroll zone when not fully scrolled to the top, similar to what was done on the bottom edge.

There seems to be a rendering bug when an element utilizing `backdrop-filter` is masked, resulting in the masking effect stopping at that element. This means that whenever the selected tab is repainted onscreen, it and any tabs below it do not have the mask effect applied on that repaint. Any individual non-selected tabs that are then updated by hovering over them will have the mask applied since then they do not share a repaint with the selected tab. According to [Chromium bug report 40284528](https://issues.chromium.org/issues/40854528), this may be due to how the specification for those two styles are defined.


mr-cheffy:
Zen doesnt run chromium though

thatcomputerguy0101:
I'm aware, but I'm assuming they both share some common logic as defined in the CSS specification. I was unsuccessful in finding any reports for that problem on Firefox.

PR #2771, opened 2024-11-09
Add workspace navigation via mouse back/forward buttons

kristijanribaric:
Adds the ability to navigate workspaces using the mouse back/forward buttons when hovering over the sidebar.  This improves the user experience by providing a more intuitive way to switch between workspaces.

This feature is only enabled when workspaces are enabled. It intercepts the "AppCommand" events to handle back/forward navigation.

### DIFF
--- a/src/browser/base/zen-components/ZenWorkspaces.mjs
+++ b/src/browser/base/zen-components/ZenWorkspaces.mjs
@@ -162,7 +162,6 @@ var ZenWorkspaces = new (class extends ZenMultiWindowFeature {
       this._swipeState.direction = this._swipeState.cumulativeDelta > 0 ? 'right' : 'left';
     }
 
-    console.log('MozSwipeGestureUpdateEND', this._swipeState);
   }
 
   async _handleSwipeEnd(event) {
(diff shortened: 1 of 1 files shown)


olerap:
Was very excited when I saw this feature being added, unfortunately not working for me right now on macOS. Using a Logitech Master 3S, forward/backward buttons are set to default. Apart from this feature they function as expected.

kristijanribaric:
Hm that's interesting.
Can you check in `about:config` page if you have preferences `mousebutton.4th.enabled` and `mousebutton.5th.enabled` set to true?

Just to be sure, you are pressing  forward/backward buttons **while** hovering the sidebar, right?

Also, can you confirm that the page goes forward/backward in history when pressing those buttons while hovering the page content?

olerap:
Yes, these are enabled and yes, I hover over the bar while doing so. Weird thing is, when I use a "swipe" gesture using two fingers on my Mac's touchpad is does work.

kristijanribaric:
The swipe gesture is not related to this as it uses a different logic for detecting workspace switching.

You didn't confirm that forward/backward buttons when hovering the page content navigate the tab history back or forward.

olerap:
Hereby confirmed, this does work.

PR #2790, opened 2024-11-10
Fix: file manager not opening in many scenarios

marat0n:
# Bug fix. File manager not opening if Zen is installed via flatpak

On my machine (and many others), the file manager won't open even after granting read/write permissions to all files.
The fix is ​​to add `org.freedesktop.portal.Desktop` to Session Bus's Talks.
I'm new to flatpak, so please check if I've done everything correctly.

Thank you!


mr-cheffy (approved):
Thanks!

PR #2828, opened 2024-11-10
Feature: Add workspace switching by horizontally scrolling sidebar

kristijanribaric:
This PR adds the ability to switch workspaces by horizontally scrolling the sidebar.  A scroll cooldown and threshold are implemented to prevent accidental workspace changes.  The `_setupHoverDetection` method is renamed to `_setupSidebarHandlers` as it now handles both hover and scroll events.

### DIFF
--- a/src/browser/base/zen-components/ZenWorkspaces.mjs
+++ b/src/browser/base/zen-components/ZenWorkspaces.mjs
@@ -12,6 +12,7 @@ var ZenWorkspaces = new (class extends ZenMultiWindowFeature {
     direction: null
   };
   _hoveringSidebar = false;
+  _lastScrollTime = 0;
 
   async init() {
     if (!this.shouldHaveWorkspaces) {
@@ -57,7 +58,7 @@ var ZenWorkspaces = new (class extends ZenMultiWindowFeature {
 
   initializeWorkspaceNavigation() {
     this._setupAppCommandHandlers();
-    this._setupHoverDetection();
+    this._setupSidebarHandlers();
   }
 
   _setupAppCommandHandlers() {
@@ -90,7 +91,7 @@ var ZenWorkspaces = new (class extends ZenMultiWindowFeature {
     }
   }
 
-  _setupHoverDetection() {
+  _setupSidebarHandlers() {
     const toolbox = document.getElementById('navigator-toolbox');
 
     toolbox.addEventListener('mouseenter', () => {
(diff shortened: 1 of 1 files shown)


mr-cheffy (approved):
Thanks!

iskaa02:
wouldn't be better to make the switching animation a non-linear animation.
my idea would be rendering the next and previous workspaces, when switching animate the position  (also animate background) by interpolating gesture/scroll, and when the animation is finished switch the workspace, this would look much better and smoother, also would eliminate the need for entering animation for tabs.


```javascript
// example pseudo code
// State
let currentWorkspace = 0;
let transitioning = false;
let scrollPos = 0; // Scroll progress (-1 to 1)

// Scroll event (hypothetical)
onScroll(progress => { 
  scrollPos = progress;
  updateWorkspaceTransforms(currentWorkspace, progress); // Non-linear visual update
});

onScrollEnd(() => {
  if (Math.abs(scrollPos) > 0.5) {
    goToWorkspace(Math.sign(scrollPos));
  } else {
    animateTo(currentWorkspace); // Snap back
  }
  scrollPos = 0;
});


// Switching
function goToWorkspace(direction) {
  if (transitioning) return;
  transitioning = true;
  const nextWorkspace = currentWorkspace + direction; //... bounds checking
  animateTo(nextWorkspace, () => {
    currentWorkspace = nextWorkspace;
    transitioning = false;
  });
}

// Animation (hypothetical functions)
function updateWorkspaceTransforms(current, progress) {
  // ... apply non-linear transforms based on progress, to current and next ...
}

function animateTo(targetWorkspace, onComplete = () => {}) {
  // ... animate workspaces to target positions (with easing if needed) ...
  onComplete(); // Callback when animation finished.
}



```

kristijanribaric:
That would require significant changes to Firefox source code where you wouldn't just create a tab in a tab container, you would have a different tab container depending on a workspace... All in all it's not that trivial and it would require major refactoring for a small gain

sinbadthepoet:
Please add a toggle for this, my mouse (Logitech G502 Hero) tends to accidentally scroll horizontally while middle clicking which I use to close tabs. Horizontal Scrolling cannot be disabled while running the mouse in memory mode without the Logitech Software installed, so I keep accidentally switcing workspaces. Probably a very selfish ask, but I have to imagine I'm not the only one with this issue of accidentally switching workspaces while closing tabs!

PR #2853, opened 2024-11-10
Fix: Pinned tab state and favicon handling

kristijanribaric:
Fixes several issues related to pinned tab restoration and favicon display:

- Ensures pinned tabs are not loaded in private windows.
- Improves favicon caching and setting for pinned tabs.
- Fixes an issue where the tab state was not properly restored, leading to incorrect icons and titles.
- Uses base64 encoded favicon data to prevent issues with favicon loading and updates the favicon loading logic to be more robust.
- Adds a call to `gBrowser._updateTabBarForPinnedTabs` to ensure the tab bar is updated correctly after restoring pinned tabs, which fixes a UI issue where the tab bar wouldn't refresh properly until interacted with.
- Improves tab initialization by explicitly calling `newTab.initialize()` after restoring a pinned tab. This fixes an issue with tabs not rendering icon and label after being restored.

The changes improve the reliability of pinned tab
restoration and ensure that favicons are displayed correctly.

### DIFF
--- a/src/browser/base/zen-components/ZenPinnedTabManager.mjs
+++ b/src/browser/base/zen-components/ZenPinnedTabManager.mjs
@@ -39,7 +39,7 @@
   class ZenPinnedTabManager extends ZenPreloadedFeature {
 
     init() {
-      if (!this.enabled) {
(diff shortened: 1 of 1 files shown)


mr-cheffy (approved):
Thanks!

Issue #2243, opened 2024-10-20
For Flathub, use the latest org.freedesktop.Platform runtime


fxzxmicah:
### Captchas

- [X] I have read the instructions.
- [X] I have searched existing issues and avoided creating duplicates.
- [X] I am not filing an enhancement request.

### What happened?

The latest runtime is not used.

### Reproducible?

- [ ] I have checked that this issue cannot be reproduced on Mozilla Firefox.

### Version

1.0.1-a.11

### What platform are you seeing the problem on?

Linux

fxzxmicah:
Upstream: https://bugzilla.mozilla.org/show_bug.cgi?id=1925839

PR #2889, opened 2024-11-11
chore: Update flatpak manifest runtime versions to 24.08

JamiKettunen:
Seems to work as per my local testing (test build also available on the now closed https://github.com/flathub/io.github.zen_browser.zen/pull/54)


mr-cheffy:
Thanks! This will take effect next release!

fxzxmicah:
#2243 Wow, before upstream.

fxzxmicah:
@mauro-balades https://github.com/zen-browser/desktop/discussions/1381
Although it is a digression as a feature request, I still hope to pay attention to this. There is no need for complicated adaptation, just need to enable it on linux.

I used it on my computer for a few days and found no problem.

PR #2899, opened 2024-11-11
Add a "Discussions" link to issue template

Lenni-builder:
This commit adds a second button when creating a new issue that tells the user to go to GitHub discussions for feature requests with a link to the Discussions page. I think this will lead to less confusion and clutter the issues tab less.


mr-cheffy:
Thanks!

PR #2904, opened 2024-11-11
Feature: Add reset to pinned URL button to pinned tabs

kristijanribaric:
This PR introduces a "Reset" button to pinned tabs, allowing users to quickly revert a tab to its stored base URL.  The button is visible on hover for non-essential pinned tabs and hidden for essentials.  Clicking the button triggers a
reset action managed by `gZenPinnedTabManager`.

The visual style of the button is defined using the "reload" icon and follows the styling of other tab buttons.

Depends on:
https://github.com/zen-browser/l10n-packs/pull/99

https://github.com/user-attachments/assets/e73ac78d-5193-4f3c-9f8b-a595bbd76e09

### DIFF
--- a/src/browser/base/content/zen-styles/zen-tabs/vertical-tabs.css
+++ b/src/browser/base/content/zen-styles/zen-tabs/vertical-tabs.css
@@ -683,6 +683,18 @@
   #tabbrowser-tabs {
     & .tabbrowser-tab {
       &[pinned] .tab-close-button {
+        display: none !important;
+      }
+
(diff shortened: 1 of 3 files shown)


mr-cheffy:
Thanks!!! <3

Shockshwat:
Would it be possible to instead unload the tab when you are already at the pinned URL? So essentially 1st click takes you to the pinned URL and the second click would unload the tab and the third click unpin and close the tab?

PR #2908, opened 2024-11-11
Adds zen profile item on-click feedback

Tanay-Kar:
Provides a change in background colour for the profile selector items on click (active) , similar to the main menu style


mr-cheffy:
THanks a lot!

PR #2941, opened 2024-11-12
Add Unload All Tabs

GunGunGun:
🐷

[Add Unload All Tabs](https://github.com/zen-browser/l10n-packs/pull/100)


mr-cheffy:
We want to do this in a different way that doesn't involve context menus. Sorry 🐷

GunGunGun:
> We want to do this in a different way that doesn't involve context menus. Sorry 🐷

🐷

PR #2947, opened 2024-11-12
Feature: container specific Essentials

kristijanribaric:
Added a pref to make essentials container specific, i.e. if a workspace has a container assigned to it, on that workspace only essentials opened in that container will be visible.


mr-cheffy (approved):
Thankssss!!!

PR #2978, opened 2024-11-13
Improve sidebar icon placement consistency

AnthemV:
With this change, the leftmost icon at the bottom of the sidebar (the one in the outermost corner) will now remain in the outermost corner when the sidebar is collapsed. Previously, it would have jumped to the top of the stack.

**Before Change**

https://github.com/user-attachments/assets/f7812079-4361-48c6-996b-36f71c079731

**With Change**

https://github.com/user-attachments/assets/da59e6ac-f481-4a71-8536-8eb98177a008


mr-cheffy:
I dont know about the reversed part.. How does it feel when having "expand on hover" enabled?

PR #3035, opened 2024-11-15
Feature: Add shortcut to enable Rearrange View

maxthyron:
Read the comments about adding a new shortcut, hope this is how it should be done.
Added some checks for opened split view menu and that split view is enabled.

I don't know how to properly add the name for the shortcut in the menu and translations for it, since they seem to be in a separate repo.

Additionally while testing I needed to create a new profile for new shortcut to show up, is this intended behavior? Or do I need to remove some cached files for changes to apply?

There are also a typos:
1) in the function's name `afterRearangeAction` -> `afterRearrangeAction`
2) `this._thumnailCanvas` -> `this._thumbnailCanvas`

Not sure if it's suitable to add here too.

Lastly, I had to disable Prettier for the project because applying the default prettier config from the repo messed up some files I opened, at least `ZenViewSplitter.mjs` has some inconsistencies(e.g. `_swapField` function has some strange whitespaces)


mr-cheffy:
Hi! Could this be updated into version 5 instead of 4? Since it's already taken...😅

maxthyron:
Done. Was familiarizing myself with the codebase and forgot to move this to a separate version clause.

mr-cheffy:
I was wondering if it should also close the rearange view if it's enabled?

maxthyron:
As far as I could understand at the moment the only way to toggle rearrange view on is with the button inside split view menu:
https://github.com/zen-browser/desktop/blob/12aa51c551d3616ec4523b183e958c49d321cf1d/src/browser/base/content/zen-popupset.inc.xhtml#L43-L46
And disabling it by pressing ESC:
https://github.com/zen-browser/desktop/blob/12aa51c551d3616ec4523b183e958c49d321cf1d/src/browser/base/zen-components/ZenViewSplitter.mjs#L244-L251
Since my intention was to add a key shortcut for enabling the rearrange view without the need to click 3 times, I'm just using functions already present and just skipping a popup.

If the rearrange view was already toggled on through regular means, shortcut won't do anything because there's already a check for that:
https://github.com/zen-browser/desktop/blob/12aa51c551d3616ec4523b183e958c49d321cf1d/src/browser/base/zen-components/ZenViewSplitter.mjs#L220-L222
With all that in mind, I can't quite get the intention of your suggestion and what's needed of me.

mr-cheffy:
Hmm.. This woudnt be needed anymore, sorry

PR #3093, opened 2024-11-18
Feature: Windows ARM64

omove:
Add ARM64 Windows builds. Requires surfer changes in zen-browser/surfer#9.


omove:
I just finished working on the Linux aarch64 build as well so I'm going to close this and do a new PR for both together since most of the changes are shared between them.

PR #3135, opened 2024-11-21
Linux/Windows ARM64 Builds

omove:
Add Linux and Windows ARM64 builds, requires surfer changes in https://github.com/zen-browser/surfer/pull/9.


configs/windows/mozconfig
```diff
@@ -12,34 +10,46 @@ if test "$ZEN_CROSS_COMPILING"; then
...
+  ac_add_options --enable-optimize="-O2 -mtune=cortex-a55 -march=armv8.2-a+simd"
+
+  export CFLAGS="-O2 -march=armv8.2-a+simd -mtune=cortex-a55"
+  export CPPFLAGS="-O2 -march=armv8.2-a+simd -mtune=cortex-a55"
+  export CXXFLAGS="-O2 -march=armv8.2-a+simd -mtune=cortex-a55"
+  export LDFLAGS="-Wl,-O2 -march=armv8.2-a+simd -mtune=cortex-a55"
+  export RUSTFLAGS="-C target-feature=+v8.2a -C codegen-units=1 -Ctarget-cpu=cortex-a55"
 else
+  ac_add_options --target=x86_64-pc-windows-msvc
+  ac_add_options --enable-eme=widevine,wmfcdm
```
mr-cheffy: Why this extra flag?

configs/windows/mozconfig
```diff
@@ -12,34 +10,46 @@ if test "$ZEN_CROSS_COMPILING"; then
 
   export CROSS_BUILD=1
   CROSS_COMPILE=1
-  TOOLS=$(echo ~)/win-cross
-
-  export WIN_UCRT_REDIST_DIR="$(echo ~)/win-cross/vs2022/Windows Kits/10/Redist/10.0.22621.0/ucrt/DLLs/x64"
-  export WIN32_REDIST_DIR="$(echo ~)/win-cross/vs2022/VC/Redist/MSVC/14.38.33135/x64/Microsoft.VC143.CRT"
 
-  export MIDL="$TOOLS/wine/bin/widl"
```
mr-cheffy, line 20: Why is this being removed?

configs/macos/mozconfig
```diff
@@ -1,17 +1,15 @@
...
 
 export MOZ_MACBUNDLE_ID=${appId}
 export MOZ_MACBUNDLE_NAME="Zen Browser.app"
 
-if test "$SURFER_COMPAT" = "true"; then
+if test "$SURFER_COMPAT" = "x86_64"; then
     export MOZ_PGO=1
     ac_add_options MOZ_PGO=1
     ac_add_options --target=x86_64-apple-darwin
+    ac_add_options --enable-eme=widevine
```
mr-cheffy: Please remove this

configs/common/mozconfig
```diff
@@ -89,7 +89,6 @@ ac_add_options --enable-raw
 ac_add_options --enable-webrtc
 ac_add_options --enable-jxl
 ac_add_options --enable-av1
-ac_add_options --enable-eme=widevine
```
mr-cheffy, line 92: Why is this being removed?

mr-cheffy (changes requested):

configs/windows/mozconfig
omove: Wine's MIDL can't build ARM64, it fails with a keyerror on the cpu type. Need to use the Windows SDK MIDL compiler (which is what's used by default when not specified).

configs/windows/mozconfig
omove: Oh, that's leftover from when I was trying to check if PlayReady would work since Firefox 132 added it. That can be removed.

configs/windows/mozconfig
mr-cheffy: Why not just add it when it's not for ARM then?

configs/common/mozconfig
omove: DRM support varies by architecture and OS. Linux aarch64 doesn't support widevine so it can't be enabled in the common mozconfig.

configs/windows/mozconfig
mr-cheffy: Oh, and does it work?

configs/common/mozconfig
mr-cheffy: Linux does support it, that's why it's in the common.. If not, people wont be able to play netflix on linux

configs/windows/mozconfig
omove: There's no reason to use different compilers versions by architecture since the windows sdk one works for all of them but sure it can be moved to the platform section.

configs/common/mozconfig
omove: No, you can't compile with it enabled on Linux aarch64, it was moved to the individual os/arch mozconfigs because it had to be.

omove: Specifically, the Linux aarch64 build will fail with 
``mozbuild.configure.options.InvalidOptionError: --enable-eme is not available in this configuration``

Tor browser also had to change theirs because of this
https://gitlab.torproject.org/tpo/applications/tor-browser/-/issues/42501

configs/windows/mozconfig
omove: Unfortunately it didn't, I tried on https://bitmovin.com/demos/drm/ but it still only detected Widevine.

configs/macos/mozconfig
omove: Is Widevine not supported on macOS? It was in the common file so it should work, it doesn't say enable-eme is an invalid option like aarch64 Linux does.
Edit: Yeah, it should definitely be there, OSX has widevine support listed under it.
https://searchfox.org/mozilla-central/source/toolkit/moz.configure#914

mr-cheffy: Remove all of them and add it how it was on the common mozconfig

omove: Okay, I moved it outside the platform block, I assume that's what you meant.

Botspot:
Watching this. Interested in adding Zen Browser to [the Pi-Apps store.](https://github.com/Botspot/pi-apps)
Is there a way to be notified when the first ARM64 linux build is released?

mr-cheffy:
Are you in the discord? Ill ping everyone when the update is out... Itll be a pretty huge update, so everyone will know lol

PR #3138, opened 2024-11-21
some adjustment for "-moz-gtk-csd-reversed-placement"

sapphired18:
adds padding to titlebar-buttonbox-container when "-moz-gtk-csd-reversed-placement" is true

before
> image

after
> image

### DIFF
--- a/src/browser/base/content/zen-styles/zen-browser-ui.css
+++ b/src/browser/base/content/zen-styles/zen-browser-ui.css
@@ -105,3 +105,16 @@
     margin-inline-end: 8px;
   }
 }
+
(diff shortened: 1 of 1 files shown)


mr-cheffy:
Sorry, but I believe these values should be hard coded but rather something more like: `margin-top: auto`, also, I dont know why but the code seems to be duplicate

sapphired18:
> Sorry, but I believe these values should be hard coded but rather something more like: `margin-top: auto`, also, I dont know why but the code seems to be duplicate

no problem ;) the value is hard-coded because I didn't check with other methods in my mind, I can't wait for new update.

PR #3182, opened 2024-11-24
Faster macOS Github Actions

omove:
Try switching the x86_64 macOS build to cross compile on macOS aarch64, makes the x86_64 building go from ~3.5-4 hours to ~1.5 hours from my actions test run https://github.com/omove/desktop/actions/runs/11993727306/job/33435040303


mr-cheffy:
Will it be able to do PGO runs though?

omove:
That actions run finished building, it just stopped at the code signing part since I don't have a apple developer cert.

mr-cheffy:
Nice! maybe we could also get arm64 PGO

PR #3194, opened 2024-11-25
Hidden container label when URL bar is narrow.

kristijanribaric:
https://github.com/user-attachments/assets/bdc5dd7d-4b5e-4ef1-8e11-94faf5eb8b47


mr-cheffy:
Thanks!!!

PR #3196, opened 2024-11-25
Improved sidebar icon placement redux

AnthemV:
Updated version of #2978. With this PR, icons at the bottom of the sidebar should be consistently placed in their logical positions. This takes into account whether expand on hover is enabled, as well as whether the sidebar is set to the left or right.

https://github.com/user-attachments/assets/df28ee89-c58d-4436-b53d-fc90a76abec8


mr-cheffy:
Could this PR be updated to the new refactor we've did? Thanks!

PR #3202, opened 2024-11-25
Enable PGO for macOS aarch64

omove:
Building the aarch64 version with PGO seems to work from my github actions test run.


thatcomputerguy0101:
This seems to break my local build of Zen, with the following error:
`../../../../config/nsinstall: cannot access desktop/engine/obj-aarch64-apple-darwin/instrumented/_tests/xpcshell/toolkit/mozapps/update/tests/data//TestAUSHelper: No such file or directory`

Issue #3162, opened 2024-11-22
Hand Gesture Swipe on Sidebar Switches Multiple Spaces Instead of One


DhananjayPorwal:
### Captchas

- [X] I have read the instructions.
- [X] I have searched existing issues and avoided creating duplicates.
- [X] I am not filing an enhancement request.

### What happened?

When using the hand gesture to swipe the sidebar in the browser, it unexpectedly switches 2-3 spaces at once. The expected behavior is that a single swipe should change only one space at a time, not multiple. This issue makes it difficult to navigate the sidebar accurately, as it skips over spaces.

[Screencast from 2024-11-22 17-40-28.webm](https://github.com/user-attachments/assets/62d35def-2d86-4307-bcd9-650e44a65de3)

### Reproducible?

- [X] I have checked that this issue cannot be reproduced on Mozilla Firefox.

### Version

1.0.1-a.19

### What platform are you seeing the problem on?

Linux

neurokitti:
this should hopefully be fixed by this https://github.com/zen-browser/desktop/pull/3206

DhananjayPorwal:
Hi @neurokitti, getting this issue after the beta update. Previously it was fixed.

Browser Version: 1.0.2-b.1 (64-bit)
OS: Ubuntu 24.04

[Screencast from 2024-12-14 23-01-10.webm](https://github.com/user-attachments/assets/e53c87d1-2711-4599-ac6a-7729492b7e3f)

### Suggested Fix

```js
(function () {
    const workspaceManager = {
        currentWorkspaceIndex: 0,
        workspaces: [],

        init() {
            // Initialize workspace list from toolbarbutton elements
            this.workspaces = Array.from(
                document.querySelectorAll('toolbarbutton.subviewbutton')
            );
            this.currentWorkspaceIndex = this.workspaces.findIndex(
                (workspace) => workspace.getAttribute('active') === 'true'
            );

            // Listen for wheel events (or replace with gesture events if supported)
            window.addEventListener('wheel', this.handleSwipe.bind(this), { passive: false });
        },

        handleSwipe(event) {
            if (event.deltaX === 0) return; // Skip if not a horizontal swipe

            event.preventDefault(); // Prevent default behavior

            const direction = event.deltaX > 0 ? 1 : -1; // Determine swipe direction
            this.switchWorkspace(direction);
        },

        switchWorkspace(direction) {
            // Calculate new workspace index, ensuring it stays in bounds
[...]

PR #3206, opened 2024-11-26
Restricted scrolling beyond the first and last workspaces and fixed bug with Sidebar Switches Multiple Spaces Instead of One

neurokitti:
I fixed this bug https://github.com/zen-browser/desktop/issues/3162

I made is to that if you keep swiping at the end of a workspace it no longer just wraps around and I lowered the sensitivity for counting swipes cuz it felt better.


neurokitti:
im adding an option for the restricting scrolling beyond the first and last workspaces

PR #3217, opened 2024-11-26
Restricted scrolling beyond the first and last workspaces and fixed bug with Sidebar Switches Multiple Spaces Instead of One

neurokitti:
I fixed this bug https://github.com/zen-browser/desktop/issues/3162

I made is to that if you keep swiping at the end of a workspace it no longer just wraps around.

IMPORTANT:
ive added some code for scrolling through workspaces with the scroll wheel and a modifer key but my zen install is having issues and i cant run a full build to test if the changes work because this patch. i checked and it has nothing to do with this PR but this feature deffinatly needs to be tested lol
Screenshot 2024-11-26 214943


src/browser/base/zen-components/ZenWorkspaces.mjs
kristijanribaric, line 1: Can you remove all the unnecessarily added empty rows?

kristijanribaric (changes requested):

src/browser/base/zen-components/ZenWorkspaces.mjs
StudioMovieGirl: this should work hopefully

mr-cheffy:
Ready for merge? @neurokitti

neurokitti:
Should be. I haven't been able to test the prefs because my zen install is still having the patch issue

neurokitti:
I'm going to do more testing

PR #3223, opened 2024-11-27
Add '.nvmrc' and '.python-version' files to ensure system dependency (node and python) is documented

vraravam:


mr-cheffy:
Thanks!

PR #3226, opened 2024-11-27
Add auto-release for twilight homebrew cask

vraravam:
Since we now have a homebrew cask for the twilight builds, we can auto-update that as well on each release of the twilight branch


mr-cheffy:
Thanks! Btw, do I need to add some sort of key or something?

vraravam:
> Thanks! Btw, do I need to add some sort of key or something?

no, this uses the same key that's used for the alpha from GH secrets. nothing extra to be done

PR #3320, opened 2024-11-30
makes the workspace button resize better

neurokitti:
i made it so that the workspaces width will fill 100% of the space so it looks like this:
image

and not this:

https://github.com/user-attachments/assets/ebe91f75-548f-4dfe-b7a1-44c3bf40ef9a

it also resizes like this:

https://github.com/user-attachments/assets/1a59b4fa-f146-4249-9d60-4af6143ae8a1


neurokitti:
nah i want to make this work better

PR #3322, opened 2024-12-01
README(fix): installation command

owljacob:
I removed the `sudo` because the installation doesn't work with it.


mr-cheffy:
Thanks!

PR #3518, opened 2024-12-10
Updated old mod marketplace dead link with working new one

op-cs:
Changed button referencing "https://zen-browser.app/themes/" in settings page which doesnt work anymore to the new link "https://zen-browser.app/mods/"


mr-cheffy:
Thanks!

PR #3547, opened 2024-12-11
Make sidebar compact mode animation smoother

Axenide:
Hi, I noticed that the sidebar in compact mode appears in a kinda linear animation. This is a really small tweak that makes it a lot smoother.
I just changed it from `ease-in-out` to `ease`.
The initial animation after the change is something that happens just one time while debugging and changing this value, it doesn't actually happen.

https://github.com/user-attachments/assets/a395c064-01b3-48b3-8355-bc68aea861f8


mr-cheffy:
Thanks!

PR #3709, opened 2024-12-13
Update dependencies, fix CI, use our own free disk space script

larziwau:
Updates dependencies and fixes something I forgot about the CI. Also uses a better script for cleaning up space (which is faster and empties more space, ALSO ITS LOCAL, means we no longer need to download the action and waste time.)


larziwau:
Also I did setup dependabot for dependencies (runs weekly, can be modified to run monthly)

larziwau:
Btw thanks to https://github.com/rust-lang/rust/pull/134151

mr-cheffy:
Sorry but a lot of versions here are changed and we dont really need our own disk freeing script. Sorry

larziwau:
Okay thanks for interest though

PR #3717, opened 2024-12-13
Removed hardcoding of python version so that maintenance can be done in a single place (.python-version)

vraravam:
According to the [basic usage](https://github.com/actions/setup-python?tab=readme-ov-file#basic-usage) of the `setup-python` action, rather than hardcoding the same python version in multiple places (which can lead to maintenance issues), its better to use the `.python-version` file.


vraravam:
Can this be merged please?

PR #3789, opened 2024-12-15
Update BetterFox

larziwau:
update betterfox to 133


mr-cheffy:
Sorry, but we arent using betterfox anymore

PR #4023, opened 2024-12-27
Re-enable the built-in password manager

different55:
Having the password manager enabled is a good default, because the alternative for the average person isn't adopting a third party manager, it's either having no manager or bouncing off Zen entirely.

I don't hold the [Principle of Least Astonishment](https://en.wikipedia.org/wiki/Principle_of_least_astonishment) as sacred law or anything, but it definitely seems applicable here, especially for a browser with the stated goals of being featureful and taking the backseat.

### DIFF
--- a/src/browser/app/profile/zen-browser.js
+++ b/src/browser/app/profile/zen-browser.js
@@ -385,7 +385,6 @@ pref("browser.formfill.enable", false);
 pref("security.insecure_connection_text.enabled", true);
 pref("security.insecure_connection_text.pbmode.enabled", true);
 pref("network.IDN_show_punycode", true);
-pref("signon.rememberSignons", false);
 
 // Telemetry
 pref("datareporting.policy.dataSubmissionEnabled", false, locked);


stergion:
Yeah, I agree with @different55.

If you want to discourage users from using the default password manager, 
I thing it would be better to do that during the initial setup. 
There could be a list of suggested alternatives for the user to pick from, or keep using the default.

mr-cheffy:
I'm not too sure, because it isn't only unsafe but also some sites crash while having it enabled (e.g. people reported a lot of crashes on Google signin)

different55:
first off, sorry for being long winded, thanks for replying!

might be better to address those issues more directly, if they're not in upstream then whatever's causing that could cause other mayhem.

wrt safety, that's a legit concern! although the mitigation isn't quite as easy as just disabling this setting, the data still exists on disk. we should also check if it's still synced to disk through ff sync if password syncing is enabled but password autofill isn't, because that would be a massive footgun *if* that's the case.

anyway, the reason we may not want to disable it even if there are legitimate issues with it, is that there's other competing concerns. without a password manager, people (unfortunately) don't become computer literate. they become worse, actually, adopting the most insane security hygiene. they use weak, easy-to-remember passwords, write them down on post-it notes, heck they often don't even use "passwords" in the plural, they'll reuse one password everywhere they can.

out of all the browsers out there, including a good swarm of explicitly privacy-minded projects, only Tor (an extremely specialized tool) disables its built in password manager because it can count on its users coming in with the proper expectations for an extremely specialized tool. that's not the image and expectation that's been set up for zen.

mr-cheffy:
You are right, im sorry

mr-cheffy (approved):

different55:
No worries, thank you so much for being open to discussion!

antonioyon:
I think simply _re-enabling_ had some unintended consequences. I use a third-party password manager which has an extension that is enabled. After installing 1.6b and creating a new password field, the default PM appeared and offered to save the new password. I went to disable this in settings and found the "Ask to Save Passwords" option selected but disabled. The only way I found to enable it so I could unselect it was to disable the third-party extension.

MohamedElashri:
The problem is that currently with the last release you cannot disable the built in passwords because the option is greyed out in settings. We should allow for use case of external password managers. 

edit: As workaround for people using third party extensions for password management and are annoyed by this. They can disable it using `about:config`

change both of these values to `false`

```
signon.rememberSignons
signon.rememberSignons.visibilityToggle
```

different55:
The commit removes a line that was added. It doesn't change "false" to "true".

different55:
The commit removes a line that was added. It doesn't change "false" to "true".

PR #4024, opened 2024-12-27
Enable bottom rounded corners for Linux

alosarjos:
On Linux, the bottom window corners are not rounded. This setting enables them as shown on the screenshots:

Before:

Before

After: 

After


mr-cheffy:
I will not enable this as it doesnt work for all enviroments

alosarjos:
> I will not enable this as it doesnt work for all enviroments

Is there more info on what environments this doesn't work to look at it?

fxzxmicah:
https://github.com/zen-browser/desktop/commit/4d46f94975137b5d525af9d06113fa9a5f1c6cf0

PR #4034, opened 2024-12-28
resistFingerprinting

Unchanged4352:
It works this way right?


mr-cheffy:
We won't enable this flag

Unchanged4352:
> We won't enable this flag

Why?

mr-cheffy:
It can break sites and not only that, but you LOCKED the preference

Unchanged4352:
> It can break sites and not only that, but you LOCKED the preference

ok thanks

PR #4088, opened 2024-12-31
Fix split view splitters being offset by sidebar

### LINKED ISSUE #2761: Split view can't be resized vertically when bookmarks sidebar is open
### Captchas

### What happened?

As the title says. Vertical resizing of split tabs doesn't seem to work when firefox's native sidebar is open
[Screencast From 2024-11-09 13-31-42.webm](https://github.com/user-attachments/assets/1961daf8-9dfb-4961-aafa-cff85984911a)

### Reproducible?

### Version

1.0.1-a.17, 1.0.1-t.18

Linux

### Relevant log output

_No response_

vintydong:
Fixes #2761

Splitters were calculated using the entire split view overlay which results in an incorrect inset % when a sidebar is present.

https://github.com/user-attachments/assets/79505396-d1fc-4a3b-9775-d7104e0e758e

A padding-left or padding-right (depending on sidebar position) equal to the sidebar width is added to the overlay for the inset % to be correct. This padding is recalculated under the following actions
- Sidebar resize
- Sidebar opening/closing
- Sidebar reverse position (left/right)

https://github.com/user-attachments/assets/fab94662-7fa7-4b46-9f65-da1c6ac0cd4a

### DIFF
--- a/src/browser/components/sidebar/browser-sidebar-js.patch
+++ b/src/browser/components/sidebar/browser-sidebar-js.patch
@@ -1,8 +1,35 @@
 diff --git a/browser/components/sidebar/browser-sidebar.js b/browser/components/sidebar/browser-sidebar.js
-index 4a124003976684e014435854aef69ce29da541d2..61ce44751c36eea3e5ae2ddcc62e42c01459629b 100644
+index e95aa93e912d97aca2f60bf094562ea43a8d179b..166497ed26b075995b2151c5f841dc5dd6d20699 100644
 --- a/browser/components/sidebar/browser-sidebar.js
 +++ b/browser/components/sidebar/browser-sidebar.js
-@@ -578,7 +578,7 @@ var SidebarController = {
+@@ -246,6 +246,8 @@ var SidebarController = {
+   _localesObserverAdded: false,
+   _mainResizeObserverAdded: false,
+   _mainResizeObserver: null,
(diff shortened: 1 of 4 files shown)


src/browser/components/sidebar/browser-sidebar-js.patch
```diff
@@ -1,8 +1,35 @@
...
+ 
+   /**
+@@ -334,6 +336,17 @@ var SidebarController = {
+       sidebarBox.style.maxWidth = `calc(75vw - ${entry.contentBoxSize[0].inlineSize}px)`;
+     });
+ 
++    if (this._sidebarResizeObserver) {
++      this._sidebarResizeObserver.disconnect();
++    }
++    this._sidebarResizeObserver = new ResizeObserver(async () => {
```
mr-cheffy: I dont think we should add an observer, I believe there are better ways...

vintydong: I found an easier solution by simply moving the splitview-overlay and wrapper into the tabbrowser so the inset % will be correct based on the tabpanels width.

I think it makes more sense for splitview overlay to be nested within the tabpanel so its always the correct size vs. doing extra calculations. I also removed the xhtml file since it seemed to be the occurrence

menganatha:
This issue https://github.com/zen-browser/desktop/issues/4680
Is probably part of the issue discussed here

PR #4095, opened 2024-12-31
Add entitlements for SmartCardServices and WebAuthn support in codesi…

mr-cheffy:
…gn configuration


Inbestigator:
Yes! This is the only big issue that I've had with Zen so far on MacOS

PR #4158, opened 2025-01-03
fix: don't flash compact sidebar when opening glance

### LINKED ISSUE #3351: Glance in compact view opens tabs bar
### Captchas

### What happened?

Hello,

When in compact view and having both tab and top bars hidden, when using glance it'll show the tab bar as if I hover over there with my mouse. It happens on both CTRL + left click & Shift + left click. I tried using glance outside of compact view and the bug doesn't persist there.

Not really familiar with coding lingo or Github, just wanted to report this bug. I apologize if there's anything missing. 

Thanks

### Reproducible?

### Version

1.0.1-a.22

Windows

### Relevant log output

_No response_

flexagoon:
Flashing the sidebar when opening glance is unnecessary, as the user explicitly opens the glance tab so there are no new information presented to the user in the sidebar.

On the other hand, the sidebar partially covers the glance tab, meaning that the user has to wait for a while before seeing the tab content, which breaks the flow.

Closes #3351

### DIFF
--- a/src/browser/components/tabbrowser/content/tabs-js.patch
+++ b/src/browser/components/tabbrowser/content/tabs-js.patch
@@ -76,7 +76,8 @@ index f7c39fe804182e2bdf53045ba3b6a5ba17079fc3..93be23da305a5e2e51bff1c5b5f28743
 -      }
 +      this.arrowScrollbox.ensureElementIsVisible(selectedTab, aInstant);
  
-+      if (!aInstant && Services.prefs.getBoolPref('zen.view.compact')
++      if (!aInstant && !gZenGlanceManager.animatingOpen
++          && Services.prefs.getBoolPref('zen.view.compact')
 +          && Services.prefs.getBoolPref('zen.view.compact.toolbar-flash-popup')) {
 +        gZenCompactModeManager.flashSidebar();
 +      }


src/browser/components/tabbrowser/content/tabs-js.patch
```diff
@@ -76,7 +76,8 @@ index f7c39fe804182e2bdf53045ba3b6a5ba17079fc3..93be23da305a5e2e51bff1c5b5f28743
 -      }
 +      this.arrowScrollbox.ensureElementIsVisible(selectedTab, aInstant);
  
-+      if (!aInstant && Services.prefs.getBoolPref('zen.view.compact')
++      if (!aInstant && !gZenGlanceManager.animatingOpen
++          && Services.prefs.getBoolPref('zen.view.compact')
```
mr-cheffy, line 80: This patch is getting a bit long. Should we move it to a function called `gZenCompactModeManager.flashSidebarIfAllowed(aInstant)`?

mr-cheffy (changes requested):

mr-cheffy:
This is an invalid patch btw... Just take that in mind for the next PR because it breaks everything...

flexagoon:
@mr-cheff should I open a new PR with what you suggested (a `flashSidebarIfAllowed` function)? Also, why is the patch invalid? Is it because of the line numbers? I wanted to adjust them, but they didn't match the ones I found in upstream Firefox code anyway, so I got confused

mr-cheffy:
No it's ok, i've already moved it into https://github.com/zen-browser/desktop/commit/76cc2fdb25040fc2ad3a8d5d4188bef15c6dade4

mr-cheffy:
> Also, why is the patch invalid?

You've done an illegal patch, you can't just edit the patch and let it be. I've added `surfer export file` so patching could be more easy. Did you test this btw?

flexagoon:
> Did you test this btw?

@mr-cheff no, I didn't find any documentation on how to run tests on the codebase. I saw that PRs are automatically tested in CI, but it seems like that's not triggered on patch file changes. Is there any place where I can find documentation on the proper development process?

PR #4187, opened 2025-01-05
Make workspaces icons randomly select only non-flag emojis

zabojeb:
I create new workspaces very often, and every time I see a randomly generated icon. The problem is that among the icons ≈25% are flags of different countries, this means that about every fourth time I see a flag of a completely random country, I think this is strange. At the same time, users may want to put some country flag as an icon, so we don't want to change the array with icons, we can just trim the range of random selection.

ofc, this is a very minor thing, but so annoying; fix takes less than 1 line of code (XS size), so please accept it


zabojeb:
Like this
image

mr-cheffy:
Thanks!

PR #4198, opened 2025-01-06
Added collapsable header for about:support in issue template

thibaultmol:
currently if you don't do this, you get issues where the first entry is massive because of this block. doing this will make it so that the output is in a collapsed block in markdown and takes up less space


thibaultmol:
(this isn't working fully correctly yet, but I can't figure out how to make it work. You get the idea though)

mr-cheffy (approved):

mr-cheffy:
Thanks!

PR #4601, opened 2025-01-24
Add WebAuthn support in entitlements and fix provisioning profile cop…

mr-cheffy:


configs/macos/mozconfig
mr-cheffy, line 1: Remember to remove this before merging

PR #4653, opened 2025-01-27
Added feature: flag for showing full uri in tab hover preview

y-haidar:
This is a shameful PR for a feature that I personally want, and probably only me.


mr-cheffy:
Sorry, but I dont want to be making extra-patches/configurations for such small features. Please use any sort of auto-config managers to add this feature for your self

PR #4669, opened 2025-01-28
fix: make 'Always floating' setting work on new tabs

VialFlorian:
## Description
The "Always floating" setting wasn't being applied when opening new tabs.  
Fixed by making the setting properly apply to newly opened tabs.

Note: This is my first PR as I'm discovering the codebase - any feedback is welcome if something needs to be improved!

### Before
https://github.com/user-attachments/assets/906621d6-d4af-4357-a453-ed659034f82d

### After
https://github.com/user-attachments/assets/bd456858-0737-4bc0-88e2-9020b22b31b9

### DIFF
--- a/src/browser/base/content/browser-commands-js.patch
+++ b/src/browser/base/content/browser-commands-js.patch
@@ -1,8 +1,32 @@
 diff --git a/browser/base/content/browser-commands.js b/browser/base/content/browser-commands.js
-index 352de44dda36e3f6672eb353f42978ede0cd2681..d6956a318c34bfb12b0ba957edab1166e1a4edaf 100644
+index 352de44dda36e3f6672eb353f42978ede0cd2681..916876ed2924cada725dc68e42cfac0c3abedae0 100644
 --- a/browser/base/content/browser-commands.js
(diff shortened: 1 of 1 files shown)


neoapps-dev:
i hope this get merged :)

Anthonyy232:
There's also some bugs in general regarding floating tabs -- notably you have to double click outside of the bounds or press ESC twice to clear it, where the regular URL bar only requires one click or one ESC. Maybe this can be another PR or you want to look into it? 😄

mvargasmoran:
Hey, can this also eliminate the "new tab" part, can we open the floating thingy right on top of what we are already seeing?

afonsofrancof:
> Hey, can this also eliminate the "new tab" part, can we open the floating thingy right on top of what we are already seeing?

Zen already has a keybinding for that.
By default, it's CTRL + L  (or CMD + L), but you can change it in the settings

Ruszero01:
Really more reasonable, click floating preview, middle key to open from new tab，Floating preview is really convenient in some scenes, especially for pages like issues.

PR #4792, opened 2025-01-31
Added a preference to toggle the sidebar opening on hover in compact mode

afonsofrancof:
This addresses #3760.
I keep accidentally opening the sidebar even after a few weeks of using zen, so It would be nice to have the option of not opening the sidebar when hovering near it.
This would make it so it would only open with the keyboard shortcut to toggle compact mode.

### DIFF
--- a/src/browser/base/zen-components/ZenCompactMode.mjs
+++ b/src/browser/base/zen-components/ZenCompactMode.mjs
@@ -276,7 +276,10 @@ var gZenCompactModeManager = {
   },
 
   get hoverableElements() {
-    const showSidebarOnHover = Services.prefs.getBoolPref('zen.view.compact.show-sidebar-on-hover', true);
+    if (typeof this._showSidebarOnHover === 'undefined') {
+      this._showSidebarOnHover = Services.prefs.getBoolPref('zen.view.compact.show-sidebar-on-hover', true);
+    }
+    const showSidebarOnHover = this._showSidebarOnHover;
     return [
       ...(showSidebarOnHover
         ? [
--- a/src/browser/base/zen-components/ZenCompactMode.mjs
+++ b/src/browser/base/zen-components/ZenCompactMode.mjs
@@ -276,19 +276,20 @@ var gZenCompactModeManager = {
   },
 
   get hoverableElements() {
-    const panels = [];
-    if (Services.prefs.getBoolPref('zen.view.compact.show-sidebar-on-hover', true)) {
-      panels.push({
-        element: this.sidebar,
-        screenEdge: this.sidebarIsOnRight ? 'right' : 'left',
-        keepHoverDuration: 100,
-      });
-    }
-    panels.push({
-      element: document.getElementById('zen-appcontent-navbar-container'),
-      screenEdge: 'top',
-    });
-    return panels;
+    const showSidebarOnHover = Services.prefs.getBoolPref('zen.view.compact.show-sidebar-on-hover', true);
+    return showSidebarOnHover
+      ? [
+          {
+            element: this.sidebar,
+            screenEdge: this.sidebarIsOnRight ? 'right' : 'left',
+            keepHoverDuration: 100,
+          },
+          {
+            element: document.getElementById('zen-appcontent-navbar-container'),
+            screenEdge: 'top',
+          },
+        ]
+      : [];
   },
 
   flashSidebar(duration = lazyCompactMode.COMPACT_MODE_FLASH_DURATION) {
--- a/src/browser/base/zen-components/ZenCompactMode.mjs
+++ b/src/browser/base/zen-components/ZenCompactMode.mjs
@@ -282,19 +282,24 @@ var gZenCompactModeManager = {
         true
       );
     }
-    return !this._showSidebarAndToolbarOnHover
-      ? []
-      : [
-          {
-            element: this.sidebar,
-            screenEdge: this.sidebarIsOnRight ? 'right' : 'left',
-            keepHoverDuration: 100,
-          },
-          {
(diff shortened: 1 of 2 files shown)


m3kka (approved):

afonsofrancof:
@mauro-balades have you had the time to look at this yet?
Thanks :)

joshdonnell:
@afonsofrancof @mauro-balades 

Any chance this can get looked at, would make a great edition for people who like to run in Compact made with shortcuts for toggling between tabs etc.

luke-concannon:
Agree - I'm finding it to be the one major pain point - accidentally triggering the sidebar on hover when interacting with elements on the side of a page. I want to just be able to manage its visibility manually.

mr-cheffy:
Yeah, but could everything be unified into a single statement? Or at least define the transitions in `not (-pref...)` for better readability? Thanks

joshdonnell:
@afonsofrancof 

I have pushed a PR to your fork, please review and if you are happy can you update this PR

afonsofrancof:
> I have pushed a PR to your fork, please review and if you are happy can you update this PR

@joshdonnell 

Your changes also add 
```js
{
     element: document.getElementById('zen-appcontent-navbar-container'),
     screenEdge: 'top',
}
```
 to the array based on the condition. Shouldn't it be added regardless?

Also, maybe we should invert the condition like @mauro-balades  asked.

In that case, it would be something like this 

```js
get hoverableElements() {
  return [
    ...(!Services.prefs.getBoolPref("zen.view.compact.show-sidebar-on-hover", true)
      ? []
      : [
          {
            element: this.sidebar,
            screenEdge: this.sidebarIsOnRight ? "right" : "left",
            keepHoverDuration: 100,
          },
        ]),
    {
      element: document.getElementById("zen-appcontent-navbar-container"),
      screenEdge: "top",
    },
  ];
}
```

Here we only use the condition on the sidebar itself.
What do you think?

joshdonnell:
> debar itself.
> W

I have updated my PR I miss read the code so sorry for my oversight here. I have updated, however I don't think we should reverse the check as this goes against how other ternary's are working in the project

joshdonnell:
> > I have pushed a PR to your fork, please review and if you are happy can you update this PR
> 
> @joshdonnell
> 
> Your changes also add
> 
> ```js
> {
>      element: document.getElementById('zen-appcontent-navbar-container'),
>      screenEdge: 'top',
> }
> ```
> 
> to the array based on the condition. Shouldn't it be added regardless?
> 
> Also, maybe we should invert the condition like @mauro-balades asked.
> 
> In that case, it would be something like this
> 
> ```js
> get hoverableElements() {
>   return [
>     ...(!Services.prefs.getBoolPref("zen.view.compact.show-sidebar-on-hover", true)
>       ? []
>       : [
>           {
>             element: this.sidebar,
>             screenEdge: this.sidebarIsOnRight ? "right" : "left",
>             keepHoverDuration: 100,
>           },
>         ]),
>     {
>       element: document.getElementById("zen-appcontent-navbar-container"),
>       screenEdge: "top",
>     },
>   ];
> }
> ```
> 
> Here we only use the condition on the sidebar itself. What do you think?

I have updated my PR I miss read the code so sorry for my oversight here. I have updated, however I don't think we should reverse the check as this goes against how other ternary's are working in the project

afonsofrancof:
@mauro-balades
Is it better now? :)

src/browser/base/zen-components/ZenCompactMode.mjs
```diff
@@ -276,12 +276,17 @@ var gZenCompactModeManager = {
   },
 
   get hoverableElements() {
+    const showSidebarOnHover = Services.prefs.getBoolPref('zen.view.compact.show-sidebar-on-hover', true);
```
mr-cheffy: Why are you doing this?

mr-cheffy (changes requested):

src/browser/base/zen-components/ZenCompactMode.mjs
mr-cheffy: Nevermind, I was confusing PRs...

mr-cheffy (approved):

joshdonnell:
@mauro-balades 

Can you re-approve when you get 5 now the formatting has been fixed for the linting rules

joshdonnell:
@mauro-balades 

Any ideas when we can expect this in twilight?

src/browser/base/zen-components/ZenCompactMode.mjs
mr-cheffy: Can you make this a lazy preference? For performance

mr-cheffy (changes requested):

src/browser/base/zen-components/ZenCompactMode.mjs
afonsofrancof: I have changed that.
Can you check if what I did is what you wanted?

mr-cheffy: Great! And last thing, could u also make it apply to the top toolbar?

afonsofrancof: If it works for both the sidebar and topbar, should I name the preference differently? Maybe "show-sidebar-and-toolbar-on-hover" or "show-bars-on-hover" instead of "show-sidebar-on-hover", or something similar?

mr-cheffy: That would be awesome! Thanks a lot

afonsofrancof: Done :)

mr-cheffy (approved):

Issue #576, opened 2024-08-24
Add additional "copy URL" shortcuts


ahgraber:
### What feature would you like to see?

One of the features I use most in Arc is the 'copy URL as markdown' which allows me to copy the page title and url into markdown formatting

On mac, that default shortcut is `cmd+shift+option+v`

When I use it on Zen Browser's homepage, here is what it pastes: `[Zen Browser](https://www.zen-browser.app/)`

enesbala5:
This is very important to users switching from Arc.

I would recommend Ctrl + Shift + C - that's the Arc default on Windows.

Hopeful we can remap it in settings.

enkvadrat:
Ctrl + Shift + C is used to open the inspector.

enesbala5:
Well so is Ctrl + Shift + I & F12 - maybe it's a bit redundant at this point.

anatasiajp:
I think I can add this feature, but later.

enesbala5:
@anatasiajp Would be much appreciated! Thank you

enkvadrat:
Sorry Ctrl + Shift + C toggles "Pick an element for page" and the inspector

mauro-balades:
This issue should be discussed further. You can view and continue the discussion here: https://github.com/zen-browser/desktop/discussions/794

 ~ This is an automatic operation to better enhancements requests organization.

PR #4801, opened 2025-01-31
Add feature: copy current URL as markdown

EmilyxFox:
A copy URL shortcut a la Arc was recently added, but my preference "Copy URL as Markdown" wasn't! As a lot of people are currently moving over from Arc I thought it would be good to have this option as well.

I also updated both shortcuts to copy all urls in a split view.

Video demo:

https://github.com/user-attachments/assets/b6d3b44d-cfae-4c5d-aea5-a4657e6bcfb2

The default shortcut `CTRL` + `SHIFT` + `ALT` + `C` was chosen as that is the same shortcut used in Arc.
image

I choose not to add a specific notification string as Arc uses the same notification for both `Copy current tab URL` and `Copy current tab URL as Markdown`
Arc_Ce23EqJBHe

This feature was discussed in #576 and https://github.com/zen-browser/desktop/discussions/794

Depends on: https://github.com/zen-browser/l10n-packs/pull/122

### DIFF
--- a/src/browser/base/zen-components/ZenCommonUtils.mjs
+++ b/src/browser/base/zen-components/ZenCommonUtils.mjs
@@ -60,23 +60,10 @@ class ZenPreloadedFeature {
 
 var gZenCommonActions = {
   copyCurrentURLToClipboard() {
-    let urlStringToCopy = "";
-    if (Services.prefs.getBoolPref('zen.keyboard.shortcuts.copy-current-url.copy-all-in-split-view', false)) {
-      const currentTabs = gZenViewSplitter.getTabsInCurrentView();
-      if (currentTabs) {
-        const stringArray = currentTabs.map(t => `${t.linkedBrowser.currentURI.spec}`);
-        urlStringToCopy = stringArray.join("\n");
-      }
-    } else {
-      const currentUrl = gBrowser.currentURI.spec;
-      if (currentUrl) {
-
-        urlStringToCopy = `${currentUrl}`
-      }
-    }
-    if (urlStringToCopy) {
+    const currentUrl = gBrowser.currentURI.spec;
+    if (currentUrl) {
       let str = Cc['@mozilla.org/supports-string;1'].createInstance(Ci.nsISupportsString);
-      str.data = urlStringToCopy;
+      str.data = currentUrl;
       let transferable = Cc['@mozilla.org/widget/transferable;1'].createInstance(Ci.nsITransferable);
       transferable.init(getLoadContext());
       transferable.addDataFlavor('text/plain');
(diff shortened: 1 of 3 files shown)


EmilyxFox:
Redrafted because I want to add this feature: https://github.com/zen-browser/desktop/discussions/794#discussioncomment-11008448

src/browser/base/zen-components/ZenCommonUtils.mjs
```diff
@@ -60,10 +60,27 @@ class ZenPreloadedFeature {
 
 var gZenCommonActions = {
   copyCurrentURLToClipboard() {
-    const currentUrl = gBrowser.currentURI.spec;
-    if (currentUrl) {
+    const currentTabs = gZenViewSplitter.getTabsInCurrentView()
+    if (currentTabs) {
+      const stringArray = currentTabs.map(t => `${t.linkedBrowser.currentURI.spec}`)
```
mr-cheffy: Id rather like copy url to just copy current view and copy markdown accept all split views

mr-cheffy (changes requested):

src/browser/base/zen-components/ZenCommonUtils.mjs
EmilyxFox: Would it be worth it to have a user pref setting for it? I don't know if I would personally use it (I didn't even know it worked like that in Arc before looking at issues related to this). Maybe its niche enough that it could be off by default but enabled with something hidden in `about:config`. I also think it might be a bit confusing to have the two copy shortcuts work differently in that way.

Another suggestion might also be to change the "copied url" confirmation message to be plural if it copied multiple tabs, but that might just be a ui qol.

EmilyxFox: I threw together a demonstration if that helps. It doesn't include plural confirmation hints though. I also didn't put a lot of thought into the setting name and scope. I assume you have some standard for that. I also don't know if you want a separate setting for each shortcut.
https://github.com/user-attachments/assets/b6d3b44d-cfae-4c5d-aea5-a4657e6bcfb2

Edit: It seems like gh doesn't want to embed the video in this image but you can watch it by clicking the link 😭

EmilyxFox:
After seeing the conversation about user choice in the Discord I gained the confidence to push this proposed compromise 🙏

I think as a lot of people are looking for an Arc replacement supporting both the Arc way of doing it and the more expected way of doing it is a good choice here and an improvement on the original PR. Here are the comments that inspired me to implement it: https://github.com/zen-browser/desktop/discussions/794#discussioncomment-11008448

EmilyxFox:
After our conversation in the Discord I've removed the copy all in split view functionality... It needs more time in the oven, and it might be better to wait till tab groups and other functionality like that before looking into it.

mr-cheffy (approved):

mwsundberg:
I noticed this could output mangled Markdown if the title has special characters in it. The Markdown spec is so fuzzy and it's such a minor thing I don't know if it's worth it, but a quick edit on L77 in [src/browser/base/zen-components/ZenCommonUtils.mjs](https://github.com/zen-browser/desktop/blob/dev/src/browser/base/zen-components/ZenCommonUtils.mjs) to the following ought to fix it:

`const tabTitle = gBrowser.selectedTab.label;`

to

    let tabTitle = gBrowser.selectedTab.label;
    "\\`*_[]".split("").forEach((c)=>{tabTitle = tabTitle.replaceAll(c, "\\" + c)});


Citing [this](https://daringfireball.net/projects/markdown/syntax#backslash) for `\` escaping working, I know it gets mangled anyways in some flavors yet better than nothing I hope. If Arc does it differently I defer to their wisdom, same if someone wants to refactor, change the escaped characters, or add a comment explaining that one-liner.

Sorry if this was the wrong way to bring this up or if the code is too ugly. In the future should I make a new issue, try some github native code review thing, make a pull of my own, or what? Loving the browser's rambunctious dev pace and fun features so far!

Edit I'm not sure if the URL comes properly encoded already yet if not that might also be worth hitting with %etc encoding.

EmilyxFox:
You're absolutely right! I didn't really think of doing this as I modelled it off Arc's implementation and they do not care 😂

I think it would be a good improvement though.
As for encoding I think it just comes out in the proper format already.

logicsec:
@mwsundberg I ran a couple of tests on this with @EmilyxFox and we are not getting issues with link parsing that I can see. Can you provide an example where the link creation had failed?

mwsundberg:
@logicsec Coded up a little demo to show how it can mess things up, here ya go:

https://mws.is/scratch/zen-copy-url-demo.html

Also @EmilyxFox, you were totally right the URL is escaped properly, no worries on that end.

PR #4944, opened 2025-02-04
Setup Dependabot and fix some actions

blu3berryys:


blu3berryys:
I'm also going to use a custom free disk space script

blu3berryys:
Or no actually

mr-cheffy:
Have you tested this?

LICENSE
Bashamega, line 1: This is an unnecessary change

blu3berryys: Going to revert

PR #4979, opened 2025-02-05
[Feature] Tab renaming

jso8910:
This is a pull request to implement tab renaming. It isn't yet complete. It works by adding a button to the right click context menu which then replaces the tab name with an input field to rename the tab. It requires one string to be added in the l10n repo (I haven't made that PR yet).

This pull request is not yet complete, as the tab name does not persist when the page reloads, loads another URL, or the browser restarts. I'm willing to take suggestions for how to do that (or contributions from other maintainers). I see files like [ZenPinnedTabsStorage.mjs](https://github.com/zen-browser/desktop/blob/dev/src/browser/base/zen-components/ZenPinnedTabsStorage.mjs) and wonder if that's a way to do it? Not quite sure. Or modifying the part of the browser engine that changes the title of a tab and stopping it from changing it if some attribute indicating the tab has been renamed is set as well as changing the part of the tab title which is saved to persist a session.

Also, I created a new file in the zen-components folder. I figured there wasn't any other place for this code to fit, but if I'm wrong, I'll refactor.

Pinging @mauro-balades to get feedback/help on the rest of this.


devxan:
What about this? 
> to dev, why not add a special folder to bookmark and use it for store tab data that user can also rename tab or add folder or sync it, make ui to combine tab and the bookmark that store in special folder to make it look like using normal tab.
> for example.
/ Bookmark Menu / ZEN-TAB / WORKSPACE-1 / <all tab here, store as bookmark >
/ Bookmark Menu / ZEN-TAB / WORKSPACE-2 /< user dir >/ <all tab here, store as bookmark >

jso8910:
That feels like it's a complete overhauling of the tabs system.... which is a bit beyond my skill level, familiarity with the code, or what I *should* be doing.

devxan:
I personally thought that would only apply to pinned tabs, but I can understand that that could still be crazy hard to make.

Anthonyy232:
If this and the bookmark menu is in direct conflict, would it not be better to wait and see how the tab system will work going forward? Folders are coming for sure, so there's bound to be some change

mr-cheffy:
Instead of overflowing the context menu, how about double clicking on pinned tabs only?

12th-devs:
Thank you so much for this PR! I was going to make it but I cannot get the browser to build.

mr-cheffy:
> This pull request is not yet complete, as the tab name does not persist when the page reloads, loads another URL, or the browser restarts. I'm willing to take suggestions for how to do that (or contributions from other maintainers). I see files like [ZenPinnedTabsStorage.mjs](https://github.com/zen-browser/desktop/blob/dev/src/browser/base/zen-components/ZenPinnedTabsStorage.mjs?rgh-link-date=2025-02-05T16%3A01%3A09Z) and wonder if that's a way to do it? Not quite sure. Or modifying the part of the browser engine that changes the title of a tab and stopping it from changing it if some attribute indicating the tab has been renamed is set as well as changing the part of the tab title which is saved to persist a session.

By just changing `tabLabel.textContent = ...` already saves on session store iirc. I can help you with this commit if you'd like. But please move this inside `gZenVerticalTabsManager`

jso8910:
@mauro-balades by `tabLabel.textContent = ...` do you literally mean the HTML element that contains the text of the tab label? Or is it a javascript object. Because I tried changing the `<label>` element and it didn't persist. And neither does changing `gBrowser.tabs[i].label`.

And either way, this wouldn't stop the page title changing from changing the renamed title. I feel like this renaming should override any reloads, page navigation, etc.

jso8910:
I found a good solution for persisting tab renames in the same session. Still not sure about persisting them across sessions. Here's an example with some dummy values.

```js
let newName = "asdf";
let tab = gBrowser.tabs[5];
let observer = new MutationObserver((mutations) => {
  mutations.forEach((mutation) => {
    if (tab.label !== newName) {
      tab.label = newName;
    }
  })
})
observer.observe(tab, {attributeFilter: ["label"]})
```

I think you would want to add an attribute to each tab that is `tab.renamedLabel` which is undefined or null (not sure about JS-isms... which is ideal?) by default and then each pinned tab would have a listener like this (which would not do anything if it is null).

jso8910:
@mauro-balades implemented your suggested changes. Now uses `zen-has-static-label` as an attribute. When you implement freezing the tab label based on that attribute, just make sure that `browser.contentTitle` (the `contentTitle` attribute in the `browser` object for the tab) is still able to change, because that's what's used for resetting the tab label.

I realized we never discussed what we should do to not only persist the fact that a tab is renamed, but also persist the renamed title (which currently doesn't restore on restart, as you'll see when you test the PR). Should I add something to the session store to store that?

src/browser/base/content/ZenUIManager.mjs
```diff
@@ -593,4 +596,95 @@ var gZenVerticalTabsManager = {
     }
     target.appendChild(child);
   },
+
+  _insertDoubleClickListenerPinnedTabs() {
```
mr-cheffy: This woudn't work if we add pinned tabs AFTER setting the listeners. I recommend adding `gBrowser.tabContainer.addEventListener()` and check if the originalTarget is the tab label

mr-cheffy (changes requested):

src/browser/base/content/ZenUIManager.mjs
jso8910: No need. It adds this listener to every single tab. `renameTabStart` checks to make sure the tab is pinned.

mr-cheffy: Even if the event listener is added to the tabContainer, event.target will still point to the tab element

jso8910: Done :+1:

mr-cheffy:
> I found a good solution for persisting tab renames in the same session. Still not sure about persisting them across sessions. Here's an example with some dummy values.
> 
> ```js
> let newName = "asdf";
> let tab = gBrowser.tabs[5];
> let observer = new MutationObserver((mutations) => {
>   mutations.forEach((mutation) => {
>     if (tab.label !== newName) {
>       tab.label = newName;
>     }
>   })
> })
> observer.observe(tab, {attributeFilter: ["label"]})
> ```
> 
> I think you would want to add an attribute to each tab that is `tab.renamedLabel` which is undefined or null (not sure about JS-isms... which is ideal?) by default and then each pinned tab would have a listener like this (which would not do anything if it is null).

@jso8910, add a patch [here](https://searchfox.org/mozilla-central/source/browser/components/tabbrowser/content/tabbrowser.js#1750) to simply return if we have the attribute. And make sure to do a `git pull`

src/browser/base/content/ZenUIManager.mjs
```diff
@@ -583,4 +585,95 @@ var gZenVerticalTabsManager = {
     }
     target.appendChild(child);
   },
+
+  //_insertDoubleClickListenerPinnedTabs() {
```
mr-cheffy: Please remove this if it's not going to be used

src/browser/base/content/ZenUIManager.mjs
```diff
@@ -583,4 +585,95 @@ var gZenVerticalTabsManager = {
...
+      document.removeEventListener('click', this.renameTabHalt.bind(this));
+
+      this._tabEdited = null;
+    } else if (event.key === 'Escape') {
+      let label = this._tabEdited.querySelector('.tab-label-container-editing');
+      this._tabEdited.querySelector('.tab-editor-container').remove();
+
+      label.style.display = '';
+      label.className = label.className.replace(' tab-label-container-editing', '');
+      document.removeEventListener('click', this.renameTabHalt.bind(this));
```
mr-cheffy: `removeEventListener` doesn't actually remove anything if you are using `bind`... Id do on the init function `this._renameTabHalt = this.renameTabHalt.bind(this)` and then `removeEventListener(..., this._renameTabHalt)` and same for `addEventListener`

src/browser/base/content/ZenUIManager.mjs
```diff
@@ -583,4 +585,95 @@ var gZenVerticalTabsManager = {
...
+    const container = window.MozXULElement.parseXULToFragment(`
+      <vbox class="tab-label-container tab-editor-container" flex="1" align="start" pack="center"></vbox>
+    `);
+    label.after(container);
+    const containerHtml = this._tabEdited.querySelector('.tab-editor-container');
+    const input = document.createElement('input');
+    input.id = 'tab-label-input';
+    input.value = this._tabEdited.label;
+    input.addEventListener('keydown', this.renameTabKeydown.bind(this));
+    input.style['white-space'] = 'nowrap';
```
mr-cheffy: Please avoid to add inline styles as it's bad for customization and debugging. Please add it to any CSS file, e.g. `vertical-tabs.css`

src/browser/base/content/ZenUIManager.mjs
```diff
@@ -583,4 +585,95 @@ var gZenVerticalTabsManager = {
...
+    }
+    if (!this._tabEdited) {
+      return;
+    }
+    this._tabEdited.querySelector('.tab-editor-container').remove();
+    const label = this._tabEdited.querySelector('.tab-label-container-editing');
+    label.style.display = '';
+    label.className = label.className.replace(' tab-label-container-editing', '');
+
+    document.removeEventListener('click', this.renameTabHalt.bind(this));
```
mr-cheffy: Same thing here with `removeEventListener`

src/browser/base/content/ZenUIManager.mjs
```diff
@@ -583,4 +585,95 @@ var gZenVerticalTabsManager = {
...
+        this._tabEdited.label = gBrowser.getBrowserForTab(this._tabEdited).contentTitle || this._tabEdited.label;
+        // If the page had a title, reset the zen-has-static-label attribute
+        if (gBrowser.getBrowserForTab(this._tabEdited).contentTitle) {
+          this._tabEdited.removeAttribute('zen-has-static-label');
+        }
+      }
+
+      this._tabEdited.querySelector('.tab-editor-container').remove();
+      label.style.display = '';
+      label.className = label.className.replace(' tab-label-container-editing', '');
```
mr-cheffy: Why are you using these weird JS hacks? Why not just use `className.remove(...)`? It's bad for readability

src/browser/base/content/ZenUIManager.mjs
```diff
@@ -583,4 +585,95 @@ var gZenVerticalTabsManager = {
...
+  renameTabStart(event) {
+    if (this._tabEdited) return;
+    this._tabEdited = event.target.closest('.tabbrowser-tab');
+    if (!this._tabEdited || !this._tabEdited.pinned) {
+      this._tabEdited = null;
+      return;
+    }
+    const label = this._tabEdited.querySelector('.tab-label-container');
+    label.style.display = 'none';
+    label.className += ' tab-label-container-editing';
```
mr-cheffy: Same here with the weird JS class manipulations, please use `className.add()`. And if you are adding the `display: none;`, why not just add it to the css file?

src/browser/base/content/ZenUIManager.mjs
```diff
@@ -583,4 +585,95 @@ var gZenVerticalTabsManager = {
...
+    // Ignore click event if it's clicking the input
+    if (event.target.closest('#tab-label-input')) {
+      return;
+    }
+    if (!this._tabEdited) {
+      return;
+    }
+    this._tabEdited.querySelector('.tab-editor-container').remove();
+    const label = this._tabEdited.querySelector('.tab-label-container-editing');
+    label.style.display = '';
```
mr-cheffy: please try to avoid inline CSS as much as possible

src/browser/base/content/ZenUIManager.mjs
```diff
@@ -583,4 +585,95 @@ var gZenVerticalTabsManager = {
...
+
+    document.addEventListener('click', this.renameTabHalt.bind(this));
+  },
+
+  renameTabHalt(event) {
+    // Ignore click event if it's clicking the input
+    if (event.target.closest('#tab-label-input')) {
+      return;
+    }
+    if (!this._tabEdited) {
```
mr-cheffy: These 2 if statements could be merged together

mr-cheffy (changes requested):

src/browser/base/content/ZenUIManager.mjs
```diff
@@ -583,4 +585,95 @@ var gZenVerticalTabsManager = {
...
+      label.className = label.className.replace(' tab-label-container-editing', '');
+      document.removeEventListener('click', this.renameTabHalt.bind(this));
+      this._tabEdited = null;
+    }
+  },
+
+  renameTabStart(event) {
+    if (this._tabEdited) return;
+    this._tabEdited = event.target.closest('.tabbrowser-tab');
+    if (!this._tabEdited || !this._tabEdited.pinned) {
```
mr-cheffy: make sure essentials can't be renamed as well, or at least, add support for them

src/browser/base/content/ZenUIManager.mjs
```diff
@@ -583,4 +585,95 @@ var gZenVerticalTabsManager = {
...
+      let input = this._tabEdited.querySelector('#tab-label-input');
+      let newName = input.value.trim();
+
+      // Check if name is blank, reset if so
+      if (newName) {
+        this._tabEdited.label = newName;
+        this._tabEdited.setAttribute('zen-has-static-label', 'true');
+      } else {
+        // If the page is loaded, get the title of the page. Otherwise, keep name as is
+        this._tabEdited.label = gBrowser.getBrowserForTab(this._tabEdited).contentTitle || this._tabEdited.label;
```
mr-cheffy: Please use `gBrowser.setTabTitle(tab)` here

src/browser/base/content/ZenUIManager.mjs
```diff
@@ -583,4 +585,95 @@ var gZenVerticalTabsManager = {
...
+        // If the page had a title, reset the zen-has-static-label attribute
+        if (gBrowser.getBrowserForTab(this._tabEdited).contentTitle) {
```
mr-cheffy: Every page has a title, id just remove this if-statement

src/browser/base/content/ZenUIManager.mjs
```diff
@@ -583,4 +585,95 @@ var gZenVerticalTabsManager = {
...
+    input.addEventListener('keydown', this.renameTabKeydown.bind(this));
+    input.style['white-space'] = 'nowrap';
+    input.style['overflow-x'] = 'scroll';
+    input.style['margin'] = '0';
+
+    containerHtml.appendChild(input);
+    input.focus();
+    input.select();
+
+    document.addEventListener('click', this.renameTabHalt.bind(this));
```
mr-cheffy: Id add a `blur` even to the input instead of click, so in any sort of occasion where the input is blurred, we always make sure to exit the state

mr-cheffy (changes requested):

mr-cheffy:
Have fun!

src/browser/base/content/ZenUIManager.mjs
jso8910: Pages that don't have `document.title` generally use the URL as the page label in the tab bar. Does the browser capture this under `contentTitle`?

kakato10:
Will the custom tab name be matched in url bar suggestions?

src/browser/base/content/ZenUIManager.mjs
mr-cheffy: No, but if we are cancelling the name (leaving it empty), we should always remove the attribute to make sure there are no hanging references to it. Or rather, no unexpected behaviors after "clearing". Just remove the attribute and call `gBrowser.setTabTitle(this._tabEdited)`

mr-cheffy:
Last thing remaning @jso8910 is that they shoudnt be able to be renamed if:
1. the option `zen.tabs.rename-tabs` is `false`
2. we have `browser.tabs.closeTabByDblclick` on `true`
3. we are in single toolbar mode

jso8910:
That's not quite the last thing remaining. The renamed tab name still isn't saved after restart. Currently, if a tab is renamed and then the initial focused tab after restart, it'll be named "New tab" and if it's renamed but not the initial focused tab after restart, its name is blank

mr-cheffy:
That's due to the pinned tab manager, I already told u I can handle that part

ifdskfjsdkfjsd (approved):

mr-cheffy:
Please do let me know when it's ready for review. Also, there seems to be some git conflicts

jso8910:
Ready for the one final change you're doing then review.

Also merged and resolved conflicts. A question for the future, I found it quite difficult to merge the patch file, so the way I did it was by applying the `dev` branch patch then manually adding back my changes to `tabbrowser.js` and exporting the patch which included all changes. Is there a better workflow to resolve conflicts in the patch file? It seems non-ideal to have to edit the patch file itself rather than the original file in the `engine` directory.

ifdskfjsdkfjsd:
@mauro-balades

src/browser/base/content/ZenUIManager.mjs
```diff
@@ -608,4 +610,75 @@ var gZenVerticalTabsManager = {
...
+        this._tabEdited.removeAttribute('zen-has-static-label');
+        gBrowser.setTabTitle(this._tabEdited);
+      }
+
+      this._tabEdited.querySelector('.tab-editor-container').remove();
+      label.classList.remove('tab-label-container-editing');
+
+      this._tabEdited = null;
+    } else if (event.key === 'Escape') {
+      let label = this._tabEdited.querySelector('.tab-label-container-editing');
```
mr-cheffy: Id move this to `renameTabHalt` to avoid duplication

src/browser/base/content/ZenUIManager.mjs
```diff
@@ -608,4 +610,75 @@ var gZenVerticalTabsManager = {
...
+
+      label.classList.remove('tab-label-container-editing');
+      this._tabEdited = null;
+    }
+  },
+
+  renameTabStart(event) {
+    if (
+      this._tabEdited ||
+      !Services.prefs.getBoolPref('zen.tabs.rename-tabs') ||
```
mr-cheffy, line 647: Add the `zen.tabs.rename-tabs` check on top of `gBrowser.tabContainer.addEventListener('dblclick', this.renameTabStart.bind(this));` better, to be more efficient

mr-cheffy: And replace it with checking for `gZenVerticalTabsManager._prefsSidebarExpanded`

src/browser/base/content/ZenUIManager.mjs
```diff
@@ -608,4 +610,75 @@ var gZenVerticalTabsManager = {
...
+
+  renameTabKeydown(event) {
+    if (event.key === 'Enter') {
+      let label = this._tabEdited.querySelector('.tab-label-container-editing');
+      let input = this._tabEdited.querySelector('#tab-label-input');
+      let newName = input.value.trim();
+
+      // Check if name is blank, reset if so
+      if (newName) {
+        this._tabEdited.label = newName;
```
mr-cheffy: Replace with `gBrowser._setTabLabel(tab, title)`

src/browser/base/content/ZenUIManager.mjs
```diff
@@ -242,11 +242,13 @@ var gZenVerticalTabsManager = {
 
     window.addEventListener('DOMContentLoaded', updateEvent, { once: true });
 
-    const tabs = document.getElementById('tabbrowser-tabs');
-
     if (!this.isWindowsStyledButtons) {
       document.documentElement.setAttribute('zen-window-buttons-reversed', true);
     }
+
+    this._tabEdited = null;
```
mr-cheffy: No need to initialize this to null

mr-cheffy (changes requested):

src/browser/base/content/ZenUIManager.mjs
jso8910: Unless ZenUIManager is reinstantiated any time the preferences change, the sidebar expands or contracts, etc, I don't think I should wrap the listener in an if statement. Because otherwise you'll need to restart the browser after a preferences change for tab renaming to work.

Though, using `_prefsSidebarExpanded` is a good idea, this feature shouldn't do anything in compact mode.

georg-stone:
Shouldn't there be a rename tab button in the context menu? It's really important for accessibility purposes, as some people can't use a mouse

jso8910:
@georg-stone I think Mauro had a good point about cluttering the context menu... I'm thinking a keyboard shortcut to rename the current tab could be the solution?

Anthonyy232:
> Shouldn't there be a rename tab button in the context menu? It's really important for accessibility purposes, as some people can't use a mouse

Definitely needs to be

georg-stone:
@jso8910 Yeah a keyboard shortcut should exist. Honestly @mauro-balades should remove the clutter from the context menu because there's a lot. After that the Rename tab should be added to the context menu.

Anthonyy232:
> @jso8910 Yeah a keyboard shortcut should exist. Honestly @mauro-balades should remove the clutter from the context menu because there's a lot. After that the Rename tab should be added to the context menu.

A keyboard shortcut would be good too, but shouldn't be in replacement of a context window button. Very few people actually use keyboard shortcuts other than the typical copy and paste

mr-cheffy:
> Shouldn't there be a rename tab button in the context menu? It's really important for accessibility purposes, as some people can't use a mouse

You don't have a cursor available?

georg-stone:
@mauro-balades I do have a cursor, but lots of people with disabilities don't. So design with accessibility in mind

mr-cheffy:
In what scenario would users not be able to click on buttons? If you can't click on buttons I doubt context menu would do much anyways

mr-cheffy:
@jso8910 ready for merge?

jso8910:
@mauro-balades yeah I'm done with my part. Other than persisting the name after restart, it's completely done.

georg-stone:
> @mauro-balades yeah I'm done with my part. Other than persisting the name after restart, it's completely done.

I don't think you need to persist it on restart, that is probably worse. Up to @mauro-balades though.

jso8910:
@georg-stone You definitely do. It's very trivial to restore the original name of the page if you don't want it to persist (just type a blank name and press enter) and I most use-cases I can think of would be permanent tab renames. It's very non-destructive to have it persist and it would make the feature much more useful.

georg-stone:
At the very least it needs to be an about:config option. Otherwise it's okay

kakato10:
Are there any plans to make the name that was set by renaming the tab to be suggested in the URL Bar?

Garfounkel:
Great feature however I would love for this to also work on non-pinned tabs like in Arc. Skimming through the comments it seems to be by design but I don't quite understand why.

niderhoff:
I am so glad this feature is finally here! But I have to say the UX was not intuitive for me.

things I would change:

- there is no right click -> rename option. renaming is only triggered by double click
- when in "rename-mode" (i.e. insert mode), pressing tab will not trigger renaming, only enter.

cgen01:
> I am so glad this feature is finally here! But I have to say the UX was not intuitive for me.
> 
> things I would change...

Thank you for adding this!

Also, on MacOS, none of the common text navigation shortcuts work (i.e. option + arrow to move around works, CMD + arrow to move to start/end).

gillpeter:
but it only works on pinned tabs and not on unpinned tabs !

PR #5026, opened 2025-02-07
Respect reduced motion and animate-sidebar in compact mode

sjwall:
Currently the compact mode is not respecting the system reduced motion setting.

This PR adds overrides to `transition` to disable animations when `prefers-reduced-motion` is set and for the sidebar it also overrides when `zen.view.compact.animate-sidebar` is set to false

I have noticed some other places that need reduced motion handling and can update those in follow up PRs if happy with how this change works


mr-cheffy:
Framer already respects motion preferences though

JSGRD22:
#5048

sjwall:
> Framer already respects motion preferences though

That's my understanding as well. It looks like these animations are CSS rather than Framer. I tested by removing all the `transition` from `zen-compact-mode.css` and the animations didn't happen, regardless of motion preference

CSS being used rather than Framer is also shown by the `zen.view.compact.animate-sidebar` config option. In the Framer animation code it is used to disable the animation but changing this value doesn't stop the animation

I don't know the zen code well enough yet to understand the usage of Framer vs CSS

mr-cheffy (approved):

Issue #3692, opened 2024-12-13
Search bar collapsed when ibus or fcitx5 input panel appears


first-storm:
### Captchas

- [X] I have read the instructions.
- [X] I have searched existing issues and avoided creating duplicates.
- [X] I am not filing an enhancement request.

#### **Description**
When using the IBus input method, the floating search bar behaves abnormally. Specifically:
- When typing in **English**, the search bar works as expected.
- However, when activating the IBus input method (e.g., for non-English input), the floating search bar **disappears** as soon as the input panel appears and only **reappears** after the input panel is closed.

#### **Steps to Reproduce**
1. Open the floating search bar.
2. Type in English to verify normal behavior.
3. Switch to a non-English input method using IBus (e.g., Pinyin, Japanese, etc.).
4. Observe the search bar disappearing when the input panel appears.
5. Close the input panel and observe the search bar reappearing.

#### **Expected Behavior**
The search bar should remain visible and active when using the IBus input method, regardless of the input panel's state.

#### **Actual Behavior**
The search bar collapses and disappears when the IBus input panel appears. It only reappears after the input panel is closed.

#### **Environment**
- **Build:** Flatpak build on Flathub
- **Input Method Framework:** IBus
- **Operating System:** openSUSE Slowroll December Version

#### **Attachments**
- **Image:**  
  ![Issue Screenshot](https://github.com/user-attachments/assets/8535950a-9767-45e3-8877-a06aff713013)
- **Video Demonstration:**  
  [Video Link](https://github.com/user-attachments/assets/0a90d146-d684-43ad-87ba-c1d3bb415f21)

#### **Additional Information**
~~- It is unclear whether this issue is specific to the Flatpak version or affects other builds as well.~~

- It can also be reproduced on AUR builds.

### Reproducible?

- [X] I have checked that this issue cannot be reproduced on Mozilla Firefox.

### Version

1.0.2-b.1

### What platform are you seeing the problem on?

Linux

ma-chengyuan:
Having the same problem with Arch Linux AUR build, both under IBus and fcitx

ma-chengyuan:
Turns out that this is a feature in upstream Firefox that explicitly closes the search bar to prevent the IME panel from overlapping with the search results. See https://firefox-source-docs.mozilla.org/browser/urlbar/preferences.html

[...]

PR #5257, opened 2025-02-11
Keep URL bar open during IME composition to prevent URL bar disappearing

JSGRD22:
This default configuration addresses #3692 and related issues by preventing the floating URL bar from disappearing when composing with IME.


JSGRD22:
Just noticed this was already fixed in a commit 17 hours ago

Issue #5114, opened 2025-02-08
Reset pinned tab button is not working


seriaati:
### Captchas

- [x] I have read the instructions.
- [x] I have searched existing issues and avoided creating duplicates.
- [x] I am not filing an enhancement request.

### What happened?

This button:

![Image](https://github.com/user-attachments/assets/0e2cd5db-3ac1-4fc5-93b0-98b2d2b9a9b1)

Pinned tab settings:

![Image](https://github.com/user-attachments/assets/91b8a0d5-74e0-4f42-b0dc-8f03b4fe33bc)

Behavior: Clicking on the button closes the tab, but does not reset the tab to its original URL. Doing ctrl+W does reset the URL
Expected behavior: Clicking on the button will reset the URL.

### Reproducible?

- [x] I have checked that this issue cannot be reproduced on Mozilla Firefox.

### Version

1.7.6b (Firefox 135.0) (64-bit)

### What platform are you seeing the problem on?

Windows

### Relevant log output if applicable

```shell

```

auckz:
I'm having the same issue, using the same version as well

Tyarel8:
Same problem here after updating

Anthonyy232:
Still occurs, please don't close this like the other one maubg, it happens even with the setting set to reset url. The keyboard shortcut works but the button doesn't. @JSGRD22

liuzhaowei55:
The same issue, and I want the pre version the refresh button, when click reset url to pined url and refresh

Anthonyy232:
Maybe has to do with this?
+        gZenPinnedTabManager._onCloseTabShortcut(event, this, 'unload-switch');

Should be left blank to lazy fetch to match the function signature
  _onCloseTabShortcut(event, selectedTab = gBrowser.selectedTab, behavior = lazy.zenPinnedTabCloseShortcutBehavior)

SennaSanzo:
Same here. It was working fine before the last update...

Anthonyy232:
Still not fixed, seems like it was an intended change.. Not sure why..

JSGRD22:
In the next update, this will be the default behaviour of this button, and there will be another toggle to reset the URL by clicking on the favicon.

seriaati:
Can confirm that the new behavior described above by Jupi is working in 1.8.1b

Issue #4754, opened 2025-01-31
Pinned Tabs - URL does not reset when tab is closed, only when returning to the tab


apotenza92:
### Captchas

- [x] I have read the instructions.
- [x] I have searched existing issues and avoided creating duplicates.
- [x] I am not filing an enhancement request.

### What happened?

When using the 'Close Tab Shortcut Behaviour' option called 'Reset URL and switch to next tab', I have noticed that the tab only resets back to the pinned URL after _returning_ to the pinned tab, not on close. This means if I return to a pinned tab I need to wait for the URL to reset before I can use it.

### Reproducible?

- [x] I have checked that this issue cannot be reproduced on Mozilla Firefox.

### Version

1.7.4b

### What platform are you seeing the problem on?

macOS - aarch64

### Relevant log output if applicable

```shell

```

tejassharma96:
this is no longer an issue in 1.8.1b (at least for me)

PR #5322, opened 2025-02-13
Fix URL reset button for pinned tabs

### LINKED ISSUE #5276: Pinned Tabs Do Not Retain Original URL
### Captchas

### Description

In the latest version of Zen Browser, it does not preserve the original URL of pinned tabs. Instead, it appears to save the last known URL before the tab is offloaded.

### Steps to Reproduce

1.     Open a webpage and pin the tab.
2.     Navigate to a page that requires VPN access or authentication.
3.     If the connection fails (e.g., due to missing VPN access), the pinned tab updates to the failed authentication page instead of keeping the original URL.
4.     Similarly, if a session times out on a web app, the pinned tab updates to the logout redirection URL rather than maintaining the originally pinned route.

### Expected Behavior

Pinned tabs should retain the URL that was set when the tab was originally pinned, regardless of session timeouts, redirects, or authentication failures.

### Actual Behavior

The pinned tab updates to the last loaded URL, losing the original intent of the pinned tab.
Impact

-     Users relying on pinned tabs for quick access to specific dashboards or authenticated pages find them unreliable.
-     Workflows requiring VPN access can be disrupted, as the pinned tab no longer points to the intended destination.

### Suggested Fix

Ensure that when a tab is pinned, its original URL is stored and does not change due to session redirects, authentication failures, or offloading events.

### Reproducible?

### Version

1.7.5b (Firefox 135.0) (aarch64)

macOS - aarch64

```shell

```

Anthonyy232:
Fixed the reset button for pinned tabs where it would unload and switch tabs but not reset the pinned tab to its original URL, even when the correct setting was selected in settings. Now defaults to lazy fetch the preference set.

Originally: `gZenPinnedTabManager._onCloseTabShortcut(event, this, 'unload-switch');`
Now: `gZenPinnedTabManager._onCloseTabShortcut(event, this);`

Per the function signature, it should now get the correct behavior by not specifying a behavior
`_onCloseTabShortcut(event, selectedTab = gBrowser.selectedTab, behavior = lazy.zenPinnedTabCloseShortcutBehavior)`

Fixes https://github.com/zen-browser/desktop/issues/5276, https://github.com/zen-browser/desktop/issues/5114, https://github.com/zen-browser/desktop/issues/4754

### DIFF
--- a/src/browser/components/tabbrowser/content/tab-js.patch
+++ b/src/browser/components/tabbrowser/content/tab-js.patch
@@ -41,7 +41,7 @@ index d41c486c02a6f09dcff5741a59ad8b617294c481..abaccd1935fc117924c44dd22cae0b32
        }
 +
 +      if (event.target.classList.contains("tab-reset-button")) {
-+        gZenPinnedTabManager._onCloseTabShortcut(event, this, 'unload-switch');
++        gZenPinnedTabManager._onCloseTabShortcut(event, this);
 +        gBrowser.tabContainer._blockDblClick = true;
 +      }
      }


mr-cheffy:
Did you just manually change the patch?

Anthonyy232:
The patch? Dunno what you're referencing, just changed one line of code in this PR. 


Edit*
Oh I see, is this a generated file lol.

PR #5342, opened 2025-02-13
Fix text rendering issues by decreasing gfx.canvas.accelerated.cache-size back to Firefox default

### LINKED ISSUE #3541: Font Rendering Error in Google Docs
### Captchas

### What happened?

Screenshot 2024-12-11 094646
Screenshot 2024-12-11 100116
Screenshot 2024-12-11 094452
Screenshot 2024-12-11 094512
Screenshot 2024-12-11 094534

Just updated to the beta version, and this happened. I don't have the same issue on Firefox Nightly 135.0a1. This is not generic version.

### Reproducible?

### Version

1.0.2-b.0

Windows

### Relevant log output

_No response_

JSGRD22:
Fixes #3541. This change is not tested as I do not experience the issue, but a large number of people have reported that changing the value manually works.

### DIFF
--- a/src/browser/app/profile/zen-browser.js
+++ b/src/browser/app/profile/zen-browser.js
@@ -425,7 +425,7 @@ pref("browser.aboutwelcome.enabled", false);
 
 // ---- Experimental settings to try make zen faster
 pref("gfx.canvas.accelerated.cache-items", 32768);
-pref("gfx.canvas.accelerated.cache-size", 4096);
+pref("gfx.canvas.accelerated.cache-size", 256);
 pref("gfx.content.skia-font-cache-size", 80);
 
 pref("media.memory_cache_max_size", 1048576);


mr-cheffy:
Why don't we just use the default?

JSGRD22:
According to https://github.com/zen-browser/desktop/issues/3541#issuecomment-2655638336, the default value in zen, likely introduced by some external factors like betterfox configs, is much higher than the default of firefox, and is the direct cause of the text rendering issues. Every individual facing the issue in #3541 and in the discord server who was given this solution so far has responded positively, indicating that setting the default value back to 256 should resolve the bug.

mr-cheffy:
Any performance hits?

JSGRD22:
> Any performance hits?

Possibly, we can see if there are noticeable effects in twilight; I haven't experienced any

PR #5435, opened 2025-02-16
Improved color picker

neurokitti:
this is not done but ive run out of time to finnish.
things that are still issues:
- we need to add the old floating dots version how the old color picker works
- I didnt add UI because im not sure how you want it 
- currently you press enter in the costom color picker to add a new dot lol (cuz i didnt mess with UI)
- custom colors need a bit added to work with the new dots system
- animations dont work on first click (i could probably fix just out of time rn)


JSGRD22:
Is it possible to add input validation to custom colours along the way to make sure people don't enter long/broken strings inside it? It can be fun to exploit but irreversibly breaks the theme for the workspace.

neurokitti:
> Is it possible to add input validation to custom colours along the way to make sure people don't enter long/broken strings inside it? It can be fun to exploit but irreversibly breaks the theme for the workspace.

God fucking damn it I just want to sleep.

neurokitti:
> Is it possible to add input validation to custom colours along the way to make sure people don't enter long/broken strings inside it? It can be fun to exploit but irreversibly breaks the theme for the workspace.

Yeah probably but people also use RGB as an input and hsl. So the logic would become very dense.

mr-cheffy:
Ill merge it into https://github.com/zen-browser/desktop/tree/new-color-picker so I can work on it 👍🏽

PR #5440, opened 2025-02-16
New color picker implementation

mr-cheffy:


neurokitti (approved):
looks good I think. the diff looked wrong for a bit in create dot but it turns out it just generated the dif weird.

Issue #5200, opened 2025-02-10
CMD+Enter inconsistently opens URLs in new tab on macOS


sambokai:
### Description

Before the recent update that removed the new tab page, CMD+Enter consistently opened the entered URL in a new tab. Now, this behavior is inconsistent.

**Expected Behavior:**

Holding CMD while pressing Enter should *always* open the URL in a new tab, regardless of how the address bar was focused.

**Actual Behavior:**

Sometimes CMD+Enter opens the URL in the current tab.

### Reproduction Steps

1.  **Works Consistently:**
    *   Press CMD+L (focus address bar).
    *   Type a URL.
    *   Press CMD+Enter.
    *   *Result:* URL opens in a new tab.

2.  **Fails Consistently:**
    *   Press CMD+T (new tab).
    *   Type a URL.
    *   Press CMD+Enter.
    *   *Result:* URL opens in the *same* tab.

### Additional Notes

*   The Windows equivalent shortcut is likely ALT+Enter. I'm not sure.

### Platform

macOS - aarch64

### Version

1.7.6b (Firefox 135.0) (aarch64)

### Reproducible?

- [x] I have checked that this issue cannot be reproduced on Mozilla Firefox.

KailashGanesh:
Hi, I checked this in both Mac and Windows - on a new tab `cmd` / `alt` + `enter` opens the URL in the same tab

is this a good first issue? if so I would like to work on it @mauro-balades

mr-cheffy:
> Hi, I checked this in both Mac and Windows - on a new tab `cmd` / `alt` + `enter` opens the URL in the same tab
> 
> is this a good first issue? if so I would like to work on it @mauro-balades 

Hi! If you'd like, sure, try find a solution!

Don't worry if you can't though, don't feel pressured

KailashGanesh:
Hi @mauro-balades , I've come up with a solution and made the fix to `UrlbarInput.sys.mjs` 

I'm unsure about the contribution process - should I create a .patch file using `pnpm run surfer patch create browser/components/urlbar/UrlbarInput.sys.mjs` before committing, or is there a different preferred workflow for contributions? 

Thanks for guidance!

sambokai:
still relevant

teal-bauer:
For me, it just stopped opening new tab for "search", instead it appends .com to my search string and opens in current tab. Only for full URLs does Cmd-Enter currently open a new tab. Firefox 136.0.1 (aarch64).

[...]

PR #5449, opened 2025-02-16
Fix: Ensure Cmd + Enter in URL bar always opens in a new tab

KailashGanesh:
**Fix for [Issue #5200](https://github.com/zen-browser/desktop/issues/5200#event-16249388177)**  

### Problem  
Pressing `Cmd + Enter` in the URL bar behaved inconsistently depending on how the URL bar was selected:  
- `Cmd + L` → Opened in a new tab  
- `Cmd + T` → Opened in the same tab  

**Bug demonstration:**  
gif of issue  

### Fix  
Now, `Cmd + Enter` will consistently open the URL in a new tab, regardless of whether the URL bar was selected using `Cmd + L` or `Cmd + T`.  

**Fix demonstration:**  
gif of fix  

Let me know if any improvements are needed


gulech:
I think your bug is already fixed on twilight but not in zen browser 1.7.6b beta.

`cmd + L` => opens in the same tab
`cmd + T` => opens a new tab

this is expected behaviour and currently works as it is in twilight. 

why do you need an extra step to open a new tap by adding `cmd + enter`?

KailashGanesh:
> `cmd + L` => opens in the same tab 
> `cmd + T` => opens a new tab
> 
> this is expected behaviour and currently works as it is in twilight.

You are right about `cmd + L` and `cmd + T` being for URL bar selection and new tabs.


> why do you need an extra step to open a new tap by adding `cmd + enter`?

The `cmd + enter` bug fix is about consistent behavior with browsers like Chrome / Arc.  pressing `cmd + enter` after typing a URL in the address bar will always open it in a new tab in those browsers.

In Zen, if you use `cmd + L`, type url and then press `cmd + enter` it would open URL in new tab
but if you use `cmd + T`, type url and then press `cmd + enter` it would open URL in the same (already occupied) tab

KailashGanesh:
Hi @mauro-balades Gentle reminder to check this PR, Thanks!

Issue #4322, opened 2025-01-13
Web panel pin button has wrong icon when clicking for the first time


mokazemi:
### Captchas

- [X] I have read the instructions.
- [X] I have searched existing issues and avoided creating duplicates.
- [X] I am not filing an enhancement request.

### What happened?

When you open the web panel, there's a button to keep it open (pin it). When you start the browser, it's shown as the "minimize" button. When you click it twice, it's now shown correctly as a "pin" icon. 

Here's video of the issue:
[Screencast_20250113_123527.webm](https://github.com/user-attachments/assets/c497d58d-cd74-470a-a57d-c5a0bb734769)

### Reproducible?

- [X] I have checked that this issue cannot be reproduced on Mozilla Firefox.

### Version

1.6b

### What platform are you seeing the problem on?

Linux (Tarball)

### Data from about:support if applicable (click on the "Copy text to clipboard" button)

<details>
  <summary>about:support</summary>
  

  ```
  Select this line and paste your about:support clipboard
  ```
</details>

mokazemi:
Yes @dosubot, it's still relevant and there's a pending pull request on it.

PR #5467, opened 2025-02-17
keep the "pinned" attribute of the side web panel and the button in sync

mokazemi:
I just added `pinned="true"` attribute to web panel pin icon, so that its initial state matches with the panel state.
Also added a function to keep the pinned state of the side web panel and the button in sync.  

**fixes:** #4322

### DIFF
--- a/src/browser/base/content/zen-sidebar-panel.inc.xhtml
+++ b/src/browser/base/content/zen-sidebar-panel.inc.xhtml
@@ -10,7 +10,7 @@
       </hbox>
       <hbox>
         <toolbarbutton id="zen-sidebar-web-panel-home" class="toolbarbutton-1 chromeclass-toolbar-additional" oncommand="gZenBrowserManagerSidebar.home();"/>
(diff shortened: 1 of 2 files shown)


mr-cheffy:
What if the initial state isn't pinned? I feel like this is a cheap solution to a bigger underlying problem. Where the check isnt being ran on initialization

mokazemi:
> What if the initial state isn't pinned? I feel like this is a cheap solution to a bigger underlying problem. Where the check isnt being ran on initialization

Oh, you're right. I should work more on it to see if I can do anything about it. I'm not much a JS developer but I guess I know where the issue is comming from. Maybe defining an init function that syncs the state of the panel with the icon state would solves the issue.

mokazemi:
I changed it to change the pin icon based on if the panel is pinned or no. @mauro-balades 
I checked all three states (initial run, when the panel is unpinned last time, when the panel is pinned last time). It seems to work in all these conditions.

mr-cheffy:
Ready for merge?

mokazemi:
> Ready for merge?

Yeah it's ready :+1:

PR #5538, opened 2025-02-19
Better aligment for the texture input

ErdemGKSL:
it was weirdly floating and overflows from the menu like this:
Image
Now it looks like this:
Image

I would rather adding the old "Texture" label actually, but i saw that it was purposely deleted so i just aligned it.


mr-cheffy:
I'm removing this input

ErdemGKSL:
oh i see

PR #5543, opened 2025-02-19
Revert "Add Bat to ZenEmojies"

mr-cheffy:
Reverts zen-browser/desktop#4734


WilliamWelsh:
Why remove an emoji from workspace options?

parkeraddison:
I'm a bit confused as well. The emoji list is clearly incomplete, but why? Where is the chosen subset of emoji coming from? @mauro-balades should we make this into an issue?

The 🦇 is a great example. According to [Emojipedia](https://emojipedia.org/bat):
> Bat was approved as part of Unicode 9.0 in 2016 and added to Emoji 3.0 in 2016.

Why aren't we supporting emoji as far back as 2016? For reference, by the time Zen was released in 2024, the world was currently at [Emoji 15.1](https://emojipedia.org/emoji-15.1). I'm a bit lost.

mr-cheffy:
The PR was badly done, it broke the search function

PR #5544, opened 2025-02-19
Finish the "Show sidebar on hover" PR

afonsofrancof:
In #4792 I made the `zen-has-hover` property be added to the sidebar and toolbar only if `zen.view.compact.show-sidebar-and-toolbar-on-hover` is true.

There was an oversight, though. The animation for the sidebar and toolbar was being triggered both by `zen-has-hover` and `hover`, like this

```css
#navigator-toolbox:hover,
#navigator-toolbox[zen-has-hover],
```

This meant that `:hover` was overriding `[zen-has-hover]`.

This PR makes both bars depend on `zen-has-hover` and removes `hover`.

I have tested this with the Browser Toolbox and it works perfectly :)

### DIFF
--- a/src/browser/base/content/zen-styles/zen-compact-mode.css
+++ b/src/browser/base/content/zen-styles/zen-compact-mode.css
@@ -113,7 +113,6 @@
       }
     }
 
-    #navigator-toolbox:hover,
     #navigator-toolbox[zen-has-hover],
     #navigator-toolbox[zen-user-show],
     #navigator-toolbox[flash-popup],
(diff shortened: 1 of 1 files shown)


mr-cheffy:
Does it still work?

afonsofrancof:
> Does it still work?

What do you mean?
It doesn't work with just my old PR.
It needs this one.

I missed that the default css hover was applied to the sidebar and toolbar. I thought it only had zen-has-hover

PR #5614, opened 2025-02-22
Use user's prefered search engine as private search by default

0xda157:
When going through setup if ddg is selected as preferred search engine private windows will still use google before this change. After this the "Use this search engine in Private Windows" checkbox will be toggled on by default, so the user's choice of search engine on the setup screen will be respected in private windows.


0xda157:
@mauro-balades could I get a review on this?

mr-cheffy:
I dont think we should do this to be honest, mainly because it might affect already changed's users preference and we can opt for a more private search engine for private mods. Sorry

PR #5914, opened 2025-02-28
Allow having full url without https in sidebar

zwamdurkel:
Before a certain update, the urlbar would show the full url without https. This is no longer possible. 

Before:
416575944-ff6ec01e-0632-430d-979d-599482a57231

Now with "trimURLS disabled and trimHTTPS enabled":
417371889-96467bed-6e4e-47f8-8651-3872200f76fc

Having https:// in front of the url is a waste of space, that's why I propose an additional flag that returns the old behavior. To the best of my knowledge, the changes in this pull request will fix this, but I could not manage to build the browser and test.

With this new flag disabled and trimURLS enabled, the behavior should return to the old look.

### DIFF
--- a/src/browser/app/profile/zen-browser.js
+++ b/src/browser/app/profile/zen-browser.js
@@ -120,6 +120,7 @@ pref('zen.view.compact.show-sidebar-and-toolbar-on-hover', true);
 pref('zen.urlbar.replace-newtab', true);
 pref('zen.urlbar.behavior', 'floating-on-type'); // default, floating-on-type, float
 pref('zen.urlbar.wait-to-clear', 45000); // in ms (default 45s)
+pref('zen.urlbar.show-domain-only-in-sidebar', true);
 
 #ifdef XP_MACOSX
 // Disable for macos in the meantime until @HarryHeres finds a solution for hight DPI screens
--- a/src/browser/base/content/ZenUIManager.mjs
+++ b/src/browser/base/content/ZenUIManager.mjs
@@ -9,6 +9,7 @@ var gZenUIManager = {
     XPCOMUtils.defineLazyPreferenceGetter(this, 'sidebarHeightThrottle', 'zen.view.sidebar-height-throttle', 500);
     XPCOMUtils.defineLazyPreferenceGetter(this, 'contentElementSeparation', 'zen.theme.content-element-separation', 0);
     XPCOMUtils.defineLazyPreferenceGetter(this, 'urlbarWaitToClear', 'zen.urlbar.wait-to-clear', 0);
+    XPCOMUtils.defineLazyPreferenceGetter(this, 'urlbarShowDomainOnly', 'zen.urlbar.show-domain-only-in-sidebar', true);
 
     gURLBar._zenTrimURL = this.urlbarTrim.bind(this);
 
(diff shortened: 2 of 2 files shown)


mr-cheffy (approved):

busybox11:
Hey, circling back on this. I just spent way too long trying to figure out why `trimHttps` wasn't doing anything on the sidebar.

I naively assumed the Zen sidebar would respect the `browser.urlbar` props, unaware of this `zen.urlbar.show-domain-only-in-sidebar`.

If this behavior is intentional, no worries and feel free to ignore this!

However if you believe this could be a bug, I've already dug into the code and fixed it on my end to see what was going on. Let me know if you'd like me to file a PR!

(apologies for the messy troubleshooting haha)

zwamdurkel:
> However if you believe this could be a bug, I've already dug into the code and fixed it on my end to see what was going on. Let me know if you'd like me to file a PR!

Can you explain to me what you "fixed"? Zen currently has 3 ways to show the address:

1. Domain only, no URL path, no https (`zen.urlbar.show-domain-only-in-sidebar: true`, `browser.urlbar.trimHttps: ignored` )
2. Full URL including path and https (`zen.urlbar.show-domain-only-in-sidebar: false`, `browser.urlbar.trimHttps: false` )
3. URL including path, but no https (`zen.urlbar.show-domain-only-in-sidebar: false`, `browser.urlbar.trimHttps: true` )

Have you added the option for domain only, no URL path, but WITH https? That is to my knowledge the only non-working configuration at the moment.

busybox11:
Didn't fix anything yet, I'm not sure if this is intended behavior.

Those combinations do indeed work - forgot to explicitly mention the `trimURLs` prop which was causing me trouble. I expected it to control the sidebar URL behavior too.

For reference, I have `trimHttps` set to true, `trimURLs` to false, and everything else by default. In this situation the sidebar still shows the entire URL with the https prefix.

After setting `zen.urlbar.show-domain-only-in-sidebar` to true, there is different behavior. Can confirm it shows URL+path without https if the prop enabled and everything else set to default.

I was expecting the sidebar URL bar to respect `browser.urlbar` config values, but was unaware of `zen.urlbar.show-domain-only-in-sidebar`. This is mostly what I'm wondering here.

zwamdurkel:
I think the point here is that the sidebar has domain only by default. Meanwhile, if you use the full urlbar (having it at the top of the screen), more of the url is shown. It is default behaviour of Zen to have this distinction. I have added the custom flag to disable the domain only mode in the sidebar urlbar. 

If we rely on just the `browser.urlbar` config values, I don't think this distinction can be made. That is why we need a Zen specific flag. Maybe a whole copy of `browser.urlbar` values for the sidebar would be better for customizability. If you have a different idea regarding this, I think you could start a discussion or make a pull request or poll interest in the Discord server / Reddit.

If you feel like I still don't understand what you are saying. I'm sorry, but I'm sure there is someone else who does. Probably not in the comments of this pull request though.

busybox11:
No this makes perfect sense! Sorry I didnt phrase my comments well.

I guess I was expecting the sidebar to use the same config props as the regular URL bar - I don't think there is any state in which the user can see both using a single config option, either only sidebar or only regular bar? lmk if i'm mistaken.

If this is however possible then yes this config distinction has to be made.

My issue mostly came down to the fact that I wasn't able to understand easily why I kept having the full URL, despite having `trimURLs` and `trimHttps` set to true. This alone is not intuitive, but if we have to make the distinction then I understand.

I still believe that having unique properties producing different end results based on the layout is confusing, maybe it would be better to duplicate them in this case.

I expect `trimURLs` to trim URLs everywhere and `trimHttps` to trim HTTPS everywhere basically!

Hope I was able to properly share what I had in mind - English isn't my first language and communication is not my strongest. Also the horrible heat isn't making it any easier haha.

Would be happy to discuss over Discord if you want, don't want to take too much time though. Again very sorry..!

PR #5634, opened 2025-02-23
Implement a more complete JXL features

noaione:
Based on this following phabricator patch/commits:
- https://phabricator.services.mozilla.com/D119700
- https://phabricator.services.mozilla.com/D122158
- https://phabricator.services.mozilla.com/D122159

Rebased for FF 136.0 RC code

Which add the following:
- Proper color profiles support for JPEG XL
- Animated JPEG XL
- Progressive decoding

Also added the following: https://src.fedoraproject.org/fork/eclipseo/rpms/firefox/blob/90702891fefc97ff9e347dbab0091136822a0f3b/f/0002-Bug-1539075-Register-.jxl-extension-in-Firefox.-r-tn.patch

Fix #726

---

https://github.com/user-attachments/assets/fbf8a16f-bf48-470b-9ebc-264509d19ac4

Tested on: https://jpegxl.info/resources/jpeg-xl-test-page.html and https://google.github.io/attention-center/ (Progressive demo)

Demo build: https://nightly.link/noaione/zen-desktop/actions/runs/13485116765

PR #5999, opened 2025-03-03
Report JXL frame counts properly

noaione:
FF 129 introduces some new function to report the frame count of an animated images.
Address issue talked here: https://github.com/zen-browser/desktop/pull/5634#issuecomment-2692802618

Although I genuinely don't know if this the correct behavior on the progressive decoding side (I don't want to pay CHF 65 for the ISO docs to verify if this actually correct or not)

https://github.com/user-attachments/assets/821aa31d-f16d-4d65-b922-730b9d5de197

Demo build: https://github.com/noaione/zen-desktop/actions/runs/13619191467


mr-cheffy:
@noaione if you have more cool patches like these, let us know!

mr-cheffy:
I'll remove these patches now from Zen once it updates to FF 149 as firefox seems to be back at it, should take a load off from us and I trust them more than I do myself!

image

PR #6156, opened 2025-03-07
remove unnecessary declaration of several variables

### LINKED ISSUE #2748: Opening links in Zen AppImage causes 100% CPU usage
### Captchas

### What happened?

When I set the default browser as Zen Browser, it sets this desktop file: `userapp-Zen Browser-08XPW2.desktop`, which has this content, and after every restart of the zen browsers, it creates different desktop files.
Here:
```
❯ fd zen
userapp-Zen Browser-08XPW2.desktop
userapp-Zen Browser-5XDIW2.desktop
userapp-Zen Browser-8SKIW2.desktop
userapp-Zen Browser-AKFWW2.desktop
userapp-Zen Browser-NLUTW2.desktop
userapp-Zen Browser-QV8UW2.desktop
userapp-Zen Browser-T9OPW2.desktop
zen-alpha.desktop
zen-browser-AM.desktop
```
I installed zen through [AppMan](https://github.com/ivan-hc/AppMan).
I tried to manually set the default desktop file using this command: `xdg-mime default zen-browser-AM.desktop x-scheme-handler/https`, but when something opens through, it will have 100% CPU usage all the time.
### Reproducible?

### Version

1.0.1-a.17

Linux

### Relevant log output

_No response_

Samueru-sama:
Likely fixes #2748

The reason is that the AppRun is setting a lot of env variables that don't need to be set. More importantly it is setting `LD_LIBRARY_PATH` which would make everything being launched by the browser try the libraries in the AppImage first, which likely causes that issue with `xdg-open` since env variables are inhereted by child processes. 

See https://github.com/pkgforge-dev/Citron-AppImage/issues/3 for an example of such issue.

Since the AppImage is already based off the portable tar build. there is no need to  be setting those env variables, since the binary already knows to look for the libraries bundled with it.

Another change is that the `AppRun` will not look into the `.desktop` to find the name of the zen binary, since it is always going to be zen right?

If there are plans for the binary name to change in the future let me know, that way I can revert that change.

### DIFF
--- a/AppDir/AppRun
+++ b/AppDir/AppRun
@@ -1,15 +1,6 @@
 #!/bin/sh
-SELF=$(readlink -f "$0")
-HERE=${SELF%/*}
-export PATH="${HERE}:${HERE}/usr/bin/:${HERE}/usr/sbin/:${HERE}/usr/games/:${HERE}/bin/:${HERE}/sbin/${PATH:+:$PATH}"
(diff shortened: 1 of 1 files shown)


Samueru-sama:
Why is the linter complaining? 😅 I tested the thing and it is 100% compliant POSIX shell. 

edit: Wao I butchered the PR title xd

mr-cheffy (approved):

PR #6244, opened 2025-03-09
Disable AI chat stuff by default

noaione:
Fix https://github.com/zen-browser/desktop/discussions/6236

### DIFF
--- a/src/browser/app/profile/zen-browser.js
+++ b/src/browser/app/profile/zen-browser.js
@@ -437,6 +437,13 @@ pref("browser.newtabpage.activity-stream.asrouter.userprefs.cfr.features", false
 pref("browser.preferences.moreFromMozilla", false, locked);
 pref("browser.aboutwelcome.enabled", false);
 
+// AI/ML stuff
+pref("browser.ml.chat.enabled", false);
+pref("browser.ml.chat.shortcuts", false);
+pref("browser.ml.chat.shortcuts.custom", false);
+pref("browser.ml.chat.sidebar", false);
+pref("browser.ml.enable", false);
+
 // ---- Experimental settings to try make zen faster
 pref("gfx.canvas.accelerated.cache-items", 32768);
 pref("gfx.canvas.accelerated.cache-size", 256);


mr-cheffy (approved):

matezoltanfarkas:
I personally love this feature since it allows me to be more productive by summarizing web pages, giving me quick access to information and sources of inspiration (e.g. boilerplates for emails) and also helps me to provide information about stuff I need to check in articles I read.
For me, banning AI stuff from zen is a dealbreaker and I still very much would like to have an option to turn it on manually.

12th-devs:
@matezoltanfarkas toggle these configs to true in `about:config`

```
pref("browser.ml.chat.enabled", false);
pref("browser.ml.chat.shortcuts", false);
pref("browser.ml.chat.shortcuts.custom", false);
pref("browser.ml.chat.sidebar", false);
pref("browser.ml.enable", false);
```

matezoltanfarkas:
thanks, I've figured it out by myself already from the PR!
Just wanted to leave a note for other users as well.

hkhelif:
An answer was already provided, but I want to touch on this point

> For me, banning AI stuff from zen is a dealbreaker and I still very much would like to have an option to turn it on manually.

I think such features should always be set as an opt-in option rather than an opt-out option, not everyone can use these tools and enabling them by default can create a security and privacy risk to the end user.

Anthonyy232:
> An answer was already provided, but I want to touch on this point
> 
> > For me, banning AI stuff from zen is a dealbreaker and I still very much would like to have an option to turn it on manually.
> 
> I think such features should always be set as an opt-in option rather than an opt-out option, not everyone can use these tools and enabling them by default can create a security and privacy risk to the end user.

This is a good point, but such a feature should be easily accessed in settings, not about:config. Hope we can agree about that.

noaione:
You need to ask Mozilla to do that, best place to put in the the settings would be in one of their page and they need proper localization text or it won't show up properly.

matezoltanfarkas:
> not everyone can use these tools

That's fair. Most of us are probably only browsing the web. However that does not mean such useful features should be hidden in about:config.
It's almost like hiding the DNS configuration or the security certificate settings or the privacy options as well. Not everyone knows they're there, but those who do are heavily relying on them and want them to be easily accessible.

> enabling them by default can create a security and privacy risk to the end user.

That's a point I see often but don't understand. Would you mind to elaborate?

Anthonyy232:
> You need to ask Mozilla to do that, best place to put in the the settings would be in one of their page and they need proper localization text or it won't show up properly.

Well since it is in the Zen fork, it should just be under regular settings, or removed completely. As for the localization, I'm pretty sure Zen introduces new text strings as well, it can use the same localization process. Essentially, own it or let it go. It should be in the settings for such a used feature in this day and age -- but if the concern is privacy then why not remove it entirely?

matezoltanfarkas:
> As for the localization, I'm pretty sure Zen introduces new text strings as well, it can use the same localization process. Essentially, own it or let it go. 

May I add some humour to the discussion?
Just use the goddamn sidebar to translate the string for any supported language! 😂
Jokes aside: I see the added work and am looking forward to the devs opinion on implementability.

Ivan-Kim:
Only discovered this thread after a whole month kept in the dark about this change. I am not interested in debating the rationale but would have been nice to list this under `Breaking Changes` section in the release notes like the [removal of web panels](https://github.com/zen-browser/desktop/releases/tag/1.11b), at least transparency with that removal allowed me to quickly adapt my workflow.

For about a month I just assumed it was Firefox upstream messing up support as it is an experimental feature after all, but turns out it was a silent feature drop. I understand Zen is still at beta stage and all but more transparency regarding removal/disabling of user facing features would be appreciated.

ioc2e3:
> ... ，我仍然非常希望有一個手動打開它的選項。

Agree ...

Hope there is an option to turn it on or off ...

.

kelvinchincc:
I concur that if privacy is a concern for the Zen team and the target audience, removing the feature entirely may be the best course of action.  Currently, enabling this feature requires manually adjusting about:config, which is inconvenient and potentially risky, as users may not always make the correct adjustments.  If the Zen team decides to remove this feature and prevent users from enabling it, please clearly communicate this decision so users do not attempt to enable it. Alternatively, adding the setting to the preferences would be a more responsible approach; directing users to about:config is not ideal.

mr-cheffy:
We'll just leave it as an option. We don't have plans on locking it

PR #6254, opened 2025-03-09
Restore AI chat settings default while keeping shortcuts disabled

JSGRD22:
The main issue with the AI chat is the unnecessary visual clutter added by the shortcut that appears on text selection, controlled by `browser.ml.chat.shortcuts`. Hence, this will stay disabled by default. Everything else stays out of the way and does not interfere with any actions unless the user seeks them. The AI chatbot sidebar only loads on opening and only uses 3rd-parties, so it does not compromise on users' privacy except by intentional action.
Disabling these configs by default may pose an annoyance for those seeking to use this feature, as they will need to seek multiple configs to toggle manually.
This commit ensures a more balanced default, where the chatbot features remain easily accessible but do not interfere with normal interactions like highlighting text.

### DIFF
--- a/src/browser/app/profile/zen-browser.js
+++ b/src/browser/app/profile/zen-browser.js
@@ -438,11 +438,7 @@ pref("browser.preferences.moreFromMozilla", false, locked);
 pref("browser.aboutwelcome.enabled", false);
 
 // AI/ML stuff
-pref("browser.ml.chat.enabled", false);
 pref("browser.ml.chat.shortcuts", false);
-pref("browser.ml.chat.shortcuts.custom", false);
-pref("browser.ml.chat.sidebar", false);
-pref("browser.ml.enable", false);
 
 // ---- Experimental settings to try make zen faster
 pref("gfx.canvas.accelerated.cache-items", 32768);


matezoltanfarkas:
hey @JSGRD22 what happened to this PR? I like this compromise!

JSGRD22:
The [privacy policy](https://zen-browser.app/privacy-policy/) states: "We do not allow third-party trackers or analytics tools to operate within Zen Browser." After an internal discussion, it was decided that disabling the AI chat was the best course of action to fulfill this promise.

matezoltanfarkas:
I don't get it. Inference happens on the server-side, while the chatbots opened in the sidebar are treated like normal webpages.
In other words: what difference does it make to open ChatGPT on a separate webpage vs opening it via the sidebar?
Both actions are triggered by the user's intention to use a feautre knowing the input data will be transmitted to a distant server. Hence, there is no "unusual privacy violation" happening here, only the ones, which would happen visiting the service manually anyway -- or am I missing something?

Edit the privacy policy [exempt](https://www.mozilla.org/en-US/privacy/firefox/#notice) from mozilla sums it up:

> If you choose to enable a chatbot in the sidebar and/or through a shortcut, Mozilla does not have access to your conversations or the underlying content you input into the selected chatbot. We do collect technical and interaction data on how this feature is used to help improve Firefox, such as how often each third-party chatbot provider is chosen, how often suggested prompts are used, and the length of selected text.

So basically, telemetry has to be killed (already done) and then we are good to go -- aren't we?

PR #6460, opened 2025-03-14
Add feature: Split New tab on Right with URL focus using shortcut (As seen on Arc Browser)

saikanam:
# Split New Tab on Right with Keyboard Shortcut

   This PR adds a new keyboard shortcut (Ctrl+Shift+'+') that:
   - Creates a vertical split view with a new tab on the right
   - Automatically focuses the URL bar for immediate navigation
   - Allows adding multiple tabs to the right by repeatedly pressing the shortcut

   ## Why This Feature?

   This enhances the Split View functionality by providing a quick keyboard-driven workflow for creating and navigating split tabs. It's particularly useful for users who prefer keyboard shortcuts over mouse interactions.

   ## Implementation Details

   - Added a new method `createSplitNewTabToRight()` to ZenViewSplitter.mjs
   - Registered the shortcut in ZenKeyboardShortcuts.mjs
   - Added support for adding multiple tabs to the right side
   - Ensured proper URL bar focus behavior

   ## Localization Strings

   This feature requires the following localization string to be added to the appropriate files in the `l10n` submodule:
``` zen-split-view-shortcut-url-focus = Split View with URL Focus ```

Since these files are in a submodule, I've not included them in this PR.

## Image demo
image


mr-cheffy:
Id rather it opens an empty tab instead of about:blank

saikanam:
Changed it to add newtab instead of blank, thanks.

One thing I forgot to mention is that max split tabs are 4, same as this.MAX_TABS

mr-cheffy:
Sorry, I meant as it opens this empty state:

image


and please do fix linter errors: https://github.com/zen-browser/desktop/actions/runs/13865573260/job/38942146067?pr=6460

saikanam:
I fixed the linter errors.
That is strange if it opens the browser in that state, atleast that's what I think you mean.

Let me investigate further on a new build.

saikanam:
It works as intended on my end, please let me know if I am missing something.

The intent was to open a new split tab to the right which is empty.


That's what happens in Arc as well, when you click control + shift + '+', it opens a new empty tab on the right alongside the url bar.

https://github.com/user-attachments/assets/d90fbb7c-031d-4f04-89da-acd46091348e

saikanam:
https://github.com/user-attachments/assets/06f14dad-5c6b-4409-93a0-29588011fd89

A demo of the intended workflow

mr-cheffy:
Ok, it looks good. But @saikanam could you please always open about:blank and add `_forZenEmptyTab` to the parameters?

saikanam:
If I add _forZenEmptyTab in the parameters,it breaks the functionality by unsplitting the tabs. This is most likely due to this attribute being heavily used in other parts of zenViewSplitter to make certain filtration.

Is this attribute absolutely necessary to add? I might have to find a workaround then.

mr-cheffy:
Yeah sorry... I can take a look at it some time this week if you want

saikanam:
That would be great, please share any insights here so I can take a look at it as well, I will do the same.

mr-cheffy:
Will do! Once I release the new update with firefox 137 ill start taking a look

saikanam:
Ok there were a lot of changes upstream but I had a chance to update the repo and take a look. 

The Issue:
  When _forZenEmptyTab is used, it sets the zen-empty-tab
   attribute on the tab, which has these effects:
  1. The tab gets display: none in CSS (completely
  hidden)
  2. Split view operations filter out tabs with
  zen-empty-tab attribute in multiple places.

  The Result:
  Using _forZenEmptyTab makes the tab invisible and
  unable to participate in split view operations,
  breaking the core functionality.

  Possible Solutions:
  1. Use about:blank without _forZenEmptyTab - Gives
  empty state but tab remains functional
  2. Modify split view filtering to allow empty tabs in
  split contexts (but this might have unintended side
  effects)
  3. Create a new parameter like _forZenSplitEmptyTab
  that gives empty state without the problematic
  attribute

  Question:
  What specific empty state behavior are you looking for?
   Is it:
  - The visual empty state of about:blank?
  - Some special UI behavior that _forZenEmptyTab
  provides?
  - Something else?
  
  This will help me on how to proceed

mr-cheffy:
@saikanam sorry for not answering. In that case yeah, we can just set about:blank... or make it an empty tab and add an attribute `allow-split` to the tab

mr-cheffy:
Sorry, this has been implemented in https://github.com/zen-browser/desktop/commit/e6882a42d60468eeb5392069d8379f53e9dde82e as I believed there was a more "intuitive" way of tackling this issue...

PR #6466, opened 2025-03-14
Respect prefers-reduced-motion in zen-styles

TimStewartJ:
Currently there are some animations in the browser which do not respect prefers reduced motion.

My particular use case is Zen over RDP - the animations end up pretty choppy in this scenario so it'd be nice if they followed the OS level preference for reduced motion.


src/browser/base/content/zen-styles/zen-panel-ui.css
```diff
@@ -3,6 +3,16 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  */
+@media (prefers-reduced-motion: reduce) {
```
mr-cheffy, line 6: Why not just use `not` for the selectors at the bottom to avoid duplication?

mr-cheffy (changes requested):

src/browser/base/content/zen-styles/zen-panel-ui.css
TimStewartJ: Oh sorry I didn't mean to open this PR early, I was just trying out the new Copilot Workspaces feature - I'll continue working on this as time allows

PR #6486, opened 2025-03-16
Fix bash install command in documentation on how to install for Linux using tarball

pewpewnor:
While copying and pasting the Bash install command from the README for other Linux distributions using the tarball, I found that the command did not work. It turns out that there is an unnecessary backtick at the end of the command?

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,7 +101,7 @@ yay -S zen-browser-bin
 - `Tarball` install:
 
 ```sh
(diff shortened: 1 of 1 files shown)


mr-cheffy (approved):

eduardoleolim:
There is a pull request in `updates-server` repo to check the architecture and install the most suitable tarball.

Issue #5022, opened 2025-02-07
Essentials tabs are automatically unloaded on browser restart


Sjoertjuh:
### Captchas

- [x] I have read the instructions.
- [x] I have searched existing issues and avoided creating duplicates.
- [x] I am not filing an enhancement request.

### What happened?

When restarting the browser, the essentials tabs are unloaded by default. I believe that they should be loaded by default and there should be an option to start with them unloaded. Enabling `prevent tab unload` on them doesn't do anything on restart.
In my case this would solve the issue of me having to open every one of them when starting up the browser, since I need most tabs for sending notifications, which requires them to be active.

### Reproducible?

- [ ] I have checked that this issue cannot be reproduced on Mozilla Firefox.

### Version

original issue: 1.7.4b
current: 1.9.1b

### What platform are you seeing the problem on?

Windows

### Relevant log output if applicable

```shell

```

cyb3rd4d:
I am also facing this issue. The fact that essential tabs are not loaded on startup defeats the purpose of the feature IMO.

Edit: Firefox 1.9b, official Flatpak, ArchLinux

standerksen:
Facing the same issue. Zen Browser 1.9b. Windows 10.

mr-cheffy:
Turn `browser.sessionstore.restore_pinned_tabs_on_demand` to false

cyb3rd4d:
Thank you @mauro-balades 
Is it planned to set this config as false by default in a future release? I really think that this should be the default behavior of Essential Tabs.

mr-cheffy:
We could do it on by default only for essentials, but it's not really a priority right now. Open for PRs though!

cyb3rd4d:
> We could do it on by default only for essentials

Yes I think it's more important for essentials than for "normal" pinned tabs 👍 

> Open for PRs though!

OK I'll try to check this out as soon as I can!

jhuang0:
Seems like this is still an issue with no resolution.

PR #6653, opened 2025-03-20
Disable restoring pinned tabs on demand by default

cyb3rd4d:
Currently, Essentials are not loaded when the browser starts. The issue has already been reported: https://github.com/zen-browser/desktop/issues/5022

I think that not loading Essentials on startup is not what we want when we use such a feature. I personally have my email client in those tabs and I expect it to be loaded when I open my browser.


Anthonyy232:
Can't this be another config? Would like to keep the behavior for pinned tabs and I'm sure many other people would appreciate this behavior not being changed out of the blue

cyb3rd4d:
A separate config would be ideal. But I'm not sure to be able to implement it... I would really appreciate if someone could guide me :pray:

PR #6667, opened 2025-03-21
ci: Use the default github.token if secrets.DEPLOY_KEY is not available

TimStewartJ:
In github actions, `github.token` is available by default. By using this as a fallback, people will be able to generate twilight builds from their forks without needing to set up any extra secrets.

Just a little convenience update for people developing in their forks since it can be a lot set up a dev enviornment. It's a lot easier to just kick off a build and check on it when it is done.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -98,7 +98,7 @@ jobs:
         uses: actions/checkout@v4
         with:
           submodules: recursive
-          token: ${{ secrets.DEPLOY_KEY }}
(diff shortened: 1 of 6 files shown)


mr-cheffy:
GitHub.token still would fail on the actions though because it doesn't have enough permissions

TimStewartJ:
In case this information is helpful to anyone else, it still has enough permissions to generate binaries for all platforms. You can see an example run here: https://github.com/TimStewartJ/desktop/actions/runs/13951534800

PR #7394, opened 2025-04-06
Fix sidebar always stays left

sporocyst:
change the order to honor `positionend` attribute in #sidebar-box while maintaining dragable by splitter


mr-cheffy:
Could you please run "npm run pretty"?

Issue #7429, opened 2025-04-07
Sidebar padding takes space in full screen


Amnoshield:
### Captchas

- [x] I have read the instructions.
- [x] I have searched existing issues and avoided creating duplicates.
- [x] I am not filing an enhancement request.
- [x] I have checked that this issue cannot be reproduced on Mozilla Firefox.
- [x] I have checked that this issue can be reproduced once I removed all my Mods and Custom CSS.

### What happened?

When the sidebar is open (on bookmarks for example) and I full screen on a page (like a YouTube video) the padding from the side bar (space in between the sidebar and the edge of the screen) is still visible.
This does not happen if the sidebar is not open.

This may be related to #7260 as I also have that issue but I don't know.

Here is an img:
![Image](https://github.com/user-attachments/assets/256f5cbe-1abf-4eeb-8cc7-c8526c8c015c)

### Version

1.11.1b (Firefox 137.0) (64-bit)

### What platform are you seeing the problem on?

Windows - x64

### What component is this issue related to?

Bookmarks

### Relevant log output if applicable

```shell

```

Propheticus:
Can recreate. See the same small bar on the left. (Also on Win x64)

sporocyst:
Did this happen on prveious ver? say 1.9 or 1.8b

Propheticus:
Good question to help find the possible code diff to blame for the issue. I don't remember seeing this before, but also don't have the sidebar open all the time...

Amnoshield:
> Did this happen on prveious ver? say 1.9 or 1.8b

I only noticed this in the latest version after I updated.

Propheticus:
I can reproduce this issue in the latest twilight (Version 1.11.4t Build ID 20250417103312) on a clean profile.

Workaround: popout the video PiP and then fullscreen.

\* updated with latest restest

Amnoshield:
Still happening on version 1.11.2b (Firefox 137.0.1) (64-bit)

4NXIE7Y:
Still happening on version 1.11.4b (64-bit) Windows.

Amnoshield:
Still happening on the latest version 1.11.5b (Firefox 137.0.2)(64-bit) and it is happening on Ubuntu and Windows 11

Propheticus:
Unless closed or otherwise updated (or in release notes) you can assume, yes, it's still happening.
Latest is 1.12.2b btw

SO9010:
[...]

PR #7449, opened 2025-04-08
Fix padding when full screen by setting gap value to 0

sporocyst:
Somehow, gap in #tabbrowser-tabbox is still taking effect even when its child elements(sidebar and splitter) are set with `visibility:collapse`. This cause unwantted space when entering fullscreen as #7429 mentioned.

### DIFF
--- a/src/browser/base/content/zen-styles/zen-browser-ui.css
+++ b/src/browser/base/content/zen-styles/zen-browser-ui.css
@@ -155,6 +155,9 @@
   flex-direction: row;
   position: relative;
   gap: var(--zen-element-separation);
+  :root[inDOMFullscreen] & {
+    gap: 0;
+  }
 }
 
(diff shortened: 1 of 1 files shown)


mauro-balades:
You realize this is used under a specific `:root:not([inDOMFullscreen='true'])` right?

src/browser/base/content/zen-styles/zen-browser-ui.css
```diff
@@ -155,6 +155,9 @@
   flex-direction: row;
   position: relative;
   gap: var(--zen-element-separation);
+  :root[inDOMFullscreen] & {
```
mr-cheffy, line 158: You realize this is used under a specific `:root:not([inDOMFullscreen='true'])` right?

mr-cheffy (changes requested):

PR #7477, opened 2025-04-09
Fix #7429 `Sidebar padding takes space in full screen`

### LINKED ISSUE #7429: Sidebar padding takes space in full screen

sporocyst:
Limiting the presence of gap to non-fullscreen mode

### DIFF
--- a/src/browser/base/content/zen-styles/zen-browser-ui.css
+++ b/src/browser/base/content/zen-styles/zen-browser-ui.css
@@ -150,7 +150,7 @@
   }
 }
 
-#tabbrowser-tabbox {
+:root:not([inDOMFullscreen='true']) #tabbrowser-tabbox {
   display: flex;
   flex-direction: row;
   position: relative;


mr-cheffy:
It's also removing the fact that it's a flexbox though? I think this is a hacky solution, maybe we should look into adding a `collpase` visibility to the correct element on fullscreen. I dont understand why a gap would make much difference either, I can't reproduce the issue so im not sure...

sporocyst:
Or we can only add `:root:not([inDOMFullscreen='true'])` on `gap: var(--zen-element-separation);`
And... I can reproduce this even on native Firefox by adding `gap`. That is why I believe this is the culprit

PR #7578, opened 2025-04-13
Add min width to split-view vertical splitter

sporocyst:
Currently , the width of split-view vertical splitter is defined by:
```
width: var(--zen-split-row-gap);
--zen-split-row-gap: var(--zen-element-separation);
```

The width of splitter becomes 0 and is not accessible when `--zen-element-separation` is manually set to 0 in `about:config`
A minimal width of 1 px is given in this pr to avoid the above condition.

===
Note: 
In comparision, the height of horizontal splitter is `calc(var(--zen-element-separation) + 1px)`, which also avoids the aforementioned condition.
I chose to add min-width because I believe that has least overall impact.


mr-cheffy:
Whoops, sorry about that. I literally pushed a fix about that a few minutes ago...

PR #7672, opened 2025-04-17
fix: closes #7274 - Fixed mouses simulating swipes with Forward and B…

### LINKED ISSUE #7274: MX Master 3 Forward/Back Mouse Buttons Don't Work For Switching Workspaces While Logi Options+ is Installed
### Captchas

### What happened?

I use the MX Master 3 mouse, and the forward and back buttons do not work for switching workspaces while Logi Options+ is installed. I found issue #4554, which was closed. However, it does not seem that that issue was completely related, as even the most recent comment still mentions that Logi Options+ is still causing issues. So, I thought I would open a new issue to get further clarification, as this is really messing with my workflow. Plus, I cannot uninstall Logi Options+, as that would make my mouse unusable for what I need. Thanks!

Just for some clarification:
- MX Master 3 buttons work fine if Logi Options+ is uninstalled, it's just when it's installed
- Having Logi Options+ installed does not prevent the issue on other mice. I tried my Logitech G502, and it worked fine. It's only my MX Master 3 (and probably other mice that run through Logi Options+) that have the issue
- zen.workspaces.scroll-modifier-key is currently set to control, but this does not work with this mouse either. Works fine with my G502.

If there's anything else I can provide, please let me know!

### Version

1.11b

macOS - aarch64

Workspaces

```shell

```

mathieutu:
Hi, 

Firstly, thanks for this amazing work! This browser is a 💎, and it's an honour to contribute to such a project!

On several mouses, and mouses customization softwares (like Logitech Option +, or Better Mouse), the Back and Forward event are emulated by sending a Left or Right SwipeEvent.

This blocks the usage of these buttons to switch workspace due to the way the events were previously handled. 

This PR fixes the issue.

Tested with a Logitech MX Vertical, and a MacBookPro Trackpad.

Thanks!

fix #7274

### DIFF
--- a/src/zen/workspaces/ZenWorkspaces.mjs
+++ b/src/zen/workspaces/ZenWorkspaces.mjs
@@ -472,19 +472,15 @@ var ZenWorkspaces = new (class extends ZenMultiWindowFeature {
   }
 
   async _handleSwipeEnd(event) {
-    if (!this.workspaceEnabled || !this._swipeState?.isGestureActive) return;
+    if (!this.workspaceEnabled) return;
     event.preventDefault();
     event.stopPropagation();
     const isRTL = document.documentElement.matches(':-moz-locale-dir(rtl)');
-    const moveForward = (this._swipeState.direction === 'right') !== isRTL;
+    const moveForward = (event.direction === SimpleGestureEvent.DIRECTION_RIGHT) !== isRTL;
 
-    let rawDirection = moveForward ? 1 : -1;
-    if (this._swipeState.direction) {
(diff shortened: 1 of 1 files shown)


mr-cheffy (approved):
Works great from my side!

Blendertom:
@mathieutu 
Is this still working for you?

I just did a clean install and this no longer works for me consistently. It will sometimes randomly work if I start scrolling with my trackpad, and then right after that I scroll with the thumb wheel.

mathieutu:
@Blendertom I did not reinstall the mouse software, so I can't say right now, sorry.

PR #7675, opened 2025-04-17
feat: improve tab switching with trackpad

mathieutu:
Hi,

When switching from Arc, I directly felt an issue when switching the workspaces with the trackpad.
Most of the time, the swipe is canceled, and reverted.

This is because the `MozSwipeGesture` is not triggered until a quite big threshold.

This PR improves the behavior, by triggering the workspace switch directly from the `MozSwipeGestureUpdate`.

| Before  | After |
| ------------- | ------------- |
| <video src="https://github.com/user-attachments/assets/061097b7-22c0-4e4c-b348-ea30e9826de7">  | <video src="https://github.com/user-attachments/assets/8ed76673-ed00-4e5a-b9db-cd3fddc59d92"> |

Thanks!
Mathieu.

### DIFF
--- a/src/zen/workspaces/ZenWorkspaces.mjs
+++ b/src/zen/workspaces/ZenWorkspaces.mjs
@@ -466,6 +466,10 @@ var ZenWorkspaces = new (class extends ZenMultiWindowFeature {
       this._swipeState.direction = delta > 0 ? 'left' : 'right';
     }
 
+    if (Math.abs(translateX) >= 40) {
+      return this._handleSwipeEnd(event);
+    }
+
     // Apply a translateX to the tab strip to give the user feedback on the swipe
     const currentWorkspace = this.activeWorkspace;
     this._organizeWorkspaceStripLocations({ uuid: currentWorkspace }, true, translateX);


mr-cheffy:
Loving these! Thanks a lot, me and trackpads aren't good friends

mr-cheffy (approved):

mr-cheffy:
Hmm... I was trying it out and it ended up being a bit weird. It changes workspace even before I finished my scrolling, sorry. But I feel like this is the right path

mathieutu:
>before I finished my scrolling

What do you mean? Could you record it? (It feels pretty right on Mac)

mathieutu:
Arc way to handle it is way better, actually.
They switch workspaces with just a little "swipe tap", but let you keep the swipe and follow it if you keep it:

https://github.com/user-attachments/assets/2ad7ac81-a921-4b6f-ae19-e9823b0cab86

I'll investigate, so see if I can manage to make something similar.


# Edit:
I've managed to have something cool using "MozSwipeGestureEnd". However, I need to handle the animation. Any hint why the workspace goes _back_?

https://github.com/user-attachments/assets/02ea3f70-3548-4b1e-a254-401ef548a176

I'm trying with the wheel event instead of swipe. (I'd fix another issue: not being able to use the scroll and thumb wheel to change workspaces too)

mr-cheffy:
The issue is that it changes workspace BEFORE I release my fingers. No, im not sure why it goes back

joshstobbs:
I saw this PR was reverted pretty quickly after being merged.  What is the work that needs to be done to push this over the line?

mathieutu:
@joshstobbs If you read the previous messages, you'll see that the behavior was not liked. You also have a video of the desired behavior (Arc). Feel free to play with the tabs to see what you can achieve.

Issue #7716, opened 2025-04-18
Zen seems to crash randomly when switching tabs


skoved:
### Captchas

- [x] I have read the instructions.
- [x] I have searched existing issues and avoided creating duplicates.
- [x] I am not filing an enhancement request.
- [x] I have checked that this issue cannot be reproduced on Mozilla Firefox.
- [x] I have checked that this issue can be reproduced once I removed all my Mods and Custom CSS.

### What happened?

Occasionally when I go to open a new tab with ctrl+t and type in a url, zen will freeze as it create the new and load the page. Tab seems to get created but the page does not load at all. zen completely locks up and after 15-30 seconds, zen crashes completely. This start after updating to 1.11.3b and has happened at least once a day

### Version

1.11.3b (Firefox 137.0.1) (64-bit)

### What platform are you seeing the problem on?

Linux (Tarball)

### What component is this issue related to?

Other

### Relevant log output if applicable

```shell
Apr 16 09:46:07 thinkpad zen-browser[12848]: libva info: VA-API version 1.22.0
Apr 16 09:46:07 thinkpad zen-browser[12848]: libva info: Trying to open /usr/lib64/dri-nonfree/radeonsi_drv_video.so
Apr 16 09:46:07 thinkpad zen-browser[12848]: libva info: Trying to open /usr/lib64/dri-freeworld/radeonsi_drv_video.so
Apr 16 09:46:07 thinkpad zen-browser[12848]: libva info: Found init function __vaDriverInit_1_22
Apr 16 09:46:07 thinkpad zen-browser[12848]: libva info: va_openDriver() returns 0
Apr 16 10:38:26 thinkpad zen-browser[17865]: Exiting due to channel error.
Apr 16 10:38:26 thinkpad zen-browser[17997]: Exiting due to channel error.
Apr 16 10:38:26 thinkpad zen-browser[13476]: Exiting due to channel error.
Apr 16 10:38:26 thinkpad zen-browser[17823]: Exiting due to channel error.
Apr 16 10:38:26 thinkpad zen-browser[17926]: Exiting due to channel error.
Apr 16 10:38:26 thinkpad zen-browser[14206]: Exiting due to channel error.
Apr 16 10:38:26 thinkpad zen-browser[17926]: Crash Annotation GraphicsCriticalError: |[C0][GFX1-]: CompositorBridgeChild receives IPC close with reason=AbnormalShutdown (t=112.826)
[...]

PR #7761, opened 2025-04-20
Fix tab switching crash when creating new tabs (Issue #7716)

mbergo:
Fix tab switching crash when creating new tabs (Issue #7716)

    Problem:
    The Zen browser was experiencing random crashes when switching tabs, particularly
    when creating a new tab with Ctrl+T and typing a URL. The crash logs showed
    "CompositorBridgeChild receives IPC close with reason=AbnormalShutdown" errors,
    indicating issues with the graphics compositor during tab switching operations.
    This was caused by race conditions in the tab switching code and lack of proper
    error handling when dealing with browser elements that might be in an invalid state.

    Solution:
    Instead of just catching errors, we've implemented a more robust solution that addresses
    the root causes of the race conditions in the tab switching code:

    1. State Management System:
       - Added a queue-based system for tab operations to ensure they execute in sequence
       - Implemented debouncing to prevent rapid tab switching that could cause race conditions
       - Added state validation before operations to ensure browser elements are in a valid state

    2. Asynchronous Operation Handling:
       - Used Promises and async/await for proper sequencing of operations
       - Added proper timing controls with small delays to ensure DOM is ready
       - Implemented a mutex-like approach to prevent concurrent workspace changes

    3. Robust Tab State Validation:
       - Added comprehensive checks for tab validity before operations
       - Implemented proper cleanup of resources to prevent memory leaks
       - Added validation for browser window state to prevent operations on closed windows

    4. Improved Error Recovery:
       - Added fallback mechanisms when operations fail
       - Implemented proper cleanup of state even when errors occur
       - Added more informative error messages for better debugging

These changes make the browser much more robust when handling tab switching operations,
    preventing the crashes reported in issue #7716.

### DIFF
--- a/src/zen/common/ZenUIManager.mjs
+++ b/src/zen/common/ZenUIManager.mjs
@@ -175,25 +175,136 @@ var gZenUIManager = {
   _clearTimeout: null,
   _lastTab: null,
 
+  // Track tab switching state to prevent race conditions
+  _tabSwitchState: {
+    inProgress: false,
+    lastSwitchTime: 0,
+    debounceTime: 100, // ms to wait between tab switches
+    queue: [],
+    processingQueue: false,
+  },
+
+  // Queue tab switch operations to prevent race conditions
+  async _queueTabOperation(operation) {
+    // Add operation to queue
+    this._tabSwitchState.queue.push(operation);
+
+    // If already processing queue, just return
+    if (this._tabSwitchState.processingQueue) {
+      return;
+    }
+
+    // Start processing queue
+    this._tabSwitchState.processingQueue = true;
+
+    try {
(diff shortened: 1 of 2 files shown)


mr-cheffy:
Shoudnt we rather tackle down the race condition rather than adding error handlers?

mbergo:
> Shoudnt we rather tackle down the race condition rather than adding error handlers?

Yes, sorry about it. I will get back to you with a better proposal. Thanks @mauro-balades

mbergo:
@mauro-balades force rebased with updated description, now we order things. let me know

mr-cheffy (approved):
Looks good! Have you tested it?

PR #7796, opened 2025-04-21
Feature: Add checks for sound before auto toggling PiP

cbackas:
Coming from Arc, I really liked using auto PiP. Enabling it in Zen has been a bit annoying because even muted videos on web pages (such as the muted stream on twitch's home page) are activating auto PiP when you change tabs. 

I've added a simple patch in the `actors/PictureInPictureChild.sys.mjs` file that checks if the video player itself is muted and if it is it does not activate PiP when you switch away from that tab.

I then added another small patch in `components/PictureInPicture.sys.mjs` which checks whether the tab itself is muted and prevents auto PiP if it is in fact muted. 

Maybe this behavior should be behind a preference?

### DIFF
--- a/src/toolkit/components/pictureinpicture/PictureInPicture-sys-mjs.patch
+++ b/src/toolkit/components/pictureinpicture/PictureInPicture-sys-mjs.patch
@@ -1,8 +1,18 @@
 diff --git a/toolkit/components/pictureinpicture/PictureInPicture.sys.mjs b/toolkit/components/pictureinpicture/PictureInPicture.sys.mjs
-index 5da0404b2672ba8cce7bcf808bf2373474776654..3b93217b38f25f54d7ef44d151e314bc1c5e5ce3 100644
+index 5da0404b2672ba8cce7bcf808bf2373474776654..6a22fce1212bd6ac6bdd4962b8247c3db956f0eb 100644
 --- a/toolkit/components/pictureinpicture/PictureInPicture.sys.mjs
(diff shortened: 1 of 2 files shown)


src/toolkit/components/pictureinpicture/PictureInPicture-sys-mjs.patch
```diff
@@ -1,8 +1,18 @@
...
+index 5da0404b2672ba8cce7bcf808bf2373474776654..6a22fce1212bd6ac6bdd4962b8247c3db956f0eb 100644
 --- a/toolkit/components/pictureinpicture/PictureInPicture.sys.mjs
 +++ b/toolkit/components/pictureinpicture/PictureInPicture.sys.mjs
-@@ -494,7 +494,7 @@ export var PictureInPicture = {
+@@ -126,6 +126,9 @@ export class PictureInPictureToggleParent extends JSWindowActorParent {
+         if (browser.ownerGlobal.gBrowser.selectedBrowser == browser) {
+           break;
+         }
++        if (browser.ownerGlobal.gBrowser.getTabForBrowser(browser).muted) {
++          break;
```
Slowlife01, line 10: There should be `browser.audioMuted`?

cbackas: I'll try that out, it took me a while to figure out the way I did it, I wasn't able to find docs about this so I just stumbled through reading other code and looking at objects

PR #7832, opened 2025-04-23
Designate opacity status of urlbar in compact mode to fix #7711

### LINKED ISSUE #7711: URL bar flashes when exiting fullscreen mode
### Captchas

### What happened?

the whole sidebar used to peek out when exiting fullscreen mode to compact mode -- this was a bug that was fixed, now everything seems to be working fine except the URL bar pops in and out with nothing else

### Version

1.11.4b

Windows - x64

Compact Mode

```shell

```

sporocyst:
Somehow, the urlbar remains visible even when `#navigator-toolbox` has `opacity: 0` in single toolbar mode
This leads to situation described in #7711

This PR designates `opacity` value specifically for urlbar, so it is now visible/hidden at right timing.

### DIFF
--- a/src/zen/compact-mode/zen-compact-mode.css
+++ b/src/zen/compact-mode/zen-compact-mode.css
@@ -66,6 +66,10 @@
             height: calc(100% - var(--zen-element-separation));
           }
 
+          & #urlbar {
+            opacity: 0;
+          }
+
           & #zen-sidebar-top-buttons {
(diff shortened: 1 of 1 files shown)


mr-cheffy:
This will make the urlbar disappear when in compact mode and clicking on ctrl+T...

sporocyst:
True.. since I've never used ctrl+T.
Ok, this is a tricky problem.

PR #7846, opened 2025-04-24
Homepage button on Zen Mods settings page.

Pdzly:
I added that the settings button has a pointer on hover to better indicate that it is clickable.

grafik


mr-cheffy (approved):

Pdzly:
@mauro-balades sry to ping but i have a question, i tried to add the readme to it too but i cant seemingly find out the path of the fitting icon ( i would have used the "info.svg" from the zen icons, but i couldnt get it to work). What is the procedere? What is the path of the icons?

PR #7901, opened 2025-04-27
Better privacy and less connections on startup

### LINKED ISSUE #7000: Unwanted Connections, Transparency and Default Search Engine
### Captchas

### What happened?

<a name="unwanted-connections"></a>This issue is a continuation of the previously locked issue #5947 and relates to the following closed issues #6787, #3417 and #236. It has been created per the suggestion of @mauro-balades in the related discussions about privacy [here](https://github.com/zen-browser/desktop/discussions/5907#discussioncomment-12589085).

## 1. Unwanted connections

When the browser is first run right after installation, it will attempt to communicate with various tracking and telemetry endpoints without user consent. There are several DNS queries made, including requests to domains that seem unnecessary and could potentially compromise user privacy. These domains include (but are not limited to):

```
incoming.telemetry.mozilla.org
location.services.mozilla.com
firefox.settings.services.mozilla.com
detectportal.firefox.com
merino.services.mozilla.com
push.services.mozilla.com
services.addons.mozilla.org
zen-browser.app
updates.zen-browser.app
example.org
ipv4only.arpa
www.google.com
s2.googleusercontent.com
calendar.google.com
trello.com
web.whatsapp.com
www.notion.so
www.youtube.com
github.githubassets.com
a.slack-edge.com
abs.twimg.com
```

Ideally, a browser should only make outgoing connections necessary for functionality. Privacy should matter even when no telemetry data is transmitted, the requests will still expose critical metadata, such as IP addresses, user-agent strings and request timings, which can be used to track, profile and infer browsing habits. This _metadata leakage_, which occurs with plain HTTP requests, can compromise user privacy by revealing geographic locations, identifying devices and associating users with specific services, even in the absence of directly sending telemetry data.

The suggestion here is to minimize or remove all (if possible) of any outgoing connections when the browser is first run, as it is done in other privacy focused browsers such as:

```
Tor Browser
Mullvad Browser
Ungoogled-Chromium
```

During the onboarding process, despite deselecting the 3 essential tabs (_Discord_, _Google Calendar_ and _Notion_) that are selected by default, it will _always_ connect to these 4 different web pages upon completion of setup:

```cpp
1. https://zen-browser.app/welcome/
2. https://zen-browser.app/privacy-policy/
3. https://x.com/zen_browser
4. https://www.reddit.com/r/zen_browser
```

I'm not sure why we can't prevent these from connecting automatically during the first run.

<a name="privacy-concerns-by-celenityy"></a>Also, when attempting to browse a website and the DNS lookup fails to resolve the domain, I'm not sure why (haven't looked into it), but a DNS request will be made to:

```
support.mozilla.org
```

## 2. Privacy Concerns by Celenityy

In [pull/927](../pull/927#issuecomment-2745338453), @celenityy raised significant concerns about the privacy and security of Zen Browser, stating the following:

> Zen Browser still has major [privacy and security deficiencies](https://codeberg.org/celenity/Phoenix/wiki/Comparison) when compared to other projects of a similar nature, and it's even less private and secure than stock Firefox. For instance, you [exclude random social trackers from ETP](https://github.com/zen-browser/desktop/blob/109eb96c089607de1bb88e479a3fa6539bcfb415/src/browser/app/profile/zen-browser.js#L280) and [allow unsigned extensions by default](https://github.com/zen-browser/desktop/blob/109eb96c089607de1bb88e479a3fa6539bcfb415/src/browser/app/profile/zen-browser.js#L258).

While only one of the issues mentioned was addressed in commit [bd5d3c0](https://github.com/zen-browser/desktop/commit/bd5d3c00f1e38bffd5689aebf09453a2aba4e811#diff-75081f01057ca31220ccb991c4999f13917b64961a878527cf848bca2e22d1e0), there are other remaining privacy issues. These can be observed by @celenityy's [Comparison Table](https://codeberg.org/celenity/Phoenix/wiki/Comparison), which outlines the default privacy and security configurations of various browsers, including Zen Browser.

Users should be provided with an option to select predefined configurations during the onboarding process, allowing them to choose between different privacy levels (e.g., strict, moderate, or basic). This would enable users to make an informed choice regarding their browsing privacy without sacrificing too much convenience.

However, I understand that certain privacy changes might not be as simple as toggling preferences. Some might require deeper changes to the underlying browser features and could impact the browsing experience in ways that users need to be made aware of in the process of selecting predefined configurations.

Addressing these privacy concerns in a balanced way would make Zen Browser a more secure and privacy-respecting choice for users, without causing major disruptions to normal web browsing. We should aim to incorporate changes that align with privacy best practices while keeping usability in mind. You can still maintain the current privacy and security defaults as "_basic level_", but don't deprive users of the ability to make their own choices regarding privacy. Providing an option to select between predefined configurations, ranging from _basic_ to _stricter_ privacy during onboarding would go a long way. These choices should clearly inform users that _stricter_ settings may disable some browser features, impact browsing experience or break certain websites, as a tradeoff for improved privacy.

<a name="privacy-policy-and-faq"></a>It's actually best to assume that users may not have the necessary knowledge or incentives to make the best privacy choices for themselves, so always try to default to improved privacy settings and give clear, understandable options for users to make informed decisions. Providing users with informed choices is essential to building trust and ensuring transparency.

## 3. Privacy Policy and FAQ Transparency

<a name="default-search-engine"></a>Update the _Privacy Policy_ or include an _FAQ_ that discloses all outgoing connections made without user interaction, what data is sent (if any) and its purpose. This should also include Zen Browser's landing page that contains the `Cloudflare Web Analytics` JavaScript beacon for tracking visitors, since the web page is always opened in the browser after first run. This could be similar to LibreWolf's FAQ "[_Does LibreWolf make any outgoing connections?_](https://librewolf.net/docs/faq/#does-librewolf-make-any-outgoing-connections)", but with more detailed information and explicit disclosure of all background connections made without user interaction.

## 4. Bring Back Default Search Engine Selection

In the version `1.7.6b` of Zen Browser, users had the option to select a default search engine during the onboarding process. In version [1.8b](https://github.com/zen-browser/desktop/commit/425d0ea95151fba24eba03bc04276276df5de021) this was removed and I'm not sure why. This change seems like a step backwards in regards to privacy and user choice as well, it engages in anti-competitive behavior by favoring one service over others. If the goal is to advocate privacy, it would be more consistent to default to a privacy-focused search engine, while still allowing users to change it if they prefer. This would send a clearer message of prioritizing privacy right from the start.

Image

### Version

1.11.1b

Linux (AppImage), Linux (Flatpak), Linux (Tarball), macOS - aarch64, macOS - Intel, Windows - x64, Windows - aarch64, Other

Privacy

```shell

```

mr-cheffy:


g3eorge:
Might wanna update your privacy policy after this to make it more complete

PR #7958, opened 2025-05-01
Oklch to Oklab (even better gradients)

BlueFox1616:
Not 100% sure it will work as intended but it should give way better gradients

### DIFF
--- a/src/zen/workspaces/ZenGradientGenerator.mjs
+++ b/src/zen/workspaces/ZenGradientGenerator.mjs
@@ -909,12 +909,12 @@
       } else if (themedColors.length === 1) {
         return this.getSingleRGBColor(themedColors[0], forToolbar);
       } else if (themedColors.length !== 3) {
-        return `linear-gradient(in oklch ${this.currentRotation}deg, ${themedColors.map((color) => this.getSingleRGBColor(color, forToolbar)).join(', ')})`;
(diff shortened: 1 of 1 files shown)


mr-cheffy (approved):
"100% sure"?

BlueFox1616:
yeah i meant im not sure i didnt add a period . I meant :  Im not 100% sure. If it works ......

BlueFox1616:
so ig that means it doesn't work and needs further implementation @mauro-balades  ?

mr-cheffy:
I think oklhc works good enough

BlueFox1616:
@mauro-balades  There is a hugeeee difference Arc uses oklab and thats how i noticed there gradients look better . Its actually crazy how much better it is if you play around with the color interpolations you can see the difference https://colorbeta.com/#t=0;0,1&c=fd30c3ff,0;89f759ff,100&i=3 . If it is actually a matter of just changing oklch to oklab it would be a huge improvement . Pls just consider it. I too thought it was the same thats why i added oklch at the start

mr-cheffy:
@BlueFox1616 oklab doesnt seem to play really well with firefox

image

BlueFox1616:
@mauro-balades it could be a dark mode issue maybe . I've noticed it adds a dark shade to gradients. It is well supported it seems like something is interfering with the gradient . It doesn't use R G B so it might not get translated well . It might be a conversion issue it uses three color-space coordinates L, a, and b.
L – perceived lightness

a – how green/red the color is

b – how blue/yellow the color is

Issue #7973, opened 2025-05-02
[Twilight 1.12] Compact mode sidebar twitches when hovering


gordonn:
### Captchas

- [x] I have read the instructions.
- [x] I have searched existing issues and avoided creating duplicates.
- [x] I am not filing an enhancement request.
- [x] I have checked that this issue cannot be reproduced on Mozilla Firefox.
- [x] I have checked that this issue can be reproduced once I removed all my Mods and Custom CSS.

### What happened?

https://github.com/user-attachments/assets/c8472830-4117-4c42-b02f-de48d58cdd97

Also hovering near the edge causes the sidebar to spasm:
https://github.com/user-attachments/assets/8a3440b8-dc09-49fc-9bf0-47965cebcc84

### Version

1.12t

### What platform are you seeing the problem on?

Windows - x64

### What component is this issue related to?

Compact Mode

### Relevant log output if applicable

```shell

```

sporocyst:
To replicate, you can
1. In maximized window, move mouse quickly and directly to left/right screen border.
2. In normal window, approach left/right window border from outside slowly until the tip of mouse is exactly on the border

It seems that the `zen-has-hover` attribute is released too quickly or too early

Edit: set `zen.theme.content-element-separation` to 0 may be easier to trigger

1.12t https://github.com/zen-browser/desktop/actions/runs/14785205338

gordonn:
Also hovering near the edge causes the sidebar to spasm:
https://github.com/user-attachments/assets/8a3440b8-dc09-49fc-9bf0-47965cebcc84

gordonn:
Another vid:

https://github.com/user-attachments/assets/4c8217ed-c7e8-4286-91dc-5fbc6eaf1445

mr-cheffy:
Could you please ping me if it still happens once https://github.com/zen-browser/desktop/actions/runs/14791350412 is out?

gordonn:
@mauro-balades still have the issues unfortunately 

https://github.com/user-attachments/assets/eff5a6b3-5474-4f0d-992d-4e3498867c70

https://github.com/user-attachments/assets/b5b9694d-6d2f-4f74-8f6d-6537b5ba31d8

mr-cheffy:
True, i've tested on my machine, new twilight will have this 100% fixed. Thanks!

Issue #7950, opened 2025-05-01
Compact mode - 1 pixel line on sidebar


Anthonyy232:
### Captchas

- [x] I have read the instructions.
- [x] I have searched existing issues and avoided creating duplicates.
- [x] I am not filing an enhancement request.
- [x] I have checked that this issue cannot be reproduced on Mozilla Firefox.
- [x] I have checked that this issue can be reproduced once I removed all my Mods and Custom CSS.

### What happened?

Seems like there has been a regression recently. There is now a ~1 pix line on the side that the compact mode sidebar is set on.
https://github.com/user-attachments/assets/2eeac909-3829-44f5-b420-6488b1bcd86f

### Version

1.12t

### What platform are you seeing the problem on?

Windows - x64

### What component is this issue related to?

Compact Mode

### Relevant log output if applicable

```shell

```

Anthonyy232:
Tested with all mods off, content separation = 0

sporocyst:
Can confirm. Before 1.12t, tab sidebar has opacity: 0 when tucked. Now it remains visible always, so a 1px width intended to detect mouse hovering can be seen.

mr-cheffy:
Thanks @Anthonyy232, fixed!

Anthonyy232:
@mauro-balades It is visually gone, but the sidebar interaction is broken as far as I can tell. It works if the sidebar is on the right side, but NOT the left side. Setting content separation back to 8 (default) doesn't fix it.

https://github.com/user-attachments/assets/4aed36a1-aee0-405f-85cc-f6b7cff3ac1a

@sporocyst Can you confirm/deny?

Using latest Twilight (just updated), no mods, content

Anthonyy232:
Just did some more testing, the sidebar only opened on the right side because I have a monitor setup to be a right side monitor. If I changed it to the left side, then the sidebar would open on the left side. If I unplugged/disabled it, neither side works. Seems like it's another hitbox thing? Maybe this old thread would be relevant, such as using 0.1 instead of 1pix
https://github.com/zen-browser/desktop/issues/7694

mr-cheffy:
Right, we can do the same trick we did with the top toolbar of adding 0.01px

PR #7981, opened 2025-05-02
Add left paading to right tab sidebar in compact mode to avoid regression of #7973

sporocyst:
The added padding would be the unhided area of sidebar and is invisible since `#navigator-toolbox` has a transparent background.

Regression seen in 1.12t   20250502102900. Note the right orange line
未命名


sporocyst:
I just realized i mentioned wrong issue, it is actually #7950 . The patch is right, but I will close this

PR #7989, opened 2025-05-03
fix: use list-style rather than menu for appMenu items

alvinsjoy:
When using `--menu-image` the list appeared cluttered.
Before:
image


Shockshwat:
where is the after?

mr-cheffy (approved):

mr-cheffy:
> where is the after?

image

PR #8168, opened 2025-05-08
Fix toast alignment in right sidebar

ChenghaoMou:
This changes how the toast is aligned when sidebar is on the right.

Previously, it was out of view.


mr-cheffy:
Sorry! But i've already fixed this a long time ago and it's now on stable... But thanks for trying!

PR #8253, opened 2025-05-12
Add patch fix #7701

### LINKED ISSUE #7701: about:protections advertises "Zen for Android and iOS," links to Firefox
### Captchas

- [ ] I am not filing an enhancement request.

### What happened?

Repeat of #4445, but also on macOS and also marked as stale? So you can close this if you want because I don't know how GitHub "manners" works, but yeah anyways:

Image

This should be an easy fix? If you click on them it links to Firefox for mobile (app store or google play based on which you picked). Even if you _are_ going to make a Zen for iOS and Android (which would be awesome), it still says "Zen Browser Browser" which is wrong. 

### Version

1.11.3b

macOS - aarch64

Other

```shell

```

SO9010:
This fixes #7701 by just setting it to removed in the preferences automatically, meaning that the users will never see it. This made more sense than having it rewritten just for this page. But if we think linking to Firefox mobile here is necessary, please let me know.


mr-cheffy:
Why cant we just set `browser.contentblocking.report.show_mobile_app` to false and locked in the preferences file in order to avoid extra-patches?

SO9010:
That is true! I'll change that now :)

SO9010:
Done :)

SO9010:
Ok finished it now, it took me longer than it should to figure out that I just needed to add locked to the end of it to make it well locked! I am still very new to this 😂

mr-cheffy:
@SO9010 it can just be added here: https://github.com/zen-browser/desktop/blob/dev/src/browser/app/profile/disablemozilla.inc

to avoid patches

SO9010:
Cool thank you so much, I hope that is correct now :)

mr-cheffy (approved):

Issue #7810, opened 2025-04-22
No keyboard shortcut assigned, yet still error 'conflict with another shortcut'


Stevenpolak:
### Captchas

- [x] I have read the instructions.
- [x] I have searched existing issues and avoided creating duplicates.
- [ ] I am not filing an enhancement request.
- [ ] I have checked that this issue cannot be reproduced on Mozilla Firefox.
- [x] I have checked that this issue can be reproduced once I removed all my Mods and Custom CSS.

### What happened?

I played around with keyboard shortcuts, but now even when no keyboard shortcut is assigned, I still see the message "conflict with another shortcut"

![Image](https://github.com/user-attachments/assets/dc73323e-ed09-4dec-89f6-1f5356e1b0d6)

### Version

1.11.2b (Firefox 137.0.1)

### What platform are you seeing the problem on?

Windows - x64

### What component is this issue related to?

Settings

### Relevant log output if applicable

```shell

```

mr-cheffy:
Did you save it correctly? Could you send a video reproducing it?

Stevenpolak:
https://github.com/user-attachments/assets/17083584-e65b-4e24-92ca-74ba716f664c

I confirm the edit by pressing the escape key. Alt+1 is obviously in use by another, standard is on workspace switch.
After restarting the settings tab (same Zen window, just closing the tab and reopening it) the confict message is gone.

SO9010:
I can confirm this issue too on nightly on Arch Linux. It goes if you enter a correct value, then unset it.

SO9010:
This is completed now :)

PR #8256, opened 2025-05-12
Add fix for #7810

SO9010:
To fix this, I just had to add a small line of code which takes care of removing the error message when the user deletes the key binding.

I also added some fixes to small visual bugs, for example, when clicking escape on anything, even when not changed from unset, it would give the impression that it has changed. I also changed it so that when you close and reopen, if the shortcut has not been set, it will have the name 'Not set' as before; it would have the unset changed name in grey if that makes sense.


mr-cheffy (approved):
We should write tests for these haha

PR #8261, opened 2025-05-12
Fix for #7910

### LINKED ISSUE #7910: When using Tabs on the Right, the Glance controls are on the wrong side
### Captchas

### What happened?

To my understanding, Glance controls were moved from the right side of the screen to the left side of the screen to make them closer to the tab bar. However, when using tabs on the right, the glance controls stay on the left side - to make the behavior consistent, they should also move to the right side. As of right now, it feels strange having them on the left when no other browser controls are over there.

### Version

1.11.5b

Linux (Tarball)

Glance

```shell

```

SO9010:
This fixes #7910 by adding a conditional check for whether the sidebar is on the right and adjusting the CSS accordingly.

### DIFF
--- a/src/zen/glance/ZenGlanceManager.mjs
+++ b/src/zen/glance/ZenGlanceManager.mjs
@@ -50,6 +50,10 @@
       });
     }
 
+    #areTabsOnRightSide() {
+      return Services.prefs.getBoolPref('zen.tabs.vertical.right-side');
+    }
+
     get #currentBrowser() {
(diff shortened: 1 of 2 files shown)


mr-cheffy (approved):
Thanks a lot for these fixes!

SO9010:
Of course! I love zen and I would like to contribute to a larger and larger scale to the project over the summer time (I have an unreasonable amount of free time now!).

Issue #8341, opened 2025-05-15
Cant Add more Essential tabs


magicman3211:
### Captchas

- [x] I have read the instructions.
- [x] I have searched existing issues and avoided creating duplicates.
- [x] I am not filing an enhancement request.
- [x] I have checked that this issue cannot be reproduced on Mozilla Firefox.
- [x] I have checked that this issue can be reproduced once I removed all my Mods and Custom CSS.

### What happened?

Last couple of updates, I no longer have the ability to add a tab to essentials tabs
Only way I can add is to drag a tab there, but it doesnt stay there
on AUR (Arch) version of Zen

### Version

1.12.5b

### What platform are you seeing the problem on?

Other

### What component is this issue related to?

Other

### Relevant log output if applicable

```shell

```

Issue #8315, opened 2025-05-14
Unable to add tabs to essentials after most recent update


Alex-Wilson:
### Captchas

- [x] I have read the instructions.
- [x] I have searched existing issues and avoided creating duplicates.
- [x] I am not filing an enhancement request.
- [x] I have checked that this issue cannot be reproduced on Mozilla Firefox.
- [x] I have checked that this issue can be reproduced once I removed all my Mods and Custom CSS.

### What happened?

After updating to Version 1.12.4b (64 bit), I can no longer add tabs to Essentials. I was able to reproduce the issue by downloading Zen on another user account and attempting to add any tab to Essentials. The problem does not appear to be related to containers, pins, or tabs as these functionalities all work as normal. Only the Essentials section is affected. Both drag and drop and the context menu option “Add to Essentials” fail to work.

### Version

Version 1.12.4b (64-bit)

### What platform are you seeing the problem on?

Windows - x64

### What component is this issue related to?

Other

### Relevant log output if applicable

```shell

```

Issue #8308, opened 2025-05-14
Adding new essentials is not working


mowael01:
### Captchas

- [x] I have read the instructions.
- [x] I have searched existing issues and avoided creating duplicates.
- [x] I am not filing an enhancement request.
- [x] I have checked that this issue cannot be reproduced on Mozilla Firefox.
- [x] I have checked that this issue can be reproduced once I removed all my Mods and Custom CSS.

### What happened?

I just tried adding some new websites to my essentials, but they didn't add. It just works when removing essentials. I tried restarting the browser and the troubleshooting mode, but nothing worked. 

### Version

1.12.4b

### What platform are you seeing the problem on?

Linux (Tarball)

### What component is this issue related to?

Customizable UI / Toolbars

### Relevant log output if applicable

```shell

```

henry-fisher:
Just want to add that this is an issue on macOS as well.

Propheticus:
To cross reference with other reports: have you got 12 essentials or more?

henry-fisher:
Yes, currently 17 and cannot add the 18th

MagicCD:
I also encountered this problem. [Add to Essentials" Area Tab Limit (Max 12 Tabs)?](https://github.com/zen-browser/desktop/discussions/8238) Very annoying🥲

Propheticus:
@mauro-balades is the 12 essentials limit done on purpose or do we have a bug here?

mr-cheffy:
It's on purpose

magicman3211:
I have 32 essential tabs atm, but cant add more

blakegearin:
What is the argument for removing support for more than 12 essential tabs? Given the amount of threads generated by this change it appears more than 12 is a common use case.

henry-fisher:
Yeah not to be dramatic, but my entire use-case for Zen depended on using the essential section with 12+ services. Now I have 17 and am using a pinned tab for the new tab I couldn't add, which is not a long term solution. Would be great to hear the reasoning for this, or suggested alternative workflows. Otherwise I'll be migrating away.

Regardless, the lack of user feedback is a critical issue. The reason people are opening issues about this and reporting it as a bug is because there's no reason to think it isn't. When a user tries to add more than 12, there should be a warning that tells the user more than 12 isn't supported. No?

cfuendev:
@Propheticus 
> is the 12 essentials limit done on purpose or do we have a bug here?

@mauro-balades
> It's on purpose

[...]

Issue #107, opened 2024-07-30
Zen won't start - Couldn't load XPCOM


burnout426:
### What happened?

Windows 11 23H2 64-bit + Zen 1.0.0-a.8 installed with "zen.installer.exe" from <https://github.com/zen-browser/desktop/releases/tag/1.0.0-a.8>. Also tried zen.win64.zip.

When I try to start Zen, I get the attached error dialog titled "Firefox" that says "Couldn't load XPCOM."

![zenxpcom](https://github.com/user-attachments/assets/f04f7ad5-17d1-45d2-90f0-ac5b15581590)

CPU: Intel Core i5 2520m: <https://www.intel.com/content/www/us/en/products/sku/52229/intel-core-i52520m-processor-3m-cache-up-to-3-20-ghz/specifications.html>. See "Extensions and technologies" at <https://www.cpu-world.com/CPUs/Core_i5/Intel-Core%20i5%20Mobile%20I5-2520M%20FF8062700840017%20(BX80627I52520M).html> for list of supported features ( SSE and AVX version support). 8GB RAM. Intel HD3000 GPU on CPU.

Firefox 128.0.3 (64-bit) loads fine.

### Reproducible?

- [X] I have checked that this issue cannot be reproduced on Mozilla Firefox.

### Version

1.0.0-a.8

### Severity impact

Critical

### What platform are you seeing the problem on?

Windows

A-Lamia:
Have the same issue.

detyso:
I have the same error xpcom error

nero-0:
It doesn't open here either

mauro-balades:
Just to be clear, did it open on the previous versions?

detyso:
Yea it open when I installed the first release of and  there was no error

nero-0:
For me, no version after 1.0.0-a.4 opens

burnout426:
Same here. a.4 is the last build that works for me.

a.6 and a.7 give me "The application was unable to start correctly (0xc0000142)".

a.8 gives me the XPCOM error.

In the changelog for a.6, it mentions "Updated CPU requirements for x86_64-v3", which I suspect is the culprit. But, I couldn't find what the requirements were before or what they are now. That's why I provided my CPU specs though.

mauro-balades:
We are trying to solve it for 1.0.0-a.9

mauro-balades:
[...]

PR #8441, opened 2025-05-20
feat: Add config to support more than 12 essential tabs

blakegearin:
## Summary

- Added `zen.tabs.essentials.max` preference with a default of `12`

- Added `zen.tabs.essentials.height` preference with a default of `0` (height expands to fit)

## Screenshots

- Higher max + default height = height expands to fit (legacy behavior)

  image

- Higher max + lower height = scrolls

  image

  image

- Higher max + higher height = expands

  image

## Notes

- Zen supported more than 12 essential tabs in version `1.12.2b` and before leading to a number of threads

  - #8238

  - #8308 

  - #8315

  - #8341


mr-cheffy:
How does it handle overflows of the window?

hatimthayyil:
Do you mean in collapsed mode? Essentials overflowing downwards?

univalis:
Is it possible to re-open the pull-request ?

mr-cheffy:
I'll reopen it but won't be merged until fundamental issues are solved.

1. Issue with overflowing over the window when there are too many
2. Issue with wrong usage of essentials. (People misusing this feature as pinned tabs rather than highly important apps)

shaqaruden:
> I'll reopen it but won't be merged until fundamental issues are solved.
> 
>     1. Issue with overflowing over the window when there are too many
> 
>     2. Issue with wrong usage of essentials. (People misusing this feature as pinned tabs rather than highly important apps)

The more I think about this the more I am leaning in your direction. I guess my main issue was with closing the PR, what I felt was too early. Which I know "what is too early?" is different for everyone. Is there an explanation in the docs that defines the intended use of each kind of tabs? Reason I ask is, "Essential Tabs" has a different meaning to different people and the delineation between Essential and Pinned tabs is a little blurry at least.

Maybe what we need here is a clearer definition of the intended use for each in the docs (if it does not already exist. I could not find it myself) and not to allow a configurable amount of Essential Tabs.

If the look is what people are after there is the Super Pins Mod which can give them that.

mr-cheffy:
> Is there an explanation in the docs that defines the intended use of each kind of tabs

not in the docs, but in the future, im planning on making essentials not being able to be unloaded, only very explicitly through the context menu. Why? Because essentials are supposed to be apps that are "essential", always running on the background. You do not need 38 apps running constantly on the background. If you want them to always stay there, that's the purpose of pinned tabs.

ErdemGKSL:
btw you should probably do it a getter, or assign listener for it. otherwise the changes you do will not be shown on the fly i guess, probably you would need to restart the browser to see the changes.

shaqaruden:
> > Is there an explanation in the docs that defines the intended use of each kind of tabs
> 
> 
> 
> not in the docs, but in the future, im planning on making essentials not being able to be unloaded, only very explicitly through the context menu. Why? Because essentials are supposed to be apps that are "essential", always running on the background. You do not need 38 apps running constantly on the background. If you want them to always stay there, that's the purpose of pinned tabs. 

Ahhh okay. Knowing that I would have to fully agree with you on this. 

@blakegearin any thoughts?

blakegearin:
Sorry for the delay. I was out of town and then came back to some build issues (XPCOM, mentioned in #107). So I'm going to try on a different machine to examine overflow more closely. I don't use collapsed mode personally.

> You do not need 38 apps running constantly on the background.

I don't think a browser should be prescriptive in this domain. People have lots of tabs open, and they're going to organize them in a way that makes the most sense to them. Maybe out there, somewhere, a security guard is using Zen to flip through 20 cameras 😄

The information density of essentials compared to pinned tabs can be as high as 4 to 1. That's a large difference. I don't think 38 pinned tabs can be displayed in one view without scroll. It's not obvious how essential and pinned tabs are different besides their densities. Even if it was well-documented, I presume most users who rub up against this limit aren't going to read it to understand the philosophical differences between essentials and pins. They're going to report a bug. Recent examples: #8689, #8900

While not a perfect comparison, I'm able to get 18 extensions pinned to the toolbar before space runs out. There is no explicit, hard limit as far as I can tell, and that's a good thing. Customization is important, and many extensions run constantly.

I understand there is a maintenance aspect to this, but I haven't heard any specifics on what makes 12 the best "magic number" as compared to 16, 20, or 24. Personally I don't use more than 24.

hatimthayyil:
The Essentials tab is useful for having easy and quick access to a frequently-used set of sites from any workspace context. I keep a number of workspaces with their own focused set of Pinned tabs. I have more than 12 tabs that is workspace-agnostic and would like to be able to access from any context.

I use the the SuperPins extension to allow for more than 4 tabs per line of essentials. Increases the density and compactness of Essentials. A similar option for essentials (allowing more than four tabs per line) could help with the overflow issue in expanded mode.. It doesn't help with collapsed mode though. I guess people who use collapsed mode would be happy with a smaller number of essentials, and have found workflows that work best for them.

The combination of workspaces, workspace-local "Pinned" tabs, and global "Essential" tabs allows for a flexible workflow that I've found very useful in Zen.

src/zen/tabs/zen-tabs/vertical-tabs.css
```diff
@@ -1303,6 +1315,48 @@
...
+    }
+  }
+
+  &[overflowing]:not([scrolledtostart])::before {
+    opacity: 1;
+  }
+
+  &[overflowing]:not([scrolledtoend])::after {
+    opacity: 1;
+  }
```
blakegearin, line 1389: Nearly identical to `src/zen/workspaces/zen-workspaces.css` ([source](https://github.com/zen-browser/desktop/blob/dev/src/zen/workspaces/zen-workspaces.css#L526-L561))

ErdemGKSL:
This looks great especially when using container based essentials. Because for different workspace/containers i have different essentials and hard limit 12 forces me to have 3 essentials per workspace... (i have 4 workspaces)

PauloDaniel1993:
> I'll reopen it but won't be merged until fundamental issues are solved.
> 
>     1. Issue with overflowing over the window when there are too many
> 
>     2. Issue with wrong usage of essentials. (People misusing this feature as pinned tabs rather than highly important apps)

Isn't a bit problematic saying that people are using the feature wrong? It was added there and people would use as they want no? And, okay that before there wasn't a limit and probably was a bug or missing some planning, but as there was no limit, there is people that came to the browser because of that feature.

mr-cheffy:
> Isn't a bit problematic saying that people are using the feature wrong?

Why would it be? I've seen people try to drive cars underwater and I wouldn’t say the car is poorly designed just because it doesn’t turn into a submarine. Same goes for features — just because someone misuses them doesn’t mean the feature itself is flawed.

erdoganersinkeskin:
Hi,

When I tried to change "zen.tabs.essentials.max" to 24 or 30, for example, I still can't add new essentials in the essentials area.

PauloDaniel1993:
> > Isn't a bit problematic saying that people are using the feature wrong?
> 
> Why would it be? I've seen people try to drive cars underwater and I wouldn’t say the car is poorly designed just because it doesn’t turn into a submarine. Same goes for features — just because someone misuses them doesn’t mean the feature itself is flawed.

Sorry, I didn't get your point. I didn't say that the feature is flawed. What I said is that the feature was there in a way and it changed behavior already established. It is pretty common that something that was designed in a way to be used in another way. And it is not wrong, it is just another way to use it.

blakegearin:
> When I tried to change "zen.tabs.essentials.max" to 24 or 30, for example, I still can't add new essentials in the essentials area.

@erdoganersinkeskin This code isn't merged or released. Are you building locally?

erdoganersinkeskin:
I tried to use the local version for it, but I gave up. I will wait for stable version updates.

chrisshieh-direct:
One way to look at it is that the users are "misusing" the feature. Another way is to recognize that the users are giving you information about how they would love to use the feature. They're giving you specific UX feedback, unprompted and motivated. Much better than user testing.

I'd like to echo @blakegearin's sentiment... is there a technical reason why the number is 12? When it was unlimited, I had 16, which doesn't seem like a crazy number.

PauloDaniel1993:
I agree, I had 20. Given the nature of my work, I need to access a lot of apps a lot of times. Essentials is magical for that.

braccae:
Just adding my use case for more than 12 essential tabs. Before Zen/Arc I used to keep about 2 dozen PWA's open for notifications and quick access. These were things like webmail, chat clients, news readers, social media, etc.  Since moving to Zen, I've replaced those with the essential tabs. This limitation is basically telling me "You can't have more than 12 apps open on your computer at a time." It doesn't feel great.

src/browser/app/profile/features.inc
```diff
@@ -4,6 +4,8 @@ pref('zen.welcome-screen.seen', false, sticky);
 pref('zen.tabs.vertical', true);
 pref('zen.tabs.vertical.right-side', false);
 pref('zen.tabs.rename-tabs', true);
+pref('zen.tabs.essentials.max', 12);
+pref('zen.tabs.essentials.height', 0); // In vh units
```
mr-cheffy, line 15: Please remove `zen.tabs.essentials.height`, let's try to keep PRs centralized on fixing a specific problem

mr-cheffy (changes requested):

src/browser/app/profile/features.inc
blakegearin: This addresses the problem identified...

> Issue with overflowing over the window when there are too many

Did you have something else in mind for handling overflow?

mr-cheffy: Maybe just add a max height for the essentials container and make it overflowable with CSS

blakegearin: I think we encounter the same issue. There's no specific magic number for the height that makes sense for all users. It's highly dependent on the number of essentials, rows/width, and screen size.

PauloDaniel1993: But if make it overflow, it can have a "random" number specified until we put this `zen.tabs.essentials.height` back, no?

SalahAdDin:
Why is this even required? In previous versions we didn't have a limit for essentials?

SomeDudeBrosky:
Also adding my use case for more than 12 essential tabs: I am a medical student. I need more than 12 essentials in order to complete my daily work. 
Also; 12 is a very arbitrary number to limit the amount of essentials to.

mr-cheffy:
Is this pr done? There seems to be some git conflicts

src/browser/app/profile/features.inc
blakegearin: @PauloDaniel1993 I'm not sure I understand your question. 

By default `zen.tabs.essentials.height` is not active. Essentials are added normally, and the height and rows adjust based on the width chosen by the user. Once it's changed to something other than `0`, you get either expansion or scrolling depending on how many essentials there are and the screen height (since the number is `vh` units). There are screenshots in the description. 

Because of this behavior, I don't think it makes sense to pick an arbitrary number. This would cause the code change from being passive and opt-in to something that will visually change how the essentials look for many users.

mr-cheffy:
Continuation in https://github.com/zen-browser/desktop/pull/9189

Issue #2911, opened 2024-11-11
Can't install or run appimage on Fedora 41


thalskarth:
### Captchas

- [X] I have read the instructions.
- [X] I have searched existing issues and avoided creating duplicates.
- [X] I am not filing an enhancement request.

### What happened?

When i try to install the appimage via que script, it ends with an error:

```
[~]$ bash <(curl https://updates.zen-browser.app/appimage.sh)
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 12498  100 12498    0     0  24193      0 --:--:-- --:--:-- --:--:-- 24174
[i] ╔════════════════════════════════════════════════════╗
[i] ║                                                    ║
[i] ║    (ﾉ◕ヮ◕)ﾉ*:･ﾟ✧  Zen Browser Installer            ║
[i] ║                                                    ║
[i] ║    CPU: AVX2 Supported (Optimized Version)         ║
[i] ║    Status: Zen Browser Not Installed               ║
[i] ║    zsync: Not Installed (Needed for Updates)       ║
[i] ║                                                    ║
[i] ╚════════════════════════════════════════════════════╝
[i] 
[i] (★^O^★) What would you like to do?
[i]   1) Install
[i]   2) Uninstall
[i]   0) Exit
Enter your choice (0-3): 1
[i] 
[w] Auto detecting AVX2 support...
[w] Downloading Zen from https://github.com/zen-browser/desktop/releases/latest/download/zen-specific.AppImage
[i] 
[i] Installing Zen Browser...
##################################################################################################################################### 100.0%
[i] Install completed successfully!
/usr/bin/AppImageLauncher: /lib64/libcurl.so.4: no version information available (required by /usr/bin/../lib/x86_64-linux-gnu/appimagelauncher/libappimageupdate.so)
QSocketNotifier: Can only be used with threads started with QThread
This doesn't look like a squashfs image.
Failed to open squashfs image
[~]$
```

If I download manually the appimage file, and run it gives the a similar error:

```
[~]$ ./zen-specific.AppImage
/usr/bin/AppImageLauncher: /lib64/libcurl.so.4: no version information available (required by /usr/bin/../lib/x86_64-linux-gnu/appimagelauncher/libappimageupdate.so)
QSocketNotifier: Can only be used with threads started with QThread
Squashfs image uses (null) compression, this version supports only xz, zlib.
[...]

PR #8562, opened 2025-05-23
Deploy  dependencies in AppImage

### LINKED ISSUE #7967: Appimage does not work on distros that depend on musl instead of glibc
### Captchas

### What happened?

The appimages from the releases use glibc, making it impossible to run on distros like Void, Alpine etc. AFAIK it is possible to build appimages statically making it work with all distros. Could it be possible that zen does this too?

Linux (AppImage)

Other

Samueru-sama:
fixes #7967 and likely other similar issues. 

So far Zen has just used the portable tarball build repacked as an AppImage, while this is usually good enough™ It means that the AppImage can't really work everywhere since the portable tar build depends on the host glibc.

These changes add the needed dependencies to guaranteed operation anywhere, such as musl distros and distros as old as ubuntu 14.04. 

--------------------------------------------------------------------

Considerations: 

This change will break integration with appimaged and appimagelauncher since I switched the AppImage to dwarfs to reduce the final size (now 180 MiB). I could go back to squashfs but that makes the AppImage **+250 MiB** and I don't think it is worth it 👀

Note that this is unrelated to the [infamous execv error](https://github.com/zen-browser/desktop/issues/2911) people will still be able to launch it even on appimagelauncher, just that integration is broken due to it not being able to read dwarfs.

AM, dbin, Gearlever, soar, etc already support dwarfs so it is not an issue there.

**Note** I'm not able to figure out how the CI works, so I'm not sure if I made a mistake in the steps. 😅 I did testing [here](https://github.com/Samueru-sama/zen-appimage-test) instead by using the tarball directly.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -419,40 +419,40 @@ jobs:
 
           # deploy deps so that the appimage can work on musl systems as well as super old distros: 
           SYSLIBS="/usr/lib/$ARCH-linux-gnu"
-          cd $APPDIR && (
-            wget "https://github.com/VHSgunzo/sharun/releases/latest/download/sharun-$ARCH-aio" -O ./sharun-aio
-            chmod +x ./sharun-aio
-            ./sharun-aio l -p -v -s -k            \
-              ./bin/zen*                          \
-              ./bin/glxtest                       \
-              ./bin/vaapitest                     \
-              ./bin/pingsender                    \
-              ./bin/lib*                          \
-              "$SYSLIBS"/lib*GL*                  \
-              "$SYSLIBS"/dri/*                    \
-              "$SYSLIBS"/libpci.so*               \
-              "$SYSLIBS"/libxcb-*                 \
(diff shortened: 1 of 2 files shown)


mr-cheffy (approved):

mr-cheffy:
Please run npm run pretty

Samueru-sama:
> Please run npm run pretty

I'm not sure how to do that 😅

JosueGalRe:
@Samueru-sama in the console run "npm run pretty"

Samueru-sama:
> @Samueru-sama in the console run "npm run pretty"

First time I use npm, hopefully it is all ok now.

xlc-dev:
bumping thread to see if this can be merged. Is everything ok with this PR @mr-cheffy? I'd appreciate it very much if this gets to main

codeCraft-Ritik (commented):
Great improvement! Bundling the required dependencies should make the AppImage much more portable across older and non-glibc distributions.
The tradeoff with dwarfs vs squashfs seems well considered given the size savings 👍

PR #8718, opened 2025-05-31
Fix misaligned split view favicons

kritishd8:
Before:
image

After:
image


mr-cheffy:
How does the group look when selected?

kritishd8:
image

looks the same ig

kritishd8:
image

didnt notice the padding on the sides, my bad.

new one should be good

mr-cheffy:
Right, but now it looks like this:

image


sorry, but there's no possible way to align them. They have different spacings and sizes, it's not possible to align them without making it look weird

Issue #8863, opened 2025-06-04
Vimium new tab opens glance view


paulwyszynski:
### Preliminary Checks

- [x] I have read and understood the important section above.
- [x] I have searched existing issues and avoided creating duplicates.
- [x] I am not filing an enhancement request.
- [x] I have checked that this issue cannot be reproduced on Mozilla Firefox.
- [x] I have checked that this issue can be reproduced once I removed all my Mods and Custom CSS.

### What happened?

After the update to 1.12.10b on Mac, I have now the issue that each time I want open a new tab with Vimium, it opens it in a glance window. Even after deactivating glance it still opens in this popup view.

Has anyone else this issue using latest zen and vimium extension?

Zen v.1.12.10b
MacOS 15.4.1
Vimium: 2.2.1

### Expected behavior

Default new tab should be opened

### Actual behavior

Opening a new tab with Vimium extension, opens it in a glance popup window instead of default tab

### Steps to reproduce

1. Install latest Zen v.1.12.10b
2. Install latest Vimium 2.2.1
3. Open a new tab with shift+o
4. Enter only text to trigger search engine

### Version

1.12.10b

### What platform are you seeing the problem on?

macOS - aarch64

### What component is this issue related to?

Glance

### Relevant log output if applicable

```shell

```

Ernestas-t:
Adding in as i have the same issue on windows 11 version 24H2 (OS Build 26100.4061)

after updating to 1.12.10b vimium opens in a new tab, although it isn't a 100% and i can't pinpoint to when it choose one over the other. And glance view is disabled in the settings.

paulwyszynski:
@Ernestas-t does it also happen only when you try to search with your search engine? When you enter a domain it should open a new tab…

Ernestas-t:
Yep, navigating to a url opens a new tab, searching through the search engine uses the glance view. doesn't matter which search engine it is google/brave/duckduckgo

Ernestas-t:
Not super familiar with how zen and vimium interact but from source i think this is whats happening.

The code responsible for opening new tabs with vimium is here:
https://github.com/philc/vimium/blob/master/background_scripts/tab_operations.js


```
  if (urlStr == null) {
    // The requested destination is not a URL, so treat it like a search query.
    //
    // The chrome.search.query API lets us open the search in a new tab, but it doesn't let us
    // control the precise position of that tab. So, we open a new blank tab using our position
    // parameter, and then execute the search in that tab.

[...]

PR #8890, opened 2025-06-05
Respect glance enabled preference before opening Glance on search

### LINKED ISSUE #8752: Search for X still opens Glance despite it being disabled in about:config
### Preliminary Checks

### What happened?

When selecting text, right clicking and selecting "Search X for Y" (where X is search engine and Y is the selected text), Glance thingy opens. BUT! I set `zen.glance.enabled` to false in about:config.

I tried searching the about:config for a new field that controls this specific search feature but couldn't find anything.

### Expected behavior

"Search X for Y" context menu item opens a new tab instead of Glance if `zen.glance.enabled` (and/or some other specific param) is set to false

### Actual behavior

"Search X for Y" context menu item opens Glance (seemingly) regardless of the config

### Steps to reproduce

1. go to `about:config`
2. set `zen.glance.enabled` to false
3. try selecting any text
4. right click on the selection
5. click "Search DuckDuckGo/etc for X"
6. witness Glance opening instead of a new tab

### Screenshots and videos

All Glance-related is disabled in the configs
Image

### Version

1.12.9b

Linux (Tarball)

Glance

```shell

```

Ernestas-t:
This pull request addresses #8863   where opening a new tab via the Vimium extension would always trigger the Glance popup view, even when Glance was disabled in preferences.

The fix was to update the onSearchSelectCommand function to check the zen.glance.enabled user preference before attempting to open Glance.

### DIFF
--- a/src/zen/glance/ZenGlanceManager.mjs
+++ b/src/zen/glance/ZenGlanceManager.mjs
@@ -747,6 +747,10 @@
     }
 
     onSearchSelectCommand(where) {
+      // Check if glance is enabled in user preferences
+      if (!Services.prefs.getBoolPref('zen.glance.enabled', false)) {
+        return;
+      }
       if (where !== 'tab') {
         return;
       }


mr-cheffy (approved):

mr-cheffy:
@Ernestas-t thanks a lot! Would be cool as well if you added another config to handle this case specificly `zen.glance.enable-contextmenu-search` for example in a new PR!

PR #8917, opened 2025-06-07
Add the ability to scroll on the workspace indicator and icons container.

CosmoCreeper:
In this pull request, I have consolidated the scrolling ability in ```_setupSidebarHandlers``` into a new function labeled ```scrollInDirection``` and have utilized that in another function labeled ```initScrollEvent``` (which may be unnecessary.) I have then utilized the scroll event function to add the ability to scroll on the workspace indicator and icons container for a much easier way for users to switch workspaces without clicking or using the Ctrl key.

**Possible wanted changes:** You may want to change the names of the new functions and/or remove the ```initScrollEvent``` function (and replace by directly adding the listeners to each case where it is used).


mr-cheffy:
Please handle this on the Zen workspace icons and make sure to only change if ctrl is pressed

CosmoCreeper:
@mauro-balades What do you mean? The purpose of this is that you can scroll on the workspace indicator without pressing Ctrl and on the icons container at the bottom of the screen. (#8917)

mr-cheffy:
This is not a good idea as the icons can actually be overflowable and therefore having a chance of the user not being able to locate it's desired icon

CosmoCreeper:
@mauro-balades Oh wait, I see. Okay, I'll update it.

CosmoCreeper:
@mauro-balades I have updated this pr by adding the event to the individual icons rather than the container itself. Do you want me to only trigger this event if the icon hovered is active or is what is here fine?

CosmoCreeper:
@mauro-balades Just in case you're wondering what I've changed, I have prevented an edge case in which the user is scrolling horizontally on touchpad and the workspace indicator scrolling event would interfere. This new version will ensure that the scrolling method is not a touchpad and that the user is not scrolling horizontally, otherwise, it will scroll just like before.

src/zen/workspaces/ZenWorkspaces.mjs
```diff
@@ -467,6 +467,21 @@ var gZenWorkspaces = new (class extends ZenMultiWindowFeature {
...
           );
+          this.initIndicatorContextMenu(workspaceWrapper.indicator);
+          workspaceWrapper.indicator.addEventListener('wheel', (event) => {
+            let isTrackpad = false;
+            if (
+              (event.wheelDeltaY && event.wheelDeltaY === event.deltaY * -3) ||
+              event.deltaMode === 0
+            ) {
+              isTrackpad = true;
+            }
```
correcthorsebatterystaple:
```suggestion
            const isTrackpad =
              (event.wheelDeltaY && event.wheelDeltaY === event.deltaY * -3) ||
              event.deltaMode === 0;
```

src/zen/workspaces/ZenWorkspaces.mjs
```diff
@@ -467,6 +467,17 @@ var gZenWorkspaces = new (class extends ZenMultiWindowFeature {
...
             tabs
           );
+          this.initIndicatorContextMenu(workspaceWrapper.indicator);
+          workspaceWrapper.indicator.addEventListener('wheel', (event) => {
+            const isTrackpad =
+              (event.wheelDeltaY && event.wheelDeltaY === event.deltaY * -3) ||
+              event.deltaMode === 0;
+
+            const verticalScroll = Math.abs(event.deltaY) > Math.abs(event.deltaX);
+            if ((isTrackpad && verticalScroll) || !isTrackpad) {
```
correcthorsebatterystaple:
```suggestion
            if (!isTrackpad || verticalScroll) {
```

PR #8990, opened 2025-06-12
feat: add preference to toggle URL bar autofocus on startup

Ernestas-t:
Add zen.startup.focus-urlbar preference to control whether the URL bar is automatically focused when switching tabs during startup. Defaults to true to maintain current behavior while allowing users to disable this feature if desired.

It would also be good to have a toggle in the UI settings, but i think it's best to leave it for a separate commit.


src/zen/workspaces/ZenWorkspaces.mjs
```diff
@@ -233,7 +233,11 @@ var gZenWorkspaces = new (class extends ZenMultiWindowFeature {
         gZenVerticalTabsManager._canReplaceNewTab
       ) {
         // Only set up URL bar selection if we're switching to a different tab
-        if (gBrowser.selectedTab !== this._emptyTab && selectURLBar) {
+        if (
+          gBrowser.selectedTab !== this._emptyTab &&
+          selectURLBar &&
+          Services.prefs.getBoolPref('zen.startup.focus-urlbar', true)
```
mr-cheffy, line 239: This isnt the correct call as it woudnt just be on startup. Rather dont call openTab on `selectStartpage`

mr-cheffy (changes requested):

mr-cheffy:
@Ernestas-t not sure if you still want to implement this, but there are some git conflicts. I'll close this and please do re-open with the conflicts solved if you are interested ^^

PR #9126, opened 2025-06-22
fix: Fix tab drag indicator orientation to horizontal on collapsed toolbar

p-sw:
This little change makes drag indicator always horizontal in collapsed toolbar mode, and make tab movement logic to match with drag indicator.

https://github.com/user-attachments/assets/1ef0aa59-4f47-451e-ae35-3e1d3ce11b81

https://github.com/user-attachments/assets/adfa0d50-c69e-4bff-9dec-e986961df47f

### DIFF
--- a/src/zen/tabs/ZenPinnedTabManager.mjs
+++ b/src/zen/tabs/ZenPinnedTabManager.mjs
@@ -857,7 +857,7 @@
               const rect = targetTab.getBoundingClientRect();
               let newIndex = targetTab._tPos;
 
-              if (isVertical) {
+              if (isVertical || !this.expandedSidebarMode) {
                 const middleY = targetTab.screenY + rect.height / 2;
                 if (!isRegularTabs && event.screenY > middleY) {
                   newIndex++;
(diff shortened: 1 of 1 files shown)


mr-cheffy:
Why not just set `isVertical ||= this...`? @p-sw

p-sw:
> Why not just set `isVertical ||= this...`? @p-sw

From what I understand and tested in the original code, `isVertical` is always false since `this.expandedSidebarMode` is false.

But `isVertical` should be false only when `expandedSidebarMode` is true and user dragged the tab into essential area since the only horizontal element of the sidebar is the essential in expanded toolbar, and essential in collapsed toolbar is vertical.

so I added `!this.expandedSidebarMode` in `if (isVertical)` to make it always true in collapsed toolbar.

I don't know if I understand the `||=` operator well enough, but my understanding is that it is a short-circuit of `x || (x = y)`. Can you tell me more about how this operator can solve the above logic error?

mr-cheffy:
> isVertical is always false since

No, isVertical is always false unless dragging into an essential. I mean, if it works, I can merge it right now, I was just suggesting a cleaner solution, I dont mind personally.

> I don't know if I understand the ||= operator well enough, but my understanding is that it is a short-circuit of x || (x = y). Can you tell me more about how this operator can solve the above logic error?

Basically, yeah. If you would like to clean it up, you can make a pr if you want, otherwise it's perfectly fine. Thanks a lot!

mr-cheffy (approved):

Issue #7865, opened 2025-04-25
Ctrl+Tab Cycling through tabs crashes if pressed without recent tabs


pymmog:
### Captchas

- [x] I have read the instructions.
- [x] I have searched existing issues and avoided creating duplicates.
- [x] I am not filing an enhancement request.
- [x] I have checked that this issue cannot be reproduced on Mozilla Firefox.
- [x] I have checked that this issue can be reproduced once I removed all my Mods and Custom CSS.

### What happened?

When opening Zen without any active tabs and trying the Cycle with Ctrl + Tab the Application Crashes.

### Version

1.11.5b

### What platform are you seeing the problem on?

Windows - x64

### What component is this issue related to?

Keyboard Shortcuts

### Relevant log output if applicable

```shell

```

PR #9148, opened 2025-06-23
Fix browser-ctrlTab-js patch for Firefox 140.0 compatibility

markzakharyan:
## Summary
This PR fixes the browser-ctrlTab-js.patch file to work with Firefox 140.0, resolving the init script failure during patch application. Note that this is my first contribution to this repo so please be kind if what I did doesn't make any sense :).

## Changes Made
- Updated patch to work with Firefox 140.0 changes
- Removed obsolete open function modification (already included in Firefox 140.0)
- Updated line numbers to match current ctrlTab.js structure
- Fixes the npm run init command that was failing due to patch application errors

## Testing
- npm run init now completes successfully without patch errors
- All 166 patches apply correctly
- Language pack update works properly

## Issue
This fixes the build setup issue where the init script was failing with patch application errors. The patch was trying to modify code that had already been updated in Firefox 140.0, causing conflicts during the build setup process. This also fixes ctrl+shift+tab behavior when the "Ctrl+Tab cycles through tabs in recently used order" setting is enabled.

### DIFF
--- a/src/browser/components/tabbrowser/content/browser-ctrlTab-js.patch
+++ b/src/browser/components/tabbrowser/content/browser-ctrlTab-js.patch
@@ -21,16 +21,7 @@ index 6753641cb579032306453be3f5054d7bc7661e8c..bf21f6c14f825fbe2d322900595cd344
        return;
      }
  
-@@ -483,7 +484,7 @@ var ctrlTab = {
-   },
- 
-   open: function ctrlTab_open() {
--    if (this.isOpen) {
-+    if (this.isOpen || !this.tabCount) {
-       return;
-     }
- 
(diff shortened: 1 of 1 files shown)


mr-cheffy:
Tab count is necessary though, why remove it?

markzakharyan:
Removing tabCount was how I got ctrl+shift+tab to work when the "Ctrl+Tab cycles through tabs in recently used order" setting is enabled. ideally, i would want this to just cycle back in the tab index instead of just not doing anything, which was the previous behavior.

mr-cheffy:
It crashes the browser if there are no tabs on some occasions. See https://github.com/zen-browser/desktop/issues/7865

PR #9183, opened 2025-06-25
fix: Changed 'KeyShortcutModifiers' to reflect proper class name 'nsKeyShortcutModifiers'

### LINKED ISSUE #9170: Keyboard combo broken to switch between workspaces
### Preliminary Checks

### What happened?

The keyboard navigation seems to be broken in the twilight starting from 06/20

I used to use `Cmd+Option+Down/Up` for traversing the tabs in the current workspace, and `Cmd+Option+Left/Right` to traverse the different workspaces. The former is working, but the latter keyboard combo also cycles only across the tabs in the current workspace (ie doesn't switch workspaces). This was working till the latest update.

Image

I am also attaching a screenshot of the kayboard preferences section (but as a side note, I am unable to change any of these any longer!)

### Expected behavior

The keyboard preferences should be respected

### Actual behavior

Workspace switching using keyboard shortcuts doesn't work even though the keyboard combination has been registered.

### Steps to reproduce

1. Upgrade to any of the latest twilight builds.
2. Setup the keyboard combination as per my screenshot
3. Setup at least 2 workspaces
4. Try to use the keyboard combination to switch workspaces.

### Screenshots and videos

_No response_

### Version

1.14t (Firefox 140.0) (2025-06-24) (aarch64)

macOS - aarch64

Workspaces

```shell

```

vraravam:
Fixes #9170

### DIFF
--- a/src/browser/components/preferences/zen-settings.js
+++ b/src/browser/components/preferences/zen-settings.js
@@ -987,7 +987,7 @@ var gZenCKSSettings = {
     let input = document.querySelector(
       `.${ZEN_CKS_INPUT_FIELD_CLASS}[${KEYBIND_ATTRIBUTE_KEY}="${this._currentActionID}"]`
     );
-    const modifiers = new KeyShortcutModifiers(
+    const modifiers = new nsKeyShortcutModifiers(
       event.ctrlKey,
       event.altKey,
       event.shiftKey,


mr-cheffy (approved):

vraravam:
This does not fix the original problem about keyboard combos to switch workspaces @mr-cheffy

mr-cheffy:
Don't mark it as "fixes" then please

Issue #9220, opened 2025-06-29
The input focus state CSS styles are wrong


Hephaest:
### Preliminary Checks

- [x] I have read and understood the important section above.
- [x] I have searched existing issues and avoided creating duplicates.
- [x] I am not filing an enhancement request.
- [x] I have checked that this issue cannot be reproduced on Mozilla Firefox.
- [x] I have checked that this issue can be reproduced once I removed all my Mods and Custom CSS.

### What happened?

When the dialog’s input field gains focus, both border and outline disappear, making the UI look a bit awkward.

I have checked the Firefox behavior, it works as what user expected.

https://github.com/user-attachments/assets/da529b89-4224-4ff6-88b1-48214e571c0a

### Expected behavior

The outline should be displayed

### Actual behavior

Neither the border nor the outline is displayed.

### Steps to reproduce

1. Go to **Settings** page (cmd + ,)
2. Click **Search** tab
3. Navigate to **Search Shortcuts** section
4. Click the **Add** button in the bottom

### Screenshots and videos

https://github.com/user-attachments/assets/59532141-1f21-404f-aa3b-4d2cccc5305c

### Version

1.13.2b

### What platform are you seeing the problem on?

macOS - aarch64

### What component is this issue related to?

Settings

### Relevant log output if applicable

```shell

```

Propheticus:
Good find and how nice of you to fix the issue you raised yourself :)

PR #9221, opened 2025-06-29
fix: align with Firefox UI styles & improve accessibility

Hephaest:
## Description
This PR aims to solve the issue (#9220) I raised in the community. I discovered this bug about two weeks ago but hadn’t had time to fix it. Since it’s still unresolved, I decided to take a look.

## Root Cause
Apparently, Zen Browser insert the custom CSS theme into the `common-shared.css`.
```diff
@@ -5,6 +5,8 @@
 @import url("chrome://global/skin/design-system/tokens-brand.css");
 @import url("chrome://global/skin/design-system/text-and-typography.css");
 
+@import url("chrome://browser/content/zen-styles/zen-theme.css");
```
And there's one thing in `zen-theme.css` caught my eye:
```css
/* This is like the secondary button */
--in-content-button-background: transparent !important;
--button-bgcolor: var(--in-content-button-background) !important;
--focus-outline-color: var(--button-bgcolor) !important; // <-- Here! Why make focus outline transparent? 🤔
```
So I inspected the Firefox implementation:
image
And also look into what `--in-content-focus-outline` stands for:
image
image

Firefox’s approach seems more reasonable for a broader range of users. I adopted it with Zen’s custom theme accordingly—the UI now looks great.

## Demo

https://github.com/user-attachments/assets/50de01ff-3fd0-4297-b9d1-13aebf8f7e6e

### DIFF
--- a/src/zen/common/styles/zen-theme.css
+++ b/src/zen/common/styles/zen-theme.css
@@ -84,7 +84,7 @@
   --button-bgcolor: var(--in-content-button-background) !important;
   --button-hover-bgcolor: var(--in-content-button-background-hover) !important;
   --button-hover-color: var(--in-content-button-text-color-hover) !important;
-  --focus-outline-color: var(--button-bgcolor) !important;
+  --focus-outline-color: var(--color-accent-primary) !important;
 
   --toolbarbutton-icon-fill-attention: var(--zen-primary-color) !important;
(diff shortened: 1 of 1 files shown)


Propheticus:
Think this one is ready to go and can be sent for review?

PR #7359, opened 2025-04-05
Initial implementation for folders and tree-styled tabs

mr-cheffy:
Functionality:
- [x] Make firefox groups work only for pinned tabs
- [x] Correctly restore folders to their designated workspaces after a restart
- [x] Allow folders to stack up together
   - [x] Split view groups should be able to fit inside folders

UI/UX:
- [ ] Rename input when creating a folder
- [x] Open / Close animations (excluding selected tab)
- [ ] Allow drag and dropping folders
- [x] Allow droping tabs inside folders

PR #9355, opened 2025-07-07
feat: Zen Folders

octaviusz:
Continued work on https://github.com/zen-browser/desktop/pull/7359

TODO:

- [x] Drag-and-drop tabs / other folders into collapsed folders.
- [x] Empty folders
- [x] Fix overall drag and dropping when having sub folders
- [x] Add right click context menu to perform folder actions (Instead of firefox's native menu)
- [x] Allow folder name inline editing
- [x] Figure out what to do in collapsed mode
- [x] Restore subfolders after restart session

Few extras:

- [x] Folder -> workspace conversion and the other way around
- [x] "welcome page" should create a `zen basics` folder

Known bugs:
- [x] When moving a split group from a normal section to the pinned section, the drop indicator does not appear.
- [x] When moving a tab from a normal section into a zen-folder, it can overlap the zen-folder label due to identical `elementIndex` values.
- [x] Need to adapt split groups to handle nested groups.
- [x] If a nested zen-folder is placed after a split group, the zen-folder is ejected outside on restart.
- [x] Fix the collapsed animation when a nested zen-folder contains a selected tab.
- [x] Fix the collapsed animation when dragging a zen-folder that has a selected tab.
- [x] Fix the bug where moving a selected tab out of a zen-folder fails if the zen-folder is both collapsed and has-active.
- [x] You shoudn't be able to add essentials into folders

Features:

- [x] Decide what to do with `zen-empty-tabs`, as after a session restart they end up outside.
- [x] We need to open a collapsed folder when moving a tab, zen-folder, or split-group into
- [x] If the user selects from the popup tabs list, we should open the group only up to the selected tab, not fully expand it.
- [x] Find a more robust way to mark the selected folder and animate its icon to prevent the animation from replaying repeatedly.
- [x] Move folders to different workspaces


12th-devs:
Not sure if this has been patched out yet, but an issue with pinned groups right now is middle clicking to open a new tab opens it in the pinned group (as a regular tab). Just incase it was not on the radar :)

mr-cheffy:
@octaviusz i've been thinking about the changes done to the mutation observers we've talked before. Is there really no possible workaround for this? This patch is going to be hell to maintain in the long run.

octaviusz:
@mr-cheffy  I tried to preserve the original logic, but the call to `#observeTabChanges` needs to be placed at the end so that `.tab-group-container` has time to initialize.

src/zen/workspaces/ZenWorkspaces.mjs
```diff
@@ -2700,7 +2700,7 @@ var gZenWorkspaces = new (class extends ZenMultiWindowFeature {
           if (glance) {
             tabs.push(glance);
           }
-        } else if (tab.tagName == 'tab-group') {
+        } else if (tab.tagName == 'tab-group' || tab.tagName == 'zen-folder') {
```
mr-cheffy: Here, please use `gBrowser.isTabGroupLabel`

src/zen/workspaces/ZenWorkspaces.mjs
```diff
@@ -2718,7 +2718,7 @@ var gZenWorkspaces = new (class extends ZenMultiWindowFeature {
   get allTabGroups() {
     if (!this._hasInitializedTabsStrip) {
       let children = this.tabboxChildren;
-      return children.filter((node) => node.tagName == 'tab-group');
+      return children.filter((node) => node.tagName == 'tab-group' || node.tagName == 'zen-folder');
```
mr-cheffy: Also use `gBrowser.isTabGroupLabel` here please

mr-cheffy (changes requested):

src/zen/folders/zen-folders.css
```diff
@@ -198,3 +198,123 @@ tab-group:not([split-view-group]) {
...
+  --zen-workspace-color-bg: var(--zen-primary-color);
+  --zen-workspace-color-fg: var(--zen-colors-primary-foreground);
+  --zen-workspace-color-stroke: var(--toolbox-textcolor);
+}
+
+zen-folder {
+  display: flex !important;
+  flex-direction: column !important;
+
+  @media not (-moz-bool-pref: 'zen.view.sidebar-expanded') {
```
mr-cheffy: Instead of using the pref, please use `:root[zen-sidebar-expanded]`

src/zen/folders/zen-folders.css
```diff
@@ -198,3 +198,123 @@ tab-group:not([split-view-group]) {
 .tab-group-line {
   display: none !important;
 }
+
+* {
```
mr-cheffy: Why are these veriables being set to *all* elements instead of just root? And why use variables when they are just holding a single possible value? Id replace these instances with their value. If you know what I mean

mr-cheffy (changes requested):

src/zen/folders/ZenFolders.mjs
```diff
@@ -103,11 +129,56 @@
...
+          `);
+      document.getElementById('context_moveTabToGroup').before(contextMenuItems);
+    }
+
+    #initTabsPopup() {
+      const popupSet = document.querySelector('#mainPopupSet');
+      if (popupSet.querySelector('#zen-folder-tabs-popup')) return;
+      const popup = window.MozXULElement.parseXULToFragment(
+        `
+        <panel id="zen-folder-tabs-popup" type="arrow" orient="vertical">
```
mr-cheffy: Instead of adding this dynamically, it can simply be added into https://github.com/zen-browser/desktop/tree/dev/src/browser/base/content/zen-panels for example

src/zen/workspaces/ZenWorkspaces.mjs
```diff
@@ -2718,7 +2718,8 @@ var gZenWorkspaces = new (class extends ZenMultiWindowFeature {
   get allTabGroups() {
     if (!this._hasInitializedTabsStrip) {
       let children = this.tabboxChildren;
-      return children.filter((node) => node.tagName == 'tab-group' || node.tagName == 'zen-folder');
+      console.log(children);
```
mr-cheffy: Did this `console.log` slip off? 😅

mr-cheffy (changes requested):

src/zen/workspaces/ZenWorkspaces.mjs
octaviusz: oh my bad

src/browser/components/tabbrowser/content/tabbrowser-js.patch
```diff
@@ -617,11 +617,18 @@ index 37570c54b0..e25a26f56b 100644
...
-         }
+@@ -6095,21 +6219,34 @@
        }
--
+       if (this.isTabGroupLabel(element)) {
+         element = element.group;
+-        if (targetElement?.group) {
+-          targetElement = targetElement.group;
+-        }
++        // if (targetElement?.group) {
```
mr-cheffy: Instead of commenting, just remove

mr-cheffy (changes requested):

mr-cheffy:
Some things left:

- [x] Drag-and-drop tabs / other folders into collapsed folders.

I believe there's to ways to tackle this. 

1. Show an indicator (e.g. when dragging a tab to the workspace name), to show an action can be done. This can be as simple as just showing a hover effect on the folder when dragged onto. Once the item is dropped, it'll be added to the start of the folder
2. Open the folder so it can be properly inserted

- [x] Empty folders.

src/zen/folders/ZenFolders.mjs
```diff
@@ -121,6 +401,246 @@
...
+
+      const activeGroup = event.target.parentElement;
+
+      this.#populateTabsList(activeGroup);
+
+      const search = this.#popup.querySelector('#zen-folder-tabs-list-search');
+      search.placeholder = `Search ${activeGroup.name}...`;
+      const tabsList = this.#popup.querySelector('#zen-folder-tabs-list');
+
+      search.addEventListener('input', () => {
```
mr-cheffy: This adds infinite amount of event handlers, it should just be added when initializing

src/zen/folders/zen-folders.css
```diff
@@ -198,3 +198,221 @@ tab-group:not([split-view-group]) {
...
+}
+
+.tabs-list-item-background {
+  position: absolute;
+  inset: 0;
+  border-radius: inherit;
+  margin-inline: var(--tab-block-margin);
+}
+
+.tabs-list-item-content {
```
mr-cheffy: These class names should be changed to something more specific before merge

src/zen/folders/ZenFolders.mjs
```diff
@@ -121,6 +401,246 @@
       }
       return this._piningFolder;
     }
+
+    openTabsPopup(event) {
+      event.stopPropagation();
```
mr-cheffy, line 621: Nothing should open if there's no tabs to search for

adrian-aaen:
>     1. Show an indicator (e.g. when dragging a tab to the workspace name), to show an action can be done. This can be as simple as just showing a hover effect on the folder when dragged onto. Once the item is dropped, it'll be added to the start of the folder
> 
>     2. Open the folder so it can be properly inserted

What about both? I don't remember specific examples, but I've seen multiple apps using option 1, then switching to option 2 if the user keeps hovering for longer.

mr-cheffy (approved):

mr-cheffy:
> What about both? I don't remember specific examples, but I've seen multiple apps using option 1, then switching to option 2 if the user keeps hovering for longer.

Oh, very true!

mr-cheffy:
@octaviusz do you have working code for the folder dots on the icon or should I start working on it?

edit: New firefox RC version just dropped, ill be focusing on that for the next days if you don't mind, sorry

mr-cheffy:
I'd also like to discuss here about limiting the amount of subfolders.

why? My main concern is space. Each sub folder has around 4 pixels of padding to the left. Meaning that if you have 10 sub folders, they would overflow outside the sidebar. This is highly noticable specially inside collapsed mode. There is simply no space for having more than 5 sub folders, visually. What do you guys think? Any solutions to this or would you guys simply not use or woudn't find the use for more than 5 sub folders?

DobbiKov:
> I'd also like to discuss here about limiting the amount of subfolders.
> 
> why? My main concern is space. Each sub folder has around 4 pixels of padding to the left. Meaning that if you have 10 sub folders, they would overflow outside the sidebar. This is highly noticable specially inside collapsed mode. There is simply no space for having more than 5 sub folders, visually. What do you guys think? Any solutions to this or would you guys simply not use or woudn't find the use for more than 5 sub folders?

My maximum was 3 subfolders, in my opinion 5 is enough.

aimryoui:
My suggestion

# Not Collapsed Toolbar

- First of all, the number of sub-folders is limited to 6 in Arc (1 parent folder and 6 sub-folders), let's take that as a reference.

Screenshot 2025-07-15 192958

- For Zen, we should have a fixed `min-width` for the sidebar (mainly for Multiple Toolbar mode. As for Single Toolbar mode, the navigation buttons are already blocking it - so it's quite wide). For example, with Arc, I don't know how it is on Mac, but on Windows, the `min-width` of the sidebar will look like this:

Screenshot 2025-07-15 192304

- You can see 2 important things in the image above:
	- The name of the deepest sub-folder still has a visible first character.
	- The tab inside the deepest sub-folder still has a favicon and a visible first character of the title (tab name).

Screenshot 2025-07-15 193518

**That's the key point to determine the `min-width` of the sidebar. And it's always the same regardless of whether the folder exists or not.**

But what if users change their `font-size` or `font-family`. Well, we always do that, so just make a variable for each parameter so we can override it :D.

Simply determine the `min-width` for the sidebar and everything is considered done. **Please don't even think about bringing `horizontal scrollbar` to the sidebar, it will reduce the overall performance and also conflict with the `swipe to change workspace` feature.**

Finally, what should be the limit for the number of sub-folders? I trust the eyes of Zen devs :)

# Collapsed Toolbar

Why not this?

Screenshot 2025-07-15 194810

- Just display the whole folder as an icon, and it can be expanded just like a menu with sub-menu, and replace icon with favicon, label with tab title.
- And then replace the inside selected tab's favicon with the previous folder icon (still expandable too).

Bellisario:
> I'd also like to discuss here about limiting the amount of subfolders.
> 
> why? My main concern is space. Each sub folder has around 4 pixels of padding to the left. Meaning that if you have 10 sub folders, they would overflow outside the sidebar. This is highly noticable specially inside collapsed mode. There is simply no space for having more than 5 sub folders, visually. What do you guys think? Any solutions to this or would you guys simply not use or woudn't find the use for more than 5 sub folders?

My suggestion is to allow folders to become a temporary "sub-space" context.

Once Arc twitter/x published a video of this prototype _(video below)_.\
I think that if studied a bit could become a super-useful feature of Zen.

---

https://github.com/user-attachments/assets/10464cae-8359-4cc5-a00a-2a4cff4886e2

> from: https://x.com/i/status/1664680163513356288

---

### My ideal use case

- I need to search "phone case" on Amazon
- I create a "phone cases" folder and open it as "sub-space"
- all tabs I open in the context are directly pushed to the folder (or sub-folder)
- I end the search and get back to normal folder behavior with a (back button) click
- all the tabs I had opened are now pinned inside the folder for future usage

12th-devs:
> > I'd also like to discuss here about limiting the amount of subfolders.
> > why? My main concern is space. Each sub folder has around 4 pixels of padding to the left. Meaning that if you have 10 sub folders, they would overflow outside the sidebar. This is highly noticable specially inside collapsed mode. There is simply no space for having more than 5 sub folders, visually. What do you guys think? Any solutions to this or would you guys simply not use or woudn't find the use for more than 5 sub folders?
> 
> My suggestion is to allow folders to become a temporary "sub-space" context.
> 
> Once Arc twitter/x published a video of this prototype _(video below)_. I think that if studied a bit could become a super-useful feature of Zen.
> Arc.-.Here.at.the.browser.co.Folders.are.kind.of.controversial.Our.l.1664679547567235072.mp4
> 
> > from: [x.com/i/status/1664680163513356288](https://x.com/i/status/1664680163513356288)
> 
> ### My ideal use case
> 
>     * I need to search "phone case" on Amazon
> 
>     * I create a "phone cases" folder and open it as "sub-space"
> 
>     * all tabs I open in the context are directly pushed to the folder (or sub-folder)
> 
>     * I end the search and get back to normal folder behavior with a (back button) click
> 
>     * all the tabs I had opened are now pinned inside the folder for future usage

I can see this being a solution but I personally believe that @hoangnhan2ka3 has a better idea. @Bellisario's method seems confusing to new users and like it would cause issues. While I do believe that this would be ideal (with a mix of nested folders like Arc currently has) for power users. I do not see me being able to explain that to someone I am sharing the browser to.

shokinn:
I currently use Sidebery and there I've now a 5 level deep tree.
But I know that I had up to 8 levels.

grafik

mr-cheffy:
> Why not this?
> Screenshot 2025-07-15 194810
> 
>     * Just display the whole folder as an icon, and it can be expanded just like a menu with sub-menu, and replace icon with favicon, label with tab title.
> 
>     * And then replace the inside selected tab's favicon with the previous folder icon (still expandable too).

The problem is how would you reorganize tabs inside it? Or any sort of tab management really. My only guess would be to not let you uncollapse it and just let you open the "search" dialog or don't provide any "spacing" to the left, or at least a really small difference

Bellisario:
> I can see this being a solution but I personally believe that @hoangnhan2ka3 has a better idea. @Bellisario's method seems confusing to new users and like it would cause issues. While I do believe that this would be ideal (with a mix of nested folders like Arc currently has) for power users. I do not see me being able to explain that to someone I am sharing the browser to.

Yeah, I also like a lot the "menu-like" implementation proposed by @hoangnhan2ka3, it could make the folder experience really snappy.\
My suggestion of "sub-spaces" could be an extension of the "menu-like" folders like this:

- to open a folder tab directly, hover on the folder, a "menu" is shown and can click on the pinned tab (original idea)
- to open a folder as "sub-space", when hovering on the folder, instead of clicking on the "menu", the user clicks on the folder itself, switching to its context. "Sub-space" folders act again "menu-like".

This is more intuitive 'cause with the hover effect the user can still "feel" a click (and subsequent action) on a folder is possible, but is pushed to use the snappy folder menu when a "focus on the folder itself" ("sub-folder" opening) it's not needed.

> I don't know if you can relate but I imagine it similar to the current Arc's implementation, but where the folder menu can also expand to sub-folders and where a click on the folder itself won't expand the folder (as it's not needed due to the menu) but actually creates the "sub-space" context, decluttering the current space during usage.

mr-cheffy:
> Yeah, I also like a lot the "menu-like" implementation proposed by @hoangnhan2ka3, it could make the folder experience really snappy. My suggestion of "sub-spaces" could be an extension of the "menu-like" folders like this:
> 
>     * to open a folder tab directly, hover on the folder, a "menu" is shown and can click on the pinned tab (original idea)
> 
>     * to open a folder as "sub-space", when hovering on the folder, instead of clicking on the "menu", the user clicks on the folder itself, switching to its context. "Sub-space" folders act again "menu-like".
> 
> 
> This is more intuitive 'cause with the hover effect the user can still "feel" a click (and subsequent action) on a folder is possible, but is pushed to use the snappy folder menu when a "focus on the folder itself" ("sub-folder" opening) it's not needed.

This is a great idea, but please, let's stay on target because github discussions can get really messy really fast hah

mr-cheffy:
@octaviusz When dragging from a normal tab to a pinned tab, i've changed it so it uses `elementIndex` instead of `_tPos`. That's why it has merge issues. Just letting you know hah

aimryoui:
> The problem is how would you reorganize tabs inside it? Or any sort of tab management really

I have 2 cases:

## I. Easier

### 1. Spreading tabs
#### First try:
- Spreading all tabs inside the parent folder when changing to Collapsed Toolbar (and sticky parent folder icon?), whatever any sub-folders inside it, just spreading all the tabs out.

image

#### Problems:
1. Drag a tab to it while in Collapsed mode, what will happen?
	- Put it as a direct child of Parent folder, then users can reorder it when they're back to Expanded mode?
	- What if they drop the tab in the center of other tabs?

2. Ordering when spreading tabs? Yes, that's a problem:

image

> Can I call the algorithm: From outside to inside, from top to bottom, each sub-folder one by one (?).

### 2. `display: hidden !important` the folder for Collapsed Toolbar 🙂

- Best suggestion if you ask me...
- Although it's a bit unreasonable to hide folders in Collapsed mode without a notification, or rather no way to notify - maybe use the toast with "Turn off Collapsed mode to see folders"?
- I'm not sure if anyone actually uses Collapsed mode, but if they do, they probably don't want to cram a bunch of tabs into it. Sticky essential tabs are enough.
- Case study? Discord? Discord's sidebar only has 1 level of folders, but users have never been happy with it :). So is it worth making things more complicated?

*Edit: 1-level folders may sound great with Collapsed mode, but it will be inconsistent compared with multiple-level folders of the Expanded one?

## II. Harder

- We can't reuse the context-menu component, right? So what about implementing a brand new `reorganizable menu` for that kind of stuff?

*Edit 2:

### Auto-expanding toolbar when drag and drop tab (into folder's zone?):

Based on the menu and sub-menu thing I've said
> Just display the whole folder as an icon, and it can be expanded just like a menu with sub-menu, and replace icon with favicon, label with tab title.

#### First try:
- Act like Compact mode, temporarily expand the toolbar when dragging a tab, so we don't need to implement `drag and hold and wait` function for the new `reorganizable menu` to drop the tab in the right place.

#### Problems:
- Layout shift. Yes, the position of A element in Expanded mode is different from A in Collapsed mode :(

12th-devs:
@octaviusz or @mr-cheffy could we possibly get a To-Do list of this PR so the community can watch the progress?

mr-cheffy:
*now in main post*

bpavuk:
> - [ ] "import bookmarks" should create folders instead of actually importing bookmarks (this needs to be discussed)

my 5 cents on this: pinned tabs must have been bookmarks since the very beginning of Zen. that way, importing existing bookmarks and synchronization of pinned tabs (and once this PR is ready and merged, folders) would be ± painless.

mr-cheffy:
> pinned tabs must have been bookmarks since the very beginning of Zen

What do you mean? Firefox already can sync tabs and groups iirc

MrModest:
@mr-cheffy @octaviusz can you please put the To-Do list into the description of the PR, so people don't need to search it through comments? 🙏

bpavuk:
> > pinned tabs must have been bookmarks since the very beginning of Zen
> 
> What do you mean? Firefox already can sync tabs and groups iirc

the "Synced Devices" section is intended for *open* tabs, yet pinned tabs are closed by default, so Firefox Sync does not pick them up. to work that around, it would be much easier to represent these as bookmarks to Firefox base. also, this way, my pinned tabs will be accessible on mobile and structured as intended, ~~although the UX of getting there is atrocious at least~~

for the outsider like myself, pinned tabs as bookmarks is extremely intuitive.

bpavuk:
plus, I could pin the new tab on mobile just by adding it to the, say, "Zen Spaces/My Space" folder.

CrypticVader:
> > pinned tabs must have been bookmarks since the very beginning of Zen
> 
> What do you mean? Firefox already can sync tabs and groups iirc

The Firefox implementation has been counterintuitive in my experience, since the tab synchronization is device dependent. So if I hop between Zen on multiple devices signed in with the same Mozilla account, then I will not be able to bring over the same open tabs and that leads to a disjointed experience.

z1lV3r:
> "import bookmarks" should create folders instead of actually importing bookmarks (this needs to be discussed)

> pinned tabs must have been bookmarks since the very beginning of Zen

I disagree with this approach. As an Arc user, I rely on having a complete and separate space for my bookmarks. The key distinctions between the concepts are:

- Pinned Tab: Handy tab to load a page _within_ a space.
- Pinned Tab folders: Group open tabs by topic _within_ a space.
- Bookmark: Shortcut to open tabs in _any_ space, independently.
- Bookmark folders: Group bookmarks by topics to open tabs in _any_ space.

An  improvement to evaluate would be to allow users to quickly generate a space from a bookmark folder preserving its URLs and nested folder structure. This would bridge both concepts without removing the flexibility of bookmarks.

bpavuk:
> > "import bookmarks" should create folders instead of actually importing bookmarks (this needs to be discussed)
> 
> > pinned tabs must have been bookmarks since the very beginning of Zen
> 
> I disagree with this approach. As an Arc user, I rely on having a complete and separate space for my bookmarks. The key distinctions between the concepts are:
> 
>     * Pinned Tab: Handy tab to load a page _within_ a space.
> 
>     * Pinned Tab folders: Group open tabs by topic _within_ a space.
> 
>     * Bookmark: Shortcut to open tabs in _any_ space, independently.
> 
>     * Bookmark folders: Group bookmarks by topics to open tabs in _any_ space.
> 
> 
> An improvement to evaluate would be to allow users to quickly generate a space from a bookmark folder preserving its URLs and nested folder structure. This would bridge both concepts without removing the flexibility of bookmarks.

is `Zen Browser` root folder in bookmarks *that* eyestrain?

bartbog:
> "import bookmarks" should create folders instead of actually importing bookmarks (this needs to be discussed)

It would be nice to offer this feature, maybe in an optional way (checkbox while importing?)
For new users used to having only bookmarks, this would be a timesaver

(but I would not consider this to be a requirement for the current PR)

bdbch:
Coming in here as well regarding the Tab vs Bookmark discussion.

I think bookmarks could be a great way to solve problems like syncing essential and pinned tabs. Currently since Firefox sync doesn't differentiate between your pinned tabs, you can't sync those to multiple devices with Zen.

Using bookmarks would allow Zen to basically sync pinned & essential tabs using just default Firefox bookmarking behavior. Instead of pinning a tab, what would happen is that it creates a bookmark in a folder (`Pinned Tabs`, `Essential Tabs` (or `Essential Tabs - WORKSPACE_NAME`) which can then easily be synced by Firefox Sync and reused on other devices.

On top of that we could just use nested bookmark folders for tab folders. In general I see "pinned" and "essentials" more like bookmarks then tabs. The only difference to normal bookmarks is the fluid transformation from "static" bookmark into an active tab.

Tl;dr

### Bookmarks

**Pro Bookmarks**

- Already exists and 100% reflects what "pinned" tabs in for example Arc were meant to be
- Easily syncable via Firefox Sync
- Already supports deep nesting
- Can be used for both pinned and essential tabs
- Can be renamed, have descriptions, etc.
- Even when users move off Zen back to Firefox, they'll be able to enjoy their synced tabs

**Contra Bookmarks**

- Needs extra work to add dynamic replacement of the bookmark in UI with the open tab
- Pollute Firefox bookmarks of sync users with Zen bookmark folders (which can easily be mitigated by putting those bookmarks all into a "Zen" parent bookmark

I created a [discussion](https://github.com/zen-browser/desktop/discussions/9559) related to this.

MrModest:
@bdbch If going with the bookmark approach, then each workspace should have its own dedicated folder so its pinned tab can be living there. Otherwise, how Zen would know which workspace the just synced pinned tab belongs to. 

I'm not arguing, just want to emphisize on the crucial part since people that don't use workspaces tend to ignore these nuances while for others they could be crucial. 

Complimenting to your suggestion, I can see the Bookmarks structure like this:
```
.
├── Zen/
│   ├── Essential Tabs/
│   │   ├── Bookmark1
│   │   └── Bookmark2
│   ├── Workspace1/
│   │   ├── Folder1
│   │   └── PinnedTab2
│   └── Workspace1/
│       ├── Folder1
│       └── PinnedTab2
└── Another Firefox root bookmark
```

**Note:** `Essential Tabs` doesn't belong to any workspace since they should be available across all workspaces

mr-cheffy:
People, please stop suggesting tabs as bookmarks, that's not going to happen. We can already have a separate sync directory just for zen's sync, it doesn't have to be a hacky bookmakrs solutions.

https://github.com/zen-browser/desktop/blob/dev/src/zen/workspaces/ZenWorkspacesSync.mjs

bdbch:
> hacky bookmarks solutions

I'm not sure what's more hacky - tabs behaving like bookmarks or bookmarks behaving like tabs and somehow forcing nested folders into it.

I know you already have an opinion on it yet I still believe bookmarks are a better fit for tab folders. Arc tabs are literally bookmarks designed like tabs.

I'd rather have actual bookmarks that I can take with me and import/export from other browsers and have a already battle-tested solution for a nested folder structure and persistence than something that MAY change again 12 beta versions later.

Why is using Bookmarks not a possibility because it's hacky but building all this extra stuff around pinned tabs to in the end basically replicate bookmarks again?

In the end it's your project and your call, we can just talk and discuss about other solutions that could be possible.

**Edit**

Sorry, just to drive my point home - I tried to replicate a possible structure in Firefox's Sidebar View, how is this a lot different to tab folders except those currently are links and not tabs? Just think away the "meta" stuff I included and you'd end up with something that looks very similar to a tab folder sidebar from Arc.

image

Sorry for being persistent here - you'll probably go with what you already have in mind. I just hope that I could maybe bring in some extra thoughts about this feature when it comes to staying open and using a standard like Bookmarks that a lot of browsers can understand.

mr-cheffy:
I get where you're coming from, but honestly, tabs and bookmarks aren't really comparable, they serve totally different purposes. Tabs are meant to be transient, tied to your current session or workflow. Bookmarks, on the other hand, are a persistent, curated list of saved links. Trying to fuse the two creates more confusion than convenience.

The idea of treating pinned or grouped tabs like bookmarks (or vice versa) just introduces unnecessary complexity. You end up forcing tabs to behave like a storage system, or bookmarks to act like live sessions, neither of which is their intended function.

Just because they look similar in a sidebar doesn’t mean they’re meant to do the same thing.

TL;DR: it just doesn't make sense what you are suggesting.

But if you'd like continue talking about this, lets do it somewhere else, because this PR is just strictly for folders and many people are subscribed to it. They are probably getting emails on every comment.

HenryKleinschmidt:
@mr-cheffy 
I think this idea comes from the way the Arc browser handles tabs/bookmarks
> https://resources.arc.net/hc/en-us/articles/19400407903767-Where-are-Bookmarks

I personally like the idea because I'm a tab hoarder but I get that it does not appeal to everyone.

bdbch:
I already opened a discussion, also don't want to pollute the PR with unrelated feature discussion: https://github.com/zen-browser/desktop/discussions/9559

tiredcisadmin:
> @mr-cheffy I think this idea comes from the way the Arc browser handles tabs/bookmarks
> 
> > https://resources.arc.net/hc/en-us/articles/19400407903767-Where-are-Bookmarks
> 
> I personally like the idea because I'm a tab hoarder but I get that it does not appeal to everyone.

It does come from Arc and is one of the things that made me move away from it. It seemed cool, until you want to export your bookmarks to another browser or do any heavy management/pruning. IMO tab folders should be separate from bookmarks, but have a context menu option to save a tab folder as a bookmark folder, and another to open a bookmark folder as a tab folder.

locales/en-US/browser/browser/zen-folders.ftl
```diff
@@ -4,3 +4,10 @@ zen-folders-search-placeholder =
 
 zen-folders-panel-rename-folder =
     .label = Rename Folder
+
+zen-folders-panel-ungroup-folder =
+    .label = Ungoup Folder
```
pregnor: **Recommended**: Typo **SHOULD** be fixed.

```suggestion
    .label = Ungroup Folder
```

mr-cheffy:
I'll start writing the unit tests for all new features and bug fixes for folders 👍🏽.

Let's hope we can start releasing to twilight this week!

z1lV3r:
images

12th-devs:
This is so unreal. I've waited sense September when I started using Zen for this. Thank you Mauro!!!

mr-cheffy:
This PR will also contain an emojis picker revamp to allow for SVGs.

image

So we can have SVGs inside folders and SVG icons for workspaces along side with the emoji selection:

image

I'll add it to the TODO

12th-devs:
> This PR will also contain an emojis picker revamp to allow for SVGs.
> 
> image
> 
> So we can have SVGs inside folders and SVG icons for workspaces along side with the emoji selection.
> 
> I'll add it to the TODO

A globe, home, and briefcase icon would be good to include as well

zJaaal:
> > This PR will also contain an emojis picker revamp to allow for SVGs.
> > image
> > So we can have SVGs inside folders and SVG icons for workspaces along side with the emoji selection.
> > I'll add it to the TODO
> 
> A globe, home, and briefcase icon would be good to include as well

What about a GitHub Icon? 👀

bpavuk:
any ideas on what to do in collapsed mode?

mr-cheffy:
> any ideas on what to do in collapsed mode?

Right now we are just planning on flatting the folders out. Not definitive, we'll push to twilight and see how it feels. There's not that many options when you have 30px to work with

MiraiDevv:
> Right now we are just planning on flatting the folders out. Not definitive, we'll push to twilight and see how it feels. There's not that many options when you have 30px to work with

A tip would be to use a popover in collapsed mode or make the popover in collapsed mode and instead of expanding downwards, it expands to the side of the sidebar to show the existing folders (I don't know if they are currently using this) if they are not I can implement this if you want......this would be an example of a popover in case anyone doesn't know what it is: [popover](https://ui.shadcn.com/docs/components/popover)

dev-hanzala:
It's done!!!

pseudodionysius:
Come on guys I want folders rn

CrypticVader:
> Come on guys I want folders rn

Why don't you chip in and help the devs who are volunteering to work on this then? One way to do that would be to be patient.

LosFarmosCTL:
> Come on guys I want folders rn

Well then go do it yourself, these are people working on it in their own time, completely for free and there are still people that feel like they're entitled to something and on top of that be disrespectful about it, I genuinely cannot understand wtf is going on in your head...

MiraiDevv:
> Come on guys I want folders rn

I have an idea, let's take out all the volunteer devs who are doing this feature, let's leave only you doing it and we'll all say "come on, do it already", damn it, honestly, stop talking shit, if you want this now go and do it yourself.

sayrusgit:
Idk why people above bitching about [NathOrmond](https://github.com/NathOrmond)'s comment. Obviously he just wanted express how excited and hyped he is about this feature (as we all are). 

Thanks to Mauro and other contributors for making this possible ❤️

mr-cheffy:
Sorry guys, last thing remaining would be to open previous folders on new windows... then it'll get merged. I'll do this first thing tomorrow

12th-devs:
> Sorry guys, last thing remaining would be to open previous folders on new windows... then it'll get merged. I'll do this first thing tomorrow

Will you start a twilight build after it is merged please?

MiraiDevv:
> Sorry guys, last thing remaining would be to open previous folders on new windows... then it'll get merged. I'll do this first thing tomorrow

Don't worry about it, Mauro! All of your work and that of the other devs is completely voluntary, so we understand that it's hard to find time to do things, especially when they're a little more difficult than normal, like folders. Don't worry, and finish when you can and are 100% ready! Anyone who criticizes or rushes things has nothing to contribute!

mr-cheffy (approved):
LGTM, @octaviusz any remaining things before merge?

octaviusz:
@mr-cheffy No, I think it's all done.

mr-cheffy (approved):

PR #9408, opened 2025-07-11
Fix incorrect tab count in closing confirmation dialog

Jai-JAP:
See #9143


mr-cheffy:
Is there a reason firefox did it to be like this?

Jai-JAP:
Not sure about that but that's surely annoying for zen users as most of them actively use essentials and pinned tabs and this issue causes windows to close when it shouldn't.

PR #9588, opened 2025-07-23
fix: Fix copy button not working in calc and unit conversion results

### LINKED ISSUE #9519: Copying from URL Bar Calculator Not Working
### Preliminary Checks

### What happened?

When typing some kind of calculation into the URL bar, the result of the calculation appears as a suggested result and implies that clicking on it will copy the result, however this doesn't happen.

### Expected behavior

Close URL Bar and copy result to clipboard

### Actual behavior

Close URL Bar but result not copied to clipboard

### Steps to reproduce

1. Open the URL Bar
2. Enter a calculation (e.g. "122+22")
3. Click on the suggested result which has a "copy" icon on it

### Screenshots and videos

Image
The suggested result which should copy the result to clipboard

### Version

1.14.5b

Windows - x64

URL Bar

```shell

```

aenriq:
Resolves race condition where urlbar view close was interrupting engagement events, preventing onEngagement from being called.

Add `setTimeout(..., 0)` to resolve race condition

Fixes https://github.com/zen-browser/desktop/issues/9519

### DIFF
--- a/src/zen/common/ZenUIManager.mjs
+++ b/src/zen/common/ZenUIManager.mjs
@@ -398,17 +398,19 @@ var gZenUIManager = {
       }
 
       if (gURLBar.focused) {
-        gURLBar.view.close({ elementPicked: onSwitch });
-        gURLBar.updateTextOverflow();
+        setTimeout(() => {
+          gURLBar.view.close({ elementPicked: onSwitch });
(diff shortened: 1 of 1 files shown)


mr-cheffy (approved):
We should start writing tests for the URLbar... Thanks!

PR #9685, opened 2025-07-29
Translate: To Turkish

yyamandevrim:
English fields that were left empty have been translated into Turkish.
Çalışma alanları (which means Workspaces) are translated to Alanlar (which means Spaces)
Fixed some typos.


mr-cheffy:
Sorry, we only accept translations directly from crowdin. https://docs.zen-browser.app/contribute/translation

Issue #9565, opened 2025-07-21
theme colors change & become lighter in original window after opening a new private window


wesleynw:
### Preliminary Checks

- [x] I have read and understood the important section above.
- [x] I have searched existing issues and avoided creating duplicates.
- [x] I am not filing an enhancement request.
- [x] I have checked that this issue cannot be reproduced on Mozilla Firefox.
- [x] I have checked that this issue can be reproduced once I removed all my Mods and Custom CSS.

### What happened?

I've noticed since 1.14 (don't remember which patch), that the colors of the browser lighten in the original window after opening a private window.

The colors don't reset to their original state until the browser is closed and reopened.

The first screenshot is when the browser is newly opened, and second is after "cmd+shift+p" to open private window.  "cmd+shift+n" does not result in the same issue. The issue also occurs if using the context menu "New private tab".

<img width="337" height="231" alt="Image" src="https://github.com/user-attachments/assets/7847faca-3d7a-4b98-9571-911e2945d2c1" />
<img width="337" height="211" alt="Image" src="https://github.com/user-attachments/assets/b4fe3617-ef69-4dd8-8093-cb3e75691139" />

Another example:

<img width="332" height="214" alt="Image" src="https://github.com/user-attachments/assets/642343a1-2645-452f-b9c9-13ddfd8c7a0a" />
<img width="331" height="206" alt="Image" src="https://github.com/user-attachments/assets/080c912b-6b34-447a-bb22-1eebe07b7cfb" />

### Expected behavior

Theme colors remain stable when opening private space

### Actual behavior

Theme colors change when opening private space until restart

### Steps to reproduce

1. Open Zen, not theme color
2. Open private space
3. Alt-tab back to original window, see that theme color has changed.

### Version

1.14.5b

### What platform are you seeing the problem on?

macOS - aarch64

### What component is this issue related to?

Mods / Themes

### Relevant log output if applicable

```shell

```

wesleynw:
Update: looks like the theme colors also become _darker_ when changing when compact mode is toggled.

First screenshot is with compact mode off, second is with compact mode on:

<img width="326" height="279" alt="Image" src="https://github.com/user-attachments/assets/c037e940-ddb3-4b39-a34b-0ca30ac549a8" />
<img width="344" height="307" alt="Image" src="https://github.com/user-attachments/assets/32d6ff60-986a-4c06-ac47-0378672a4521" />

mr-cheffy:
[...]

PR #9722, opened 2025-08-02
fix: legacy gradients setting being ignored after opening a private window

ImDarkTom:
Fixes theme colors becoming brighter when opening a new private window if the user has legacy gradients on. Addresses #9565.

## Technical: 
In `ZenGradientGenerator.mjs` when `isLegacyVersion` is initialized, it is set to true if `zen.theme.gradient-legacy-version` is 0, however after a new private window is opened, `isLegacyVersion` is becomes undefined, causing the theme colors to change and become brighter if the user has the legacy gradients pref enabled. This addresses the problem by instead reading directly from the pref value directly anytime a workspace change is triggered (such as when opening a private window), giving a correct value.

I plan to investigate further why `isLegacyVersion` becomes undefined when opening a private window but for now I’ve implemented this fix to resolve this UI bug that has been affecting me and others.

### DIFF
--- a/src/zen/workspaces/ZenGradientGenerator.mjs
+++ b/src/zen/workspaces/ZenGradientGenerator.mjs
@@ -1448,7 +1448,11 @@
         if (dominantColor) {
           const primaryColor = this.getAccentColorForUI(dominantColor);
           browser.document.documentElement.style.setProperty('--zen-primary-color', primaryColor);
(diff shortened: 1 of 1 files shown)


mr-cheffy (approved):

mr-cheffy:
@ImDarkTom could you please run `npm run pretty`?

Issue #9746, opened 2025-08-04
Option+Cmd+E shortcut is assigned twice in standard settings


simon-mnk:
### Preliminary Checks

- [x] I have read and understood the important section above.
- [x] I have searched existing issues and avoided creating duplicates.
- [x] I am not filing an enhancement request.
- [x] I have checked that this issue cannot be reproduced on Mozilla Firefox.
- [x] I have checked that this issue can be reproduced once I removed all my Mods and Custom CSS.

### What happened?

The shortcut Option+Cmd+E is assigned for both 'Forward Workspace' and 'Toggle Network Monitor'. These settings are standard and stay when the shortcuts are reset.

### Expected behavior

Option+Cmd+E should switch to next workspace

### Actual behavior

Option+Cmd+E toggles the network monitor

### Steps to reproduce

Use the Option+Cmd+E shortcut

### Version

1.14.9b (64-bit)

### What platform are you seeing the problem on?

macOS - aarch64

### What component is this issue related to?

Settings

### Relevant log output if applicable

```shell

```

anirban6996:
I am changing the default shortcut for forwarding workspace

PR #9769, opened 2025-08-06
Fix: Resolve shortcut conflict for forwarding workspace 

anirban6996:
I changed the modifier key from 'E' to 'A' as in issue #9746 there was the same default key for dev network tool and forward workspace shortcuts.

### DIFF
--- a/src/zen/kbs/ZenKeyboardShortcuts.mjs
+++ b/src/zen/kbs/ZenKeyboardShortcuts.mjs
@@ -668,7 +668,7 @@ class nsZenKeyboardShortcutsLoader {
     newShortcutList.push(
       new KeyShortcut(
         'zen-workspace-forward',
-        'E',
+        'A',
         '',
         ZEN_WORKSPACE_SHORTCUTS_GROUP,
         nsKeyShortcutModifiers.fromObject({ accel: true, alt: true }),


mr-cheffy:
What about just alt+E or just ctrl+E? for easier access

anirban6996:
> What about just alt+E or just ctrl+E? for easier access


I have to check if they are already occupied or used by other shortcuts then I will update it.

anirban6996:
> What about just alt+E or just ctrl+E? for easier access

Alt + E is a standard system level shortcut it may conflict...Or might cause unexpected behavior in macOS and other operating system....So Opt+Cmd+A is a safer option

anirban6996:
We can experiment with it tho, the options are open

anirban6996:
Also Ctrl + E moves the cursor to end of line I guess. It might conflict?

mr-cheffy:
Do you happen to know which ones arc uses?

anirban6996:
> Do you happen to know which ones arc uses?

Cmd + Opt + Right Arrow (For forward workspace in ARC)

mr-cheffy:
Why not use that then?

anirban6996:
I will check if it's used somewhere and update it accordingly

anirban6996:
https://github.com/zen-browser/desktop/issues/9779

Also there is this

anirban6996:
Do i need to change it to rightarrow & leftarrow?

mr-cheffy:
I think that would be the best option

PR #9821, opened 2025-08-09
add Shortcut for Spliting, Expanding and Closing Glance

dkr91:
Adds Customizable Shortcuts as seen in  #5079

### DIFF
--- a/src/zen/kbs/ZenKeyboardShortcuts.mjs
+++ b/src/zen/kbs/ZenKeyboardShortcuts.mjs
@@ -650,6 +650,40 @@ class nsZenKeyboardShortcutsLoader {
         'zen-compact-mode-shortcut-show-toolbar'
       )
     );
+    // Glance mode's KeySet
+    newShortcutList.push(
+      new KeyShortcut(
+        'zen-glance-expand',
+        'Enter',
+        '',
+        '',
+        '',
+        'cmd_zenGlanceExpand',
+        'zen-glance-shortcut-expand'
+      )
+    );
+    newShortcutList.push(
+      new KeyShortcut(
+        'zen-glance-split',
+        'S',
+        '',
+        '',
+        '',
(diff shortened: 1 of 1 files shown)


src/zen/kbs/ZenKeyboardShortcuts.mjs
```diff
@@ -650,6 +650,40 @@ class nsZenKeyboardShortcutsLoader {
...
+        'zen-glance-split',
+        'S',
+        '',
+        '',
+        '',
+        'cmd_zenGlanceSplit',
+        'zen_glance-shortcut-split'
+      )
+    );
+    newShortcutList.push(
```
mr-cheffy, line 676: Esc is already a shortcut, purposefully done on the content process. I'd do the same with the other shortcuts

mr-cheffy (changes requested):

src/zen/kbs/ZenKeyboardShortcuts.mjs
dkr91: What content process. were would I need to add 'ESC'ape instead? thanks for the input.

mr-cheffy: The ESC key is arleady added, im saying that the other shortcuts should be added there as well.

peteygao: Are you referring to [this line](https://github.com/zen-browser/desktop/blob/48ab77ef5279e856e73ad5a7204d19351a7b6d14/src/browser/base/zen-components/ZenGlanceManager.mjs#L41)? The Esc key is hardcoded in this case.

How would you add the other hotkeys while making the hotkeys customiseable in `ZenGlanceManager.mjs`?

dkr91:
closed with PR10271

PR #9838, opened 2025-08-10
Fix: New Tab Menu Not Displayed During Fullscreen Video Playback

### LINKED ISSUE #5229: New Tab Menu Not Displayed During Fullscreen Video Playback
### Captchas

### What happened?

**Description:**
When attempting to open a new tab while playing a video in fullscreen mode with the new feature enabled, the new tab menu does not appear, and the video continues playing uninterrupted. The expected behavior is for the browser to either exit fullscreen mode to display the new tab menu or remain in fullscreen while rendering the new tab menu overlay.

**Steps to Reproduce:**
1. Enable the new feature in Zen Browser settings. (it should be enabled by default with the latest update 1.7.6b)
2. Navigate to a website with video content.
3. Play a video and switch to fullscreen mode.
4. Attempt to open a new tab using the designated shortcut (CTRL+T).

**Expected Behavior:**
Upon initiating the action to open a new tab during fullscreen video playback, the browser should either:
- Exit fullscreen mode and display the new tab menu.
- Maintain fullscreen mode and render the new tab menu as an overlay.

**Actual Behavior:**
The new tab menu does not appear (I belive it is rendered in the background and not visible due to the video being in fullscreen), and the video continues playing in fullscreen mode without any indication of the menu being opened (if you type the keypresses get registered by the website not by the menu).

**Environment:**
- **Operating System:** *Arch Linux x86_64 Kernel: 6.13.1-arch2-1*
- **Extensions/Add-ons Installed:** *Dark Reader | uBlock Origin*

### Reproducible?

### Version

1.7.6b

Other

```shell

```

luisbettencourt:
Fixes https://github.com/zen-browser/desktop/issues/5229 by exiting fullscreen mode when opening the new tab menu.


mr-cheffy (changes requested):
I dont think exiting fullscreen is the best option to be honest. Maybe we should see why it isnt showing in the first place.

luisbettencourt:
> I dont think exiting fullscreen is the best option to be honest. Maybe we should see why it isnt showing in the first place.

I tweaked the css that hides the sidebar when the browser is in fullscreen mode. Let me know if you think this is the correct approach.

[Screencast From 2025-08-11 15-39-43.webm](https://github.com/user-attachments/assets/68ee5adc-6a58-442b-aabd-f77bd6634760)

luisbettencourt:
The URL Bar placement looks a bit weird when the Behavior setting is 'Normal' instead of 'Floating only when typing' or 'Always floating', as it slightly clips the top border. Should we always center it when in fullscreen mode or try to fix its position?

image

mr-cheffy (changes requested):
It should always open in floating mode if it's fullscreen

PR #9865, opened 2025-08-13
fix: Shift navbar down when accesing it on macOS in fullscreen (#8757)

### LINKED ISSUE #8757: Macos fullscreen doesn't allow to access the content at the top
### Preliminary Checks

### What happened?

The top bar (that includes all extensions) is almost impossible and very frustrating to access when in compact mode in OSX.

This is the same as #1903 and #3058 (both closed without any response despite being a legitimate bug that is 1-zen only, 2-workflow breaking, 3-frustrating)

The behavior is perfect in Firefox while it is broken in zen. Here's a side by side video of zen and Firefox.

### Expected behavior

I can access the top bar with all my pinned extensions without turning off fullscreen

### Actual behavior

Almost impossible to access the top bar without turning off fullscreen

### Steps to reproduce

1. Use mac
2. enable compact mode ("hide both" option)
3. try to hover your mouse over any extension

### Screenshots and videos

This first video shows me using Zen in compact mode and how the top bar can't be accessed despite hovering over it

https://github.com/user-attachments/assets/cf6934f0-025c-48ef-9f5e-32e32c05b550

This second video shows me using Firefox in full screen and being able to access the top bar when hovering over it

https://github.com/user-attachments/assets/53cbdb08-a877-4536-9ae1-7c3151284a72

### Version

1.12.8b

macOS - aarch64

Compact Mode

```shell

```

npinochet:
Shift nav bar down when accessing it on macOS while in fullscreen, allowing it to be shown when the macOS top bar appears.
I tried to keep this PR simple, but the nav bar kept auto hiding when the macOS top bar appears making it difficult to access. The changes made in `ZenCompactMode.mjs` fixes that.

Fixes https://github.com/zen-browser/desktop/issues/8757

EDIT: This is how it currently works
https://github.com/user-attachments/assets/5ad31673-366a-4c65-b6ba-2d2b2a2036ff

### DIFF
--- a/src/browser/base/content/browser-fullScreenAndPointerLock-js.patch
+++ b/src/browser/base/content/browser-fullScreenAndPointerLock-js.patch
@@ -2,18 +2,18 @@ diff --git a/browser/base/content/browser-fullScreenAndPointerLock.js b/browser/
 index bef746dc594ad974c91075cf3657c199f8f8830f..bb671341b6715c42df76f7298ba55e1fac73f33b 100644
 --- a/browser/base/content/browser-fullScreenAndPointerLock.js
 +++ b/browser/base/content/browser-fullScreenAndPointerLock.js
-@@ -424,10 +424,10 @@ var FullScreen = {
+@@ -423,10 +423,10 @@ var FullScreen = {
+     shiftSize = shiftSize.toFixed(2);
      gNavToolbox.classList.toggle("fullscreen-with-menubar", shiftSize > 0);
  
-     let transform = shiftSize > 0 ? `translateY(${shiftSize}px)` : "";
+-    let transform = shiftSize > 0 ? `translateY(${shiftSize}px)` : "";
 -    gNavToolbox.style.transform = transform;
 -    gURLBar.textbox.style.transform = gURLBar.textbox.hasAttribute("breakout")
 -      ? transform
--      : "";
-+    //gNavToolbox.style.transform = transform;
-+    //gURLBar.textbox.style.transform = gURLBar.textbox.hasAttribute("breakout")
-+    //  ? transform
-+    //  : "";
++    const padding = shiftSize > 0 ? `${shiftSize}px` : "";
++    const appContentNavbarWrapper = document.getElementById('zen-appcontent-navbar-wrapper');
++    appContentNavbarWrapper.style.paddingTop = gURLBar.textbox.hasAttribute("breakout")
++      ? padding
+       : "";
      if (shiftSize > 0) {
        // If the mouse tracking missed our fullScreenToggler, then the toolbox
-       // might not have been shown before the menubar is animated down. Make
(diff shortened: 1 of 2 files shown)


src/browser/base/content/browser-fullScreenAndPointerLock-js.patch
```diff
@@ -2,18 +2,18 @@ diff --git a/browser/base/content/browser-fullScreenAndPointerLock.js b/browser/
...
 -    gURLBar.textbox.style.transform = gURLBar.textbox.hasAttribute("breakout")
 -      ? transform
--      : "";
-+    //gNavToolbox.style.transform = transform;
-+    //gURLBar.textbox.style.transform = gURLBar.textbox.hasAttribute("breakout")
-+    //  ? transform
-+    //  : "";
++    const padding = shiftSize > 0 ? `${shiftSize}px` : "";
++    const appContentNavbarWrapper = document.getElementById('zen-appcontent-navbar-wrapper');
++    appContentNavbarWrapper.style.paddingTop = gURLBar.textbox.hasAttribute("breakout")
```
mr-cheffy, line 15: Shoudnt we check if it's double toolbar as well?

mr-cheffy (changes requested):

src/browser/base/content/browser-fullScreenAndPointerLock-js.patch
npinochet: Good point, I didn't consider that case, I'll do some tests and try to add it

mr-cheffy: I wonder if instead of this fullscreen we can just not hide it at all, like what chromium browsers do.

npinochet: You mean like not showing the macOS title bar that shifts down at all? That could work, the window minimize, maximize and close buttons would be lost and should be moved to the toolbar though.
As it is, this fix takes the same solution that firefox did, but I agree the chromium way is more simple. I'll try experimenting to see if I can replicate it.

karimboubouh:
Can you please merge this, its painful to work with Zen on full screen on compact mode.

mr-cheffy (changes requested):
@npinochet could you please resolve git conflicts so it can get merged into twilight?

npinochet:
~~Later today~~ Tomorrow I'll resolve the merge conflicts. I still believe the way chrome/safari handles fullscreen is a better solution, I tried making it work, but my lack of Objective C skills and compile times made me give up.

karimboubouh:
Any issues still in this PR ?

src/browser/base/content/browser-fullScreenAndPointerLock-js.patch
mr-cheffy: Has this check been done? Im not sure where that is

Gittykittymitty:
In version 1.18.5b I'm also experiencing this problem. 
I belive easiest way to fix this would be doing Firefox's approach, where the whole tab bar shifts. The best method I believe would be to not use this drop down title bar at all and just keep the traffic lights window buttons in the same place (with yellow greyed out).

mr-cheffy:
There seems to be various git conflicts, its my fault and im sorry for taking so long to review, its just a lot going on. Please do reopen if you are still interested in contributing. Thanks and sorry!

PR #6243, opened 2025-03-09
Hide sponsored and pinned shortcuts from the URL bar by default

noaione:
Added toggle (default to disabled) to remove pinned and sponsored results from the URL bar.

See: https://github.com/zen-browser/desktop/discussions/6186

Demo:

https://github.com/user-attachments/assets/8fe23e10-0bab-4294-bf40-a8ea4889a7b6

(Video has broken text since it's missing l10n stuff)

---

l10n PR: https://github.com/zen-browser/l10n-packs/pull/137

PR #9883, opened 2025-08-14
feat: add preference to enable compact mode on browser fullscreen

marciosobel:
This PR adds a new preference `zen.view.compact.enable-on-fullscreen`, which enables compact mode whenever the browser window goes on fullscreen (e.g. when pressing F11).

This is for situations I commonly find myself into, like when I have more than one window on my screen or when I'd like to use the split tabs feature (sometimes even both!), the non-compact mode takes up too much space.

https://github.com/user-attachments/assets/f2d93c1e-1d2a-409e-b2e0-ecf445c09fa2


marciosobel:
Also, I thought about creating a "Compact mode" section on the preferences page, but after seeing the comment at https://github.com/zen-browser/desktop/pull/6243#issuecomment-2708841561, I opted to leave it to the advanced preferences page. Lmk if this should have it's own section!

src/zen/compact-mode/ZenCompactMode.mjs
```diff
@@ -57,10 +57,18 @@ var gZenCompactModeManager = {
     const tabIsRightObserver = this._updateSidebarIsOnRight.bind(this);
     Services.prefs.addObserver('zen.tabs.vertical.right-side', tabIsRightObserver);
 
+    const enableOnFullscreenObserver = this._updateEnableOnFullscreen.bind(this);
+    Services.prefs.addObserver('zen.view.compact.enable-on-fullscreen', enableOnFullscreenObserver);
```
mr-cheffy: I think we should use `XPCOMUtils.defineLazyPreferenceGetter` here instead of adding our own observers.

mr-cheffy (changes requested):
Looks good! Some things:

1. I think this is important enough to have it's own checkbox on the settings page.
2. Could it be possible if a simple test is written for this specifically?

Thanks!

ekorchmar:
This would close #614 in the best possible way. I now have to follow up F11 keypress with Ctrl+Alt+C, which is frustrating.

marciosobel:
I wanted to come back to this PR, but I am unable to build the browser again (I run out of RAM every time, even if it's running in one thread). Maybe some other time!

PR #9920, opened 2025-08-18
fix: After expanding a folder, has-active folders show extra items

octaviusz:


mr-cheffy:
We should really start writing tests for these @octaviusz.

Now that we have this selected ID, I dont think we need the `has-attribute` anymore, no?

mr-cheffy (approved):

PR #9926, opened 2025-08-18
Fix ```:root:has()``` requirements with mods.

CosmoCreeper:
This pull request makes it so that mods will no longer need to have a ```:root:has()``` selector for dropdowns by injecting ```#theme-{name}``` elements into a ```#themes-container``` element, allowing mods to do this:
```css
#themes-container:has(#theme-{name}[property="value"]) ~ #element-to-style {
  property: value;
}
```
This snippet basically runs a ```:has()``` on the themes-container element, then looks for siblings of that container, and because the themes-container is at the root of the body, the mod will be able to select any element that they would be able to style and then either style it or style its child element, effectively providing the same ability that ```:root:has()``` provides currently.

Also, due to the way that this works, it is backwards-compatible with other mods, so mods that still use ```:root:has()``` will work.

### DIFF
--- a/src/zen/mods/ZenMods.mjs
+++ b/src/zen/mods/ZenMods.mjs
@@ -155,7 +155,7 @@
       for (const browser of nsZenMultiWindowFeature.browsers) {
         const container = browser.document.getElementById('themes-container');
         let hasEnabledMods = false;
-        
+
         for (const { enabled, preferences, name } of modsWithPreferences) {
           const sanitizedName = this.sanitizeModName(name);
 
--- a/src/zen/mods/ZenMods.mjs
+++ b/src/zen/mods/ZenMods.mjs
@@ -189,7 +189,7 @@
                     element.style.display = 'none';
                     element.setAttribute('id', sanitizedName);
 
-                    browser.document.body.appendChild(element);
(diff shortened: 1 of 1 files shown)


mr-cheffy:
I don't think this will work. The element would have to be a sibling. And why not just use `-moz-pref` anyways?

CosmoCreeper:
Cause ```-moz-pref``` does not support strings. Also, any element the mod would want to style would be a sibling or a child of a sibling considering the container is appended to the body element rather than in some sub-element.

CosmoCreeper:
Nvm, I'm an idiot. I found out you can do this:
```css
#theme-SuperPins[uc-show-separator="never"] ~ #zen-main-app-wrapper .pinned-tabs-container-separator {
  display: none;
}
```
rather than:
```css
:root:has(#theme-SuperPins[uc-show-separator="never"]) .pinned-tabs-container-separator {
  display: none;
}
```
meaning that this pull request is not needed.

but regardless, i think that the documentation for this could be changed, unless the user wants to style the ```html``` or ```body``` tag, which would be one of the few that wouldn't work with the sibling operation.

EDIT: oh, idk if the theme elements are appended to the end of the body tag or the beginning, so prepend would fix this.

PR #9994, opened 2025-08-25
Added a setting to disable the re-opening of the main window in the workspace where it was previously closed.

### LINKED ISSUE #8252: Normal windows always open in previously used Windows virtual desktop instead of current one
### Captchas

### What happened?

🐛 **Bug**: Zen browser always opens first window in previous desktop

🧩 **Description**
There appears to be an issue with how the normal Zen browser window chooses which virtual desktop to open in, under specific circumstances on Windows 11.

**Steps to Reproduce:**
If no Zen window is currently open, launching Zen from any desktop opens it in that current desktop ✅.

After that, if one window is open, launching Zen again opens new windows correctly in the current desktop ✅.

However, if all Zen windows are closed, and then Zen is launched again, the first new window always opens in the desktop that previously had the last active Zen window ❌.

Also, if Zen was last active in Desktop 2, then I close that window, and I then drag Desktop 2 to the position of Desktop 4 (so it is now labeled Desktop 4), the next time I launch Zen from desktop 3 or 2 for example, it will still open in that same desktop (now called Desktop 4), based on where the last active window was — not the current desktop I’m on.

If I delete/close that desktop (to which zen is defaulting to), the behavior resets. Zen will again open wherever it is launched — until the above pattern begins again.

✅ **What Works as Expected**
Zen Private Browsing windows do not have this issue — they always open in the current desktop, even if it's the first window opened.

💡 **Additional Notes**
This behavior is consistent and reproducible regardless of how the browser is launched (shortcut, Start menu, or PowerToys Run).
It seems the browser retains a reference to the desktop from the last closed window and defaults back to it when opening the next first window — except in private mode.

💻 **System Info**
Zen Version: 1.12.3b (Firefox 138.0.1) (64-bit)
OS: Windows 11 Version 24H2 (OS Build 26100.3775)
Architecture: 64-bit OS, x64-based processor
Launch Methods Tested: Desktop shortcut, Start menu, PowerToys Run
Affected: Only normal windows (not private windows)

### Version

1.12.3b

Windows - x64

Other

```shell

```

blouflab:
Added a setting to disable the re-opening of the main window in the workspace where it was previously closed.

1. **about:config**
2. Create a key : **browser.sessionstore.restore_workspace**, boolean
3. **If true or unset :** the browser main window will open on the last workspace it has been closed
4. **If false :** the browser main window will open on the current workspace

Tested on Linux/KDE.

Resolves : #8252


mr-cheffy (changes requested):
Please name it `zen.sessionstore.restore_workspace` and add it to the `prefs/` dir. I dont think it should be handled here, but rather on the actual session restore functionality.

Thanks!

blouflab:
Oh lalala ... I already cleaned everything waiting for the official update 😓

blouflab:
> Please name it `zen.sessionstore.restore_workspace` and add it to the `prefs/` dir. I dont think it should be handled here, but rather on the actual session restore functionality.
> 
> Thanks!

ok go on 👍

prefs/zen.yaml
```diff
@@ -28,3 +28,6 @@
 
 - name: zen.tabs.open-pinned-in-new-tab
   value: true
+
+- name: zen.sessionstore.restore_workspace
+  value: false
```
mr-cheffy, line 33: Should be `true` by default. Also, please name it `restore-workspace` (sorry about that)

src/firefox-patches/disable-workspace-restore.patch
```diff
@@ -0,0 +1,14 @@
+--- a/widget/gtk/nsWindow.cpp
++++ b/widget/gtk/nsWindow.cpp
+@@ -3790,6 +3790,11 @@
+   int32_t workspaceID = workspaceIDStr.ToInteger(&rv);
+ 
+   LOG("nsWindow::MoveToWorkspace() ID %d", workspaceID);
++
++  if (!mozilla::Preferences::GetBool("zen.sessionstore.restore_workspace", true)) {
```
mr-cheffy, line 8: This should rather be in https://searchfox.org/firefox-main/source/browser/components/sessionstore/SessionStore.sys.mjs#5737

mr-cheffy (changes requested):

prefs/zen.yaml
blouflab: For the refactoring diy man, i'm new there, i'm not familiar with the ways of the house

mr-cheffy: Sorry?

Propheticus:
This PR is linked to an issue about Zen starting in the wrong Virtual Desktop (OS-level).
The information above however mentions "Workspaces". 
For clarity; we're not talking about the workspaces within Zen, right?

mr-cheffy:
@Propheticus this is about OS workspaces, not the ones inside zen

blouflab:
@Propheticus 
That's it. Wait for @mr-cheffy to do it.

Propheticus:
@blouflab thank you for picking this up and the confirmation.

prefs/zen.yaml
OliverWich: > @mr-cheffy 
> Should be `true` by default. Also, please name it `restore-workspace` (sorry about that)

Just to clarify, why `true` by default?

This would be the current behavior, which I find very confusing and guess other users might as well.
If Zen switches to the last workspace by default without any clear indication (like it does today), it feels like it was not able to launch correctly as even opening it repeatedly does not show any windows since they are all moved away.

I would propose do go with a `false` default, just so that when a user launches Zen, they get Zen and don't have to go searching around for it after it jumped away.
All other applications and browsers (Firefox too) behave this way too. You open the application, it gets shown to you on your current virtual desktop / workspace.
No jumping away / hiding somewhere else by default.

Maybe this is a MacOS / Windows difference thing?
I don't know how MacOS handles workspaces, maybe it takes the focus with it so that the user does not directly notice the jump to a different workspace?
That is the main problem here, as on windows, the freshly launched Zen just vanishes.
Not sure about the different Linux DEs, but in general I feel like any behavior that does not result in a launched, in focus, ready to use and interact with Zen is not good and should be changed.

Just my 2 cents on this, a user who did think Zen was broken in some way after first install because of this (and who is constantly bothered by it every time I launch it and have to search for the window before being able to do any meaningful work :D)

Propheticus: While I agree the behaviour feels strange, the default session restore of Firefox also defaults to restore to the virtual desktop (workspace) where the window was closed.  

> All other applications and browsers (**Firefox too**) behave this way too

This is not true. (at least not in my testing on Windows)

OliverWich: Oh interesting, for me it did always open on the virtual desktop i was on, regardless of where I last closed it...

Admittedly, this is Firefox nightly, but with pretty much default settings.

Issue #6968, opened 2025-03-29
closing tab opened from essentials will leads to blank window


lolipodass:
### Captchas

- [x] I have read the instructions.
- [x] I have searched existing issues and avoided creating duplicates.
- [x] I am not filing an enhancement request.
- [x] I have checked that this issue cannot be reproduced on Mozilla Firefox.
- [x] I have checked that this issue cannot be reproduced once I removed all my Mods and Custom CSS.

### What happened?

When a tab is added within the "Essentials" section, and a further tab is opened from within that new tab, closing the second tab results in a blank browser window instead of returning to the original "Essentials" tab.

https://github.com/user-attachments/assets/50da4ea7-521f-4486-899d-d5f1a507d9db

### Version

1.10.3b but persist from 1.9ish

### What platform are you seeing the problem on?

Windows - x64

### What component is this issue related to?

Tabs

### Relevant log output if applicable

```shell

```

LeMoonStar:
I can confirm this to be an issue with Zen 1.11.1b on Fedora Workstation 41 as well.
This is also occurs if there are other non-pinned and non-essential tabs open.

Do note for reproduction:
- External pages to essentials are opened as glance popup first. This issue occurs if that glance popup is first "expanded" to be its own tab and then closed.
- If the user switches to another tab in-between of opening and closing this tab, the next tab is chosen instead.

gavink97:
This also occurs in the latest version of Zen for MacOS.

For reproduction:
1. Go to an essential tab
2. Open a new tab
3. Close new tab

Expected:
User to return to the essential tab

Result:
User is returned to a non essential / non pinned tab or a blank new page if no tabs are open.

TaeyoungLeeJack:
I guess this phenomenon is also a reason for blank preview windows of the zen browser on the taskbar of Windows OS, which is an issue that I posted yesterday. Even if the essential or pinned tabs are closed, the preview windows still show them as opened but with a blank page.

nh2seven:
This issue still persists on version 1.12.10b, Linux build. Same behavior as reported in previous comments.
Any tab that is opened as a result of expanding a glance tab is not restored in a new session. To be precise, a glance tab that is expanded from a pinned/essentials tab returns to a blank tab after reopening the browser. Happens regardless of which workspace is being used.

HeasHeartfire:
I've been checking this issue regularly and I'm sad to see no updates after all this time.

[...]

PR #10079, opened 2025-08-28
Issue: 6968 Fix essential tab owner relationships

vineeth14:
https://github.com/user-attachments/assets/46f98271-26ac-460e-acbf-0a9eee173faa

These changes are addressing issue [6968](https://github.com/zen-browser/desktop/issues/6968)
 - Protect owner tab during tab switching or tab addition if owner is essential
 - Allow essential tabs to bypass workspace exclusion as legitimate owners
 
 
This prevents Firefox from clearing owner relationships when the owner is an essential tab, ensuring that tabs opened from essential tabs will properly return to their essential parent when closed, regardless of tab switching activity.


mr-cheffy:
Why is the patch so weirdly made? Did you use `npm run export <file>`?

PR #10095, opened 2025-08-29
Supplemental Simplified Chinese Translation

FanXiangTM:


mr-cheffy:
Translations should be done through crowdin, sorry,

PR #10165, opened 2025-08-31
fix: replace essential icons with SVGs in onboarding

reizumii:
This patch replaces the .ico files with SVGs for the essentials selection during first time setup. The icons now scale without being pixelated regardless of scaling. Also, the essential container's height and icon is increased to make them stand out more in the screen.

This change indirectly corrects the Google Calendar icon as their logo is always 31.

The one and only primary motivation behind this change is the adaptive GitHub icon :D

image


mr-cheffy (approved):

mr-cheffy:
Thanks!

PR #10176, opened 2025-09-01
Update ru locale zen-folders.ftl

neohdx:
Added Russian locale for folders


mr-cheffy:
Please use crowdin for translations.

PR #10178, opened 2025-09-01
Update ru locale zen-workspaces.ftl

neohdx:


mr-cheffy:
Please use crowdin for translations.

PR #10201, opened 2025-09-02
fix: hide manage themes option in customize toolbar

reizumii:
The browser still displays the link for managing Firefox themes in Customize Toolbar so this patch fixes that. It can be shown by toggling `zen.theme.disable-lightweight` to `false` in about:config.


mr-cheffy (approved):

mr-cheffy:
Thanks a lot for fixing these small quirks zen have, I don't have time for them so these things really help me out!

PR #10233, opened 2025-09-03
feat: add tada & at icons for ZenEmojiPicker

ThatOneCalculator:
Adds "tada" and "at" icons for the emoji picker (for folders/etc).

- At: useful for email, ATProto, etc
- Tada: useful for party/event planning, celebrations, etc

Previews:

image

image


mr-cheffy:
Could you please send how they look inside folders? They look great!

@ThatOneCalculator

mr-cheffy (approved):

mr-cheffy:
Screenshot 2025-09-0 at 01 24 33

@ThatOneCalculator i've reverted the change since they look way too big compared to the rest

ThatOneCalculator:
ah, are there sizing guidelines? I tried to line them up to the other icons (using skull as a reference)

ThatOneCalculator:
for some reason inkscape did transforms on `<g>` instead of modifying the path itself, fixed that now. re-opened in #10237

PR #10241, opened 2025-09-04
fix: always show HTTP warning icon in single toolbar

reizumii:
While it does display the `http://` part in the address bar, I think it's important to always make the warning symbol always show on HTTP sites in single toolbar mode. Not everybody knows what it means.

Also changed the icon container width to `28px` to center the icon and removed the hardcoded background color as well.

Screenshot 2025-09-04 180111


mr-cheffy (approved):
<3

PR #10271, opened 2025-09-06
feat(glance): Add Ctrl/Cmd+O shortcut to expand Glance; user-configurable via Keyboard Shortcuts; tooltips

KyleKincer:
Adds a default, user-configurable shortcut to expand Glance, and improves Glance button tooltips.

Changes
- Default shortcut: Ctrl+O (Cmd+O on macOS) to expand Glance to a full tab.
- User-configurable: Shortcut is exposed in Settings → Keyboard Shortcuts (under Other).
- Context-aware: The key is enabled only while a Glance is active to avoid hijacking the Open File action when no Glance is open.
- Safety: `fullyOpenGlance()` no-ops when no Glance is active.
- Tooltips: Glance overlay buttons now display helpful tooltips with the current shortcut.
  - Close shows “Esc” even when there isn’t a <key> binding (since Escape is handled in the Glance actor).
  - Expand/Split reflect the user-configured keybinding dynamically.

Implementation
- Keyboard shortcuts: bump KBS version and add migration to include a default `zen-glance-expand` key bound to `cmd_zenGlanceExpand`.
- Dynamic enable/disable: toggle the expand key element while opening/closing Glance.
- Tooltip builder: derive shortcut text from `<key>` elements; fall back to “Esc” for Close.

Related discussion
- https://github.com/zen-browser/desktop/discussions/5079

Testing
- Manual: open Glance, press Ctrl/Cmd+O to expand. With no Glance open, Ctrl/Cmd+O continues to open the file dialog. Tooltips show the expected shortcuts.
- Automated: `python3 scripts/run_tests.py glance/browser_glance_expand.js` (and other glance tests).

### DIFF
--- a/src/zen/glance/ZenGlanceManager.mjs
+++ b/src/zen/glance/ZenGlanceManager.mjs
@@ -617,6 +617,10 @@
     }
 
     async fullyOpenGlance({ forSplit = false } = {}) {
+      // If there is no active glance, do nothing
+      if (!this.#currentGlanceID || !this.#currentTab) {
+        return;
+      }
       this.animatingFullOpen = true;
       this.#currentTab.setAttribute('zen-dont-split-glance', true);
 
--- a/src/zen/glance/ZenGlanceManager.mjs
+++ b/src/zen/glance/ZenGlanceManager.mjs
@@ -27,38 +27,8 @@
         .addEventListener('click', this.onOverlayClick.bind(this));
       Services.obs.addObserver(this, 'quit-application-requested');
 
-      // Intercept Accel+O to override Firefox's default Open File dialog when Glance is active
-      // Use capture to ensure we run before the browser default handlers
-      window.addEventListener(
-        'keydown',
-        (event) => this.onKeydownOverride(event),
-        { capture: true }
-      );
     }
 
-    onKeydownOverride(event) {
-      // Only act on Accel+O (Ctrl+O on Win/Linux, Cmd+O on macOS)
-      // and only when there is an active Glance session
-      const isAccel = event.ctrlKey || event.metaKey;
-      if (!isAccel) {
-        return;
-      }
-      if ((event.key || '').toLowerCase() !== 'o') {
-        return;
-      }
-      // If there is no active glance, do not interfere with default behavior
-      if (!this.#currentGlanceID || !this.#currentTab) {
-        return;
-      }
-      // Prevent the default Firefox Open dialog and trigger expand
-      event.preventDefault();
-      event.stopPropagation();
-      try {
-        this.fullyOpenGlance();
-      } catch (e) {
-        console.error('Error expanding Glance from Accel+O:', e);
-      }
-    }
 
     handleMainCommandSet(event) {
       const command = event.target;
--- a/src/zen/glance/ZenGlanceManager.mjs
+++ b/src/zen/glance/ZenGlanceManager.mjs
@@ -26,9 +26,7 @@
         .getElementById('tabbrowser-tabpanels')
         .addEventListener('click', this.onOverlayClick.bind(this));
       Services.obs.addObserver(this, 'quit-application-requested');
-
     }
-
 
     handleMainCommandSet(event) {
       const command = event.target;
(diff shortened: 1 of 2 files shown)


src/zen/glance/ZenGlanceManager.mjs
```diff
@@ -116,6 +119,52 @@
...
+          if (commandId === 'cmd_zenGlanceClose') {
+            if (!combo) combo = 'Esc';
+            else if (!combo.toLowerCase().includes('esc')) combo = `${combo} / Esc`;
+          }
+          const text = combo ? `${label} (${combo})` : label;
+          el.setAttribute('tooltiptext', text);
+          el.setAttribute('aria-label', text);
+        };
+
+        setTooltip(closeBtn, 'Close Glance', 'cmd_zenGlanceClose');
```
mr-cheffy: These should be translated

src/zen/glance/ZenGlanceManager.mjs
```diff
@@ -116,6 +119,52 @@
         .getElementById('zen-glance-sidebar-template')
         .content.cloneNode(true);
       const container = newButtons.querySelector('.zen-glance-sidebar-container');
+      // Set helpful tooltips, including the current shortcut if available
+      try {
+        const closeBtn = newButtons.querySelector('.zen-glance-sidebar-close');
+        const expandBtn = newButtons.querySelector('.zen-glance-sidebar-open');
+        const splitBtn = newButtons.querySelector('.zen-glance-sidebar-split');
+
+        const buildShortcutText = (keyElem) => {
```
mr-cheffy: This should rather be moved into a shortcut generic space, like on nsZenKeyboardShortcut...

mr-cheffy (changes requested):

KyleKincer:
@mr-cheffy thanks for the feedback, updated :+1:

src/zen/glance/ZenGlanceManager.mjs
```diff
@@ -26,6 +26,9 @@
         .getElementById('tabbrowser-tabpanels')
         .addEventListener('click', this.onOverlayClick.bind(this));
       Services.obs.addObserver(this, 'quit-application-requested');
+
+      // Disable expand shortcut by default when no glance is active
+      document.getElementById('zen-glance-expand')?.setAttribute('disabled', true);
```
mr-cheffy: Why is it being set as disabled if we already have the guard check at the start of "fulylExpandGlance"?

mr-cheffy (changes requested):
I'd honestly just merge the shortcut part and make a different PR for tooltips and text...

KyleKincer:
@mr-cheffy thanks again-- made those changes and reverted the tooltips and internationalizations-- Will submit that separately. 
Weirdly, it was working with these changes initially, but today in my testing the shortcut was being consumed by Firefox's default Open File command. I added some code to consume the shortcut before Firefox can and it's working again-- let me know if you prefer a different approach

mr-cheffy (changes requested):
Instead of this, we should change the `accel+O` to another keybind as this is not really used...

KyleKincer:
@mr-cheffy Change the Expand Glance shortcut, or the Open File shortcut from `accel+O`?

mr-cheffy:
Open file shortcut. Maybe add an extra combination like ctrl+shift+o? (I can't check if that's taken as well right now)

KyleKincer:
Sounds good, done. I also removed the changes to `ZenGlanceManager` as I don't think they were necessary at all

mr-cheffy (approved):
Looks good! if you try to do that shortcut without being on glance, does any error occour?

nexxai:
> Instead of this, we should change the `accel+O` to another keybind as this is not really used...

I can't speak for anyone else, but I've been looking for this keyboard shortcut since coming over from Arc, and CMD+O is what they used there, so if it can default to that (or at least have it as a configurable option), that would be great.

KyleKincer:
No user-facing errors for sure. I will confirm nothing in the console when I'm home from my day job 👍

mr-cheffy:
What about experimenting with simply removing the shortcut to open files?

KyleKincer:
@mr-cheffy Good call, I doubt people would miss it. We may need to vibe check after release. Current PR has it completely removed. There was an error triggering the shortcut with no glance active-- Added the minimal fix for that as well

mr-cheffy (approved):

mr-cheffy:
Good then! I'll merge to twilight this night and we'll see if there are any regressions. Thanks!

PR #10325, opened 2025-09-11
Enable workspace horizontal scrolling compatibility for Logitech mice

### LINKED ISSUE #8649: Horizontal Scrolling with mouse wheel is not working on Mac
### Captchas

### What happened?

I use the Logitech MX Master 3s with a Mac Mini and I cannot use the horizontal scroll wheel to switch between workspaces (Neither does CMD+Scroll or Shift+Scroll work).

It works on my Windows machine (with the same mouse) and my Macbook (via touchpad), so having to click the workspace icons manually really is a hassle when you're used to the functionality from other environments.

### Version

1.12.8b

macOS - aarch64

Workspaces

```shell

```

xxxbrian:
## Problem Description

Fixes #8649

Logitech mice horizontal scroll wheels don't work for workspace navigation on macOS.

## Root Cause Analysis

The original `deltaMode !== 1` check was **intentionally designed** to prevent macOS trackpad momentum scrolling issues:

### Trackpad Momentum Scrolling Problem
- When users swipe once on Mac trackpads, the system generates dozens of continuous `deltaMode=0` events
- These events continue for several hundred milliseconds after the user stops touching the trackpad
- Without filtering, this causes unintended multiple workspace switches from a single gesture

However, Logitech horizontal scroll wheels also send `deltaMode=0` events instead of the expected `deltaMode=1`, causing them to be filtered out along with trackpad momentum events.

### Implementation
This implements a targeted compatibility layer that:

1. **Expands accepted `deltaMode` values**: Changes from `deltaMode !== 1` to `deltaMode > 1`
2. **Adds momentum scrolling detection**: Filters consecutive `deltaY: 0` events based on timing
3. **Improves direction detection**: Uses proportional thresholds to prevent accidental diagonal scrolling

### Event Pattern Recognition
- **Logitech mice**: Send discrete `deltaY: 0` events with sufficient time gaps (>200ms) → Allowed
- **Mac trackpad momentum**: Sends rapid consecutive `deltaY: 0` events (<200ms apart) → Filtered out
- **Diagonal trackpad gestures**: Only clear directional intent (2:1 ratio) triggers workspace switching

### Long-term Solution
A more elegant approach would follow Arc browser's progressive scroll handling with gradual, responsive workspace transitions. This would require significant animation system refactoring.

Workplace behavior on Arc

### Standards Progress
The W3C has proposed adding an `isInertialScrolling` property to `WheelEvent` [W3C UI Events Issue w3c/pointerevents#587](https://github.com/w3c/pointerevents/issues/587), which would provide a native way to distinguish user-initiated vs. momentum scrolling events.

## Testing

Tested with Logitech MX Master on macOS and MacBook trackpad. Not test on Windows/Linux devices.

### DIFF
--- a/src/zen/workspaces/ZenWorkspaces.mjs
+++ b/src/zen/workspaces/ZenWorkspaces.mjs
@@ -616,30 +616,42 @@ var gZenWorkspaces = new (class extends nsZenMultiWindowFeature {
           }
         }
 
-        const isVerticalScroll = event.deltaY && !event.deltaX;
+        const absX = Math.abs(event.deltaX || 0);
+        const absY = Math.abs(event.deltaY || 0);
 
-        //if the scroll is vertical this checks that a modifier key is used before proceeding
-        if (isVerticalScroll) {
-          const activationKeyMap = {
-            ctrl: event.ctrlKey,
-            alt: event.altKey,
-            shift: event.shiftKey,
-            meta: event.metaKey,
-          };
+        const isVerticalScroll = absY > absX * 2;
+        const isHorizontalScroll = absX > absY * 2;
 
-          if (
-            this.activationMethod in activationKeyMap &&
-            !activationKeyMap[this.activationMethod]
-          ) {
-            return;
-          }
+        const activationKeyMap = {
+          ctrl: event.ctrlKey,
+          alt: event.altKey,
+          shift: event.shiftKey,
+          meta: event.metaKey,
+        };
+        const modifierActive =
+          this.activationMethod in activationKeyMap && activationKeyMap[this.activationMethod];
+
+        let delta;
+        let shouldProceed = false;
+
+        if (isVerticalScroll && modifierActive) {
+          // scroll is vertical + modifier key
+          delta = event.deltaY;
+          shouldProceed = true;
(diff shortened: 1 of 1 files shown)


src/zen/workspaces/ZenWorkspaces.mjs
```diff
@@ -596,39 +596,66 @@ var gZenWorkspaces = new (class extends nsZenMultiWindowFeature {
...
+          // scroll is vertical + modifier key
+          delta = event.deltaY;
+          shouldProceed = true;
+        } else if (isHorizontalScroll) {
+          // clear horizontal scrolling
+          delta = event.deltaX;
+          shouldProceed = true;
+        } else {
+          // diagonal scrolling or unclear direction, ignore
+          return;
```
mr-cheffy, line 651: `shouldProceed` will never be true when reaching this part, why add an extra check below this?

mr-cheffy (approved):
Really interesting solution! Not sure how this affects mac trackpad swiping though?

> Mac trackpad momentum: Sends rapid consecutive deltaY: 0 events (<200ms apart) → Filtered out

I'll merge it so we can have some testing on twilight for the next few days and we can tackle the small changes later

PR #8322, opened 2025-05-15
Fix #6972

SO9010:
Fix #6972, this took me a lot longer than it should have. I hope this is all correct, I've tested it for quite a while now and seems to be all working good. But if there is anything that I should do differently please do let me know. This was the most concise way i could find to do it!

PR #10348, opened 2025-09-13
Fixed sidebar split screen error. Fix #6972

### LINKED ISSUE #6972: Failing to drag tab for split window when sidebar is opened
### Captchas

- [x] I have checked that this issue cannot be reproduced once I removed all my Mods and Custom CSS.

### What happened?

Normally, when drag a tab to the right sdie of current page to mak split window, we can see it like:

Image

But with native sidebar panel enabled on right side, it will become this:

Image
Here, the browser will try to put tab to right side of sidebar(in this case, my bookmark tree) and end up failing to make split window

### Version

1.10.3b

Windows - x64

Split View

```shell

```

SO9010:
This uses the same solution as previously  #8322
To fix #6972

I programmed this with no AI assistance, however here is a change log written by claude :)

1. [browser-box.inc.xhtml](vscode-file://vscode-app/opt/visual-studio-code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) (via patch)
Refactored the tabbox structure:
Wrapped the tabbox in a new <vbox id="zen-appcontent-wrapper">.
Added <html:div id="zen-appcontent-navbar-container"></html:div> for navbar items.
Introduced <hbox id="zen-tabbox-wrapper"> to contain the tabbox.
Included zen-tabbrowser-elements.inc.xhtml inside the tabbox.
2. [ZenStartup.mjs](vscode-file://vscode-app/opt/visual-studio-code/resources/app/out/vs/code/electron-browser/workbench/workbench.html)
Updated sidebar relocation logic:
Changed the target for appending sidebar elements from tabbrowser-tabbox to zen-tabbox-wrapper to match the new structure.
3. [zen-sidebar.css](vscode-file://vscode-app/opt/visual-studio-code/resources/app/out/vs/code/electron-browser/workbench/workbench.html)
Adjusted sidebar styling:
Increased the margin for #sidebar-splitter.
Set margin-right and margin-left to 0 for sidebar box depending on side.
Changed ordering logic for sidebar elements to target #sidebar-box[sidebar-positionend] and its adjacent splitter, instead of nested selectors.
Summary:
You restructured the browser tabbox layout to use new wrapper elements, updated the startup logic to match this structure, and refined sidebar CSS for improved layout and compatibility with the new DOM hierarchy.

XD sorry for the confusion with the other PR, turns out i just didnt add the language when building it the second time to test which is why I thought it wasnt working! This is the correct solution :)


src/browser/base/content/browser-box-inc-xhtml.patch
```diff
@@ -1,20 +1,19 @@
...
+@@ -25,7 +25,13 @@
      </stack>
    </vbox>
    <splitter id="sidebar-splitter" class="chromeclass-extrachrome sidebar-splitter" resizebefore="sibling" resizeafter="none" hidden="true"/>
 +<vbox flex="1" id="zen-appcontent-wrapper">
-+  <html:div id="zen-appcontent-navbar-wrapper">
-+    <html:div id="zen-appcontent-navbar-container"></html:div>
-+  </html:div>
-+  <hbox id="zen-tabbox-wrapper" flex="1">
++  <html:div id="zen-appcontent-navbar-container"></html:div>
```
mr-cheffy: Why are you removing `zen-appcontent-navbar-wrapper`? Its essential for double toolbar

mr-cheffy (changes requested):

src/browser/base/content/browser-box-inc-xhtml.patch
SO9010: Sorry that must have been a mistake.

src/browser/base/content/browser-box-inc-xhtml.patch
```diff
@@ -10,7 +10,7 @@ index afa7f8e7dd74173bf2c696bd96f7e86e8b0126bc..4847c24923f673e91eb7fb65ea6b037f
 +  <html:div id="zen-appcontent-navbar-wrapper">
 +    <html:div id="zen-appcontent-navbar-container"></html:div>
 +  </html:div>
-+  <hbox id="zen-tabbox-wrapper" flex="1">
++  <hbox id="zen-tabbox-wrapper" flex="1" tabcontainer="tabbrowser-tabs">
```
mr-cheffy: What's the reason for the new `tabcontainer` attribute?

mr-cheffy (changes requested):

src/browser/base/content/browser-box-inc-xhtml.patch
SO9010: Sorry I should have reviewed the code more closely after creating a solution to make sure it was clean.

PR #10355, opened 2025-09-13
Revert "Enable workspace horizontal scrolling compatibility for Logitech mice"

mr-cheffy:
Reverts zen-browser/desktop#10325

@xxxbrian This has been reverted due to that swiping really fast with the mac track-pad switches the workspace **twice**. Is this something you could look into?


xxxbrian:
> Reverts #10325
> 
> @xxxbrian This has been reverted due to that swiping really fast with the mac track-pad switches the workspace **twice**. Is this something you could look into?

Understood. Trackpad momentum detection is indeed very tricky - native APIs (like NSEvent on macOS) provide momentum phase information that isn't exposed to JavaScript's WheelEvent interface. 

The fast swiping issue suggests my time-based filtering approach wasn't robust enough for all trackpad gesture patterns. I'll investigate alternative detection methods when I have time to revisit this.

PR #10393, opened 2025-09-16
fix: repeating right sidebar gradient in compact mode

reizumii:
Fixes the repeating gradient on the right sidebar in compact mode in displays wider than >2000 pixels

Replaces the hardcoded `2000px` value with `100vw` which adapts to the window width ^^

Before and after:
image

### DIFF
--- a/src/zen/compact-mode/zen-compact-mode.css
+++ b/src/zen/compact-mode/zen-compact-mode.css
@@ -361,7 +361,7 @@
           @media -moz-pref('zen.view.compact.color-toolbar') {
             background-attachment: fixed;
             background: var(--zen-main-browser-background-toolbar);
-            background-size: 100% 100vw;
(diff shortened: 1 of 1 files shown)


src/zen/compact-mode/zen-compact-mode.css
```diff
@@ -361,7 +361,7 @@
           @media -moz-pref('zen.view.compact.color-toolbar') {
             background-attachment: fixed;
             background: var(--zen-main-browser-background-toolbar);
-            background-size: 100% 2000px;
+            background-size: 100% 100vw;
```
mr-cheffy: Shoudn't this be the height?

mr-cheffy (changes requested):

mr-cheffy (approved):

PR #10417, opened 2025-09-18
feat: Added more urlbar commands, b=no-bug, c=workspaces

BibekBhusal0:
Added several new commands to the global actions template:
- Toggle Tabs on Right
- Add to Essentials / Remove from Essentials
- Bookmark This Page / Bookmark All Tabs / Show All Bookmarks
- Find in Page / Translate Page
- View Downloads / History
- Manage Extensions / Customize Toolbar / Clear Recent History
- Create New Profile


mr-cheffy (approved):
I've removed some of the commands as we might not want to have *too many* yet due to performance concerns. Thanks!

mr-cheffy (approved):

PR #10508, opened 2025-09-24
Improve animation triggering when moving tabs across the separator (fix #10295)

### LINKED ISSUE #10295: Cannot move pinned tab to the top of normal tabs
### Preliminary Checks

### What happened?

This issue occurs when the 'Move the new tab button to the top' option is not checked.
When I attempt to move a pinned tab across the separator to the top of the normal tabs, the animation does not trigger, and the tab cannot be positioned unless I drag it past the existing top normal tab.

### Expected behavior

When dragging a pinned tab across the separator, it should be possible to place it directly at the top of the normal tabs—that is, crossing the separator should allow the tab to move into that position.

### Actual behavior

When dragging a pinned tab across the separator to the top of the normal tabs, the tab does not move just by crossing the separator. It can only be placed below the existing top normal tab, rather than becoming the topmost normal tab.

### Steps to reproduce

1. Go to **Settings > Look and Feel > Sidebar and tabs layout > Vertical Tabs** and uncheck either `Move the new tab button to the top`.
2. Prepare at least one pinned tab and one or more normal tabs.
3. Drag one of the pinned tabs with the mouse and try to place it between the separator and the top normal tab.
4. In practice, just crossing the separator does not trigger the animation, and even if dropped, the tab remains pinned.
5. To move the pinned tab into the normal tab area, you must drag it below the existing top normal tab (i.e., into the second normal tab position).

### Screenshots and videos

https://github.com/user-attachments/assets/fef397ea-b3ab-4762-8cbc-d46f7e70af4b

### Version

1.15.4b

Windows - x64

Tabs

```shell

```

hyugabokko:
## Summary

Changed the behavior of the separator animation when dragging and dropping tabs between pinned tabs and normal tabs.
- Made the animation occur even when there are overlapping elements.
- Added thickness to the boundary that triggers the animation when crossing the separator, improving the usability of dropping tabs directly before or after the separator.

## Related Issue
fix #10295

## Changes
- [5913396](https://github.com/zen-browser/desktop/pull/10508/commits/59133962ee60cd7cfa791c1bebff41c4a7f0afaf): Remove the condition that triggers the animation only when dropElement is absent
  - When the New tab button is positioned at the bottom of the tabs, and you attempt to move a pinned tab to the normal tabs area, the target tab overlaps with the topmost normal tab at the exact moment it slightly exceeds the separator. Therefore, if there is a condition that triggers the animation only when dropElement is absent, there is a problem where the animation does not execute even after crossing the separator.
  - The animation is controlled to occur only when the tab moves to a different area using multiplier, so the condition is unnecessary.
- [8fd5afd](https://github.com/zen-browser/desktop/pull/10508/commits/8fd5afd013a9cb25340d4593d6e1f55a76f820ef): Change to add thickness to the boundary that triggers the animation
  - Modified to trigger a separator movement animation when the dragged tab overlaps with the boundary. The boundary has a thickness from the separator's Y-position: margin + half the tab height. This accounts for users typically clicking near the tab's center when dragging, so we add half the tab height to the boundary.
  - This prevents the issue where the tab accidentally straddles an existing tab immediately before or after the separator, causing it to be placed beyond that tab instead of directly before or after the separator.

## Verification

- [x] Confirmed that the build passes locally
- [x] Verified functionality in actual usage scenarios  
  | `New tab` Button Position | Video |
  | --- | --- |
  | Top | <video src="https://github.com/user-attachments/assets/4fd8d6e0-a667-483c-bfee-589af4ac02da"> |
  | Bottom | <video src="https://github.com/user-attachments/assets/fd3dd353-3383-4665-8945-f9fc483fe3e2"> |

### DIFF
--- a/src/browser/components/tabbrowser/content/tabs-js.patch
+++ b/src/browser/components/tabbrowser/content/tabs-js.patch
@@ -1,5 +1,5 @@
 diff --git a/browser/components/tabbrowser/content/tabs.js b/browser/components/tabbrowser/content/tabs.js
-index c7557dad38db9ef02b981c46de9595df77cb67db..4192dc797d5f07ece57e7e9ecb0cd0698e189659 100644
+index c7557dad38db9ef02b981c46de9595df77cb67db..4b49d647f2d780cefe23a8fcd7b76d84c2156a1d 100644
 --- a/browser/components/tabbrowser/content/tabs.js
 +++ b/browser/components/tabbrowser/content/tabs.js
 @@ -44,6 +44,9 @@
@@ -471,7 +471,7 @@ index c7557dad38db9ef02b981c46de9595df77cb67db..4192dc797d5f07ece57e7e9ecb0cd069
  
        let dropElement = getOverlappedElement();
 +      if (dropElement?.hasAttribute("split-view-group")) dropElement = dropElement.labelElement;
-+      gZenPinnedTabManager.animateSeparatorMove(movingTabs, dropElement, isPinned, event);
++      gZenPinnedTabManager.animateSeparatorMove(movingTabs, isPinned, event);
  
        let newDropElementIndex;
        if (dropElement) {
--- a/src/zen/tabs/ZenPinnedTabManager.mjs
+++ b/src/zen/tabs/ZenPinnedTabManager.mjs
@@ -1294,12 +1294,19 @@
       const draggingTabHeight = movingTabs.reduce((acc, item) => {
         return acc + window.windowUtils.getBoundsWithoutFlushing(item).height;
       }, 0);
-      let topToNormalTabs = itemsToCheck[0].screenY;
-      if (!isPinned) {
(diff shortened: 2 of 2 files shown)


mr-cheffy:
Oh well, this is uncomfortable now... Before I could see this PR, I pushed this commit: https://github.com/zen-browser/desktop/commit/9046e083eec0bbac966d913802ca41a6467971f5#diff-3e3ff293d54489f0c16553039aa8d982385b9b96c102d7dd9db9b2145ddb816f

which fixes long term concerns about drag and dropping with multiselected tabs. I was wondering if this PR could be refactored into adding support for no-newtab?

hyugabokko:
Thank you for checking. The part I was just about to change happens to overlap with yours…
I'll try if it can be refactored.

mr-cheffy:
Sorry..

mr-cheffy:
I'll close this, feel free to reopen it if you'd like!

PR #10511, opened 2025-09-24
chore: update flatpak runtime to 25.08

vixalien:


mr-cheffy (approved):

mr-cheffy:
I will revert these since it breaks some fonts

vixalien:
> I will revert these since it breaks some fonts

Would be nice to know which, so that I can probably fix it. Firefox seems to be based on this runtime

mr-cheffy:
@vixalien https://github.com/zen-browser/desktop/issues/10545

PR #10518, opened 2025-09-24
feat: add basic vim navigation to urlbar

adrianord:
Adds the keybinds ctrl+j and ctrl+k to down and up respectively when the urlbar is active for making selections.

This change adds:
- Preference for activating vim navigation for backwards compatibility
- Patch for activating the preference
- IDs in the locales for the preference
  - Localization is all in English, but I am happy to go back and try to translate a few if that helps, just not my strong suit.

Tested on MacOS and Ubuntu (Parallels)

Behavior we see:
- ctrl+j and ctrl+k take precedence when the urlbar is active
- keybinds bound to ctrl+j and ctrl+j still work while the urlbar is *not* active

https://github.com/user-attachments/assets/612c6116-c184-4a9c-879e-122603ea03c2


mr-cheffy:
Could there please be some unit tests about this to make sure future firefox versions dont break this functionality?

adrianord:
@mr-cheffy I've gone ahead and written a unit test for the vim navigation feature.

It's a little crude, so let me know if you have a better way to force multiple possible selections in the URLBar more reliably. What I've done works, but I'm afraid it might be a little brittle, so I'll leave it up to your judgement.

adrianord:
Rebased to fix conflicts, is there anything else that needs to be done in this PR before it can get reviewed?

locales/cs/browser/browser/preferences/zen-preferences.ftl
```diff
@@ -140,6 +140,8 @@ zen-urlbar-title = Zen adresní řádek
 zen-urlbar-header = Obecné nastavení adresního řádku
 zen-urlbar-description = Přizpůsobte si adresní řádek podle svých představ
 zen-urlbar-behavior-label = Chování
+zen-urlbar-vim-navigation =
+  .label = Vim Navigation
```
mr-cheffy: Could you remove this from all of the languages except `en-US`?

src/browser/components/preferences/zenLooksAndFeel.inc.xhtml
```diff
@@ -151,6 +151,10 @@
                   </menupopup>
             </menulist>
         </hbox>
+
+      <checkbox id="zenUrlBarVimNavigation"
```
mr-cheffy: I woudn't add it here to prevent having too much options, id just leave it as an about:config

src/zen/tests/urlbar/browser_vim_navigation.js
```diff
@@ -0,0 +1,41 @@
+/* Any copyright is dedicated to the Public Domain.
+   https://creativecommons.org/publicdomain/zero/1.0/ */
+
+'use strict';
+
+ChromeUtils.defineESModuleGetters(this, {
+  UrlbarTestUtils: 'resource://testing-common/UrlbarTestUtils.sys.mjs',
+});
+
+add_task(async function test_Vim_Navigation() {
```
mr-cheffy: Looks great! Maybe a test when the pref is `off` just in case?

mr-cheffy (changes requested):

src/browser/components/preferences/zenLooksAndFeel.inc.xhtml
adrianord: Sure, I'll remove it here and remove the localization since the localization is only used for the checkbox.

adrianord:
Updated the PR to remove the option from Preferences to keep it just in about:config, removed the localization since that was only used for the Preferences checkbox, and added the test for when the about:config option is turned off.

adrianord:
Is there anything else that needs to be done to get this feature out?

src/browser/components/urlbar/UrlbarController-sys-mjs.patch
```diff
@@ -2,6 +2,27 @@ diff --git a/browser/components/urlbar/UrlbarController.sys.mjs b/browser/compon
 index 36e3ab4a5a153230bb488b66dda7e3e7c763ca23..cc4ea61914a316451fa54b01a5c8c6a305e4038a 100644
 --- a/browser/components/urlbar/UrlbarController.sys.mjs
 +++ b/browser/components/urlbar/UrlbarController.sys.mjs
+@@ -302,6 +302,20 @@ export class UrlbarController {
+       return;
+     }
+
++    // Handle vim navigation bindings.
++    if (
++      Services.prefs.getBoolPref("zen.urlbar.vim-navigation.enabled", false) &&
```
mr-cheffy: Could you please move this pref below the `(event.key == "j" || event.key == "k")` check? Fetching prefs can be slow, we would really only want to do it if we are 100% certain the combination satisfies the condition

mr-cheffy (changes requested):

src/browser/components/urlbar/UrlbarController-sys-mjs.patch
adrianord: That's a great call out, made the change.

adrianord:
I see that a recent zen update upgraded to firefox 143.0.4, so I'm building locally to make sure everything in this PR still works.

adrianord:
Everything works, ready for re-review

mr-cheffy (approved):

mr-cheffy (changes requested):
Im seeing tests are failing, did you give these a try?

image

adrianord:
I did a full clean of my repo and rebuild then test and I don't see the test failure. What platform are you running on so I can try to reproduce? I ran the tests today on MacOS 26.0.1 and get one failing test that is unrelated to the PR.

image

adrianord:
I was able to reproduce the test failure on Ubuntu, will let you know when it is fixed.

KennethHoff:
Standard vim key bindings for navigation - assuming `j` and `k` (Without modifiers) is unavailable - is `ctrl-n` and `ctrl-p` (I assume they stand for `next` and `previous` respectively, but not sure about that). `ctrl-j` and `ctrl-k` is unorthodox.

adrianord:
@KennethHoff Huh, I hadn't realized this because I'm so used to using `ctrl+j` and `ctrl+k` from my own configuration for so many years. Regardless, I feel like there is a quite a few people who are looking for this either because they use this in their own (neo)vim configuration or because they are coming from Arc where this is standard.

I'll re-think the feature, maybe just give the preference a different name, or dive deeper to make this more configurable.

yatinlala:
#10753 has added CTRL-n and CTRL-p by default, though its not customizable. Perhaps this PR should be closed.

adrianord:
Will open another PR once I get around to working on making those keys customizable

PR #10536, opened 2025-09-25
feat: improve existing monochrome icons and add new ones

reizumii:
Updated existing monochrome icons to use the latest [Ionicons](https://ionic.io/ionicons) and added more from the set.

image


12th-devs:
I love this, you are amazing!

mr-cheffy:
@reizumii is this ready for testing?

JaceThings:
remove these pls – cheff

src/browser/themes/shared/zen-icons/icons.css
```diff
@@ -816,5 +816,4 @@
 .zen-workspace-context-icon img {
   -moz-context-properties: fill;
   fill: currentColor;
-  scale: 1.6;
```
mr-cheffy, line 819: Have you tested this on context menus?

reizumii: works fine on linux, also icons are updated to work at any size, and the menu icons have a fixed 16x16 size so i think it should work fine on windows too. not sure about mac tho

reizumii: as for testing, i tested the pr and everything works so far, but i recommend you test the pr first especially on mac

lolipodass:
> remove these pls – cheff 

why so? if you remove android and google icons but leave apple one, it's kinda strange move, same applies for different sport balls. And removing any of selected icons is strange, because if you don't like it, then just don't use them, I can find all of them useful in some situations

JaceThings:
The Apple one should be removed as well, actually, it's not about not liking them. It's about them being made poorly. They are just a bad icon pack in general, especially when being rendered at such small sizes

lolipodass:
> The Apple one should be removed as well, actually, it's not about not liking them. It's about them being made poorly. They are just a bad icon pack in general, especially when being rendered at such small sizes

I think those icons could be a problem at a small size,
image
However, regarding google, youtube, bookmark, rss, 9 dots icons, I don't understand why you think they are poorly made, because their design is so simple and fully match original

JaceThings:
The original ones are also a poor choice. Especially for Arc. When these were originally made for the Mac, using SF icons would have made more sense.

Especially if one would want a large library of icons made to respect small sizes.

They're poorly made for these sizes, even on Arc, regardless of them being "original" or not. Not to mention these icons were not "made" to be put into folders' aspect ratio.

What I'm doing in the screenshot is filtering out the icons either with more than one stroke, multiple changes in "depth", too wide/tall of an aspect ratio, and just overall bad design for the size (campfire as an example).

I don't really care what gets added or not, but my reasoning from a design perspective is valid, from an opinion one, that's up to anyone else.

mr-cheffy (approved):

mr-cheffy:
@reizumii whats the end result then?

reizumii:
> @reizumii whats the end result then?

Not sure if I understand your comment... but I updated the PR to remove the icons you and Jace mentioned. I also edited the image in the post to reflect that.

Icons are much more aligned but appear a bit smaller (to compensate alignment and padding inside the folder), you can test the PR before merging. I'll leave it for you to decide :)

mr-cheffy:
From the image i'd still remove:
* car
* tennis/basketball
* train
* umbrella
* headphones
* phone
* eye
* diamond

because their complexity would make them look awful inside a folder imho

reizumii:
Updated the PR to remove said icons as requested except for tennis (baseball) since it's already part of Zen.

I kept the phone and eye because they look fine in folders.

mr-cheffy (approved):

Issue #7253, opened 2025-04-03
IDN address shows punny code in URL


rafaelmr2:
### Captchas

- [x] I have read the instructions.
- [x] I have searched existing issues and avoided creating duplicates.
- [x] I am not filing an enhancement request.
- [x] I have checked that this issue cannot be reproduced on Mozilla Firefox.
- [x] I have checked that this issue can be reproduced once I removed all my Mods and Custom CSS.

### What happened?

When I insert an IDN address. ex: graduação.ufrj.br, it shows in punny code on URL bar: xn--graduao-2wa9a.ufrj.br

Thanks.

### Version

1.10.3b

### What platform are you seeing the problem on?

Linux (Tarball)

### What component is this issue related to?

URL Bar

### Relevant log output if applicable

```shell

```

Issue #5119, opened 2025-02-08
Punycode instead of special characters shown in urlbar by default


romibi:
### Captchas

- [x] I have read the instructions.
- [x] I have searched existing issues and avoided creating duplicates.
- [x] I am not filing an enhancement request.

### What happened?

By default neither up to date Firefox (135.0) nor up to date chrome (133.0.6943.60) show the punycode encoding of urls but display the special characters.
But Zen Browser has the flag `network.IDN_show_punycode` in 'about:config' enabled by default.

See for example https://www.fahrschule-fahrvergnügen.de/
Firefox and Chrome show it with the 'ü' but Zen Browser shows 'www.xn--fahrschule-fahrvergngen-vpc.de'

I understand, this probably is done for security reasons.

Personally in my opinion the best would be if the punycode would be shown while the urlbar has focus but the proper special characters would be shown while the urlbar doesn't have focus.

Whatever the decision is: to keep it as is, to change the default, or to change the behaviour entirely …
I think it's best if it is documented here in the issues list.

PS:
At least for German speaking area, most websites that support a variant with umlauts, redirect to the variant without an umlaut.
(It was a bit difficult to find an example that does not dox me and does not redirect. So it's not a big issue here.)
But I assume this default setting might be a bigger issue in languages that have more non-ascii characters.

### Reproducible?

- [x] I have checked that this issue cannot be reproduced on Mozilla Firefox.

### Version

1.7.6b

### What platform are you seeing the problem on?

Windows

### Relevant log output if applicable

```shell

```

romibi:
Yes, starting zen 1.9.1b with a new profile (using `zen.exe -p`) the default for `network.IDN_show_punycode` is still different than in Firefox and my suggestion to displaying Punycode only when the URL bar is focused is not implemented.

mr-cheffy:
## Closing this issue as part of a repo cleanup

Due to technical-debt we had in the past, we are currently clearing out old, duplicate, stale, and non-actionable issues to improve issue tracking and make it easier to focus on fixing real bugs.

If this issue is still relevant, please open a new issue using our updated issue template, which will help us categorize and address it more efficiently.

Thanks for your understanding and for helping us improve the project! 🚀

This issue can be reopened at: https://github.com/zen-browser/desktop/issues/new?template=bug_report.yml&title=Punycode%20instead%20of%20special%20characters%20shown%20in%20urlbar%20by%20default%0A

PR #10600, opened 2025-09-28
urlbar.yaml: Fix network.IDN_show_punycode to show international characters as designed

### LINKED ISSUE #4527: IDN address show punny code in URL
### Captchas

### What happened?

When I insert an IDN address. ex: `graduação.ufrj.br`, it shows in punny code on URL bar: `xn--graduao-2wa9a.ufrj.br`

Thanks.

### Reproducible?

### Version

1.7.1b

Linux (Flatpak)

```shell

```

bernhardkaindl:
All web browsers that support IDN (international domain names) should not show the ASCII punycode for IDN, but the actual international characters.

As the bug reports say, this is fixed by setting `network.IDN_show_punycode` to `false`.

Fixes: #4527, #5119, #7253:
- Punycode instead of special characters shown in urlbar by default

### DIFF
--- a/prefs/urlbar.yaml
+++ b/prefs/urlbar.yaml
@@ -54,7 +54,7 @@
   value: false
 
 - name: network.IDN_show_punycode
-  value: true
+  value: false
 
 - name: browser.urlbar.suggest.topsites
   value: true
--- a/prefs/urlbar.yaml
+++ b/prefs/urlbar.yaml
@@ -53,9 +53,6 @@
 - name: browser.formfill.enable
   value: false
 
-- name: network.IDN_show_punycode
-  value: false
-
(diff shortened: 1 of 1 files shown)


Zaczero (approved):
Showing punycode is a good security feature that makes it harder to get phished. However, in the case of Zen Browser, I believe the right choice would be to let users opt into that feature, as this goes against the "welcome to a calmer internet" branding.

P.S. I'm guessing, but if we remove the network.IDN_show_punycode setting altogether, it would fall back to the Firefox default setting, which is false. Maybe the entire section should be simply removed?

mr-cheffy (changes requested):
@bernhardkaindl this seems to be the default at firefox, could this pref just be removed completly?

PR #10606, opened 2025-09-28
fix: zen.source ignore VCS data

Zaczero:
Exclude VCS data to reduce source bundle size from ~1.9GB to ~800MB while preserving all other dotfiles. This is a low–medium priority follow-up to https://github.com/zen-browser/desktop/pull/10599. I hadn't realized the engine contained .git data - a newbie mistake. On the bright side, I confirmed that other dotfiles are preserved, and I'll proceed with packaging Zen Browser for nixpkgs :slightly_smiling_face:. I verified the new command locally on the unpacked engine data, and it appears to work perfectly.

PR #10623, opened 2025-09-29
feat: Bootstrapping of zen.source in build.yml

Zaczero:
Hello again!

https://github.com/zen-browser/desktop/pull/10606:

> Well, I'm giving it a try and will see what happens. I'll post updates here if I find or learn anything interesting.

This is my follow up. So I built the browser successfully but it's clear that it's missing the custom mozconfig ([you mentioned](https://github.com/zen-browser/desktop/pull/10606#issuecomment-3343846893)) that apparently is done via surfer bootstrap. The engine without it is pretty much unusable. The current zen.source does not contain necessary data so I figured out the best way forward would be to finish the bootstrapping of sources within the pipeline, so the produced zen.source is ready for build. I followed the instructions from: https://docs.zen-browser.app/contribute/desktop/building

But:

1. I was not able to test the changes locally due to:
   `NotImplementedError: Bootstrap support for this Linux distro not yet available: nixos`
3. I am not 100% sure whether *Update Language Packs* step is necessary but I think it is.
4. I am hoping that python3 is pre-installed in the runner image (I see you are using something custom).

Please kindly review my PR and see if my idea is correct! :rocket:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -264,6 +264,12 @@ jobs:
       - name: Import
         run: npm run import -- --verbose
 
+      - name: Bootstrap
+        run: npm run bootstrap
+
+      - name: Update Language Packs
+        run: python3 ./scripts/update_en_US_packs.py
+
       - name: Compress
         run: |
           tar \


mr-cheffy:
This would significantly increase the tarball size and bootstrapping doesn't actually change what it does. I believe this change is unnecessary and woudn't fix your original issue. I recommend solving the root issue differently rather than doing it like this, zen specific flags are ran via "npm run build", but that.. builds the browser. Why not just extract the already built tarballs for nix?

Zaczero:
> This would significantly increase the tarball size

I am curious why, I don't have necessary knowledge and I don't quite understand what bootstrapping does here. Could you please explain?

> I believe this change is unnecessary and woudn't fix your original issue.

You are right. I have now double checked it.

> Why not just extract the already built tarballs for nix?

Building from source provides more flexibility + verification. Just feels right.

---

Hmm... Do you think it would be possible to update surfer in a way to support producing ready-to-build zen.source? This would make it perfect and I also think that's the main idea behind these source bundles. I can submit a PR if we agree on some general plan. The build already works and I managed to launch and "use" the browser, it's just missing some assets and configs. I can't imagine there's so much more to be included in zen.source in terms of bundle size.



It would be possible to support building zen-browser "raw" from github, but this would require putting *a lot* more work. Simple reason: there is no internet access during the build procedure. Simple stuff like `surfer download` would need to be patched, other standard solutions like npm/cargo are already "solved".

PS. If you disagree with the above, can you explain to me the idea behind having zen.source?

PS2. The current problem I have, is that the mozconfig files are not in the engine folder that is published. If we can find another way to include all necessary build data in the zen.source upstream then I'll manage to write/patch any necessary tooling downstream. I could make a dirty copy from the GitHub repo but I'm worried it would be more complex than a proper fix upstream +solving it here would make zen.source usable to anyone.

PR #10651, opened 2025-10-01
feat: New site data popup

mr-cheffy:


12th-devs:
Can we just make every Arc feature as a mod so you are forced to add them all?

mr-cheffy:
What do you mean?

12th-devs:
> What do you mean?

You added the command bar, actions animation, and now this popup thing. All of these are in-developement mods that our modding community is working on (team Vertex, etc.)

mr-cheffy:
Im not aware of who's doing what, im just following a todo list... Are you suggesting I should stop working on zen so your mods don't get interrupted?

12th-devs:
> Im not aware of who's doing what, im just following a todo list... Are you suggesting I should stop working on zen so your mods don't get interrupted?

No lol, I just think it's funny that right after someone makes a mod it gets added (like bibek made command bar, then the omnibox got added to twilight 5 days later), I was talking to some people about it last night on discord

Keep up the great work!

mr-cheffy:
Oh, yeah not sure, my plan was to unify everything into the popup and then add the compact mode button. Then the next thing would be window syncing

12th-devs:
> Oh, yeah not sure, my plan was to unify everything into the popup and then add the compact mode button. Then the next thing would be window syncing

That's awesome! Is compact mode gonna be on the default toolbar?

12th-devs:
Also, when you have some time can you make the customization mode work with single toolbar instead of showing double toolbar?

xperr:
I have a suggestion to add a small volume slider in the Page Actions panel, similar to the “Volume Master” extension on Chrome. This is a feature I really miss on Firefox-based browsers, since existing extensions that try to do this don’t work properly. It would let users control the sound level of each tab directly from the menu, making it easier to manage media volume without extra or unreliable extensions.

Something like this (generated by AI :-D):

ChatGPT Image Oct 4, 2025, 12_27_23 PM

PR #10688, opened 2025-10-04
feat: Unload a workspace

athkdev:
- [x] Add a button in the context popup for a workspace to unload all tabs in that workspace
- [x] Add label entry for "Unload Workspace" for all l10n packs

image

### Motivation

I use the tab unload feature in Zen A LOT. Right now, I have ~50 tabs across 5 workspaces on my M1 mac mini - I do find myself needing to shift workspaces quite a bit. So I thought, although tabs auto-unload after some time, having the option to manually unload a workspace immediately would be nice.

____

If this makes sense, I had a follow-up question: after unloading a workspace, should the user stay on that unloaded workspace, or move to the last used/next workspace?


src/zen/workspaces/ZenWorkspaces.mjs
```diff
@@ -1479,6 +1479,21 @@ var gZenWorkspaces = new (class extends nsZenMultiWindowFeature {
     });
   }
 
+  async unloadWorkspace(event) {
+    const tabs = this.allStoredTabs.filter(
```
mr-cheffy: Here, you can just do `gBrowser.tabs` and remove the `workspace-id` check

mr-cheffy: Here, you can just do `gBrowser.tabs` and remove the `workspace-id` check

mr-cheffy (changes requested):
Looks great! Please remove all the translations strings except for the en-US ones. Thanks!

athkdev:
Thanks for the review!

Made the requested changes.

src/zen/workspaces/ZenWorkspaces.mjs
athkdev: Done! 

Also, removed the `event` args it wasn't getting used anywhere.

mr-cheffy (approved):

locales/en-US/browser/browser/zen-workspaces.ftl
```diff
@@ -23,6 +23,9 @@ zen-workspaces-panel-change-icon =
 zen-workspaces-panel-context-default-profile =
     .label = Set Profile
 
+zen-workspaces-panel-unload =
+    .label = Unload Workspace
```
mr-cheffy: Please call it `Unload Space` for consistency

src/zen/workspaces/ZenWorkspaces.mjs
```diff
@@ -1479,6 +1479,20 @@ var gZenWorkspaces = new (class extends nsZenMultiWindowFeature {
...
+      (tab) =>
+        !tab.hasAttribute('zen-empty-tab') &&
+        !tab.hasAttribute('zen-essential')
+    );
+
+    if (tabs.length === 0) {
+      return;
+    }
+
+    await gBrowser.explicitUnloadTabs(tabs);
```
mr-cheffy: I was wondering if there could be an unit test written about this in a future PR?

mr-cheffy (changes requested):

mr-cheffy (approved):

mr-cheffy (changes requested):
@athkdev When right clicking on the space buttons below, instead of unloading the right-clicked workspace, it unloads the current one

athkdev:
> @athkdev When right clicking on the space buttons below, instead of unloading the right-clicked workspace, it unloads the current one

My bad I missed that - I'll fix it and push real quick.

src/zen/workspaces/ZenWorkspaces.mjs
athkdev: That sounds great! I'll note it down.

mr-cheffy (approved):

FabianoLothor:
why unload spaces do not unload essentials?

can we add a feature flag in the configs to adjust the unload space behavior to also unload essentials?

mr-cheffy:
Essentials arent part of spaces

FabianoLothor:
> Essentials arent part of spaces

Really? There is a feature to have different essentials on each space though.

PR #10753, opened 2025-10-09
Allow ctrl+n/p to navigate urlbar entries

CrynTox:
This somewhat addresses this feature request: #1936.

An ideal solution would be to implement a custom keybind for this, but I'm nowhere near competent enough to do that, so I changed the code to allow using ctrl+n/p, which already worked for macos, to work anywhere. I feel like this change is fine, since it's not intrusive and simple.

### DIFF
--- a/src/browser/components/urlbar/UrlbarController-sys-mjs.patch
+++ b/src/browser/components/urlbar/UrlbarController-sys-mjs.patch
@@ -1,8 +1,16 @@
 diff --git a/browser/components/urlbar/UrlbarController.sys.mjs b/browser/components/urlbar/UrlbarController.sys.mjs
-index a1faf7e4278c66865f267283515f22052769928d..d76d19da5a3d4b9739faf3a673bb3ad693765ada 100644
+index a1faf7e4278c66865f267283515f22052769928d..297391205ab3cd96bed2e5a3e36a181c6da90098 100644
 --- a/browser/components/urlbar/UrlbarController.sys.mjs
 +++ b/browser/components/urlbar/UrlbarController.sys.mjs
-@@ -441,6 +441,8 @@ export class UrlbarController {
+@@ -297,7 +297,6 @@ export class UrlbarController {
+     const isMac = AppConstants.platform == "macosx";
+     // Handle readline/emacs-style navigation bindings on Mac.
+     if (
+-      isMac &&
+       this.view.isOpen &&
+       event.ctrlKey &&
+       (event.key == "n" || event.key == "p")
+@@ -441,6 +440,8 @@ export class UrlbarController {
              });
            }
            event.preventDefault();


mr-cheffy:
Why not just use the arrows?

src/browser/components/urlbar/UrlbarController-sys-mjs.patch
```diff
@@ -1,8 +1,20 @@
...
-index a1faf7e4278c66865f267283515f22052769928d..d76d19da5a3d4b9739faf3a673bb3ad693765ada 100644
+index a1faf7e4278c66865f267283515f22052769928d..e1380c25f792bfc7baaddabd14431f4bc93370a9 100644
 --- a/browser/components/urlbar/UrlbarController.sys.mjs
 +++ b/browser/components/urlbar/UrlbarController.sys.mjs
-@@ -441,6 +441,8 @@ export class UrlbarController {
+@@ -294,10 +294,8 @@ export class UrlbarController {
+    */
+   // eslint-disable-next-line complexity
+   handleKeyNavigation(event, executeAction = true) {
+-    const isMac = AppConstants.platform == "macosx";
```
mr-cheffy: Please dont remove this / change the comment. Just remove the mac check to avoid bigger patches

mr-cheffy (changes requested):

src/browser/components/urlbar/UrlbarController-sys-mjs.patch
CrynTox: Understood

CrynTox: I did it, would you like me to force push the change on to the first commit?

CrynTox:
> Why not just use the arrows?

Uh, it just feels more ergonomic to me, since my fingers have to move away less from the keyboard. (I use my browser in a very keyboard centric way, so it annoys me a bit to have to move to the arrow keys)

CrynTox:
Actually, when I thought about this more simply, it's just that when you're typing a search query, it's quicker to move to n and p than to the arrow keys.

Bellisario:
> Why not just use the arrows?

I often switch from macOS to linux and miss the fact I cannot use ctrl+n/p to next/previous search results and I always end up opening multiple windows.

---

P.S.: I would suggest to add a setting for this change and for it to be disabled by default 'cause I imagine some users _might_ want open a new tab from keyboard also when the URL bar is open.

CrynTox:
> P.S.: I would suggest to add a setting for this change and for it to be disabled by default 'cause I imagine some users _might_ want open a new tab from keyboard also when the URL bar is open.

@mr-cheffy can I create a config flag for this?
If I can, is this name fine? `zen.urlbar.enable-ctrl-n-p-navigation`

mr-cheffy:
I guess its fine without a pref, don't worry

mr-cheffy (approved):

Issue #10854, opened 2025-10-17
Mozilla services miscredited in user menu


MagentaManifold:
### Preliminary Checks

- [x] I have read and understood the important section above.
- [x] I have searched existing issues and avoided creating duplicates.
- [x] I am not filing an enhancement request.
- [x] I have checked that this issue cannot be reproduced on Mozilla Firefox.
- [x] I have checked that this issue can be reproduced once I removed all my Mods and Custom CSS.

### What happened?

After signing in with a Mozilla account, there are several Mozilla services (Monitor & Mozilla VPN) in the menu that shows up when clicking on the avatar button in the menu bar. They are described as "protection tools from Zen OSS Team", which is not the case. We should either remove this section completely, or replace "Zen OSS Team" with "Mozilla".

### Expected behavior

The "protection tools" section in the menu either does not appear or say "Mozilla" instead of "Zen OSS Team"

### Actual behavior

The "protection tools" section in the menu says "Try other protection tools from Zen OSS Team"

### Steps to reproduce

1. Sign in to Zen with a Mozilla account.
2. edit the toolbar to make the account button appear
3. Click on the account button on the tool bar.
4. Observe the section titled "Try other protection tools from Zen OSS Team"
### Screenshots and videos

<img width="314" height="537" alt="Image" src="https://github.com/user-attachments/assets/bb8bcad0-a6c0-43af-82d0-eb265dc1754d" />

### Version

1.16.4b

### What platform are you seeing the problem on?

macOS - aarch64

### What component is this issue related to?

Localization

### Relevant log output if applicable

```shell

```

MagentaManifold:
I am working on a fix to replace the l10n string with a custom one

PR #10856, opened 2025-10-18
fix: replace incorrect team name for Mozilla services

MagentaManifold:
This is a fix for #10854. This PR replaces The `appmenu-other-protection-header` FTL string with a custom one, so that it no longer refers to `-vendor-short-name`, but use the hardcoded "Mozilla".

A screenshot with this fix applied: 
image

### DIFF
--- a/locales/en-US/browser/browser/zen-override.ftl
+++ b/locales/en-US/browser/browser/zen-override.ftl
@@ -1,0 +1,1 @@
+zen-appmenu-other-protection-header = Try other protection tools from Mozilla:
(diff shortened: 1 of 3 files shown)


locales/en-US/browser/browser/zen-override.ftl
MagentaManifold, line 1: I'm not sure which locale file to put it into, so I created a new one. Please let me know if there's a better place to place it!

mr-cheffy (changes requested):
I think these should rather be completely removed with a simple display: none as these aren't really "our services/features", rather just for firefox.

MagentaManifold:
I have filed a bug [upstream](https://bugzilla.mozilla.org/show_bug.cgi?id=1995168), as it is affecting other Firefox forks as well. I'm closing this PR for now.

mr-cheffy:
Great! I still think these should be completely removed from the popup though, so if you still want an easy bug to tackle, there's that :)

PR #10890, opened 2025-10-21
fix: remove mozilla services from app menu

### LINKED ISSUE #10854: Mozilla services miscredited in user menu

MagentaManifold:
This is a fix for #10854. Here's a screenshot with this fix applied:

image

### DIFF
--- a/src/zen/common/styles/zen-single-components.css
+++ b/src/zen/common/styles/zen-single-components.css
@@ -33,6 +33,12 @@
   display: none !important;
 }
 
+/* hide Mozilla services from the app menu */
+#PanelUI-fxa-menu-services,
+#PanelUI-fxa-cta-menu {
+  display: none !important;
+}
+
 body > #confetti {
   z-index: 1;
 }


MagentaManifold:
I couldn't figure out how to test this, but I think it's straight forward enough to not need a test (`_debloatContextMenus` doesn't have a test either, anyway). `_debloatContextMenus`'s method won't work for this issue, because these widgets are dynamically created on click.

src/zen/common/ZenUIManager.mjs
```diff
@@ -125,6 +126,18 @@ var gZenUIManager = {
     }
   },
 
+  _debloatAppMenu() {
+    // Add CSS rules to hide Mozilla services from the app menu
+    const style = document.createElement('style');
```
mr-cheffy: Why are you doing inline styles? We've talked yesterday that adding these IDs to the function below where we set the attributes `hideen=true` would work. Otherwise just do it in "single-components.css" maybe. But this is quite hacky imo

mr-cheffy (changes requested):

src/zen/common/ZenUIManager.mjs
MagentaManifold: I tried out what you said yesterday, but this sadly doesn't work the same way, because the menu is only created in the DOM when you click on the account button, after Zen UI manager is initialized. Moreover, these widgets are already conditionally displayed, i.e., their `hidden` property is already `true`, and Firefox's code changes it to `false` to display it, that why I chose to use `display: none`.

MagentaManifold: Yeah, putting it in `zen-single-components.css` is better. I didn't do this yesterday because I didn't know which file is a good place to put it in 😅 (still very new to the codebase). I'll repush when I get time

MagentaManifold: Done!

mr-cheffy (approved):

PR #10896, opened 2025-10-21
fix: Hide 'Change Name' for Spaces in collapsed toolbar mode

nrysk:
## Description of the Bug

In "Collapse Toolbar" mode, the "Change Name" context menu item for Spaces was incorrectly appearing **when right-clicking** on a Space icon. This item is non-functional in collapsed mode (as intended by the CSS rule in `:root:not([zen-sidebar-expanded='true'])` in `vertical-tabs.css`), but the JavaScript in `updateWorkspaceActionsMenu` was overriding the `.hidden` property.

## Description of the Fix

This PR updates the `updateWorkspaceActionsMenu` function in `gZenWorkspaces.js`.
It now checks if the sidebar is collapsed (using `!Services.prefs.getBoolPref('zen.view.sidebar-expanded')` or the DOM attribute) before setting the `workspaceName.hidden` property.

This ensures the "Change Name" item is correctly hidden in collapsed mode, aligning the JavaScript logic with the existing CSS intention.

## How to Test

1.  Enable "Collapse Toolbar" mode (in Settings > Looks & Feel).
2.  Right-click on Space icon in the sidebar.
3.  Verify that the "Change Name" option is **no longer visible** in the context menu.

## Screenshot

| Before | After |
| --- | --- |
|  image | image |

### DIFF
--- a/src/zen/tabs/zen-tabs/vertical-tabs.css
+++ b/src/zen/tabs/zen-tabs/vertical-tabs.css
@@ -618,12 +618,6 @@
   --tab-min-width: 48px !important;
   --zen-toolbox-padding: 6px !important;
   --zen-toolbox-max-width: calc(var(--tab-min-width) + var(--zen-toolbox-padding) * 2);
-
-  /* We can't show the rename input properly in collapsed state,
-     so hide the workspace edit input */
-  #context_zenEditWorkspace {
-    display: none;
-  }
 }
 
 #navigator-toolbox:not([zen-sidebar-expanded='true']) {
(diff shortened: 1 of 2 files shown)


mr-cheffy (changes requested):
Looks good!

Could you please remove the CSS for part that removes the menu item and move the comment here?

Also, please run `npm run lint:fix`

nrysk:
Thank you for reviewing!
I fix the part you pointed and run `npm run lint:fix`.

mr-cheffy (approved):

PR #10939, opened 2025-10-22
Updated confusing tracking protection setting

Kwasow:
The wording on the Tracking Protection setting in the new Control Center is confusing.

Currently, there are two options:
- Tracking Protection: Allowed
- Tracking Protection: Blocked

It's likely to be confusing to many users, since Tracking Protection is a feature that **blocks** tracking when enabled. I myself was wondering if _Allowed_ meant that requests are allowed to pass through it or if TP is allowed to work on that site.

Changing the wording to _Enabled_ and _Disabled_ makes it much more clear what the switch does.

### DIFF
--- a/locales/ar/browser/browser/zen-general.ftl
+++ b/locales/ar/browser/browser/zen-general.ftl
@@ -57,8 +57,8 @@ zen-generic-manage = Manage
 zen-generic-more = More
 zen-generic-next = Next
 # These labels will be used for the site data panel settings
-zen-site-data-setting-allow = Allowed
-zen-site-data-setting-block = Blocked
+zen-site-data-setting-allow = Enabled
+zen-site-data-setting-block = Disabled
 zen-site-data-setting-cross-site = Cross-Site cookie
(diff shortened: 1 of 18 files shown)


mr-cheffy:
This also changes the text for all settings though
CleanShot 2025-10-22 at 16 48 48@2x

Kwasow:
That's an omission on my end. If you agree with the change for Tracking Protection, I can complete this PR by adding a separate string for that and replacing it in the UI, so that it doesn't change for the other settings.

mr-cheffy:
Sorry, I was a bit in a hurry so I had to push this fix: https://github.com/zen-browser/desktop/commit/be417381edd7d4a4a907d3cf1c7a31fa24263bb9#diff-6976b665de99a7c0bfacec5631cdfddf357a72d440f4f20b8e13a9706c2e4486

Other PRs are welcome though! And thanks a lot!

Kwasow:
No problem, thank you for fixing it quickly!

PR #11006, opened 2025-10-24
Fix cursor height in "Create Space" input field.

CosmoCreeper:
Currently if you put your cursor in the space of the first character in the input field of the "Create Space" form, it will have less height than if you put it after a character. This is due to the border radius of the text field which is invisible so I have removed it to fix this issue.

(first screenshot shows the issue, second shows what it should look like)
Screenshot 2025-10-24 220702
Screenshot 2025-10-24 220747

### DIFF
--- a/src/zen/workspaces/create-workspace-form.css
+++ b/src/zen/workspaces/create-workspace-form.css
@@ -109,6 +109,7 @@ zen-workspace-creation {
           padding: 0 !important;
           width: 100%;
           outline: none;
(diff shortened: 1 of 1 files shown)


mr-cheffy:
Woudn't it be better to remove the radius thats causing this issue originally?

CosmoCreeper:
Technically but the border radius is applied by a universal styling for all input fields.

mr-cheffy (approved):

PR #11071, opened 2025-10-29
feat: Add Arc's boosts implementation, b=no-bug, c=boosts

FlorianButz:
This PR aims to implement a basic version of the Boosts feature from Arc. This currently features font customization, theme customization and smart inverting.

https://github.com/user-attachments/assets/e092bacd-f3d2-41b9-ae84-a7b9a19d3e8f


12th-devs:
I love this so much!

mr-cheffy:
I've added the boilerplate code we should use in order to change how firefox renders colors for certain shells. If you could make it in a window instead of a popup, it would be great via using `Services.obs` to communicate. Would also remove the need for ZenBoosts.mjs.

6639835 (approved):

mr-cheffy:
Very excited for the implementation I came up with!

We edit firefox's rendering engine to "filter" the colors that where going to be displayed on the screen, unlike arc which I believe they use filters on the document. This means our solution is more efficient + more compatible with websites

Example:

image


And here's where zen's (1) vs arc's (2) implementation really differ:

zen:
image

arc:
image

6639835 (approved):

src/zen/common/ZenUIManager.mjs
```diff
@@ -50,6 +50,14 @@ var gZenUIManager = {
       this.onUrlbarSearchModeChanged.bind(this)
     );
 
+    window.gBrowser.addProgressListener({
+      onLocationChange: (aWebProgress) => {
+        if (aWebProgress.isTopLevel) {
+          this.onShownTabChange();
```
mr-cheffy: This would be cleaner to do inside ZenSiteDataPanel.sys.mjs

mr-cheffy (changes requested):

src/zen/common/ZenUIManager.mjs
FlorianButz: I moved the listener to ZenSiteDataPanel.sys.mjs

FlorianButz:
Bildschirmfoto 2025-11-07 um 8 42 39 PM
Changed the editor UI to look more like the one from Arc

12th-devs:
> Bildschirmfoto 2025-11-07 um 8 42 39 PM
> Changed the editor UI to look more like the one from Arc

@FlorianButz a few questions I had:

1. How does it look in dark mode?

2. How is the size controlled? Drop down list, typing, scroll over the container?

3. What is with the right margin on the size percentage

4. Could the button inside of the gradiant picker be tweaked a bit? It feels a bit out of place in my opinion, but I would love to hear your side.

This all looks so good and I'm really excited, great work man!

12th-devs:
> Bildschirmfoto 2025-11-07 um 8 42 39 PM
> Changed the editor UI to look more like the one from Arc

@FlorianButz a few questions I had:

1. How does it look in dark mode?

2. How is the size controlled? Drop down list, typing, scroll over the container?

3. What is with the right margin on the size percentage

4. Could the button inside of the gradiant picker be tweaked a bit? It feels a bit out of place in my opinion, but I would love to hear your side.

This all looks so good and I'm really excited, great work man!

FlorianButz:
@12th-devs 
1. There is no dark mode. Arc does not have it, so I left it out for the time being. This can still be added later on if necessary.
2. The size changes the zoom on the website, the dropdown list shows all available system fonts, the typing overrides the text case
3. There are still quite a few visual issues. However to not slow down the development I decided to commit the changes now. Perfecting can be done after everything else is in place.
4. As I said in 3, there's still some issues I currently don't want to prioritize

12th-devs:
@FlorianButz that sounds good, thanks for the quick response.

6639835 (approved):

6639835 (approved):

greeeeeenbean:
Hey all, any update on this?

FlorianButz:
@greeeeeenbean Yes, I'm still working on it. I didn't have a lot of time the last few weeks.

The theme of the website can now be synced to the workspace theme.

https://github.com/user-attachments/assets/17012bb7-e589-4ec6-b4b1-3da5f688b673

6639835 (approved):

Minemetero:
Bro reviews every change like he can actually do something about it...

6639835:
> Bro reviews every change like he can actually do something about it...

At least more than you could do 😃

prefs/zen/boosts.yaml
mr-cheffy, line 1: Could you please name it "boosts.yaml"?

locales/en-US/browser/browser/zen-boosts.ftl
mr-cheffy, line 1: Could you add the license here?

```
# This Source Code Form is subject to the terms of the Mozilla Public
# License, v. 2.0. If a copy of the MPL was not distributed with this
# file, You can obtain one at http://mozilla.org/MPL/2.0/.

```

src/browser/themes/shared/zen-icons/icons.css
```diff
@@ -34,6 +34,19 @@
...
+
+#PanelUI-zen-emojis-picker-none,
+#zen-emojis-picker-none {
+  list-style-image: url('trash.svg');
+}
+
+#PanelUI-zen-gradient-generator-color-remove,
+#zen-gradient-generator-color-remove {
+  list-style-image: url('unpin.svg') !important;
 #appMenu-quit-button2 {
```
mr-cheffy: These seems to be something wrong going on here?

mr-cheffy (changes requested):

src/browser/themes/shared/zen-icons/icons.css
FlorianButz: I accidentally deleted two lines, it is resolved now.

FlorianButz:
Updated the zap mode UI. The buttons on the bottom control bar are temporary

https://github.com/user-attachments/assets/b0536b7e-609f-4517-baf7-7a5c9b82c517

prefs/zen/boosts.yaml
```diff
@@ -0,0 +1,8 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+- name: zen.boosts.enabled
+  value: false
+- name: zen.boosts.dissolve-on-zap
```
mr-cheffy, line 8: Nit: add a new line between prefs

src/browser/themes/shared/zen-icons/common/selectable/brush.svg
mr-cheffy, line 1: Still under `selectable` folder. `selectable` is used for folder / tab icons btw, not for UI icons.

src/browser/themes/shared/zen-icons/icons.css
```diff
@@ -34,6 +34,21 @@
...
+#zen-emojis-picker-none {
+  list-style-image: url('trash.svg');
+}
+
+#PanelUI-zen-gradient-generator-color-remove,
+#zen-gradient-generator-color-remove {
+  list-style-image: url('unpin.svg') !important;
+}
+
 #appMenu-quit-button2 {
```
mr-cheffy: You can remove this `appMenu-quit-button2` now as its already added at the top?

src/browser/themes/shared/zen-icons/icons.css
```diff
@@ -497,13 +514,34 @@
...
+    pointer-events: none;
+  }
 }
 
 .geo-icon {
   list-style-image: url("geo.svg") !important;
 }
 
-.geo-icon.blocked-permission-icon {
+.geo-icon.clpermission-icon {
```
mr-cheffy: Revert this change

src/zen/boosts/actors/ZenBoostsChild.sys.mjs
```diff
@@ -0,0 +1,432 @@
...
+    "pointerenter",
+    "pointerover",
+    "pointerout",
+    "pointerleave",
+  ];
+
+  /**
+   * Called when the actor is destroyed. Cleans up the events.
+   */
+  didDestroy() {
```
mr-cheffy: Wouldn't it be better to simply listen for "unload"? Not sure window is available at this stage

src/zen/boosts/actors/ZenBoostsChild.sys.mjs
```diff
@@ -0,0 +1,432 @@
...
+      default:
+        break;
+    }
+  }
+
+  handleZapEvent(event) {
+    if (
+      [...ZenBoostsChild.OVERLAY_EVENTS, ...ZenBoostsChild.PREVENTABLE_EVENTS].includes(event.type)
+    ) {
+      this.#overlay.handleEvent(event, ZenBoostsChild.PREVENTABLE_EVENTS.includes(event.type));
```
mr-cheffy: This looks like it could be optimized a bit, because doing constant checks might be bad for performance

src/zen/boosts/actors/ZenBoostsChild.sys.mjs
```diff
@@ -0,0 +1,432 @@
...
+      this.document.addEventListener(event, this._handleZapEvent, true);
+    }
+    this.#preventableEventsAdded = true;
+  }
+
+  /**
+   * Removes the event listeners from the document
+   */
+  #removeEventListeners() {
+    // this.contentWindow.removeEventListener("unload", this._disableZapMode);
```
mr-cheffy: Remove comment

src/zen/boosts/actors/ZenBoostsChild.sys.mjs
```diff
@@ -0,0 +1,432 @@
...
+
+  /**
+   * Adds necessary event listeners to the document
+   * to prevent content interactions
+   */
+  #addEventListeners() {
+    this._handleZapEvent = this.handleZapEvent.bind(this);
+
+    this._disableZapMode = this.disableZapMode.bind();
+    // this.contentWindow.addEventlistener("unload", this._disableZapMode);
```
mr-cheffy: Remove comment

src/zen/boosts/actors/ZenBoostsChild.sys.mjs
```diff
@@ -0,0 +1,432 @@
...
+
+  /**
+   * Aquires the boost data for this website
+   * @returns Boost data for the current website
+   */
+  async getWebsiteBoost() {
+    const domain = this.browsingContext.topWindow?.location?.host;
+    if (!domain) return null;
+
+    return this.sendQuery("ZenBoost:GetBoostForDomain", domain);
```
mr-cheffy, line 297: Why's this function async?

src/zen/boosts/actors/ZenBoostsChild.sys.mjs
```diff
@@ -0,0 +1,432 @@
...
+  }
+
+  /**
+   * Aquires the stylesheet for this website
+   * @returns Generated stylesheet string for this website
+   */
+  async getWebsiteStyle() {
+    const domain = this.browsingContext.topWindow?.location?.host;
+    if (!domain) return null;
+    const styleSheet = (await this.sendQuery("ZenBoost:GetStyleForDomain", domain)).styleSheet;
```
mr-cheffy: Instead of repeating `sendQuery`, you can just use getWebsiteBoost()

src/zen/boosts/actors/ZenBoostsChild.sys.mjs
mr-cheffy: Also, this could be cached maybe?

src/zen/boosts/actors/ZenBoostsChild.sys.mjs
```diff
@@ -0,0 +1,432 @@
...
+  async #applyBoostForPageIfAvailable(unloadStyles = false) {
+    const browsingContext = this.browsingContext;
+    // Prevent applying boosts to iframes or non-top-level browsing contexts.
+    // It makes the tab crash if we try to load stylesheets into an iframe's
+    if (!browsingContext || browsingContext.parent !== null) {
+      return null;
+    }
+
+    const domain = browsingContext.topWindow?.location?.host;
+    if (!domain) return null;
```
mr-cheffy: Is `domain` needed here? Could just check for `!boost`

mr-cheffy (changes requested):
Im not finished, but this is a good start

src/zen/boosts/actors/ZenBoostsParent.sys.mjs
```diff
@@ -0,0 +1,117 @@
...
+export class ZenBoostsParent extends JSWindowActorParent {
+  /**
+   * Creates a new ZenBoostsParent actor instance and sets up an observer
+   * for boost update notifications.
+   */
+  constructor() {
+    super();
+
+    this._observe = this.observe.bind(this);
+    Services.obs.addObserver(this._observe, "zen-boosts-update");
```
mr-cheffy: No need to call `this._observe` as `Services.obs.addObserver(this)` should already call the `observe` function

src/browser/themes/shared/zen-icons/common/selectable/brush.svg
mr-cheffy, line 1: Why remove?

src/zen/boosts/zen-advanced-color-options.css
```diff
@@ -0,0 +1,87 @@
...
+  border: solid 1px #d1d0d5;
+  border-radius: 100%;
+}
+
+#zen-boost-advanced-color-options-panel input::-moz-range-thumb:active {
+  background-color: #cbcad0;
+}
+
+#zen-boost-advanced-color-options-container {
+  display: flex !important;
```
mr-cheffy: `!important`?

src/zen/boosts/zen-boost-editor.xhtml
```diff
@@ -0,0 +1,118 @@
...
+        <div class="zen-boost-color-picker-circle"></div>
+      </hbox>
+
+      <hbox flex="1" id="zen-boost-toolbar-wrapper">
+        <button data-l10n="zen-boost-invert" id="zen-boost-invert" class="subviewbutton mod-button small"></button>
+        <button data-l10n="zen-boost-controls" id="zen-boost-controls" class="subviewbutton mod-button small"></button>
+        <button data-l10n="zen-boost-disable" id="zen-boost-disable" class="subviewbutton mod-button small"></button>
+      </hbox>
+
+      <div id="zen-boost-font-wrapper">
```
mr-cheffy: html elements should be marked with the html namepsace, e.g. `<html:div ...`

src/zen/boosts/zen-boosts.css
```diff
@@ -0,0 +1,604 @@
...
+#zen-boost-zap {
+  & #zen-boost-zap-value {
+    text-align: right;
+    right: 0;
+    margin: auto;
+  }
+}
+
+#zen-boost-zap[hideicon="true"] {
+  list-style: none !important;
```
mr-cheffy: Its bad practice to use `!important`, try to use it as little as possible, it shouldn't be needed here

src/zen/boosts/ZenBoostsManager.sys.mjs
```diff
@@ -0,0 +1,309 @@
...
+class nsZenBoostsManager {
+  registeredBoosts = new Map();
+  #stylesManager = new nsZenBoostStyles();
+
+  #saveFilename = 'zen-boosts.jsonlz4';
+
+  #file = null;
+
+  constructor() {
+    this.#init();
```
mr-cheffy: Can just have the contents of `init` here if its just a one-liner

src/zen/boosts/ZenBoostStyles.sys.mjs
```diff
@@ -0,0 +1,115 @@
...
+
+  /**
+   * Converts a raw CSS style string into a data URI.
+   * @param {string} rawStyle - The raw CSS style string.
+   * @returns {string} The data URI representing the CSS style.
+   * @private
+   */
+  #convertStyleToDataUri(rawStyle) {
+    const encodedStyle = encodeURIComponent(rawStyle);
+    return Services.io.newURI(`data:text/css;charset=utf-8,${encodedStyle}`);
```
mr-cheffy, line 102: Not too sure of having styles applied as `url`s because it might not get cached, but i'd have to check on that

src/zen/boosts/ZenZapDissolve.sys.mjs
```diff
@@ -0,0 +1,532 @@
...
+    if(distToCenter > 0.5) discard;
+
+    float alpha = sampled.a * v_POpacity;
+    float brightnessFactor = 1.5;
+    gl_FragColor = vec4(sampled.rgb * brightnessFactor, alpha);
+}
+  `;
+
+  VERT = `
+  precision highp float;
```
mr-cheffy, line 29: Can't this be stored on a separate file?

src/zen/boosts/ZenZapOverlayChild.sys.mjs
```diff
@@ -0,0 +1,684 @@
...
+  /**
+   * Handles the addition of the current zap selector
+   */
+  #handleZap() {
+    const cssPath = this.getSelectionPath();
+
+    this.removeHighlight();
+    this.#resetHoverDiv();
+
+    const { setTimeout } = ChromeUtils.importESModule("resource://gre/modules/Timer.sys.mjs");
```
mr-cheffy: Import it lazylly at the top

src/zen/boosts/ZenZapOverlayChild.sys.mjs
```diff
@@ -0,0 +1,684 @@
...
+
+    setTimeout(() => {
+      const useDissolve = Services.prefs.getBoolPref("zen.boosts.dissolve-on-zap");
+      if (!this.window.gReduceMotion && useDissolve) {
+        const elements = this.document.querySelectorAll(cssPath);
```
mr-cheffy: Does `gReduceMotion` actually exist on `this.window`?

src/zen/boosts/ZenZapOverlayChild.sys.mjs
```diff
@@ -0,0 +1,684 @@
...
+          const dissolveEffect = new lazy.ZapDissolve(this.document);
+          await dissolveEffect.initialize();
+          dissolveEffect.dissolve(element);
+        });
+      }
+
+      this.zenBoostsChild.addZapSelector(cssPath);
+      this.#setState(ZapOverlay.STATES.SELECTING);
+      this.onZapUpdate();
+    }, 5);
```
mr-cheffy: Why `5`?

src/zen/boosts/ZenZapOverlayChild.sys.mjs
```diff
@@ -0,0 +1,684 @@
...
+      unzapButton.type = "button";
+      unzapButton.id = "unzap";
+      unzapButton.value = "X";
+      unzapButton.setAttribute("selector", selector);
+      zapList.appendChild(unzapButton);
+    });
+
+    if (boostData.zapSelectors.length == 0)
+      zapList.innerHTML += '<p class="pcenter">Click elements on the page to <b>Zap</b> them</p>';
+    else zapList.innerHTML += "<p>← Click to Unzap</p>";
```
mr-cheffy: Untranslated strings

src/zen/boosts/ZenZapOverlayChild.sys.mjs
```diff
@@ -0,0 +1,684 @@
...
+    this.zenBoostsChild.tempShowZappedElement(selector);
+
+    const { setTimeout } = ChromeUtils.importESModule("resource://gre/modules/Timer.sys.mjs");
+
+    // This has to run with delay, as the elements we are trying to highlight do not exist yet.
+    // The css has to load first and calculate the bounding boxes for the elements before we can highlight.
+    setTimeout(() => {
+      const selection = this.document.querySelectorAll(selector);
+      if (selection.length != 0) this.showHightlight(selection);
+    }, 10);
```
mr-cheffy: Timeout with un-explained milliseconds

src/zen/boosts/ZenZapOverlayChild.sys.mjs
```diff
@@ -0,0 +1,684 @@
...
+  /**
+   * Handles the mouse enter event for the unzap buttons
+   * @param {Event} event
+   */
+  #unzapButtonHover(event) {
+    const button = event.originalTarget;
+    const selector = button.getAttribute("selector");
+    this.zenBoostsChild.tempShowZappedElement(selector);
+
+    const { setTimeout } = ChromeUtils.importESModule("resource://gre/modules/Timer.sys.mjs");
```
mr-cheffy: Import lazily

src/zen/boosts/ZenZapOverlayChild.sys.mjs
```diff
@@ -0,0 +1,684 @@
...
+        top: `${elementMeasurement.top - padding}px`,
+        width: `${elementMeasurement.width + padding * 2}px`,
+        height: `${elementMeasurement.height + padding * 2}px`,
+      });
+
+      highlightContainerDiv.appendChild(highlightDiv);
+    }
+
+    this.getElementById("highlight-shadow").display = "initial";
+  }
```
mr-cheffy: Shouldn't it be `.style.display`?

src/zen/boosts/ZenZapOverlayChild.sys.mjs
```diff
@@ -0,0 +1,684 @@
...
+  /**
+   * Clears the highlight
+   */
+  removeHighlight() {
+    const highlightContainerDiv = this.getElementById("highlight-container");
+    highlightContainerDiv.style.display = "none";
+
+    // Clear all children elements
+    highlightContainerDiv.innerHTML = "";
+    this.getElementById("highlight-shadow").display = "none";
```
mr-cheffy: Same here

src/zen/boosts/ZenZapOverlayChild.sys.mjs
```diff
@@ -0,0 +1,684 @@
...
+  /**
+   * Handles the mouse move event
+   * @param {Event} event Mouse move event params
+   * @param {Boolean} isZapContent Flag if the target element is a zap related element
+   */
+  #handleMouseMove(event, isZapContent) {
+    if (this.#lastOverElement === event.target) return;
+    if (!isZapContent) this.#lastOverElement = event.target;
+
+    if (isZapContent) this.#hideHoverDiv();
```
mr-cheffy: Could just call `else` from the if statement above

src/zen/boosts/ZenZapOverlayChild.sys.mjs
```diff
@@ -0,0 +1,684 @@
...
+
+    if (isZapContent && event.originalTarget.id == "unzap") this.#unzapButtonClick(event);
+  }
+
+  /**
+   * Handles the mouse enter event
+   * @param {Event} event Mouse enter event params
+   */
+  #handleHoverDelegation(event) {
+    if (event.originalTarget.id == "unzap") this.#unzapButtonHover(event);
```
mr-cheffy: I'd simply add the event listener directly to the button to prevent any possible misunderstandings and call the event less often.

mr-cheffy (changes requested):
Looks good! Please also run `npm run lint:fix`

src/zen/boosts/actors/ZenBoostsChild.sys.mjs
FlorianButz: Haven't had issues or errors, but listening for unload would probably make more sense since we're only removing event listeners.

src/zen/boosts/actors/ZenBoostsChild.sys.mjs
FlorianButz: I changed it to use statically cached Sets, which reduces per-call memory allocation and does a more performant lookup than the linear search an array would give. It should be faster now

src/zen/boosts/actors/ZenBoostsChild.sys.mjs
FlorianButz: I don't think caching it would make much sense since it is already cached in the `ZenBoostStyles.sys.mjs`. 
I don't know why it's async, I probably wanted to add an await call for the sendQuery but forgot it. I removed the async.

src/zen/boosts/actors/ZenBoostsChild.sys.mjs
FlorianButz: I used to update the stylesheet separately for temporarily displaying the zap elements on hover, but since that's gone there's no need for this function anymore.

I removed it.

src/zen/boosts/actors/ZenBoostsChild.sys.mjs
FlorianButz: Correct, it's technically already doing that when checking if the boost is valid before applying the changes.

I removed it.

src/browser/themes/shared/zen-icons/common/selectable/brush.svg
FlorianButz: Old icon, it was replaced by the `paintbrush.svg` and `paintbrush-fill.svg` from the Nucleo icon pack.

src/zen/boosts/zen-advanced-color-options.css
FlorianButz: Removed the important and the display line entirely, added the flex instruction in the xhtml instead.

src/zen/boosts/zen-boosts.css
FlorianButz: Yes, I overused it in the editor css to simplify testing. I'll be removing as many as I can.

src/zen/boosts/ZenZapDissolve.sys.mjs
FlorianButz: Initially wanted to do that, but I wasn't too sure how I should load it in. Having it in class also saves performance as the class itself is created often. 
If the performance impact is too big, which I haven't tested but also didn't notice, I will probably have to create a pool of cached ones anyway.

src/zen/boosts/ZenZapOverlayChild.sys.mjs
FlorianButz: It does, I haven't found another way of getting to that value. Just using `gReduceMotion` is not possible here

src/zen/boosts/ZenZapOverlayChild.sys.mjs
FlorianButz: There has to be a delay in order for the highlight changes to apply. Five is an arbitrary number that worked for me while testing.

This should probably be replaced but I am not sure how to check for when the style changes are applied as 1 seemed to not be a large enough delay for the changes to apply.

src/zen/boosts/ZenZapOverlayChild.sys.mjs
FlorianButz: Same as above, I'll be looking into how to properly wait for the style update instead of waiting an arbitrary amount of time

src/zen/boosts/ZenZapOverlayChild.sys.mjs
FlorianButz: The reason I did it this way is because registering the event listener directly did nothing. I am not sure why, maybe it was an issue at the time. I'll try to debug and simplify.

FlorianButz:
Brought some much needed improvements to the CSS editor, inspired by Arc (as always). Made the `SelectorComponent` reusable and added translations for pretty much every text. 

The animation of the selector will be disabled for non macOS platforms to comply with platform specific design rules, but is enabled for the showcase.

(+) Other minor improvements in code structure as well

https://github.com/user-attachments/assets/6bc56d70-d681-4d19-8a45-727cfbdf70a7

src/zen/boosts/ZenZapOverlayChild.sys.mjs
FlorianButz: The magic number and timeout have been replaced with `window.requestAnimationFrame()`. This approach is more reliable

src/zen/boosts/ZenBoostsManager.sys.mjs
```diff
@@ -289,6 +287,90 @@ class nsZenBoostsManager {
...
+
+    // Replace if resulting filename is empty
+    if (!sanitized) sanitized = "New Boost";
+
+    fp.defaultString = sanitized;
+    fp.defaultExtension = "json";
+    fp.appendFilters(nsIFilePicker.filterAll);
+
+    return new Promise((resolve) => {
+      fp.open(async (result) => {
```
mr-cheffy, line 697: I would just do `fp.open(async (result) => {` also, the function is async for now reason?

mr-cheffy (changes requested):

src/zen/boosts/ZenBoostsManager.sys.mjs
FlorianButz: The promise is done so the editor has a callback for when the dialog action is complete. I forgot to remove the async when changing the pattern to promises. Will be removed

prefs/zen/boosts.yaml
```diff
@@ -0,0 +1,9 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+- name: zen.boosts.enabled
+  value: false
```
mr-cheffy: Should be `value: "@IS_TWILIGHT@"`

src/zen/boosts/moz.build
```diff
@@ -0,0 +1,36 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+EXTRA_JS_MODULES += [
```
mr-cheffy: I'd do `EXTRA_JS_MODULES.zen.boosts += ...`

src/zen/boosts/zen-advanced-color-options.css
```diff
@@ -0,0 +1,86 @@
...
+    margin-top: 10px;
+  }
+
+  & separator {
+    height: 24px !important;
+  }
+
+  @media (-moz-platform: macos) {
+    &[animate="open"] {
+      animation: var(--boost-spring-duration) zen-color-options-panel-animation-macos var(--boost-spring-easing) forwards !important;
```
mr-cheffy: I dont think we need such complex easing function, no? Id just use the existing zen animation

src/zen/boosts/zen-selector.css
```diff
@@ -0,0 +1,232 @@
...
+        1.0446 44.68%,
+        1.0407 47.44%,
+        1.0118 61.86%,
+        1.0027 69.52%,
+        0.9981 80.46%,
+        0.9991 99.96%
+      );
+      --boost-spring-duration: 0.6s;
+
+      animation: var(--boost-spring-duration) select-component-in var(--boost-spring-easing) forwards;
```
mr-cheffy: Same here, overly complex easing function

src/zen/boosts/zen-boosts.css
```diff
@@ -0,0 +1,806 @@
...
+
+.mod-button {
+  height: 38px !important;
+
+  border-radius: 6px;
+  @media (-moz-platform: macos) {
+    border-radius: 12px;
+  }
+
+  background-color: #ebebed;
```
mr-cheffy, line 333: Why not just keeping it transparent?

mr-cheffy (changes requested):

6639835 (changes requested):
P1: resetBoost calls gZenBoostsManager.getEmptyBoost(domain) at src/zen/boosts/ZenBoostsEditor.mjs:1067, but the manager only defines getEmptyBoostEntry at src/zen/boosts/ZenBoostsManager.sys.mjs:97. Triggering Reset can throw at runtime. Please use the existing empty-boost API and the correct payload shape, for example getEmptyBoostEntry().boostData.

FlorianButz:
Thanks! I missed that during the refactor

src/zen/boosts/zen-boosts.css
FlorianButz: What do you mean exactly? The button background?

6639835:
> Thanks! I missed that during the refactor

Also, I just want to mention this:
`RecomputeBrowsingContextDependentData` reads `aPresContext->mZenBoostsPresContextData` in `src/zen/boosts/nsZenBoostsBackend.cpp:166` before the field is guaranteed to be initialized. The field was added in `src/layout/base/nsPresContext-h.patch:9` without a default value, so the first comparison may read indeterminate data, causing undefined behavior and potentially triggering random restyle behavior.

mr-cheffy:
Auto color invert:

image

image

src/zen/boosts/actors/ZenBoostsChild.sys.mjs
```diff
@@ -0,0 +1,451 @@
...
+          0.1 + primaryGradientColor[2] * 0.6 * boostData.brightness
+        );
+
+        const rgbColor = boostData.autoTheme ? primaryGradientColor : colorWheelColor;
+        const nsColor = this.#rgbToNSColor(rgbColor, (1 - boostData.contrast) * 255);
+        browsingContext.zenBoostsData = nsColor;
+        return;
+      }
+    }
+    browsingContext.zenBoostsData = 0;
```
6639835, line 364: I think the no-boost path only sets `browsingContext.zenBoostsData = 0` and never clears `browsingContext.isZenBoostsInverted`.

It seems like we should explicitly reset `browsingContext.isZenBoostsInverted` in the no-boost path to avoid leaking state across navigations.

src/zen/boosts/ZenSelectorComponent.sys.mjs
```diff
@@ -0,0 +1,687 @@
...
+      const pathTypeElementLength = document.querySelectorAll(pathTypeElementBuilt).length;
+      if (!buildMap.has(pathTypeElementLength))
+        buildMap.set(pathTypeElementLength, pathTypeElementBuilt);
+
+      let parentExactElement = [];
+      if (isValidNode(element.parentNode)) {
+        traverse(element.parentNode, 0, parentExactElement);
+      }
+
+      const pathParentElementBuilt = build(parentExactElement);
```
6639835, line 676: In SelectorComponent.getSelectionPath, findBestExactSelector calls document.querySelectorAll(pathParentElementBuilt) even when pathParentElementBuilt is empty

6639835 (changes requested):

src/zen/boosts/actors/ZenBoostsChild.sys.mjs
FlorianButz: The `isZenBoostsInverted` has been excluded from the color boosting behavior.
The user should be able to disable color boosting and still apply an inversion filter.

src/zen/boosts/actors/ZenBoostsChild.sys.mjs
```diff
@@ -0,0 +1,493 @@
...
+          this.#startPickingOverlay();
+        } else if (this.#currentState === ZenBoostsChild.STATES.PICKER) {
+          this.disablePickerMode();
+        }
+        break;
+      case "ZenBoost:ZapModeEnabled":
+        return this.#currentState === ZenBoostsChild.STATES.ZAP;
+      case "ZenBoost:SelectorPickerModeEnabled":
+        return this.#currentState === ZenBoostsChild.STATES.PICKER;
+      case "ZenBoost:OpenInspector":
```
mr-cheffy, line 256: Why do we need this exactly? Doesn't make much more sense to send the message as we recieve it. Doesn't it make an infinite loop?

src/zen/boosts/actors/ZenBoostsChild.sys.mjs
```diff
@@ -0,0 +1,493 @@
...
+      this.#unloadCurrentStyleSheet();
+    }
+
+    if (boost) {
+      const { boostData } = boost.boostEntry;
+      if (boost.styleSheet) {
+        this.#loadStyleSheet(boost.styleSheet);
+      }
+
+      browsingContext.isZenBoostsInverted = boostData.smartInvert;
```
mr-cheffy, line 326: We are not resetting the invert if boost doesn't exist

src/zen/boosts/actors/ZenBoostsChild.sys.mjs
```diff
@@ -0,0 +1,493 @@
...
+      if (boostData.enableColorBoost) {
+        let colorWheelColor = this.#hslToRgb(
+          boostData.dotAngleDeg / 360,
+          /* already is [0, 1] */
+          boostData.dotDistance * (1 - boostData.saturation),
+          /* lightness range from [0.1, 0.7] */
+          0.1 + boostData.dotDistance * 0.6 * boostData.brightness
+        );
+
+        let primaryGradientColor = boost.workspaceGradient[0].c ?? this.#rgbToHsl([75, 75, 75]);
```
mr-cheffy: Directly use the result of `this.#rgbToHsl([75, 75, 75])` as a constant to avoid excessive computations. Also, what if there are no workspace gradients?

src/zen/boosts/actors/ZenBoostsParent.sys.mjs
```diff
@@ -0,0 +1,154 @@
...
+    ZenBoostsParent.OBSERVERS.forEach((observe) => {
+      Services.obs.addObserver(this._observe, observe);
+    });
+  }
+
+  /**
+   * Called when the actor is destroyed. Cleans up the observer.
+   */
+  async destroy() {
+    super.destroy();
```
mr-cheffy: `await`

src/zen/boosts/actors/ZenBoostsParent.sys.mjs
```diff
@@ -0,0 +1,154 @@
...
+        );
+
+        const { gDevTools } = require("devtools/client/framework/devtools");
+
+        let win = Services.wm.getMostRecentWindow("navigator:browser");
+        let tab = win.gBrowser.selectedTab;
+
+        let toolbox = gDevTools.getToolboxForTab(tab);
+
+        if (toolbox) {
```
mr-cheffy, line 84: It shouldn't matter, no? We always want to open the "inspector" tab

src/zen/boosts/actors/ZenBoostsParent.sys.mjs
```diff
@@ -0,0 +1,154 @@
...
+      case "ZenBoost:ZapSelector": {
+        const data = message.data;
+
+        if (!data.action) {
+          break;
+        }
+        if (!data.selector) {
+          break;
+        }
+        if (!data.domain) {
```
mr-cheffy, line 104: Can be simplified to `!data.action && ...`

src/zen/boosts/nsZenBoostsBackend.cpp
```diff
@@ -0,0 +1,257 @@
...
+}
+
+auto nsZenBoostsBackend::onPressShellLeave(nsPresContext* aPresContext)
+    -> void {
+  // TODO: We should set it as a null as well, but this prevents borders and
+  // shadows from being drawn into our Zen boosts modifications.
+  if (!aPresContext) {
+    return;
+  }
+  mCurrentPresContext = aPresContext;
```
mr-cheffy: Maybe I should be using a weak ref here

src/zen/boosts/zen-advanced-color-options.css
```diff
@@ -0,0 +1,52 @@
...
+    margin-top: 10px;
+  }
+
+  & separator {
+    height: 24px !important;
+  }
+
+  @media (-moz-platform: macos) {
+    &[animate="open"] {
+      animation: 0.4 zen-color-options-panel-animation-macos cubic-bezier(0.14, 1.43, 0.56, 1.01) forwards !important;
```
mr-cheffy, line 33: We don't need this anymore as we have native popups now

src/zen/boosts/zen-boost-editor.xhtml
```diff
@@ -0,0 +1,152 @@
...
+
+  <link rel="stylesheet" href="chrome://browser/content/zen-styles/zen-boosts.css" />
+  <link rel="stylesheet" href="chrome://browser/content/zen-styles/zen-buttons.css" />
+  <link rel="stylesheet" href="chrome://browser/content/zen-styles/zen-theme.css" />
+  <link rel="stylesheet" href="chrome://browser/content/zen-styles/zen-animations.css" />
+  <link rel="stylesheet" href="chrome://browser/content/zen-styles/zen-panel-ui.css" />
+  <link rel="stylesheet" href="chrome://browser/skin/zen-icons/icons.css" />
+  <link rel="stylesheet" href="chrome://browser/content/zen-styles/zen-advanced-color-options.css" />
+
+  <link rel="localization" href="browser/zen-boosts.ftl"/>
```
mr-cheffy, line 36: Wrap in `<linkset>` and mark it as html `<html:link ...`

src/zen/boosts/zen-boost-editor.xhtml
```diff
@@ -0,0 +1,152 @@
...
+    <vbox flex="1" id="zen-boost-code-editor">
+    </vbox>
+
+    <hbox id="zen-boost-code-bottom-bar">
+      <button data-l10n-id="zen-boost-css-picker" id="zen-boost-css-picker" class="subviewbutton mod-button big-button toggleable-button"></button>
+      <button data-l10n-id="zen-boost-css-inspector" id="zen-boost-css-inspector" class="subviewbutton mod-button big-button"></button>
+    </hbox>
+  </vbox>
+
+  <popupset id="mainPopupSet">
```
mr-cheffy, line 114: Move to top

src/zen/boosts/zen-boost-editor.xhtml
```diff
@@ -0,0 +1,152 @@
...
+        <p data-l10n-id="zen-bootst-color-brightness"></p>
+        <html:input id="zen-boost-color-brightness" type="range" min="0" max="1" value="0.5" step="0.01"/>
+        <separator />
+        <p data-l10n-id="zen-bootst-color-original-saturation"></p>
+        <html:input id="zen-boost-color-saturation" type="range" min="0" max="1" value="0.5" step="0.01"/>
+      </html:div>
+    </panel>
+  </popupset>
+
+  <menupopup id="zenBoostContextMenu">
```
mr-cheffy, line 129: Move to top

src/zen/boosts/zen-boosts.css
```diff
@@ -0,0 +1,804 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+#zenBoostWindow {
+  /* For the mica effect we want a white tint */
+  color-scheme: light !important;
+  appearance: none !important;
```
mr-cheffy: Why important?

src/zen/boosts/zen-boosts.css
```diff
@@ -0,0 +1,804 @@
...
+}
+
+#zen-boost-code-editor-root {
+  flex: 1;
```
mr-cheffy: Can just give it `flex="1"` to the element

src/zen/boosts/ZenBoostStyles.sys.mjs
mr-cheffy: Yeah, we might need to write these to a file and give out the file name?

src/zen/boosts/ZenZapDissolve.sys.mjs
```diff
@@ -0,0 +1,591 @@
...
+   * Helper for getting an anonymous element by id
+   */
+  getElementById(id) {
+    return this.content.root.getElementById(id);
+  }
+
+  get markup() {
+    return `
+    <template>
+      <link rel="stylesheet" href="chrome://browser/content/zen-styles/content/zen-zap.css" />
```
mr-cheffy, line 516: Shouldn't it be `scoped`?

src/zen/boosts/ZenZapOverlayChild.sys.mjs
```diff
@@ -0,0 +1,383 @@
...
+    this.window.requestAnimationFrame(() => {
+      this.#handleUnzap(selector);
+    });
+  }
+
+  /**
+   * Removes all event listeners and removes the overlay from the Anonymous Content
+   */
+  tearDown() {
+    this.#selectorComponent.tearDown();
```
mr-cheffy, line 318: Might also need to clear `this.window` and document

src/zen/common/styles/zen-animations.css
```diff
@@ -4,6 +4,29 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  */
 
+@keyframes zen-color-options-panel-animation-macos {
```
mr-cheffy: Not needed anymore

src/zen/common/styles/zen-animations.css
```diff
@@ -16,6 +39,18 @@
   }
 }
 
+@keyframes zen-boost-color-picker-dot-animation {
```
mr-cheffy: I wouldn't use this either, let it be static, less motion

mr-cheffy (changes requested):
We are so close to finishing this!

6639835:
Nice work! We’re almost finished!

src/zen/boosts/actors/ZenBoostsChild.sys.mjs
FlorianButz: These things report the state to the editor which is needed for the button color. Why would there be an infinite loop?

src/zen/boosts/actors/ZenBoostsParent.sys.mjs
FlorianButz: No, if we don't do this clicking the button multiple times will open a new inspector every time. We have to disable it if there's an open one.

src/zen/boosts/zen-boosts.css
FlorianButz: This didn't work without important if I recall correctly. I will check again though, maybe I had other styles conflicting

src/zen/boosts/actors/ZenBoostsChild.sys.mjs
mr-cheffy: Im talking about `ZenBoost:OpenInspector`

FlorianButz: No, messages received from the outside will land in the child actor. However the child actor can not open the inspector as it does not have the required references. So it has to send the message back to the Parent, which does not create a recursive loop

src/zen/workspaces/ZenGradientGenerator.mjs
```diff
@@ -1675,8 +1678,8 @@ export class nsZenThemePicker extends nsZenMultiWindowFeature {
     const gradient = nsZenThemePicker.getTheme(colors, this.currentOpacity, this.currentTexture);
     let currentWorkspace = gZenWorkspaces.getActiveWorkspace();
 
+    currentWorkspace.theme = gradient;
     if (!skipSave) {
-      currentWorkspace.theme = gradient;
       gZenWorkspaces.saveWorkspace(currentWorkspace);
     }
```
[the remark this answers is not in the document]
FlorianButz: This does not regard this PR and therefor won't be fixed here.

mr-cheffy (approved):

mr-cheffy:
https://github.com/zen-browser/desktop/pull/12586

PR #11122, opened 2025-11-02
feat: Add three command bar actions for switching the Light/Dark Mode

nrysk:
This PR adds three new actions to the Command Bar, allowing users to switch the browser's theme directory. This is useful for web developers checking appearance in different themes.

The following commands are added:
- `Switch to Automatic Appearance`
- `Switch to Light Mode`
- `Switch to Dark Mode`

These commands are dynamically hidden if they are already active.

**Note of Naming**: command naming was inspired by Arc Browser

## Screenshot
**Switch to Automatic Appearance**
image

**Switch to Light Mode**
image

**Switch to Dark Mode**
image

### DIFF
--- a/src/browser/base/content/zen-commands.inc.xhtml
+++ b/src/browser/base/content/zen-commands.inc.xhtml
@@ -61,8 +61,4 @@
 
   <command id="cmd_zenTogglePinTab" />
   <command id="cmd_zenCloseUnpinnedTabs" />
-
-  <command id="cmd_zenSwitchAutomaticAppearance" />
-  <command id="cmd_zenSwitchLightMode" />
-  <command id="cmd_zenSwitchDarkMode" />
 </commandset>
--- a/src/browser/base/content/zen-commands.inc.xhtml
+++ b/src/browser/base/content/zen-commands.inc.xhtml
@@ -61,4 +61,8 @@
 
   <command id="cmd_zenTogglePinTab" />
   <command id="cmd_zenCloseUnpinnedTabs" />
+
+  <command id="cmd_zenSwitchAutomaticAppearance" />
+  <command id="cmd_zenSwitchLightMode" />
+  <command id="cmd_zenSwitchDarkMode" />
 </commandset>
--- a/src/zen/common/zen-sets.js
+++ b/src/zen/common/zen-sets.js
@@ -131,17 +131,14 @@ document.addEventListener(
           }
           case 'cmd_zenSwitchAutomaticAppearance': {
             Services.prefs.setIntPref('zen.view.window.scheme', 2);
-            Services.prefs.setIntPref('layout.css.prefers-color-scheme.content-override', 2);
             break;
           }
           case 'cmd_zenSwitchLightMode': {
             Services.prefs.setIntPref('zen.view.window.scheme', 1);
-            Services.prefs.setIntPref('layout.css.prefers-color-scheme.content-override', 1);
             break;
           }
           case 'cmd_zenSwitchDarkMode': {
             Services.prefs.setIntPref('zen.view.window.scheme', 0);
-            Services.prefs.setIntPref('layout.css.prefers-color-scheme.content-override', 0);
             break;
           }
           default:
--- a/src/zen/common/zen-sets.js
+++ b/src/zen/common/zen-sets.js
@@ -129,18 +129,6 @@ document.addEventListener(
             gZenWorkspaces.unloadWorkspace();
             break;
           }
-          case 'cmd_zenSwitchAutomaticAppearance': {
-            Services.prefs.setIntPref('zen.view.window.scheme', 2);
-            break;
-          }
-          case 'cmd_zenSwitchLightMode': {
-            Services.prefs.setIntPref('zen.view.window.scheme', 1);
-            break;
-          }
-          case 'cmd_zenSwitchDarkMode': {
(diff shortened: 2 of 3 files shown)


src/zen/common/zen-sets.js
```diff
@@ -129,6 +129,21 @@ document.addEventListener(
...
+            break;
+          }
+          case 'cmd_zenSwitchLightMode': {
+            Services.prefs.setIntPref('zen.view.window.scheme', 1);
+            Services.prefs.setIntPref('layout.css.prefers-color-scheme.content-override', 1);
+            break;
+          }
+          case 'cmd_zenSwitchDarkMode': {
+            Services.prefs.setIntPref('zen.view.window.scheme', 0);
+            Services.prefs.setIntPref('layout.css.prefers-color-scheme.content-override', 0);
```
mr-cheffy: No need to change this if `zen.view.window.scheme` is already being changed

src/zen/urlbar/ZenUBGlobalActions.sys.mjs
```diff
@@ -153,6 +153,30 @@ const globalActionsTemplate = [
     command: 'Tools:Addons',
     icon: 'chrome://browser/skin/zen-icons/extension.svg',
   },
+  {
+    label: 'Switch to Automatic Appearance',
+    command: 'cmd_zenSwitchAutomaticAppearance',
+    icon: 'chrome://browser/skin/zen-icons/sparkles.svg',
+    isAvailable: () => {
+      return Services.prefs.getIntPref('zen.view.window.scheme') !== 2;
```
mr-cheffy: Would be nice to have this as a lazy pref for better performance

mr-cheffy (changes requested):

src/zen/common/zen-sets.js
nrysk: I've removed the unnecessary `layout.css.prefers-color-scheme.content-override` call.

src/zen/urlbar/ZenUBGlobalActions.sys.mjs
nrysk: I've implemented the `ThemeCache` for the lazy pref

nrysk:
@mr-cheffy
I've applied your feedback!

* Removed the unnecessary `layout.css.prefers-color-scheme.content-override` call.
* Implemented the `ThemeCache` for the lazy pref pattern directly within `ZenUBGlobalActions.sys.mjs`.

I wasn't 100% sure if this was the most appropriate file for this logic. If there's a better place for it, please let me know and I'd be happy to move it.

src/browser/base/content/zen-commands.inc.xhtml
```diff
@@ -61,4 +61,8 @@
 
   <command id="cmd_zenTogglePinTab" />
   <command id="cmd_zenCloseUnpinnedTabs" />
+
+  <command id="cmd_zenSwitchAutomaticAppearance" />
```
mr-cheffy: Now that I think about it, we don't need to make so much boilerplate for this. We shoudn't be using commands, we should rather use inline commands.

Like this: 
https://github.com/zen-browser/desktop/blob/96c6ec658cff4ec8b69192375ce8fbfc68737a29/src/zen/urlbar/ZenUBGlobalActions.sys.mjs#L37

src/zen/common/zen-sets.js
```diff
@@ -129,6 +129,18 @@ document.addEventListener(
             gZenWorkspaces.unloadWorkspace();
             break;
           }
+          case 'cmd_zenSwitchAutomaticAppearance': {
```
mr-cheffy: As said in the comments above, we don't really need this

src/zen/urlbar/ZenUBGlobalActions.sys.mjs
```diff
@@ -6,6 +6,26 @@ function isNotEmptyTab(window) {
   return !window.gBrowser.selectedTab.hasAttribute('zen-empty-tab');
 }
 
+const ThemeCache = {
```
mr-cheffy: Move it on top of the function and call it `lazy`.

Not sure why you are doing your own implementation when one already exists, for example, here:

https://github.com/zen-browser/desktop/blob/96c6ec658cff4ec8b69192375ce8fbfc68737a29/src/zen/urlbar/ZenUBActionsProvider.sys.mjs#L32-L37

src/zen/urlbar/ZenUBGlobalActions.sys.mjs
```diff
@@ -153,6 +173,30 @@ const globalActionsTemplate = [
...
+    label: 'Switch to Automatic Appearance',
+    command: 'cmd_zenSwitchAutomaticAppearance',
+    icon: 'chrome://browser/skin/zen-icons/sparkles.svg',
+    isAvailable: () => {
+      return ThemeCache.value !== 2;
+    },
+  },
+  {
+    label: 'Switch to Light Mode',
+    command: 'cmd_zenSwitchLightMode',
```
mr-cheffy: See my comment above, about using inline commands

mr-cheffy (changes requested):

nrysk:
Thank you for the detailed review! This was very helpful.

mr-cheffy (approved):

PR #11129, opened 2025-11-02
Fix prevent session loss when library is closed after main window (#11073)

### LINKED ISSUE #11073: Tabs lost when closing main window before Library window
### Preliminary Checks

### What happened?

When the Library window (opened with Ctrl + Shift + O) is left open and the main browser window is closed first, closing the Library window afterward causes all previously open tabs and windows to be lost.
Even if “Open previous windows and tabs” or “Continue where you left off” is enabled, reopening Zen Browser starts with no restored tabs.

Possibly related to session handling when non-main windows are closed last?

### Expected behavior

Browser should restore the previously opened tabs and windows according to session restore settings.

### Actual behavior

Browser opens with no restored tabs — session data seems to be lost if the main window is closed before the Library window.

### Steps to reproduce

1. Open Zen Browser.
2. Open several tabs.
3. Press Ctrl + Shift + O to open the Library window.
4. Close the main browser window (keep the Library window open).
5. Close the Library window.
6. Reopen Zen Browser.

### Screenshots and videos

_No response_

### Version

1.17.3b

Windows - x64

Tabs

```shell

```

Radless0:
Closing the Library window after the main window used to clear the session state completely; this patch ensures session save is skipped for non-browser windows and adds a test to cover this scenario.

Fix: #11073


src/zen/patches/fix-library-session.patch
```diff
@@ -0,0 +1,23 @@
+# HG changeset patch
+# User GitHub Copilot <copilot@github.com>
+# Date 1698940800 0
+# Node ID abc123def456
+# Parent  0000000000000000000000000000000000000000
+Fix session restore when Library window is closed last
+
+diff --git a/browser/components/sessionstore/SessionStore.sys.mjs b/browser/components/sessionstore/SessionStore.sys.mjs
```
mr-cheffy: We don't need this file, do we?

mr-cheffy (changes requested):

src/zen/patches/fix-library-session.patch
Radless0: > We don't need this file, do we?

Yes, this is an duplicate file, so I deleted it.

mr-cheffy:
Would it also be interesting to also ignore private windows?

Radless0:
> Would it also be interesting to also ignore private windows?

As far as I know, private windows are already excluded from session restore by design, so I think it’s unnecessary to handle them here as well.

mr-cheffy:
I dont understand the patch then, because this check for "navigator:browser", isn't it done before by checking if __Ssi exists on the window's attributes?

Radless0:
> I dont understand the patch then, because this check for "navigator:browser", isn't it done before by checking if __Ssi exists on the window's attributes?

In Zen, private windows are handled as a completely separate process (Zen Private Browsing), so they don’t interact with the main session store at all, that’s why I didn’t include them in this patch.

As for the navigator:browser check in Zen, secondary windows like the Library still run inside the main process and may still expose session-related attributes such as __Ssi. That’s why the extra navigator:browser check is needed: it ensures only real browser windows affect the session state when closing.

mr-cheffy:
Wouldn't it be more interesting to submit this patch to upstream firefox?

mr-cheffy:
This should've been fixed with window sync, but thanks for showing interest!

PR #11144, opened 2025-11-04
feat: add intelligent tab grouping by similarity

matteomekhail:
This feature automatically organizes workspace tabs into categories based on their content and domain.

How it works:
- Analyzes all unpinned tabs in the active workspace
- Uses keyword matching to categorize tabs by topic (Development, Social, Shopping, Video, News, etc.)
- Groups similar tabs together with visual category headers (Similar to Arc Browser)

The grouping is triggered via a new 'Group' button in the workspace toolbar.

Is not probably the best way to categorize it, a better future way would be to use AI to generate the categories based on the pages, instead of having a map with keyword, but this is definetely a cheaper way.

Localization support added for: en-US, en-GB, it (AI generated Language files, IT as is my main language)


matteomekhail:
image

This is how the "Group" button is implemented in the UI, and also how the categories are shown in the sidebar.

mr-cheffy:
Have you tested this by the way?

matteomekhail:
> Have you tested this by the way?

Yes i did tested, but my bad, I've let AI do some things and didn't realized it forced pushed, I fixed it by restoring the working version, i'm really sorry for the confusion.

Here is a video showing the functionality:

https://github.com/user-attachments/assets/c83f04a2-c331-476c-b988-ec0192c4fd7a

mr-cheffy:
I think we should be using Firefox's local llm for this, no?

ZiedYousfi:
This is noisier than it needs to be. 

Also the list feels way too subjective. The Internet is way too big to be condensed and grouped in that way.

It also means that if we want to change the list we have to release a new version.

matteomekhail:
> I think we should be using Firefox's local llm for this, no?

didn't know i could do that, will implement that now

mr-cheffy:
> This is noisier than it needs to be.
> 
> Also the list feels way too subjective. The Internet is way too big to be condensed and grouped in that way.
> 
> It also means that if we want to change the list we have to release a new version.

Yep, I believe having a small, local LLM that just takes a look at tab titles and domains could serve just as good. Good thing about Firefox's LLM is that its not intrusive, its private and doesn't really do anything unless it being explicitly called. IIRC they now use the GPU for them as well? Meaning they are faster and less resource intensive.

@matteomekhail read: https://firefox-source-docs.mozilla.org/toolkit/components/ml/api.html

matteomekhail:
Been trying to implement this LLM model solution in the last couple of days, and none of those can give me good result

BGE-base (or even small): 
Pros: stronger embeddings, fewer random merges.
Cons: still splits similar brands (Audi/Mercedes), many micro-clusters 

all-MiniLM-L6-v2:
Pros: model tailored for clustering, lightweight.
Cons: extremely sensitive threshold; based on configuration, either single-tab clusters, otherwise mixes distant domains

distilgpt2 :
Pros: basically none lol, just very light and fast
Cons: not instruction-tuned → needs heavy prompt engineering/parsing; still overuses the “Other” category, basically it can't understand how to categorize them

All the other Xenova model don't do the job as well, either they not able to generate the right categories title, or they will not categorize correctly ( ad example not merging adidas, nike, gucci or Lv under the same category)

The only approach that i've tried and it sorta work is a domain based, where i just look at the URL and merge them if they have the same based (merge all the yt tabs ad example) but that will not create a "video" category including other video provider like vimeo (as is a different domain)

I really want to implement this feature in the best possible way as it was my favorite feature on Arc, so if anyone has any idea or model to try i'll be very happy to

By "cluster" i mean a collection of tabs under a URL
Thanks for making the PR a draft! i should have done it sooner @mr-cheffy

12th-devs:
@matteomekhail I'm not sure if it could be any help, but I vibecoded a project like this that used two of the models Firefox uses to sort tabs (as I had the same issues as you with the other model) and had amazing results. I learned quite a lot about how it all worked in the process, so if I could help in any way, please feel free to reach out.

https://github.com/Vertex-Mods/Zen-Tidy-Tabs

12th-devs:
I would also look on searchfox and try to find the prompt they use for naming the tabs, as I don't remember off the top of my head

matteomekhail:
> @matteomekhail I'm not sure if it could be any help, but I vibecoded a project like this that used two of the models Firefox uses to sort tabs (as I had the same issues as you with the other model) and had amazing results. I learned quite a lot about how it all worked in the process, so if I could help in any way, please feel free to reach out.
> 
> https://github.com/Vertex-Mods/Zen-Tidy-Tabs



wow i just saw the screenshot and it looks good! i'll have a look at the code tomorrow, thank you!

mr-cheffy:
Firefox does have an auto-grouping feature, might be worth to checkout how they do it:

https://searchfox.org/firefox-main/source/browser/components/tabbrowser/SmartTabGrouping.sys.mjs

12th-devs:
> Firefox does have an auto-grouping feature, might be worth to checkout how they do it:
> 
> https://searchfox.org/firefox-main/source/browser/components/tabbrowser/SmartTabGrouping.sys.mjs

~~That's the model I use in the script~~

Ignore this, two different things.

12th-devs:
> > @matteomekhail I'm not sure if it could be any help, but I vibecoded a project like this that used two of the models Firefox uses to sort tabs (as I had the same issues as you with the other model) and had amazing results. I learned quite a lot about how it all worked in the process, so if I could help in any way, please feel free to reach out.
> > 
> > https://github.com/Vertex-Mods/Zen-Tidy-Tabs
> 
> 
> 
> wow i just saw the screenshot and it looks good! i'll have a look at the code tomorrow, thank you!

I might suggest feeding it the url along with the title when you give it to the naming model, and if possible the tabs history, as Arc seems to do that too. I can also provide the prompts that Arc uses, I will drop the link here when I find it.

This is all the prompts TBC uses in there AI stuff: https://github.com/neurokitti/Arc-genAI

matteomekhail:
> > > @matteomekhail I'm not sure if it could be any help, but I vibecoded a project like this that used two of the models Firefox uses to sort tabs (as I had the same issues as you with the other model) and had amazing results. I learned quite a lot about how it all worked in the process, so if I could help in any way, please feel free to reach out.
> > > 
> > > https://github.com/Vertex-Mods/Zen-Tidy-Tabs
> > 
> > 
> > 
> > wow i just saw the screenshot and it looks good! i'll have a look at the code tomorrow, thank you!
> 
> I might suggest feeding it the url along with the title when you give it to the naming model, and if possible the tabs history, as Arc seems to do that too. I can also provide the prompts that Arc uses, I will drop the link here when I find it.
> 
> This is all the prompts TBC uses in there AI stuff: https://github.com/neurokitti/Arc-genAI
> 

Yep, I thought about that too, I also considered feeding the metadata from each tab’s meta tag, but that feels a bit over-engineered
The only issue I’ve found before with feeding the whole URL is that, on sites like YouTube, if a video has a title like “playing bla bla bla”, it treats it as if it were a playing website rather than a video player, and puts it in a different cluster.
I’ll try your solution tomorrow and see if I get the same behavior

matteomekhail:
I've implemented the solution using 12th suggestion, and it works much better now. It’s still not a perfect solution (ad example streaming instead of clothing, see video for reference) , but given the models available at the moment, I believe it’s the best possible solution, can always be upgraded later as smaller and better models will release and become available in Firefox.

Any suggestion to improve it is appreciate it! Tomorrow i'll make the UI look cleaner & add an animation while the model works

Attached is the video showing a demo

https://github.com/user-attachments/assets/277b9c81-e502-4e95-bc9a-0aa346ab0fb5

Thanks @12th-devs again :)

12th-devs:
Yeah of course, glad it worked out!

mr-cheffy:
Please let me know when its done for review!

matteomekhail:
> Please let me know when its done for review!

Sure it is ready now, Thank you!

src/browser/themes/shared/zen-icons/icons.css
```diff
@@ -393,6 +393,18 @@
   list-style-image: url('arrow-down.svg') !important;
 }
 
+.zen-workspace-group-tabs-button {
+  list-style-image: none !important;
```
mr-cheffy: No need to add `!important` here?

src/zen/tabs/zen-tabs/vertical-tabs.css
```diff
@@ -1364,3 +1364,49 @@
 .tab-group-label-container[zen-dragtarget] {
   z-index: 9 !important;
 }
+
+/* ==========================================================================
+   Tab Category Headers (for Group function)
+   ========================================================================== */
+.tabbrowser-tab[zen-category-first="true"] {
```
mr-cheffy: This looks quite hacky to be honest, why not use something like firefox's existing tab group implementation?

src/zen/workspaces/ZenWorkspace.mjs
```diff
@@ -14,6 +14,28 @@ class nsZenWorkspace extends MozXULElement {
           <vbox class="zen-workspace-tabs-section zen-workspace-pinned-tabs-section" hide-separator="true">
             <hbox class="pinned-tabs-container-separator">
               <toolbarseparator flex="1" />
+              <toolbarbutton command="cmd_zenGroupTabs"
+                             class="zen-workspace-group-tabs-button toolbarbutton-1">
+                <div class="zen-group-icon-container" style="display: grid; place-items: center; margin-inline-end: 5px;">
+                  <svg height="16px" width="16px" class="zen-group-loading-icon" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg" style="grid-area: 1 / 1;">
```
mr-cheffy: Please don't use inline SVGs, rather use make a file and link it so it can be cached

src/zen/workspaces/ZenWorkspaces.mjs
```diff
@@ -2737,6 +2741,616 @@ class nsZenWorkspaces extends nsZenMultiWindowFeature {
     });
   }
 
+  async groupTabsBySimilarity() {
```
mr-cheffy, line 2750: I'd move this into it's own `ZenTabsTidy.sys.mjs` file and just import it lazyly. maybe a new file under `zen/ml/ZenTabsTidy.sys.mjs`

src/zen/workspaces/zen-workspaces.css
```diff
@@ -6,6 +6,16 @@
 @namespace html 'http://www.w3.org/1999/xhtml';
 @namespace xul 'http://www.mozilla.org/keymaster/gatekeeper/there.is.only.xul';
 
+.zen-group-loading-icon {
```
mr-cheffy: Instead of a loading icon, shouldn't we make the separator a squiggly line going up and down? Like arc.

mr-cheffy (changes requested):

src/zen/tabs/zen-tabs/vertical-tabs.css
matteomekhail: image

Well it would look like that and i'm not sure i like it? If you like it more i can leave it like that though

src/zen/workspaces/ZenWorkspace.mjs
```diff
@@ -1,10 +1,10 @@
 // This Source Code Form is subject to the terms of the Mozilla Public
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
-
-class nsZenWorkspace extends MozXULElement {
-  static get markup() {
-    return `
+{
```
mr-cheffy, line 4: Also, seems like something has gone wrong while migrating the branch?

mr-cheffy (changes requested):

src/zen/tabs/zen-tabs/vertical-tabs.css
mr-cheffy: We can change the styles of those. Just make the background always transparent and make it use the color we use for tab groups basically

12th-devs: @matteomekhail if you need UI/UX for the tab groups, while it does not work anymore, an old project of mine [ATG](https://github.com/Vertex-Mods/Advanced-Tab-Groups) should help with both of those

matteomekhail: image


I've just figure it out ahah,using the firefox group implementation it and changing the style it would look like that

src/zen/workspaces/zen-workspaces.css
matteomekhail: Done, i did the spinner before cause i though was harder to do, but it wasn't that hard after all

image
image

src/zen/workspaces/ZenWorkspaces.mjs
matteomekhail: Done, i moved the whole logic there

src/zen/workspaces/ZenWorkspace.mjs
matteomekhail: Yep, done, also removed the spinner one as is not needed anymore

src/zen/workspaces/ZenWorkspace.mjs
matteomekhail: i think it had something to do when i fix the conflicts, whenever you review the new changes i'll check again to merge it

src/zen/tabs/zen-tabs/vertical-tabs.css
mr-cheffy: Looks good for now

src/zen/tabs/zen-tabs/vertical-tabs.css
```diff
@@ -146,12 +146,36 @@
...
     margin: auto 4px;
+    position: relative;
+    border: none;
+    overflow: visible;
 
     &::before {
       border-inline-start: none;
     }
+
+    &::after {
```
mr-cheffy: I'd only define this element once data-grouping is enabled

src/zen/ml/ZenTabsTidy.sys.mjs
```diff
@@ -0,0 +1,647 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+const Services = globalThis.Services;
```
mr-cheffy: `Services` will always be global, we don't need this

src/zen/ml/ZenTabsTidy.sys.mjs
```diff
@@ -0,0 +1,647 @@
...
+}
+
+async function generateAiClusterLabel(clusterData) {
+  if (!Services.prefs.getBoolPref('browser.ml.enable', false)) {
+    return null;
+  }
+
+  try {
+    const { createEngine } = ChromeUtils.importESModule(
+      'chrome://global/content/ml/EngineProcess.sys.mjs'
```
mr-cheffy: Maybe lazyly load this module?

https://github.com/zen-browser/desktop/blob/dev/src/zen/urlbar/ZenUBActionsProvider.sys.mjs#L23-L30

src/zen/ml/ZenTabsTidy.sys.mjs
```diff
@@ -0,0 +1,647 @@
...
+        }
+        return desc;
+      })
+      .join('\n\n');
+
+    const titles = clusterData.map((data) => data.title).filter((title) => title.length > 0);
+    const keywords = extractKeywords(titles);
+    const domains = [...new Set(clusterData.map((data) => data.domain))].join(', ');
+
+    const prompt = `You are an expert organizer who creates concise, descriptive category names.
```
mr-cheffy, line 497: Not sure if this is the best way we should do it, specially since there might be some possible vulnerability by directly injecting the scripts. Firefox's implementation (https://searchfox.org/firefox-main/source/browser/components/tabbrowser/SmartTabGrouping.sys.mjs) doesn't do this, for example

mr-cheffy (changes requested):
Good, but instead of adding the new file to the jar manifest, better just make a new moz.build on that directory and declare it as a system module.

For example: 
https://github.com/zen-browser/desktop/blob/886b559b507438c38b640dce0657bd627acfeb51/src/zen/common/moz.build#L6

src/zen/tabs/zen-tabs/vertical-tabs.css
matteomekhail: Yep good idea, done

src/zen/ml/ZenTabsTidy.sys.mjs
matteomekhail: Removed now, thanks

src/zen/ml/ZenTabsTidy.sys.mjs
matteomekhail: Not sure i've done it correctly, please double check, but should work

src/zen/ml/ZenTabsTidy.sys.mjs
matteomekhail: I don't see any possibility to inject scripts here?
also i believe that firefox implementation is very similar,  we both import createEngine, call the model, even tho it uses a much smaller prompt

can you please tell me what worries you exactly so i can fix it? or if you have some alternatives in mind?

12th-devs:
What all still needs to happen here?

src/browser/themes/shared/zen-icons/icons.css
```diff
@@ -393,6 +393,18 @@
...
 }
 
+.zen-workspace-group-tabs-button {
+  list-style-image: none;
+
+  & .toolbarbutton-icon {
+    display: none !important;
+  }
+
+  @media -moz-pref('zen.view.show-clear-tabs-button', false) {
```
mr-cheffy, line 403: Should also be hidden if `browser.ml.enable` is disabled

src/zen/ml/ZenTabsTidy.sys.mjs
mr-cheffy: > can you please tell me what worries you exactly so i can fix it? or if you have some alternatives in mind?
Im still investigating how firefox does it as they dont have any sort of "huge text" chunk going on

mr-cheffy (changes requested):

src/browser/themes/shared/zen-icons/icons.css
matteomekhail: Yes agree, i just fixed that

src/zen/ml/ZenTabsTidy.sys.mjs
matteomekhail: > I think it's because Mozilla Firefox uses this https://huggingface.co/Mozilla/smart-tab-topic for labeling

i am also using this?

image
image


I'm actually using both smart-tab-embedding & smart-tab-topic, one to see how similar the tabs are, one to generate the label, not sure if is the best option thought

Minemetero: Firefox’s SmartTabGrouping https://searchfox.org/firefox-main/source/browser/components/tabbrowser/SmartTabGrouping.sys.mjs#1199-1210 only sends a tiny template to the model and then cleans up the result with a bunch of JS
ZenTabsTidy just puts all those rules directly into the prompt instead
So they basically achieve the same thing

matteomekhail: Yep, correct

> Firefox’s SmartTabGrouping https://searchfox.org/firefox-main/source/browser/components/tabbrowser/SmartTabGrouping.sys.mjs#1199-1210 only sends a tiny template to the model and then cleans up the result with a bunch of JS ZenTabsTidy just puts all those rules directly into the prompt instead So they basically achieve the same thing

12th-devs:
@matteomekhail from what I remember, the AI model sorted the tabs really fast, so I added a minimum time in the UI as it made it feel more natural, because of it doing the sorting instantly and seemed unsettling to me almost. Just something to consider based on your testing

matteomekhail:
> @matteomekhail from what I remember, the AI model sorted the tabs really fast, so I added a minimum time in the UI as it made it feel more natural, because of it doing the sorting instantly and seemed unsettling to me almost. Just something to consider based on your testing

That's probably a good idea, i don't want it to feel fake thought, i will try to see what it feels better

matteomekhail:
> That's probably a good idea, i don't want it to feel fake thought, i will try to see what it feels better

Increased it from 500 to 750ms, it feels a bit better now, 1s was way too much based on my testing, agreed that 500ms was maybe too fast

peteygao:
It only feels fast because you're testing with a small number of tabs. Increase the tab count to something reasonable (100+ for starters) where you'd needing a grouping feature and it'll start to crawl (transformer models slows down quadratically as context lengths increases). I wouldn't add any artificial delay to the grouping functionality.

12th-devs:
> It only feels fast because you're testing with a small number of tabs. Increase the tab count to something reasonable (100+ for starters) where you'd needing a grouping feature and it'll start to crawl (transformer models slows down quadratically as context lengths increases). I wouldn't add any artificial delay to the grouping functionality.

I agree with the context length stuff, that's a known issue with these models, but respectfully, I disagree with you about the artificial delay. If you sort 100 tabs and it takes 5 seconds, the delay is gonna be ignored, but if it's 10 tabs and it's sorted in 0.3 seconds, then the 1 second delay is gonna make it feel a bit more real. Does that clear things up a bit?

mr-cheffy:
> then the 1 second delay is gonna make it feel a bit more real. Does that clear things up a bit?

Users don't care if it feels real or not, artificial delay means users will always be waiting for X amounts of seconds and that can become tiresome in the long run

BlueFox1616:
Leaving this here because i thing Arc has a potentially better method for gathering more info about each tab 
https://github.com/neurokitti/Arc-genAI
I think Arc also utilizes the tabs history as far as i remember

jacksongoode:
> Leaving this here because i thing Arc has a potentially better method for gathering more info about each tab [neurokitti/Arc-genAI](https://github.com/neurokitti/Arc-genAI?rgh-link-date=2026-01-04T00%3A43%3A57Z) I think Arc also utilizes the tabs history as far as i remember

I don't think this needs to be included in this PR, but Arc/Dia in addition does a few nice things that could eventually be done in this implementation. This list was a result of an LLM review - not directly by me. Still worth adding to this thread given the feature could benefit from it:

   * Similarity consideration into the algorithm when two tabs share the same opener (parent-child relationship).
   * Factor the time tabs were opened into the clustering logic; tabs opened in a quick sequence are often part of the same mental task.
   * Prioritize the first path segment of a URL (e.g., github.com/zen-browser/) over the page title.
   * The local topic-generation prompt could look for a specific "Task" or "Project" name rather than just the most frequent keyword - considering intent.
   * Consider the "active tab" at the time of opening a new tab as a strong semantic anchor for the new cluster.

houmanhafez:
Any updates on this? It's a really nice feature :)

matteomekhail:
> Any updates on this? It's a really nice feature :)

i will pull all the latest changes and maybe follow up on this on the weekend, thanks for reminding me :)

JonathanXDR:
@matteomekhail Thanks for this cool feature, really nice addition!

I had a couple of small thoughts while looking at this PR. I’m not sure how the current version looks, but it would be nice to adjust the button to better match how it was implemented in Arc. I really like the name "Group" as it’s much clearer and more descriptive than "Tidy," which Arc used. That said, we could still refine the hover style and icon, and maybe add a divider to improve the overall feel. What do you think?

grafik

wuhaoyujerry:
Hi, any update on this feature? Would love to use it.

PR #11160, opened 2025-11-05
fix: #9997 New container tab popup menu displayed incorrectly

### LINKED ISSUE #9997: "New Tab" tab bar button renders Chinese text vertically in container menu
### Preliminary Checks

### What happened?

I use container tabs. When the "New Tab" button on the tab sidebar is clicked and held to bring up the container menu, Chinese text in the container name string is rendered vertically. This is undesirable for users who use Chinese language tab strings as it consumes an excessive amount of vertical display space.

### Expected behavior

For this purpose, and by default, Chinese (double-byte) text should be rendered horizontally, left to right.

### Actual behavior

Chinese text renders vertically.

### Steps to reproduce

1. Activate container tabs.
2. Create a container tab with a Chinese text string (e.g. 中文版).
3. Click and hold the "New Tab" tab bar button to bring up the container tab selection menu.

This behavior does not reproduce using the File -> New Container Tab menu item. I could not test on Firefox because this is a Zen-specific feature.

### Screenshots and videos

Image
Image

### Version

1.14.11b

macOS - aarch64

Tabs

```shell

```

Ivaon:
Fix #9997 
As stated in the issue, the context menu to create a new tab with a specific container (opened via holding the new tab button) is displayed differently than the right click's context menu on the new tab button.

# Image of the issue

image

# Image after the fix

image

### DIFF
--- a/src/zen/common/styles/zen-single-components.css
+++ b/src/zen/common/styles/zen-single-components.css
@@ -609,3 +609,13 @@ body > #confetti {
 #widget-overflow-list {
   flex-direction: column-reverse;
 }
+
+/* Contextual new tab popup menu */
+#tabs-newtab-button > .new-tab-popup > menuitem {
+  & > :is(.menu-highlightable-text, .menu-accel) {
+    display: none !important;
+  }
+  & > label {
+    opacity: 1 !important;
+  }
+}


mr-cheffy (approved):

mr-cheffy:
Why the `opacty: 1;` and why use `!important` for the items?

Ivaon:
You're right, doesn't need the second `!important`.

There are two `menupopup`. The first one (right click) `#new-tab-button-popup` and the second (hold click) is a child of the toolbar button `#tabs-newtab-button`. And the first one have opacity set at 1.

So I wanted for them to look similar.

Do you want me to make a pr to change it?

Ivaon:
For the first one `!important`, theoretically I can remove it. Just there is somewhere with a pretty high specificity on the `display`.

And it will require something like, for example, this to get rid of the first one `!important`:

```css
#navigator-toolbox #titlebar #TabsToolbar #tabs-newtab-button > .new-tab-popup > menuitem {
  & > :is(.menu-highlightable-text, .menu-accel) {
    display: none;
  }
  & > label {
    opacity: 1;
  }
}
```

mr-cheffy:
Why do we need `opacity: 1` though

Ivaon:
Huh? I think I'm becoming crazy. I swear that i saw a difference in opacity, but *apparently not*. 

> Why do we need opacity: 1 though

@mr-cheffy, yeah good catch! We don't need it.

Ivaon:
Really, I'm sorry.

So how do you want to proceed next?

mr-cheffy:
A pr would be fine

PR #11190, opened 2025-11-07
feat: Close tab and switch to recent tab when going back with no history

KyleKincer:
## Description

This PR implements a feature where navigating back on a regular unpinned tab with no browser history will close the tab and switch to the most recently visited tab, providing a more intuitive browsing experience similar to other modern browsers.

## Changes

- **Back navigation with no history now closes regular unpinned tabs**
- Automatically switches to the most recently visited tab
- Works with all input methods:
  - Back button click
  - Keyboard shortcuts (Alt+Left, Cmd+Left)
  - Trackpad swipe gestures

### Files Modified

1. **browser/base/content/browser-commands.js**
   - Added logic to check if back navigation is possible
   - If not, close the tab and switch to most recently visited tab
   - Only applies to regular unpinned tabs

2. **browser/base/content/browser.js**
   - Modified `UpdateBackForwardCommands` to keep back button enabled
   - Ensures back button is clickable even with no history for closeable tabs

3. **browser/base/content/browser-gestureSupport.js**
   - Modified `canGoBack` to allow swipe gestures
   - Enables trackpad gestures to trigger tab close

## Tab Types Excluded

The feature only affects regular unpinned tabs and excludes:
- Pinned tabs
- Essential tabs
- Empty tabs (zen-empty-tab)

## Testing

Tested with:
- ✅ Back button click
- ✅ Keyboard shortcuts
- ✅ Trackpad swipe gestures
- ✅ Verified exclusions (pinned, essential tabs)
- ✅ Tab history switching works correctly

b=no-bug, c=tabs, common, gestures

### DIFF
--- a/src/browser/base/content/browser-commands-js.patch
+++ b/src/browser/base/content/browser-commands-js.patch
@@ -1,8 +1,8 @@
 diff --git a/browser/base/content/browser-commands.js b/browser/base/content/browser-commands.js
-index 939ca497b882b3f4200141ba1b6764fb5c846f45..77abad7ace109bea5392930d6e7c06568685a46b 100644
+index 939ca497b882b3f4200141ba1b6764fb5c846f45..bdf343698bdf313a519a9c8fc4645aabd21be1cf 100644
 --- a/browser/base/content/browser-commands.js
 +++ b/browser/base/content/browser-commands.js
-@@ -14,6 +14,19 @@ var BrowserCommands = {
+@@ -14,6 +14,17 @@ var BrowserCommands = {
      const where = BrowserUtils.whereToOpenLink(aEvent, false, true);
  
      if (where == "current") {
@@ -10,11 +10,9 @@ index 939ca497b882b3f4200141ba1b6764fb5c846f45..77abad7ace109bea5392930d6e7c0656
 +      const canGoBack = gBrowser.webNavigation?.canGoBack;
 +
 +      if (!canGoBack) {
-+        const currentTab = gBrowser.selectedTab;
-+
-+        if (gZenCommonActions.shouldCloseTabOnBack(currentTab)) {
++        if (gZenCommonActions.shouldCloseTabOnBack()) {
 +          // removeTab handles tab selection (owner or most recent)
-+          gBrowser.removeTab(currentTab);
++          gBrowser.removeTab(gBrowser.selectedTab);
 +          return;
 +        }
 +      }
@@ -22,7 +20,7 @@ index 939ca497b882b3f4200141ba1b6764fb5c846f45..77abad7ace109bea5392930d6e7c0656
        try {
          gBrowser.goBack();
        } catch (ex) {}
-@@ -315,6 +328,10 @@ var BrowserCommands = {
+@@ -315,6 +326,10 @@ var BrowserCommands = {
        }
      }
  
@@ -33,7 +31,7 @@ index 939ca497b882b3f4200141ba1b6764fb5c846f45..77abad7ace109bea5392930d6e7c0656
      // A notification intended to be useful for modular peformance tracking
      // starting as close as is reasonably possible to the time when the user
      // expressed the intent to open a new tab.  Since there are a lot of
-@@ -399,6 +416,11 @@ var BrowserCommands = {
+@@ -399,6 +414,11 @@ var BrowserCommands = {
        return;
      }
  
@@ -45,7 +43,7 @@ index 939ca497b882b3f4200141ba1b6764fb5c846f45..77abad7ace109bea5392930d6e7c0656
      // Keyboard shortcuts that would close a tab that is pinned select the first
      // unpinned tab instead.
      if (
-@@ -406,8 +428,8 @@ var BrowserCommands = {
+@@ -406,8 +426,8 @@ var BrowserCommands = {
        (event.ctrlKey || event.metaKey || event.altKey) &&
        gBrowser.selectedTab.pinned
      ) {
--- a/src/browser/base/content/browser-commands-js.patch
+++ b/src/browser/base/content/browser-commands-js.patch
@@ -1,8 +1,28 @@
 diff --git a/browser/base/content/browser-commands.js b/browser/base/content/browser-commands.js
-index 939ca497b882b3f4200141ba1b6764fb5c846f45..36c830a503b004c0a332340f686c86cad68e9381 100644
+index 939ca497b882b3f4200141ba1b6764fb5c846f45..77abad7ace109bea5392930d6e7c06568685a46b 100644
 --- a/browser/base/content/browser-commands.js
 +++ b/browser/base/content/browser-commands.js
-@@ -315,6 +315,10 @@ var BrowserCommands = {
+@@ -14,6 +14,19 @@ var BrowserCommands = {
+     const where = BrowserUtils.whereToOpenLink(aEvent, false, true);
+ 
+     if (where == "current") {
++      // Check if we can go back in history
++      const canGoBack = gBrowser.webNavigation?.canGoBack;
++
++      if (!canGoBack) {
++        const currentTab = gBrowser.selectedTab;
++
++        if (gZenCommonActions.shouldCloseTabOnBack(currentTab)) {
++          // removeTab handles tab selection (owner or most recent)
++          gBrowser.removeTab(currentTab);
++          return;
++        }
++      }
++
+       try {
+         gBrowser.goBack();
+       } catch (ex) {}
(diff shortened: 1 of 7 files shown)


src/browser/base/content/browser-commands-js.patch
```diff
@@ -1,8 +1,43 @@
...
++      // Check if we can go back in history
++      const canGoBack = gBrowser.webNavigation?.canGoBack;
++
++      if (!canGoBack) {
++        const currentTab = gBrowser.selectedTab;
++
++        // Only close regular unpinned tabs (not pinned, not essential, not empty)
++        const isRegularTab = currentTab &&
++                             !currentTab.pinned &&
++                             !currentTab.hasAttribute('zen-essential') &&
```
mr-cheffy: We don't need to check for `zen-essential`, as essentials are already `pinned`

src/browser/base/content/browser-commands-js.patch
```diff
@@ -1,8 +1,43 @@
...
+index 456fcb56f40cf1936ebc614a467bc3223dd5e4d5..65cc1e81ddef443aff3031409feb119af6d682ea 100644
 --- a/browser/base/content/browser-commands.js
 +++ b/browser/base/content/browser-commands.js
-@@ -317,6 +317,10 @@ var BrowserCommands = {
+@@ -16,6 +16,34 @@ var BrowserCommands = {
+     const where = BrowserUtils.whereToOpenLink(aEvent, false, true);
+ 
+     if (where == "current") {
++      // Check if we can go back in history
++      const canGoBack = gBrowser.webNavigation?.canGoBack;
```
mr-cheffy: I'd make this into our own function and calling it from here, to avoid making big patch files. Maybe under `gZenCommonActions`?

src/browser/base/content/browser-gestureSupport-js.patch
```diff
@@ -0,0 +1,22 @@
...
+-    return gBrowser.webNavigation.canGoBack;
++    if (gBrowser.webNavigation.canGoBack) {
++      return true;
++    }
++    
++    // Allow swipe back gesture for regular unpinned tabs (to close them)
++    const currentTab = gBrowser?.selectedTab;
++    return currentTab &&
++           !currentTab.pinned &&
++           !currentTab.hasAttribute('zen-essential') &&
```
mr-cheffy: Same here with checking if its an essential. Also, from my testings on arc, I think we also need to check if there's a `currentTab.owner`? So we go back to the original tab

src/browser/base/content/browser-js.patch
```diff
@@ -10,7 +10,34 @@ index 32b67f846b9942ad3da4276bb8748fee915eb485..529ef8895d2ab01b4cbebd2992fd3b64
...
+   var forwardDisabled = forwardCommand.hasAttribute("disabled");
++  
++  // Check if we should keep back button enabled for regular unpinned tabs
++  var shouldEnableBackForTabClose = false;
++  if (!aWebNavigation.canGoBack) {
++    const currentTab = gBrowser?.selectedTab;
++    shouldEnableBackForTabClose = currentTab &&
++                                  !currentTab.pinned &&
++                                  !currentTab.hasAttribute('zen-essential') &&
++                                  !currentTab.hasAttribute('zen-empty-tab');
```
mr-cheffy: I feel like all these checks should be into a single util function, to avoid duplicating so much

mr-cheffy (changes requested):
Great work!

src/browser/base/content/browser-js.patch
mr-cheffy: We also need to check for a pref for this, as I would understand some people might not want this behavior

KyleKincer:
Thanks for the feedback-- I made those changes (except adding the preference) in the last commit. In my testing, the `tab.parent` doesn't behave exactly like Arc, but it mostly works. For example, a tab opened from a Google search page seems to have no parent, whereas if you open a subreddit from the Reddit homepage, it goes back to the parent Reddit tab as expected. 
I may just not understand how the Firefox implementation works

mr-cheffy:
`tab.parent`? It should be tab.owner

src/browser/base/content/browser-js.patch
```diff
@@ -10,7 +10,30 @@ index 32b67f846b9942ad3da4276bb8748fee915eb485..529ef8895d2ab01b4cbebd2992fd3b64
...
+   var forwardDisabled = forwardCommand.hasAttribute("disabled");
++  
++  // Check if we should keep back button enabled for regular unpinned tabs
++  var shouldEnableBackForTabClose = false;
++  if (!aWebNavigation.canGoBack) {
++    shouldEnableBackForTabClose = gZenCommonActions.shouldCloseTabOnBack(gBrowser?.selectedTab);
++  }
++  
+   if (backDisabled == aWebNavigation.canGoBack) {
+     if (backDisabled) {
```
mr-cheffy, line 24: Shouldn't we set `backDisabled` instead of making a new variable `shouldEnableBackForTabClose` here?

src/zen/common/ZenCommonUtils.mjs
```diff
@@ -107,4 +107,33 @@ var gZenCommonActions = {
...
+  },
+
+  /**
+   * Gets the tab to switch to when closing the current tab.
+   * Prioritizes the tab owner (if exists), otherwise uses most recently visited tab.
+   *
+   * @param {Tab} currentTab - The tab being closed
+   * @return {Tab|null} The tab to switch to, or null if none found
+   */
+  getTabToSwitchOnClose(currentTab) {
```
mr-cheffy: gBrowser.removeTab should already check if theres tab.owner and all of that stuff, I dont think we need to duplicate the behavior here.

mr-cheffy (changes requested):

mr-cheffy:
@KyleKincer could you please update the git conflicts?

To update firefox's engine:
`rm -rf engine .surfer && npm run download`

KyleKincer:
> @KyleKincer could you please update the git conflicts?
> 
> To update firefox's engine: `rm -rf engine .surfer && npm run download`

Fixed :rocket:  :+1:

mr-cheffy (approved):
Everything looks good, one last thing though. When exactly does tab.owner not become valid? Because I thought the entire point of this feature was to go back to the original tab when clicking "back". This however will always show "back" even though the original tab could not exist or is not valid anymore?

@KyleKincer

KyleKincer:
> Everything looks good, one last thing though. When exactly does tab.owner not become valid? Because I thought the entire point of this feature was to go back to the original tab when clicking "back". This however will always show "back" even though the original tab could not exist or is not valid anymore?
> 
> @KyleKincer

I'm really going for the Arc behavior, where a tab goes back to its "owner" when you do anything that goes "back" while it is on its origin page. What we have on this PR is _similar_, but doesn't quite work the same because of some of the cases when Firefox doesn't set `owner`. 

For example, on Arc, if you open a new tab in the background via Ctrl+Click, going back will always take you back to the tab that it was spawned from. This doesn't work in our currently implementation because in `browser/components/tabbrowser/content/tabbrowser.js` line 3016:
```
// If we're opening a foreground tab, set the owner by default.
ownerTab ??= inBackground ? null : this.selectedTab;
```

I would suggest that we set ownerTab regardless of whether the tab is opened in the background or foreground, similar to Arc's behavior. Thoughts?

mr-cheffy:
image

ctrl+click on a link does set the tab owner though? Note that it uses the `??=` operator, which just runs that condition if the owner is undefined, which it isn't when you ctrl+click. I say, make it check for the owner and lets test on twilight, see how it feels. If its not quite as expected, you can make another PR to remove that owner check. Because what would be really confusing for users is "why's the back arrow also working as a close tab button".

mr-cheffy (changes requested):
When I was testing this, the "back" arrow remains "disabled", maybe you forgot to export a patch?

mr-cheffy (changes requested):
When I was testing this, the "back" arrow remains "disabled", maybe you forgot to export a patch?

KyleKincer:
> When I was testing this, the "back" arrow remains "disabled", maybe you forgot to export a patch?

Yeah, I was missing the patch for `browser-js`, it's back in now :+1:

mr-cheffy:
Well, now its always "enabled". I'll fix it.

mr-cheffy (approved):

KyleKincer:
Thanks @mr-cheffy, I see more now that you're looking to optimize for as small of patches as possible. Will keep that in mind in the future :+1: thanks for fixing those. Will test once this is in twilight

PR #11229, opened 2025-11-10
feat: move translate to panel

Tymon3310:
Move the translate button from url bar to site data popup.

### DIFF
--- a/src/browser/base/content/zen-panels/site-data.inc
+++ b/src/browser/base/content/zen-panels/site-data.inc
@@ -31,6 +31,10 @@
                         data-l10n-id="zen-site-data-header-bookmark"
                         command="Browser:AddBookmarkAs"
                         flex="1" />
+        <toolbarbutton id="zen-site-data-header-translate"
+                        data-l10n-id="zen-site-data-header-translate"
+                        command="cmd_translate"
+                        flex="1" />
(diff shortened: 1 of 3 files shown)


mr-cheffy:
This looks like an overly-complex solution... I feel like instead of adding it to the unified panel, maybe there's another place it could be moved onto... where does arc place it, for example?

Tymon3310:
so arc on mac does it like this, but i wanted to always have an option to translate page even when firefox doesn't detect the language
Screenshot 2025-11-10 at 12 37 52
Screenshot 2025-11-10 at 12 37 42

mr-cheffy:
Hm... I can imagine having it there would make the actual clickable-area of the urlbar *really* small. I can see how it might be annoying for users when the urlbar is filled with actions (And woudn't work with double toolbar). I say, for now, lets just leave it how it is and ill think about how we can properly fix this issue without it being too annoying. Really sorry..

Tymon3310:
thats why i thought the best place is the panel, to uncluter the url bar

mr-cheffy:
Yes, but the solution here is immensely more complex than it has to be. Maybe, not sure why you are using mutation observers and all of that stuff. Why not just override the anchor the translations panel should assign to?

PR #11240, opened 2025-11-10
feat: Preserve tab owner for back-to-close

KyleKincer:
I was testing on Twilight and Firefox is aggressively removing tab owners in some cases. These changes attempt to maintain persistent tab owner relationships to enable consistent back-to-close behavior similar to Arc Browser.

## Changes

**Modified `tabbrowser.js` in 3 places:**

1. **Background tabs now set owner** - Ctrl+Click and middle-click tabs maintain owner relationship
    - It seems like this was usually happening anyway despite what this part of the code would suggest. I'm fine with removing this if you think it's unnecessary
2. **Switching tabs preserves owner** - Owner no longer cleared when changing tabs
    - If you spawned a new tab then switch tabs around, the owner is lost and back-to-close no longer works. 
3. **Related tab chains keep owner** - Multiple tabs opened from same source all maintain owner

## Testing

1. **Ctrl+Click**: Click link with Ctrl → switch to tab → press back → returns to original tab
2. **Multiple tabs**: Open several from one page → all can return to that page
3. **Tab switching**: Switch tabs multiple times → back still works

b=no-bug, c=tabs

### DIFF
--- a/src/browser/components/tabbrowser/content/tabbrowser-js.patch
+++ b/src/browser/components/tabbrowser/content/tabbrowser-js.patch
@@ -1,5 +1,5 @@
 diff --git a/browser/components/tabbrowser/content/tabbrowser.js b/browser/components/tabbrowser/content/tabbrowser.js
-index c0eafd4faf8d57b8486c5bf8917375850ec8147e..326bf96d9346aba7096d518fbf63cc349868d0d0 100644
+index c0eafd4faf8d57b8486c5bf8917375850ec8147e..3be730e772677a3c7aad3f1dd577b61969460dd2 100644
 --- a/browser/components/tabbrowser/content/tabbrowser.js
 +++ b/browser/components/tabbrowser/content/tabbrowser.js
 @@ -450,15 +450,64 @@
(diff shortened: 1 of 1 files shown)


mr-cheffy (commented):
Im not comfortable changing how owners work here since it might even lead to memory leaks. Could you please show me examples or videos where the owner gets reset and it doesn't on arc?

KyleKincer:
> Im not comfortable changing how owners work here since it might even lead to memory leaks. Could you please show me examples or videos where the owner gets reset and it doesn't on arc?

Sure. Here are the two main ones that I run into. 
- Issue 1: Navigating to any other page clears the tab owner

https://github.com/user-attachments/assets/c3a30d1e-79a2-43cc-8f82-2e9946923a0d

- Issue 2: Opening multiple tabs from one parent results in the children having no owner

https://github.com/user-attachments/assets/9a67f8a4-ed20-49aa-99b5-5748ee8ea8f0

mr-cheffy:
I'll move this into https://github.com/zen-browser/desktop/pull/11267

PR #11199, opened 2025-11-08
feat: Ctrl+Tab cycling to keep within essentials/regular tabs only

junnjiee:
## Overview
Added an option in settings (under General > Tabs) to restrict ctrl + tab cycling to essentials/regular tabs only. If the current tab is a regular tab, ctrl + tab will not cycle to essential tabs, and vice versa.

This feature is disabled if **Ctrl+Tab cycles through tabs in recently used order** is also toggled on.

## Implementation details

`tabbox-js.patch` -> `tabFilter`
- handles filtering of `"zen-essential"` tabs.
- only does the filtering if **sort by recently used** is toggled off

`main-js.patch`
- if both tab cycling behavior toggles are on, show text to notify the user that the cycling by restriction feature is not active
image

---
Issues/Discussions requesting for the feature: #4787, #6316, #9119.

Issue #4787, opened 2025-01-31
Ctrl-tab should not cycle through essential tabs


KorigamiK:
### Captchas

- [x] I have read the instructions.
- [x] I have searched existing issues and avoided creating duplicates.
- [x] I am not filing an enhancement request.

### What happened?

Cycling through tabs is a pain because essential tabs are something that should stay where they are, only when I'm navigating the essential tabs using ctrl-tab the movement through essential tabs should take place otherwise only the workspace tabs should be cycled through.

### Reproducible?

- [x] I have checked that this issue cannot be reproduced on Mozilla Firefox.

### Version

Mozilla Zen 1.7.4b

### What platform are you seeing the problem on?

Linux (Tarball)

### Relevant log output if applicable

```shell

```

lirena00:
yea I also feel this is really needed

TreyWB:
Agreed, at the very least unloaded essential tabs should be ignored when cycling with Ctrl+Tab.

mr-cheffy:
## Closing this issue as part of a repo cleanup

Due to technical-debt we had in the past, we are currently clearing out old, duplicate, stale, and non-actionable issues to improve issue tracking and make it easier to focus on fixing real bugs.

If this issue is still relevant, please open a new issue using our updated issue template, which will help us categorize and address it more efficiently.

Thanks for your understanding and for helping us improve the project! 🚀

This issue can be reopened at: https://github.com/zen-browser/desktop/issues/new?template=bug_report.yml&title=Ctrl-tab%20should%20not%20cycle%20through%20essential%20tabs%0A

PR #11242, opened 2025-11-10
feat: Ctrl+Tab cycling to keep within essentials/regular tabs only

junnjiee:
Note: This PR is a continuation from #11199

## Overview
This PR implements Ctrl+Tab cycling within Essential or Workspace tabs, depending on which tab is currently shown.

preference variable is named `cycleByAttribute` (in `browser.yaml`)

If both `sortByRecentlyUsed` and `cycleByAttribute` are both toggled on, using Ctrl+Tab will default to cycling by recently used. But, using Ctrl+PgUp/PgDn will still activate this feature, as it accesses the `advanceSelectedTab` function directly.

### `tabbox.js` (patch file)
- uses `"zen-essential"` attribute to filter tabs.

### `main.inc.xhtml` (patch file)
- checkbox to toggle the feature under Settings > General > Tabs
- warning message below the checkbox element, hidden by default

### `main.js` (patch file)
- `toggleCtrlTabFeatureWarning()` -> shows the warning message when both `sortByRecentlySelected` and `cycleByAttribute` are both toggled on
image

## Testing
- `browser_tabs_cycle_by_attribute.js` -> tests that tab cycles within essential/regular tabs when `cycleByAttribute` is true. This is achieved by checking if the tab cycles back to itself (example: 3 tabs, advance 3 times, so starting tab should cycle back to itself)
---
Issues/Discussions requesting for the feature: #4787, #6316, #9119.

### DIFF
--- a/prefs/firefox/browser.yaml
+++ b/prefs/firefox/browser.yaml
@@ -83,6 +83,3 @@
 
 - name: browser.tabs.closeWindowWithLastTab
   value: false
-
-- name: browser.ctrlTab.cycleByAttribute
-  value: false
--- a/prefs/firefox/browser.yaml
+++ b/prefs/firefox/browser.yaml
@@ -83,3 +83,6 @@
 
 - name: browser.tabs.closeWindowWithLastTab
   value: false
+
+- name: browser.ctrlTab.cycleByAttribute
+  value: false
--- a/src/browser/components/preferences/main-inc-xhtml.patch
+++ b/src/browser/components/preferences/main-inc-xhtml.patch
@@ -1,8 +1,21 @@
 diff --git a/browser/components/preferences/main.inc.xhtml b/browser/components/preferences/main.inc.xhtml
-index ff577c035ff6128f7ae948bc810082c92b3a624c..40f94fff53c63bdff28eaa68f9b0736d53c5b4f8 100644
+index ff577c035ff6128f7ae948bc810082c92b3a624c..eda462e81c448ddd3d63726d94caa6702500ab73 100644
 --- a/browser/components/preferences/main.inc.xhtml
 +++ b/browser/components/preferences/main.inc.xhtml
-@@ -141,7 +141,8 @@
+@@ -74,6 +74,12 @@
+     <checkbox id="ctrlTabRecentlyUsedOrder" data-l10n-id="ctrl-tab-recently-used-order"
+               preference="browser.ctrlTab.sortByRecentlyUsed"/>
+ 
++    <checkbox id="ctrlTabCycleByAttribute" data-l10n-id="ctrl-tab-cycle-by-attribute"
++              preference="browser.ctrlTab.cycleByAttribute"/>
++
++    <description id="ctrlTabCycleByAttributeWarning" class="indent description-deemphasized"
++              data-l10n-id="ctrl-tab-cycle-by-attribute-warning" hidden="true" />
++
+     <checkbox id="linkTargeting" data-l10n-id="open-new-link-as-tabs"
+               preference="browser.link.open_newwindow"/>
+ 
+@@ -141,7 +147,8 @@
  <!-- Browser layout -->
  <groupbox data-category="paneGeneral"
            data-subcategory="layout"
@@ -12,7 +25,7 @@ index ff577c035ff6128f7ae948bc810082c92b3a624c..40f94fff53c63bdff28eaa68f9b0736d
    <label><html:h2 data-l10n-id="browser-layout-header"/></label>
    <radiogroup id="browserLayoutRadioGroup" preference="sidebar.verticalTabs">
      <radio id="browserLayoutHorizontalTabs" data-l10n-id="browser-layout-horizontal-tabs" value="false"/>
-@@ -346,6 +347,7 @@
+@@ -346,6 +353,7 @@
                languages-customize-add.label,
              " />
    </hbox>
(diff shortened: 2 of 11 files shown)


junnjiee:
@mr-cheffy Thanks for reviewing my code in #11199! This PR includes your requested changes
- localization is moved to `zen-preferences.ftl`
- used lazy pref getter in `tabbox.js` using `XPCOMUtils.defineLazyPreferenceGetter()`

src/browser/components/preferences/main-inc-xhtml.patch
```diff
@@ -1,8 +1,21 @@
...
-index ff577c035ff6128f7ae948bc810082c92b3a624c..40f94fff53c63bdff28eaa68f9b0736d53c5b4f8 100644
+index ff577c035ff6128f7ae948bc810082c92b3a624c..eda462e81c448ddd3d63726d94caa6702500ab73 100644
 --- a/browser/components/preferences/main.inc.xhtml
 +++ b/browser/components/preferences/main.inc.xhtml
-@@ -141,7 +141,8 @@
+@@ -74,6 +74,12 @@
+     <checkbox id="ctrlTabRecentlyUsedOrder" data-l10n-id="ctrl-tab-recently-used-order"
+               preference="browser.ctrlTab.sortByRecentlyUsed"/>
+ 
++    <checkbox id="ctrlTabCycleByAttribute" data-l10n-id="ctrl-tab-cycle-by-attribute"
```
mr-cheffy: I'd move this into zen's "tab management" settings section, and avoid making patches

src/browser/components/preferences/main-js.patch
```diff
@@ -1,8 +1,16 @@
...
-index 46283d3eacfea0f419afa3bdfd2a51e411502a20..4fe229706eb0145a21042b8b3cca323682e89e76 100644
+index 46283d3eacfea0f419afa3bdfd2a51e411502a20..730058671fb0cbab0b5b5b1c01cdb912560116a2 100644
 --- a/browser/components/preferences/main.js
 +++ b/browser/components/preferences/main.js
-@@ -1117,6 +1117,11 @@ let SETTINGS_CONFIG = {
+@@ -102,6 +102,7 @@ Preferences.addAll([
+   { id: "browser.warnOnQuitShortcut", type: "bool" },
+   { id: "browser.tabs.warnOnOpen", type: "bool" },
+   { id: "browser.ctrlTab.sortByRecentlyUsed", type: "bool" },
++  { id: "browser.ctrlTab.cycleByAttribute", type: "bool" },
```
mr-cheffy: This can be done in zen-settings.js:

https://github.com/zen-browser/desktop/blob/5bd0a01d5369d3bc5cc3f2481319b997c0ca703e/src/browser/components/preferences/zen-settings.js#L1057-L1138

src/browser/components/preferences/main-js.patch
```diff
@@ -23,3 +31,40 @@ index 46283d3eacfea0f419afa3bdfd2a51e411502a20..4fe229706eb0145a21042b8b3cca3236
      false,
      undefined,
      locales
+@@ -1746,6 +1752,17 @@ var gMainPane = {
+     setEventListener("ctrlTabRecentlyUsedOrder", "command", function () {
+       Services.prefs.clearUserPref("browser.ctrlTab.migrated");
+     });
++
++    Preferences.get("browser.ctrlTab.sortByRecentlyUsed").on(
```
mr-cheffy: Same as the comment above

src/toolkit/content/widgets/tabbox-js.patch
```diff
@@ -50,3 +68,35 @@ index cfe2da6e199667bd668f117cc8972212c7f57da2..bdd7dcc26139202e6e31afde47dc4d87
...
++
+       // Handle keyboard navigation for a hidden tab that can be selected, like the Firefox View tab,
+       // which has a random placement in this.allTabs.
+       if (startTab.hidden) {
+         if (aDir == 1) {
+-          newTab = this.allTabs.find(tab => tab.visible);
++          newTab = this.allTabs.find(tabFilter);
+         } else {
+-          newTab = this.allTabs.findLast(tab => tab.visible);
++          newTab = this.allTabs.findLast(tabFilter);
```
mr-cheffy: We shouldn't really apply this filter here, when the tab is hidden we want to get out of it, even if its outside of our "rules"

prefs/firefox/browser.yaml
```diff
@@ -83,3 +83,6 @@
 
 - name: browser.tabs.closeWindowWithLastTab
   value: false
+
+- name: browser.ctrlTab.cycleByAttribute
```
mr-cheffy: Please name this zen.tabs.* and move it to the correct zen's config file where the other zen.tabs.* prefs are

mr-cheffy (changes requested):
Looks good! Some small changes

junnjiee:
Included your requested changes with the most recent commit! @mr-cheffy 

## Key Changes
- reverted patches for `main.inc.xhtml` and `main.js`, so they shouldn't show up in files changed anymore
- UI related logic is now implemented in `zenTabsManagement.inc.html` and `zen-settings.js`
- when tab is hidden, the tab filter used is `tab => tab.visible`

### Updated UI
Toggle now resides in Tab Management > Workspaces
image

mr-cheffy (approved):
LGTM, one thing though: Would it be possible for users to find it annoying if all of they sudden they can't ctrl+Tab outside of the essentials area? e.g. if they enabled it without thinking.

Shouldn't we move the user out of the essentials area, **always**? But that's for a different PR, I guess

junnjiee:
Good point! Do you think this behavior should an option, or a default, unchangeable one?

I personally would like cycling within essentials since the commands Alt+1 to 8 only covers up to 8 essential tabs.

PR #11258, opened 2025-11-11
Feature: previous/next tab shortcuts

joao-paulo-santos:
Added shortcut options for Previous / Next tab navigation and Move tab Forward / backwards.

removing the default Firefox ctrl+tab tab cycling however would require a patch and more maintenance, so i just added the option for "extra" shortcuts.


mr-cheffy:
I'll close this in favor that Firefox does have plans to add support for customizing this shortcut (https://bugzilla.mozilla.org/show_bug.cgi?id=2012210) and I think i'll just end up using firefox's system once it gets a bit more stable. Thanks a lot for the patches and your time, and im sorry it didn't end up landing, but its just less maintenance work for me.

wave84:
This has been closed, but the feature is nowhere to be seen in Firefox or Zen. It's 2 years now since these keyboard shortcuts were removed. They are highly needed, especially on a Mac where the default shortcuts for tab management are almost useless (cmd+option+arrows, ctrl+tab).

PR #11267, opened 2025-11-12
fix: Fixed tab ownership being removed when opening new tabs

mr-cheffy:


KyleKincer:
Thanks for this one @mr-cheffy. Looks pretty good in Twilight. Issue 2 from [here](https://github.com/zen-browser/desktop/pull/11240#issuecomment-3516984044) persists though. 
- From one parent tab, open more than one background tab
- None of the children will have an owner

KyleKincer:
proposed fix on https://github.com/zen-browser/desktop/pull/11274

Issue #11220, opened 2025-11-09
Browser switches to another workspace upon clicking a pinned tab


haikeyidesu:
### Preliminary Checks

- [x] I have read and understood the important section above.
- [x] I have searched existing issues and avoided creating duplicates.
- [x] I am not filing an enhancement request.
- [x] I have checked that this issue cannot be reproduced on Mozilla Firefox.
- [x] I have checked that this issue can be reproduced once I removed all my Mods and Custom CSS.

### What happened?

Computer: M3 Macbook Pro
Current OS:  MacOS Sequoia 15.6.1

My Zen browser has 7 workspaces. 

Issue: 
When I click on the bottom-most pinned tab in my 5th workspace, the current workspace would change to the 2nd workspace. This has happened before in previous settings, and as of the latest testing, appears to only affect the 5th workspace only. 
I have encountered slightly varied occurances of this bug, though the current method triggers the bug the most consistently for me now

fyi: I'm not sure if this issue could be a duplicate, I have not yet been able to search for an existing similar issue if there was one. 🙏

hope this helps! 👍

### Expected behavior

clicking and going to the pinned tab

### Actual behavior

clicking the specific bottom-most pinned tab in workspace 5, as mentioned, would change the current workspace to workspace 2. 

### Steps to reproduce

Click on the bottom-most pinned tab of workspace 5 of 7. (middle workspace + 1? just a guess)

### Screenshots and videos

https://github.com/user-attachments/assets/541fa73c-5c97-458d-9da6-1c395b0549fe

### Version

1.17.6b

### What platform are you seeing the problem on?

macOS - aarch64

### What component is this issue related to?

Workspaces

### Relevant log output if applicable

```shell

```

junnjiee:
By any chance, is your pinned tab a duplicated tab from your essentials? 

This may be a duplicate of #10457

mr-cheffy:
I'll mark this as a duplicate and done, please do ping me to reopen if it still happens after the next update!

haikeyidesu:
Oh yes it's a duplicate of #10457, though I didn't realize it before. 
I'm now using the 1.17.6b (Firefox 144.0.2) (aarch64), and the issue still pops up a bit. 
I'll update Zen and test if the issue still persists in the latest version!
Thanks! 👍

haikeyidesu:
Alright I just update Zen, the issue still persists. It doesn't switch to the last workspace where I opened the essential tab, though...
[...]

PR #11293, opened 2025-11-15
fix: duplicated essential tabs should set zenWorkspace as current active workspace

### LINKED ISSUE #10457: Clicking on a duplicated tab of an essential tab causes switching to the last workspace
### Preliminary Checks

### What happened?

When I duplicate an essential tab, then click on another random tab, and then click on the new duplicated tab again, the browser automatically switches to the last workspace. This happens regardless of which workspace I am currently in, except for the last one itself of course.
The issue does not occur when I duplicate a regular pinned tab. It appears to be a bug specifically with duplicating essentials.

### Expected behavior

Clicking on the duplicated tab should simply open that tab within the current workspace.

### Actual behavior

Clicking on the duplicated tab causes the browser to switch to the last workspace.

### Steps to reproduce

1. Open Zen browser with at least two different workspaces.
2. In any workspace except for the last one, right click on an essential tab and select "Duplicate Tab"
3. The browser will automatically open the newly created tab. So click on a different tab to switch away from the duplicate.
4. Now click on that duplicated tab again. The browser will switch to the last workspace, instead of simply opening that tab by staying in the same workspace.

### Screenshots and videos

_No response_

### Version

1.16t

Windows - x64

Tabs

```shell

```

junnjiee:
Fixes #10457 
Might also be the bug causing #11220 

## Problem
When duplicating an essential tab, `tabState.zenWorkspace` is still set to the tab's original workspace, before getting pinned to essentials. 

Thus selecting the duplicated tab causes the workspace to switch if `tabState.zenWorkspace` still exists

## Fix
Duplicated tab should always update `tabState.zenWorkspace` to current active workspace

Do you need me to write a test to ensure that this doesn't happen again?

### DIFF
--- a/src/browser/components/sessionstore/SessionStore-sys-mjs.patch
+++ b/src/browser/components/sessionstore/SessionStore-sys-mjs.patch
@@ -1,5 +1,5 @@
 diff --git a/browser/components/sessionstore/SessionStore.sys.mjs b/browser/components/sessionstore/SessionStore.sys.mjs
-index 1cdbc0f41bf5b55dfbbd850cb618c6d870f7a261..f13ed1412bb7ae6623aa2605d7691d10540eb659 100644
+index 1cdbc0f41bf5b55dfbbd850cb618c6d870f7a261..4eac2fed26d779908107ef60f1c2bd0eeba2381c 100644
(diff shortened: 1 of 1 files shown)


mr-cheffy (approved):
Nice catch!

PR #11335, opened 2025-11-18
feat: Enable original extensions panel on hide-unified-extensions-button: false

blakegearin:
## Summary

- Updated `ZenSiteDataPanel.sys.mjs` to support multiple panels and pass relevant objects to `browser-addons.js`

- Patched `browser-addons.js` to accept panel objects

- Patched `navigator-toolbox.js` to get panel objects from `window.gZenSiteDataPanel`

- Updated `zen-single-components.css` to conditionally hide extensions on unified panel

## Screenshots

- `hide-unified-extensions-button: true`

	image

- `hide-unified-extensions-button: false`

	image

	image

## Notes

- Toggling `hide-unified-extensions-button` does require a browser restart to fully take effect and switch over. This is because the list of extensions are added on browser initialization. It does not appear `browser/components/customizableui/CustomizableUI.sys.mjs` is able to natively support adding and removing extensions from multiple areas


src/zen/common/styles/zen-single-components.css
```diff
@@ -497,17 +497,23 @@ body > #confetti {
...
+    display: none;
+  }
 }
 
 #zen-site-data-header {
   gap: 8px;
   align-items: center;
   padding: 10px 9px;
-  padding-bottom: 0;
+  padding-bottom: 8px;
```
mr-cheffy, line 513: This should only be set if the preference is enabled?

src/zen/common/styles/zen-single-components.css
```diff
@@ -497,17 +497,23 @@ body > #confetti {
...
+
+  #unified-extensions-button:not([showing]) {
+    display: none !important;
+  }
 }
 
-#unified-extensions-button:not([showing]) {
-  display: none !important;
+@media not -moz-pref('zen.theme.hide-unified-extensions-button') {
+  .zen-site-data-section:has(#unified-extensions-manage-extensions) {
```
mr-cheffy: Don't use :has, better set an attribute to the section that contains the addons list

mr-cheffy (changes requested):
Instead of choosing what panel should be used every time, we could just revert these changes if the preference is off:

https://github.com/zen-browser/desktop/blob/08996215b2f55bc15a26b84e2d5585716aa3f863/src/browser/base/content/browser-addons-js.patch#L44-L52

https://github.com/zen-browser/desktop/blob/08996215b2f55bc15a26b84e2d5585716aa3f863/src/browser/base/content/browser-addons-js.patch#L27

Or at least simply override it? `gUnifiedExtensions._panel = ...`

src/zen/common/styles/zen-single-components.css
blakegearin: I'll defer to you on that.

For context, the reason I added it was for tabs like `about:config` where only the toolbar buttons are shown now when `hide-unified-extensions-button: false`

Example

image

---

It does make the spacing between sections a little more consistent. But I can add the gating if you'd like.

- Before

	image

- After

	image

src/zen/common/styles/zen-single-components.css
blakegearin: +1 Understood

Commit: 1caf0b413c5a13d355d5228836118390aa5f3229

blakegearin:
> Instead of choosing what panel should be used every time, we could just revert these changes if the preference is off:
> ...
> Or at least simply override it? gUnifiedExtensions._panel = ...

Yes, that's true. Effectively that's what `setMainPanel` was for, but that approach is cleaner. And we can safely remove the patches on `get panel()` which is a nice side effect

src/browser/base/content/browser-addons-js.patch
```diff
@@ -36,58 +36,112 @@ index f33a300eb0eed74e286da24919ad3e8505f632c3..9f400e887424693d0ecceb9db79fccf2
...
-       );
-@@ -2703,6 +2699,7 @@ var gUnifiedExtensions = {
-         // and no alternative content is available for display in the panel.
-         const policies = this.getActivePolicies();
+@@ -2391,7 +2391,7 @@ var gUnifiedExtensions = {
+         break;
+
+       case "toolbarvisibilitychange":
+-        this.onToolbarVisibilityChange(event.target.id, event.detail.visible);
++        this.onToolbarVisibilityChange(event.target.id, event.detail.visible, panel);
```
mr-cheffy: You are still passing `panel` as an argument though?

mr-cheffy (changes requested):

src/zen/urlbar/ZenSiteDataPanel.sys.mjs
```diff
@@ -21,7 +21,16 @@ export class nsZenSiteDataPanel {
     this.window = window;
     this.document = window.document;
 
-    this.panel = this.document.getElementById('zen-unified-site-data-panel');
+    this.unifiedPanel = this.#initUnifiedPanel();
+    this.unifiedPanelView = 'unified-extensions-view';
+    this.extensionsPanelView = 'original-unified-extensions-view';
+
+    if (!Services.prefs.getBoolPref('zen.theme.hide-unified-extensions-button', true)) {
+      this.window.gUnifiedExtensions._panel = this.extensionsPanel = this.#initExtensionsPanel();
```
mr-cheffy: We technically don't need to run this as the default panel for the gUnifiedExtensions is already the extensions panel

mr-cheffy (changes requested):

src/zen/urlbar/ZenSiteDataPanel.sys.mjs
mr-cheffy: @blakegearin I wonder, if you simply return past this point, how does firefox react? Does it start using the old popup?

src/browser/base/content/browser-addons-js.patch
blakegearin: I set a default so this wouldn't necessary here

Commit: [`d715921`](https://github.com/zen-browser/desktop/pull/11335/commits/d715921d2ec0932b886253f8ce724da2464af988)

src/zen/urlbar/ZenSiteDataPanel.sys.mjs
blakegearin: ~~`this.window.gUnifiedExtensions._panel =` prevents the core logic of `get panel()` from getting executed, which would create a second extensions panel outside of Zen's direct purview.~~

I was able to rework this to leverage Firefox's `get panel()` which makes this simpler and more resilient. 
Commit: [`b4c1d52`](https://github.com/zen-browser/desktop/pull/11335/commits/b4c1d52d8d500bb4afeb8fa591f32c6e774f480f)

A return here (at the end of this `if` block) prevents the unified panel from being initialized. The extensions button also doesn't work correctly because `this.extensionsPanelButton` doesn't get set, though this is easily movable. I assume this flag should **not** remove the unified panel altogether since it contains other important controls and info.

src/browser/base/content/browser-addons-js.patch
```diff
@@ -36,58 +36,103 @@ index f33a300eb0eed74e286da24919ad3e8505f632c3..9f400e887424693d0ecceb9db79fccf2
...
            triggerEvent: aEvent,
          });
        }
-@@ -2930,18 +2927,20 @@ var gUnifiedExtensions = {
+@@ -2830,7 +2828,7 @@ var gUnifiedExtensions = {
+       if (
+         notInPanel ||
+         document.querySelector("#unified-extensions-area > :first-child")
+-          ?.id === widgetId
++          ?.id === widgetIdz
```
mr-cheffy: Where's `widgetIdz` defined?

src/zen/urlbar/ZenSiteDataPanel.sys.mjs
```diff
@@ -34,22 +43,21 @@ export class nsZenSiteDataPanel {
...
+
+    this.unifiedPanelButton = this.window.gUnifiedExtensions._button = this.anchor;
+    this.extensionsPanelButton = this.document.getElementById('unified-extensions-button');
 
     this.document
       .getElementById('nav-bar')
       .setAttribute('addon-webext-overflowbutton', 'zen-site-data-icon-button');
 
-    // Remove the old permissions dialog
-    this.document.getElementById('unified-extensions-panel-template').remove();
```
mr-cheffy, line 44: This should be done if the preference is enabled, as the IDs inside the dialog conflicts. Dont worry about listening to dynamic changes, as the pref can be simply on startup

src/zen/urlbar/ZenSiteDataPanel.sys.mjs
```diff
@@ -124,6 +132,45 @@ export class nsZenSiteDataPanel {
...
+
+    return panel;
+  }
+
+  #initUnifiedPanel() {
+    const panel = this.document.getElementById('zen-unified-site-data-panel');
+    const customizationArea = panel?.querySelector('#unified-extensions-area');
+
+    // Partial implementation of "get panel()" from engine/browser/base/content/browser-addons.js
+    const { BrowserAddonUI, CustomizableUI } = this.window;
```
mr-cheffy: As I said before, this shouldn't need to be re-implemented. We can just keep using the original functionality at `get panel` and just use the preference at startup.

src/browser/base/content/browser-addons-js.patch
```diff
@@ -36,58 +36,103 @@ index f33a300eb0eed74e286da24919ad3e8505f632c3..9f400e887424693d0ecceb9db79fccf2
...
+
      return anchorID;
-@@ -2646,11 +2646,7 @@ var gUnifiedExtensions = {
-     // Lazy load the unified-extensions-panel panel the first time we need to
-     // display it.
-     if (!this._panel) {
--      let template = document.getElementById(
--        "unified-extensions-panel-template"
--      );
--      template.replaceWith(template.content);
```
mr-cheffy, line 48: Instead of removing, you should just see if the preference is enabled and then do what the patch is originally doing or firefox's addon functionality.

https://github.com/zen-browser/desktop/pull/11335/files#r2545668861

mr-cheffy (changes requested):

src/browser/base/content/browser-addons-js.patch
blakegearin: Whoops, just an extra character that slipped in. Good catch.

Commit: [`31f4d4e`](https://github.com/zen-browser/desktop/pull/11335/commits/31f4d4ebf9ffe1ca58ad1471612b527e4c01546f)

src/zen/urlbar/ZenSiteDataPanel.sys.mjs
blakegearin: +1 Yep, makes sense. 

Commit: [`ab4c812`](https://github.com/zen-browser/desktop/pull/11335/commits/ab4c812dcdbfaca303f592bae1216dbd713db3fc)

src/zen/urlbar/ZenSiteDataPanel.sys.mjs
blakegearin: I'm confused. I think we have some wires crossed here.

For reference, here's `get panel()` in Firefox: [source](https://github.com/mozilla-firefox/firefox/blob/FIREFOX_145_0_1_RELEASE/browser/base/content/browser-addons.js#L2645-L2680)

When `hide-unified-extensions-button: false` there are two panels. Both need to be run through `CustomizableUI.registerPanelNode` and the code that follows after. So I don't believe re-implementation is avoidable.

Previously the Zen panel was built by patching `get panel()`. I moved that code into the function referenced here, `#initUnifiedPanel()`. I didn't think it made sense to continue building Zen's panel in a Firefox file (via patch) and build the Firefox's panel in a Zen file.

If we leave `get panel()` alone it will only build the Firefox extensions panel when told...

```js
this.extensionsPanel = this.#initExtensionsPanel();
...
#initExtensionsPanel() {
	const panel = this.window.gUnifiedExtensions.panel;
  	...
}
```

src/zen/urlbar/ZenSiteDataPanel.sys.mjs
```diff
@@ -124,6 +135,45 @@ export class nsZenSiteDataPanel {
...
+
+    CustomizableUI.registerPanelNode(customizationArea, CustomizableUI.AREA_ADDONS);
```
mr-cheffy: Not really sure about this, as doing re-implementations might be hard to sync when updating upstream. Maybe some unit tests would help this issue?

mr-cheffy (approved):
I'll give it a test try later today, looks good overall

src/zen/urlbar/ZenSiteDataPanel.sys.mjs
blakegearin: Where do unit tests live? I could give it a shot

mr-cheffy: You can give it a try here: https://github.com/zen-browser/desktop/tree/dev/src/zen/tests

Making a new directory like "control-panel" for example

blakegearin: After thinking about this some more I was able to remove this and have the patch handle both panels cleanly

Commit: [`596ca01`](https://github.com/zen-browser/desktop/pull/11335/commits/596ca01f23a72efbf02a5fe5a015631fb0b04ee7)

mr-cheffy:
Let me know once its finished for review @blakegearin 

Please run `npm run lint:fix` as well and patches seems to fail importing? https://github.com/zen-browser/desktop/actions/runs/19818358014/job/56774645713?pr=11335

blakegearin:
@mr-cheffy Thanks. Should be good to go.

mr-cheffy:
Sorry for the delay, i'll review ASAP when im free

blakegearin:
No worries. Got the conflict resolved.

mr-cheffy:
https://github.com/user-attachments/assets/6f3bf13c-5c36-4d4f-a8a6-da73c49001ff

@blakegearin everything seems good, theres this issue that when opening, it opens it ontop of the button though

src/browser/base/content/browser-addons-js.patch
```diff
@@ -36,58 +36,152 @@ index fadcbfca95ee28140579430c0371baad0e2f216a..7454b801b4ad892d6ad122277eb7c773
...
+     );
+
+@@ -2662,37 +2662,41 @@ var gUnifiedExtensions = {
        );
-@@ -2714,6 +2710,7 @@ var gUnifiedExtensions = {
-         // and no alternative content is available for display in the panel.
-         const policies = this.getActivePolicies();
+       template.replaceWith(template.content);
+       this._panel = document.getElementById("unified-extensions-panel");
+-      let customizationArea = this._panel.querySelector(
```
mr-cheffy, line 63: Might be worth to send this patch to firefox, so we don't have to mantain it

blakegearin: Sure, I will try and submit :)

blakegearin:
> @blakegearin everything seems good, theres this issue that when opening, it opens it ontop of the button though

Good catch. Thankfully it was a pretty easy fix.

src/browser/base/content/browser-addons-js.patch
mr-cheffy: Cool! if you want, mention is for zen to update to upstream more easily, so they at least know why the change

mr-cheffy (approved):
Looks good! Should drop to twilight tomorrow and then to stable next update! Would be nice if tests could be written so we make sure this doesn't break across upstram updates

src/browser/base/content/browser-addons-js.patch
blakegearin: For visibility: https://phabricator.services.mozilla.com/D276429

PR #11343, opened 2025-11-18
Hold click to open glance

devdinc:
Added the ability to hold your mouse click button to open glance.

Don't know whether this is finished or not, so will do a draft PR.
activation ms threshold can be a preference, etc.

Also since we detect movement, it should fix the issue related to Alt + Hold Click to select links.


src/zen/glance/actors/ZenGlanceChild.sys.mjs
```diff
@@ -15,12 +20,17 @@ export class ZenGlanceChild extends JSWindowActorChild {
     }
   }
 
-  async #initActivationMethod() {
-    this.#activationMethod = await this.sendQuery('ZenGlance:GetActivationMethod');
+  #ensureOnlyOneModifierKeyIsPressed(event) {
+    return (
+      [event.ctrlKey, event.altKey, event.shiftKey, event.metaKey].filter(Boolean).length === 1
+    );
```
garrza, line 26: I think we can make this `ensureOnlyOneModifierKeyIsPressed` faster and lighter by avoiding the array allocation and filter.

Since JavaScript coerces booleans to numbers during addition, we can just sum them up such as:

`#ensureOnlyOneModifierKeyIsPressed(event) {
  return (event.ctrlKey + event.altKey + event.shiftKey + event.metaKey) === 1;
}`

This would be more lightweight for a high-frequency usage event handler. Nothing goes to the heap and nothing would need to be garbage collected.

mr-cheffy:
Could you please show me a video on how it should look?

devdinc:
> Could you please show me a video on how it should look?

There is no visual changes or an indicator. When mouse up is done after 300ms it opens glance. If you are wondering whether the code works, I have not really tested the code except some POC version that I posted into web console that logs the duration and the element.

We can consider adding visual indicator, some circular slider like thing maybe?  To cancel the visual indicator, we can create a message endpoint in parent. Adding an attribute on the top element is risky as before and after might be already used, so we need to create an element probably like a popup that has the same proportions and edit that using css.

mr-cheffy:
Sorry, but this doesn't work for me. Please send the PR once tested properly, otherwise it'll just it'll be a very painful process

devdinc:
> Sorry, but this doesn't work for me. Please send the PR once tested properly, otherwise it'll just it'll be a very painful process

That is sad to hear, as I can't really build the zen browser, since I'm having an issue with git and storage I can't really solve now, well POC script I used, which still works is here, paste it into website console as browser internal console doesn't pick up from websites:

```js
(() => {
  // Unload previous instance if it exists
  if (window.__zenGlanceDebugUnload) {
    console.log("%c[ZenGlance Debug] Unloading previous instance...", "color: orange");
    window.__zenGlanceDebugUnload();
  }

  console.log("%c[ZenGlance Debug] Loading new instance...", "color: cyan");

  const config = {
    activationMethod: "hold", // "hold", "ctrl", "alt", "shift", "meta"
    holdDuration: 500,
    moveThreshold: 8,
  };

  let downTime = 0;
  let downX = 0;
  let downY = 0;

  const handlers = {
    mousedown(e) {
      const target = e.target.closest("a");
      console.log("[ZenGlance Debug] mousedown", {
        target,
        x: e.clientX,
        y: e.clientY,
        button: e.button,
      });

      if (e.button !== 0) return;

      downTime = performance.now();
      downX = e.clientX;
      downY = e.clientY;
    },

    mouseup(e) {
      if (e.button !== 0) return;

      const target = e.target.closest("a");
      if (!target) return;

      const movedTooMuch =
        Math.abs(e.clientX - downX) > config.moveThreshold ||
        Math.abs(e.clientY - downY) > config.moveThreshold;

      const elapsed = performance.now() - downTime;

      console.log("[ZenGlance Debug] mouseup", {
        target,
        movedTooMuch,
        elapsed,
        modifiers: {
          ctrl: e.ctrlKey,
          alt: e.altKey,
          shift: e.shiftKey,
          meta: e.metaKey,
          onlyOne:
            [e.ctrlKey, e.altKey, e.shiftKey, e.metaKey].filter(Boolean).length === 1,
        },
      });

      if (movedTooMuch) return;

      const onlyOne =
        [e.ctrlKey, e.altKey, e.shiftKey, e.metaKey].filter(Boolean).length === 1;

      let shouldActivate = false;

      if (config.activationMethod === "hold") {
        shouldActivate = elapsed >= config.holdDuration;
      } else if (onlyOne) {
        shouldActivate =
          (config.activationMethod === "ctrl" && e.ctrlKey) ||
          (config.activationMethod === "alt" && e.altKey) ||
          (config.activationMethod === "shift" && e.shiftKey) ||
          (config.activationMethod === "meta" && e.metaKey);
      }

      if (shouldActivate) {
        e.preventDefault();
        console.log(
          "%c[ZenGlance Debug] Would open glance for: " + target.href,
          "color: lime"
        );
      }
    },

    keydown(e) {
      if (e.key === "Escape") {
        console.log("%c[ZenGlance Debug] Escape pressed — would close glance", "color: orange");
      }
    },
  };

  // Attach listeners
  document.addEventListener("mousedown", handlers.mousedown, true);
  document.addEventListener("mouseup", handlers.mouseup, true);
  document.addEventListener("keydown", handlers.keydown, true);

  // Provide unload function
  window.__zenGlanceDebugUnload = () => {
    document.removeEventListener("mousedown", handlers.mousedown, true);
    document.removeEventListener("mouseup", handlers.mouseup, true);
    document.removeEventListener("keydown", handlers.keydown, true);

    delete window.__zenGlanceDebugUnload;
    console.log("%c[ZenGlance Debug] Unloaded!", "color: red");
  };
})();
```

If another developer can continue and assess whats wrong with it, I would be happy.

PR #11345, opened 2025-11-18
fix: improve drag-and-drop behavior for tab folders (#11326)

### LINKED ISSUE #11326: Unintuitive drag-and-drop behavior when moving tabs into folders
### Preliminary Checks

### What happened?

I initially believed it was impossible to move a tab into a folder via drag-and-drop due to high friction in the interaction. After testing, I realized it is possible but requires very specific cursor placement that is unintuitive for the user.

### Expected behavior

Dragging a tab over a folder should feel stable and forgiving. The entire folder component should act as the drop target (drop zone).

### Actual behavior

When attempting to drag a tab into a collapsed folder, two main issues occur.  There is a layout shift on hover, when the tab is dragged over the folder, the folder briefly expands (or triggers an animation state). However, as soon as the cursor enters the specific visual area of the folder, the element shifts position. This causes the cursor to lose focus on the target, effectively "dodging" the drop action. There is also, a very small hit box, the area immediately above the folder is not registered as a "safe" drop zone. Attempting to drag a tab into a folder from the top results in a failed drop or triggers the layout shift mentioned above. The current active drop zone seems limited to a small area below or strictly adjacent to the folder, rather than the folder element itself.

### Steps to reproduce

Try to drop a tab into a folder and suffer.

### Screenshots and videos

https://github.com/user-attachments/assets/18fd1dac-ece2-427d-8d42-3735b0639526

### Version

1.17.6b

macOS - aarch64

Tab Folders

```shell

```

garrza:
- Simplify drop zone logic with 25% edge zones for better UX
- Tabs can now be inserted into folder from either above or below, more forgiving drop as well
- Remove unnecessary upper/lower threshold preferences

https://github.com/user-attachments/assets/b935b485-d3d7-4ca6-ac5c-213500f03d43

Fixes #11326


src/zen/folders/ZenFolders.mjs
```diff
@@ -899,7 +899,11 @@
       const labelContainer = group.querySelector('.tab-group-label-container');
       // Setup mouseenter/mouseleave events for the folder
       labelContainer.addEventListener('mouseenter', (event) => {
-        if (!group.collapsed || !Services.prefs.getBoolPref('zen.folders.search.enabled')) {
+        if (
+          !group.collapsed ||
+          !Services.prefs.getBoolPref('zen.folders.search.enabled') ||
+          document.getElementById('tabbrowser-tabs')?.hasAttribute('movingtab')
```
mr-cheffy: Use `gBrowser.tabContainer` and it'll never be undefined

src/zen/folders/ZenFolders.mjs
```diff
@@ -1164,24 +1164,17 @@
...
       const isRestrictedGroup = isSplitGroup || dropElementGroup.collapsed;
 
-      const shouldDropInside =
-        !dropBefore &&
-        overlapPercent >= dragDownThreshold &&
-        overlapPercent <= dragUpThreshold &&
-        !isSplitGroup;
-      const shouldDropNear = overlapPercent < dragUpThreshold || overlapPercent > dragDownThreshold;
+      // Edge zones: 25% top/bottom for dropping before/after, 50% middle for dropping inside
+      const edgeZone = 0.25;
```
mr-cheffy: I'd make this a preference

src/zen/folders/zen-folders.css
```diff
@@ -116,6 +130,10 @@ zen-folder {
           }
         }
 
+        #tabbrowser-tabs[movingtab] & {
+          transition: transform 0.15s ease, opacity 0.15s ease !important;
```
mr-cheffy: Why did you add this?

src/zen/folders/zen-folders.css
```diff
@@ -45,6 +45,20 @@ zen-folder {
     }
   }
 
+  #tabbrowser-tabs[movingtab] & {
+    contain: layout style;
```
mr-cheffy: Why's this necessary?

mr-cheffy (changes requested):

mr-cheffy:
https://github.com/user-attachments/assets/b4941751-77d6-4fe4-83d6-9c1ff738d0e0

now the folder doesn't transition though?

src/zen/folders/zen-folders.css
garrza: not necessary, there is actually no need to change css, i'll revert these.

src/zen/folders/ZenFolders.mjs
garrza: yeah this sounds good, will make it so.

src/zen/folders/ZenFolders.mjs
garrza: nice, will do this.

garrza:
simplified a bit by removing non-needed css, also added edge zone threshold as a preference loaded at component level.

added defaults for both max subfolders and edge zone threshold to handle timing issues during module initialization. module-level constants load before the preference system is fully ready, so defaults prevent `NS_ERROR_UNEXPECTED` errors that would break the entire `ZenFolders` module. 

no default for max subfolders preference could have potentially randomly broken the module for new users which do not have preferences loaded due to race condition.

src/zen/folders/ZenFolders.mjs
```diff
@@ -29,7 +29,9 @@
     return `${month} month${month === 1 ? '' : 's'} ago`;
   }
 
-  const ZEN_MAX_SUBFOLDERS = Services.prefs.getIntPref('zen.folders.max-subfolders');
+  const ZEN_MAX_SUBFOLDERS = Services.prefs.getIntPref('zen.folders.max-subfolders', 5);
+  const ZEN_EDGE_ZONE_THRESHOLD =
+    Services.prefs.getIntPref('zen.view.drag-and-drop.edge-zone-threshold', 25) / 100;
```
mr-cheffy: move these constants inside the nsZenFolders, e.g. `#ZEN_MAX_SUBFOLDERS`, ...

mr-cheffy (changes requested):

src/zen/folders/ZenFolders.mjs
garrza: gotcha, lemme do that man

mr-cheffy:
Please, don't push with `--force` as its more difficult for me to update this branch

mr-cheffy (approved):
Looks good! Lets drop this on twilight and i'll start testing on how it generally feels and different configuration options!

PR #10034, opened 2025-08-27
Full cross-window workspace syncing

mr-cheffy:

PR #11363, opened 2025-11-19
Fix: Persist favicons for pinned tabs and migrate existing pins (#9389)

### LINKED ISSUE #9389: Pinned tabs and essential tabs icons disappears when the tab is closed
### Preliminary Checks

### What happened?

The pined tabs and the essential tabs icons disappears every time 

Image

 I close them. But if I unload them then the icons do not disappear. It creates a huge problem cause I cannot just close a pinned tab .I have to right click and unload them or all the icons jut disappear.

### Expected behavior

The icons should  not disappear and it should become a bit dimmer so it is easy to detect its closed

### Actual behavior

The icons disappears everythime I close them and it becomes icon less text

### Steps to reproduce

- Load a pinned or essential tab
- Close them using the mouse or the keyboard shortcut
- Icon of the website or webpage gone

### Screenshots and videos

Image

### Version

 1.14.3b

Windows - x64

Tabs

```shell

```

Vedaant-Rajoo:
Fixes #9389

Persists favicon data for pinned tabs by storing base64 data URIs in the database. This ensures favicons are preserved across browser restarts and workspace changes.

**Changes:**
- Add `icon_data` column to store favicon data URIs
- Capture and save favicons when tabs are pinned
- Update stored favicons when tab icons change
- Lazy migration for existing pins without favicons
- Enhanced favicon fetching with network fallback when Places cache is unavailable
- Restore favicons from storage when pins are loaded

**Migration:**
Existing pinned tabs without favicons are automatically migrated in the background after initialization, prioritizing essential pins.


Vedaant-Rajoo:
There are a lot of ways to reproduce the issue #9389. @mr-cheffy but I'm not too sure if the way i went about fixing it is the best solution. If you have any qualms about it. Let me know. I know a lot of people are waiting for this bug to get fixed.

mr-cheffy:
I dont think this is the best way to tackle the issue tbh, but anyways, we are removing this implementation with window sync. https://github.com/zen-browser/desktop/pull/10034

PR #11394, opened 2025-11-22
fix: allow selection within link

### LINKED ISSUE #8391: Holding ⌥ on Mac to select the text inside a link is broken
### Captchas

### What happened?

In Firefox it is possible to hold <kbd>⌥</kbd> (or <kbd>Alt</kbd> on Linux or Windows) to select the text inside a link instead of clicking on the link. But on Zen the link is unintentionally opened at the end of the clic.

### Version

1.12.5b

macOS - aarch64

Other

```shell

```

Spissable:
fixes: #8391

The glance feature was clashing with the possibility to select text within a link. To avoid the conflict, glance will only open upon mouseup and only if the mouse hasn't moved since the mouse was pressed. That's just like a normal link click works as well, so I think it makes sense for glance to behave like this too.

https://github.com/user-attachments/assets/c4c52bbe-4e41-41bb-a2c2-ced0c2488d5d

### DIFF
--- a/src/zen/glance/actors/ZenGlanceChild.sys.mjs
+++ b/src/zen/glance/actors/ZenGlanceChild.sys.mjs
@@ -3,6 +3,7 @@
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 export class ZenGlanceChild extends JSWindowActorChild {
   #activationMethod;
+  #glanceTarget = null;
 
   constructor() {
     super();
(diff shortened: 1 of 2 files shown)


src/zen/glance/actors/ZenGlanceChild.sys.mjs
```diff
@@ -81,10 +82,22 @@ export class ZenGlanceChild extends JSWindowActorChild {
       return;
     }
     if (target) {
```
mr-cheffy: We wouldn't technically need this check anymore

src/zen/glance/actors/ZenGlanceChild.sys.mjs
```diff
@@ -99,5 +112,8 @@ export class ZenGlanceChild extends JSWindowActorChild {
 
   async on_DOMContentLoaded() {
     await this.#initActivationMethod();
+    this.contentWindow.addEventListener('mousedown', this, true);
+    this.contentWindow.addEventListener('mouseup', this, true);
```
mr-cheffy: Don't add listeners here, rather on here:

https://github.com/zen-browser/desktop/blob/0e8b8da84561d9f7d7a42b335920dc6581a10090/src/zen/common/sys/ZenActorsManager.sys.mjs#L45-L50

mr-cheffy (changes requested):

Spissable:
Hey @mr-cheffy thanks for the feedback! I made the changes as requested

mr-cheffy (approved):

PR #10731, opened 2025-10-08
feat: Add config to support more than 12 essential tabs

blakegearin:
## Summary

- Added `zen.tabs.essentials.max` preference with a default of `12`

- Added `maxEssentialTabs` lazy preference getter

- Updated locale files to have `{ $max }` slot

## Notes

- Previous PRs closed due to merge conflicts

  - #8441 

  - #9189

- Zen supported more than 12 essential tabs in version `1.12.2b` and before leading to a number of threads

  - #8238

  - #8308 

  - #8315

  - #8341

  - #8689
  
  - #8900

PR #11400, opened 2025-11-23
feat: Add "Max amount of Essential Tabs" setting to the Preferences UI

hoXyy:
# Summary

- Adds a UI representation of the `zen.tabs.essentials.max` preference added in https://github.com/zen-browser/desktop/pull/10731:
image

# Notes
It was requested in this discussion thread: https://github.com/zen-browser/desktop/discussions/10379

I wasn't sure where exactly to put the setting, so I put it in "Tab Management -> Workspaces". If requested, I'l move it.


mr-cheffy:
Thanks for the contribution! However, this setting is intentionally not exposed in the Preferences UI. The limit for Essential Tabs is meant to remain an advanced configuration option, since it can affect stability and internal behavior that most users shouldn't modify directly. For that reason, we prefer to keep `zen.tabs.essentials.max` accessible only through advanced config rather than adding it to the standard Preferences.

PR #11478, opened 2025-11-30
Add language: Australian English

Will-Georges:
Added language en-AU (Australian English) and updated supported languages file.


mr-cheffy:
New languages should be suggested through crowdin

PR #11500, opened 2025-12-01
Update Chinese translations in zen-general.ftl

CodeFreer:


mr-cheffy:
Translations should be done via crowdin

PR #11502, opened 2025-12-01
Update Translate UI labels to Chinese

CodeFreer:


mr-cheffy:
Translations should be done via crowdin

PR #11539, opened 2025-12-05
Fix bookmarks toolbar flickering on toggle

Anodev-sudo:
Accidental submit. Ignore.


Anodev-sudo:
Sorry for this accidental submit. I have a better fix.

PR #11581, opened 2025-12-11
Enable importing bookmarks into zen folders

Abdul-fattah-Tayih:
This change adds the ability to import the Firefox bookmarks into a zen folder, which allows us to achieve a few things

1. Get up and running on a new device very quickly, avoids manually adding tabs 1 by 1
2. Achieve pseudo syncs, since bookmarks are synced with Firefox

I've tested this on my windows PC (via WSL2) and my macbook, the only issue is that on the mac, the labels are missing, I've tried adding the translations to the `.ftl` files, but that didn't fix it, so any help here would be great before we go ahead with

Zen bookmark import


KiKaraage:
How about making the existing "Open All Bookmarks" / "Open All in Tabs" opens as a tab folder (pinned) instead of directly as multiple regular tabs? I imagine this will be easier for existing bookmark users.

(optionally, both strings can be changed into "Open Links in Folder")

image

image

Vice versa with making bookmark folder from existing tab folder - a new "Bookmark Folder.." option can be added to mirror the existing "Bookmark Tabs.." option:

image

image

mr-cheffy:
I think it would be better to have an option to "Convert to folder" when right clicking a bookmark folder.

Abdul-fattah-Tayih:
@mr-cheffy agreed, will update the PR when I have some time

kitawa1981:
Looking forward to it, I don't know when it will be completed

Anatharias:
Absolutely amazing!
I tested Zen, and managed to export my bookmarks from Arc to Zen without any possible way to bulk add them to the sidebar of each space, and what you're proposing is exactly it!

Alas in Zen and I assume that this is a Firefox limitation, bookmarks ARE bookmarks and TABS are tabs.
Whereas in Arc, Bookmarks ARE tabs and tabs ARE bookmarks.

My main concern over Zen sidebar is that this is not handled as a bookmark cabinet but solely tabs. The sidebar should actually display user bookmarks and once clicked on, load a tab in its stead.
This is what I thought was, this is what I was expecting, seeing this help screenshot on the help pages:
bookmarks-sidebar c08a432f


Without complexifying the process too much, I assume that a live sync of tabs <-> bookmarks could achieve the same. Though I'm no coder, just providing feedback on a feature that would be amazing if implemented.

mr-cheffy:
Sorry, but I dont think this is how I would want it to be at the end and I think its better if I just close it instead of leaving it here consume more and more git conflicts. Thanks for trying to contribute and im so sorry I have to close this

Abdul-fattah-Tayih:
> Sorry, but I dont think this is how I would want it to be at the end and I think its better if I just close it instead of leaving it here consume more and more git conflicts. Thanks for trying to contribute and im so sorry I have to close this

No worries man, i was actually planning to close it in the coming few days, sadly i had no time whatsoever to look over this 🫠

Hopefully ill reopen one when get the time

mr-cheffy:
For future reference, I think maybe we should patch this instead?

CleanShot 2026-03-20 at 17 11 28@2x

PR #11607, opened 2025-12-13
feat(downloads): add pref to control download popup position independently

salmonumbrella:
## Summary

Adds `zen.downloads.icon-popup-position` preference that allows users to control the download popup/indicator position independently from the vertical tabs position.

**Valid values:**
- `"follow-tabs"` (default): popup appears on same side as vertical tabs (current behavior)
- `"left"`: popup always appears on the left
- `"right"`: popup always appears on the right

## Use Case

This is useful for users who have vertical tabs on the right (`zen.tabs.vertical.right-side = true`) but prefer the download indicator to appear on the left side of the screen.

## Changes

- Added new preference `zen.downloads.icon-popup-position` in `prefs/zen/downloads.yaml`
- Modified `#areTabsOnRightSide()` in `ZenDownloadAnimation.mjs` to check the new pref first, falling back to tab position when set to `"follow-tabs"`

## Testing

1. Set `zen.tabs.vertical.right-side` to `true` in `about:config`
2. Set `zen.downloads.icon-popup-position` to `"left"` in `about:config`
3. Download a file
4. Verify the download popup appears on the left side despite tabs being on the right

### DIFF
--- a/prefs/zen/downloads.yaml
+++ b/prefs/zen/downloads.yaml
@@ -7,3 +7,6 @@
 
 - name: zen.downloads.download-animation-duration
   value: 1000 # ms
+
+- name: zen.downloads.icon-popup-position
+  value: "follow-tabs"
--- a/prefs/zen/downloads.yaml
+++ b/prefs/zen/downloads.yaml
@@ -9,4 +9,7 @@
   value: 1000 # ms
 
 - name: zen.downloads.icon-popup-position
-  value: "follow-tabs"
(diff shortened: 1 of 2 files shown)


mr-cheffy (approved):

mr-cheffy (approved):
Changed the preference to be numbers, so its easier to change instead of righting the full english word. See my comments to see what number means what. Thanks!

Issue #8894, opened 2025-06-05
Copy current tab URL shortcut doesn't take priority over site


uncenter:
### Preliminary Checks

- [x] I have read and understood the important section above.
- [x] I have searched existing issues and avoided creating duplicates.
- [x] I am not filing an enhancement request.
- [x] I have checked that this issue cannot be reproduced on Mozilla Firefox.
- [x] I have checked that this issue can be reproduced once I removed all my Mods and Custom CSS.

Reopening of https://github.com/zen-browser/desktop/issues/5963.

### What happened?

Using Zen's copy current tab URL shortcut (Cmd+Shift+C by default) is great, but it doesn't take priority over the current tab's site shortcuts.

### Expected behavior

Zen should take priority of the shortcut and the URL should always be copied regardless of site behavior.

### Actual behavior

The site takes priority of capturing the keyboard shortcut and Zen does not copy the URL.

### Steps to reproduce

1. Go to a file on GitHub, such as https://github.com/zen-browser/desktop/blob/dev/LICENSE.
2. Run the copy URL shortcut (Cmd+Shift+C).
3. Observe that file contents are copied (GitHub's shortcut), not the tab URL.

### Version

1.12.10b

### What platform are you seeing the problem on?

macOS - aarch64

### What component is this issue related to?

Keyboard Shortcuts

### Relevant log output if applicable

```shell

```

copilotenstar:
Noticed this on TikTok too. Kind of strange how a site can override the browser hotkeys. Feels like a big security concern.

uncenter:
> Maybe a more flexible approach would help—like using a less common shortcut or even rethinking the implementation altogether. What do you think?

This reads like you wrote this comment with AI. 🤮

> Also worth noting: `Ctrl+Shift+C` is already widely used for browser dev tools (e.g., element picker). Overriding it might not sit well with many users, including myself.

Ctrl + Shift + C *is already the URL shortcut*. There are plenty of issues and discussions about that topic, but it is completely irrelevant here. (You can literally change the shortcut in preferences now if you want iirc!)

> Some sites may implement this shortcut, but users still prefer the browser to take priority—just like in your example.

It seems relatively easy to add a setting for that as well if that is a concern. The harder part I imagine is just taking priority over the site in the first place.

IAmTahazzot:
> This reads like you wrote this comment with AI. 🤮

[...]

PR #11659, opened 2025-12-18
fix: Prevent sites from hijacking keyboard shortcuts by default

aadunn:
Fix to address https://github.com/zen-browser/desktop/issues/8894

As mentioned in this issue, there are websites that can hijack the keyboard shortcuts set by Zen for a site feature. By default this behaviour is disabled on Arc and I believe it also makes sense to have it disabled on Zen.

Enabling site shortcuts is still possible on a per-site basis in the site settings window.

### DIFF
--- a/prefs/firefox/permissions.yaml
+++ b/prefs/firefox/permissions.yaml
@@ -1,0 +1,10 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+# This file contains configuration values for site permissions
+
+# Disables site shortcuts by default.
+# To allow site shortcuts they have to be explicitly enabled per site
+- name: permissions.default.shortcuts
+  value: 2


mr-cheffy:
Interesting, but I think the most effective way would be to maybe showcase a "do the the shortcut again to do the action" message, like arc does? Maybe changing it for everyone might cause some issues but im not entirely sure how most people would react as if a site is hijacking a shortcut, it might be for a reason. @Propheticus thoughts?

Relevant:
https://searchfox.org/firefox-main/source/dom/events/GlobalKeyListener.cpp#75-87

Propheticus:
> Interesting, but I think the most effective way would be to maybe showcase a "do the the shortcut again to do the action" message, like arc does? Maybe changing it for everyone might cause some issues but im not entirely sure how most people would react as if a site is hijacking a shortcut, it might be for a reason. @Propheticus thoughts?
> 
> Relevant: https://searchfox.org/firefox-main/source/dom/events/GlobalKeyListener.cpp#75-87

I'd respect site shortcuts and try to use browser shortcut defaults that are unlikely to overlap with those often used on sites. (e.g. don't use Ctrl+s) 
Maybe offer a checkbox on the shortcut settings page to "Block sites from overriding shortcuts unless explicitly allowed via site permissions" ? 
I think defaulting that on would be unexpected by most users and annoying on many sites with comment fields or text editors that use **ctrl+b** for bold and _ctrl+i_ for italic etc. Trying to type some bold text and the bookmarks sidebar opening instead...

Starry-AXQG:
I think we should display a pop-up window when users first enter the site to ask if they want to disable the site's shortcut keys, and then save the setting in the website settings. This behavior is similar to that of Firefox and Zen's native functions. For instance, Firefox and Zen will prompt the user whether to allow websites to automatically play videos, or to grant permission for website notifications.

Edite: We should display the permission prompt window when the user presses the shortcut key, as this would be more in line with the user's intuition.

Propheticus:
After a years long debate Mozilla/ Firefox settled on a default allow, but optional disallow on a per-site basis: https://bugzilla.mozilla.org/show_bug.cgi?id=380637

Either way (default allow, default disallow) is imperfect, but I also feel getting a prompt on every (new) site will get bothersome quickly.

Propheticus:
Regarding the actual pref change proposed by this PR: Defaulting this to blocked is not a good idea in Zen at the moment. It breaks things like backspace and delete in text fields.

[noDelete_Backspace.webm](https://github.com/user-attachments/assets/d38fe208-3d23-417c-923a-8f8662caf504)

(Zen left, Firefox right)

mr-cheffy:
As @Propheticus, changing the default value would just be a hacky fix to this issue and doesn't really solve the underlining problem. I believe the best solution would be to implement "Do it again to perform the action", like how arc does it.

aadunn:
> As @Propheticus, changing the default value would just be a hacky fix to this issue and doesn't really solve the underlining problem. I believe the best solution would be to implement "Do it again to perform the action", like how arc does it.

Fair enough! I'll try to get a PR up this weekend implementing that

mr-cheffy:
That would be great!

PR #11703, opened 2025-12-23
feat(urlbar): Open multiple pasted URLs in separate tabs

salmonumbrella:
## Summary

When pasting multiple URLs into the URL bar and pressing Enter, each URL now opens in a separate tab instead of treating the combined string as a malformed single URL.

**Use case:** User copies multiple URLs from a document/spreadsheet and pastes them into Cmd+T - instead of failing to navigate to a garbled string, Zen opens each URL in its own tab.

## Features

- Supports `https://`, `http://`, `ftp://`, `file://` protocols
- Detects bare domains (e.g., `example.com`) and normalizes to `https://`
- Handles concatenated URLs (`url1https://url2`), whitespace-separated, and newlines
- Deduplicates URLs

## Security

- Validates URLs via `Services.io.newURI()` 
- Blocks dangerous schemes (`javascript:`, `data:`, `about:`, etc.)
- Limits to 20 tabs per paste to prevent DoS
- Input size capped at 100KB to prevent regex DoS

## Test plan

- [ ] Paste two URLs separated by space → both open in new tabs
- [ ] Paste URLs separated by newline → both open in new tabs  
- [ ] Paste concatenated URLs (`https://a.comhttps://b.com`) → both detected and opened
- [ ] Paste bare domains (`google.com facebook.com`) → normalized to https and opened
- [ ] Paste single URL → normal behavior (no change)
- [ ] Paste `https://safe.com javascript:alert(1)` → only safe.com opens
- [ ] Paste 25+ URLs → only first 20 open (DoS protection)

### DIFF
--- a/src/browser/components/urlbar/UrlbarInput-sys-mjs.patch
+++ b/src/browser/components/urlbar/UrlbarInput-sys-mjs.patch
@@ -44,7 +44,7 @@ index 6ad064710da20f7a13fda3517780c2f38b3d2865..58a01eef024d8b9992068a6a5b7be68f
      if (!this.#isAddressbar) {
        throw new Error(
          "Cannot set URI for UrlbarInput that is not an address bar"
-@@ -798,8 +818,16 @@ export class UrlbarInput {
+@@ -798,8 +818,20 @@ export class UrlbarInput {
          return;
        }
      }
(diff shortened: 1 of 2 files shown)


mr-cheffy:
Sorry, we don't allow AI generated code

PR #11730, opened 2025-12-25
Set owner when creating a new tab

### LINKED ISSUE #11576: Closing a recently opened tab returns to the top tab, not the previously focused tab
### Preliminary Checks

### What happened?

When opening a new tab then closing it shortly after, Zen focuses the first tab, rather than the previously focused tab.

### Expected behavior

The focused tab should be the previously focused tab before the new tab was opened

### Actual behavior

The focused tab was the tab currently on the top

### Steps to reproduce

1. Open 2 tabs
2. Focus the tab on the bottom
3. Open a new tab via Ctrl+T or pressing the new tab button and entering any website
4. Close the new tab

### Screenshots and videos

_No response_

### Version

1.17.12b

Windows - x64

Tabs

```shell

```

kjljixx:
Fix #11576 by setting the owner of the new tab to the current tab.

### DIFF
--- a/src/browser/components/tabbrowser/content/tabbrowser-js.patch
+++ b/src/browser/components/tabbrowser/content/tabbrowser-js.patch
@@ -262,9 +262,9 @@ index 42027bfa55eab8ea9298a7d425f2ded45188f7f3..a87bcaf4e9b92a9f63e025409efabca9
          UserInteraction.start("browser.tabs.opening", "initting", window);
        }
  
-+      if (!gURLBar.hasAttribute("zen-newtab")) {
        // If we're opening a foreground tab, set the owner by default.
        ownerTab ??= inBackground ? null : this.selectedTab;
++      if (!gURLBar.hasAttribute("zen-newtab")) {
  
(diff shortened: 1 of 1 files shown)


mr-cheffy:
I dont think this is the proper solution as a new tab isn't shouldn't have as owner what's currently selected

kjljixx:
Why not? Should creating a new tab and then closing it not return you to the tab that was previously selected?

mr-cheffy:
Yes, but I dont think setting the owner should be the way, as they dont have any relationship. Checkout this article https://support.mozilla.org/en-US/questions/1235742

kjljixx:
Correct me if I'm wrong, but It looks like setting the owner is [the default FF behavior](https://github.com/mozilla-firefox/firefox/blob/6a0626aaa09a87200258dbf6161fde44722c763e/browser/components/tabbrowser/content/tabbrowser.js#L2978C1-L2978C59) for new tabs?

mr-cheffy:
Yes, but in zen, we do it a bit differently as the new tabs are created from the URL bar, not from the immediate action callback

PR #11737, opened 2025-12-27
refactor: Improve animation robustness and simplify logic

octaviusz:


src/zen/folders/ZenFolders.mjs
```diff
@@ -867,6 +867,12 @@ class nsZenFolders extends nsZenDOMOperatedFeature {
       this.setFolderUserIcon(group, stateData.userIcon);
     }
 
+    // Initialize the tab style
+    group.tabs.forEach((tab) => {
+      tab.style.opacity = 1;
+      tab.style.height = '40px';
```
mr-cheffy, line 873: Why do we need this exactly?

mr-cheffy (changes requested):

PR #11770, opened 2025-12-30
Perceived Size Increase via SVG Padding Optimization

realSilasYang:
By reducing the original SVG viewBox padding by 80% (from 2px to 0.4px per side), the content proportion within the fixed 18px container is significantly increased, effectively achieving a "Perceived Size Increase." This high-fill-rate design optimizes visual weight, allowing icons to exhibit stronger impact and recognizability within complex browser UIs without altering their physical dimensions.


mr-cheffy:
Could you show how it would be different? Maybe via screenshots?

realSilasYang:
I reduced the surrounding whitespace of the original image, allowing a slightly larger icon to fit in the same space.

This is a screenshot comparison of the actual application in the Zen Browser address bar, using the same scaling ratio. I also included the adjacent refresh button to facilitate comparison.
Current version vs. Version I Submitted:
截图_2025-12-30_22-09-34 vs.
截图_2025-12-30_22-10-47

mr-cheffy:
I dont think this makes lots of sense and it goes against nucleo's guidelines. Sorry

PR #11847, opened 2026-01-09
fix: Allow user override of zen.theme.border-radius pref value

### LINKED ISSUE #7004: "zen.theme.border-radius" dosen't work on linux.
### Captchas

### What happened?

"zen.theme.border-radius" in about:config used to work fine until version 1.7.2b.
but it does not change any border radius now. 
(issue is exclusive to linux)

Image

### Version

1.10.3b

Linux (Flatpak)

Customizable UI / Toolbars

```shell

```

busybox11:
Fixes #7004 

Tested and confirmed working on Windows and Linux. Don't have any macOS platform available to test this on yet.

Changed the `zen.theme.border-radius` property to use a default of `-1` (previously `8`) that `zen-theme.css` would treat as "auto-detect". Each platform defines its own override from the CSS rather than from build prefs.

Some notes:

- I'm not sure if `zenThemeModifier.updateBorderRadius` is still useful, even without these changes.

- As discussed in #7004, some other implementations would be possible and might be more desireable.

  Maybe setting the value in JS would have less overhead than CSS, especially if it's only computed once.
  On the other hand, CSS-only is convenient and perf impact is probably negligeable. Not 100% convinced of my approach here.

- The default value pref could be dropped altogether, to make every change intentional as the users will have to create the property by themselves. Not convinced, but worth mentioning (I hope)

### DIFF
--- a/src/zen/common/zenThemeModifier.js
+++ b/src/zen/common/zenThemeModifier.js
@@ -97,7 +97,10 @@ var ZenThemeModifier = {
         document.documentElement.style.setProperty('--zen-border-radius', targetRadius + 'px');
       } else if (AppConstants.platform == 'linux') {
         // Linux uses GTK CSD titlebar radius, default to 8px
-        document.documentElement.style.setProperty('--zen-border-radius', 'env(-moz-gtk-csd-titlebar-radius, 8px)');
+        document.documentElement.style.setProperty(
+          '--zen-border-radius',
+          'env(-moz-gtk-csd-titlebar-radius, 8px)'
+        );
       } else {
         // Windows defaults to 8px
         document.documentElement.style.setProperty('--zen-border-radius', '8px');
--- a/src/zen/common/zenThemeModifier.js
+++ b/src/zen/common/zenThemeModifier.js
@@ -88,11 +88,22 @@ var ZenThemeModifier = {
 
   updateBorderRadius() {
     const borderRadius = Services.prefs.getIntPref('zen.theme.border-radius', -1);
+
     // -1 is the default value, will use platform-native values
     // otherwise, use the custom value
     if (borderRadius == -1) {
-      document.documentElement.style.removeProperty('--zen-border-radius');
+      if (AppConstants.platform == 'macosx') {
+        const targetRadius = window.matchMedia('(-moz-mac-tahoe-theme)').matches ? 15 : 10;
+        document.documentElement.style.setProperty('--zen-border-radius', targetRadius + 'px');
+      } else if (AppConstants.platform == 'linux') {
+        // Linux uses GTK CSD titlebar radius, default to 8px
+        document.documentElement.style.setProperty('--zen-border-radius', 'env(-moz-gtk-csd-titlebar-radius, 8px)');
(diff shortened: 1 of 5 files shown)


src/zen/common/zenThemeModifier.js
```diff
@@ -87,8 +87,14 @@ var ZenThemeModifier = {
   },
 
   updateBorderRadius() {
-    const borderRadius = Services.prefs.getIntPref('zen.theme.border-radius');
-    document.documentElement.style.setProperty('--zen-border-radius', borderRadius + 'px');
+    const borderRadius = Services.prefs.getIntPref('zen.theme.border-radius', -1);
+    // -1 is the default value, will use platform-native values
+    // otherwise, use the custom value
+    if (borderRadius == -1) {
+      document.documentElement.style.removeProperty('--zen-border-radius');
```
mr-cheffy: There are other elements that still use `zen-border-radius`, so I dont think we should be removing it.

Maybe we should do the nativity checks here? It would also clean up the CSS above and we would not need `--zen-native-content-radius` anymore, we could just use this one.

Checkout `AppConstants.platform` and `window.matchMedia('-moz-mac-tahoe-theme')`

mr-cheffy (changes requested):

src/zen/common/zenThemeModifier.js
busybox11: Right. Not a lot of them, just enough for me to forget about it since yesterday. My bad, sorry!

But yeah this is what I thought too. Went with CSS because it was fairly straight forward, but as I said in the notes, JS would have some advantages. 

Fixing this, thanks for the feedback!

busybox11: Hm, further into debugging it looks like `--zen-border-radius` is useless on `.folders-tabs-list-item`.

The hover radius shown to the user is actually coming from its only child, `.folders-tabs-list-item-content`.

Unsure why `&:first-child` and `&:last-child` are present in its style, as it always is the only children of its `.folders-tabs-list-item` parent. Maybe you'll know better, I can't find anything relevant.

The other remaining usage is in `.notificationbox-slack` but I'm not sure why it is not using the same radius var as the rest of the similar `zen-omnibox.css` declarations.

Regardless, I have implemented the JS-based `--zen-border-radius` and replaced all old occurencies of `--zen-native-content-radius`. Will push soon after test.

mr-cheffy: Regardless of whether its used, it would still be beneficial to remove the extra "native" arg and run all the checks on the JS side, so I wouldn't worry much about it tbh

busybox11: Yep, those won't see any change anyway. I'm not that familiar with the codebase so I took the time to investigate, but it's not a priority indeed.

lmk if I should also fix it in this PR, otherwise all good for me!

sidenote - this breaks some zen mod i forgot the name of that applies solid borders to the page frame. I guess it's expected though.

mr-cheffy (approved):

mr-cheffy (approved):

PR #11889, opened 2026-01-13
Zen Library development

JustAdumbPrsn:
- [x] Library base 
- [x] Sidebar
- [x] Animations (also i added animation for nav toolbox)
- [x] UI for spaces (Workspace cards not with tabs right now)

Does not currently have proper support for compact mode when sidebar is hidden and zen right side
I have tried to make the tabpanel translate to right instead of shrinking width but i am having problems with that (shrinking width each frame can cause drop in performance so i want to translate tabpanel like arc does)


MiraiDevv:
Could someone list and create a to-do list of the remaining tasks within this project so I can guide myself and help complete it? I know this is for creating the Zen library, but I'm still not sure exactly what your plans are for what to include or do.

mr-cheffy:
https://github.com/zen-browser/desktop/pull/12655

PR #11910, opened 2026-01-15
feat: Add option to restore closed tabs to original workspace

7ito:
Adds a new preference 'zen.workspaces.restore-tabs-to-original-workspace'. 
When enabled, restores closed tabs (Ctrl+Shift+T) to the workspace they were originally closed from (and moves you to that tab and workspace), instead of the currently active workspace.
- Adds preference toggle in Settings > Tab Management
- Modifies undoCloseTab to preserve original workspace
- Adds changeWorkspace option to addTrustedTab
- Fixes tab container placement in onTabBrowserInserted


mr-cheffy:
Thanks for the PR! Unfortunately, I fixed this before I saw the PR... The solution was much more simpler, just needed to add a missing argument to `addTab`. Again, thanks a lot!

7ito:
I see, that’s great! As long as I can have that tab re-open behaviour on Zen, I’m happy, thanks!

PR #11925, opened 2026-01-16
Adds workspace search filter

jababda:
This adds a workspace search filter to the url bar in a very similar way to how firefoxes default local search filters work. Its intent is to allow those of us with lots of workspaces to easily


mr-cheffy:
Closing as I see you opened https://github.com/zen-browser/desktop/pull/11926

PR #11933, opened 2026-01-17
feat: Return to previous tab when closing tabs

### LINKED ISSUE #11775: Closing Tab Does Not Go To Previous Tab for Pinned tabs
### Preliminary Checks

### What happened?

When closing a tab, I don't consistently go to the previous tab I was in.

### Expected behavior

Go back to essential tab

### Actual behavior

Goes to first tab

### Steps to reproduce

1. Be on a Essential Tab
2. Hit Cmd + T to open a new tab
3. Navigate to any website
4. Hit Cmd + W to close the recent tab
5. You'll end up on the first non-essential tab you have open (instead of the Essential Tab you were previously on)

### Screenshots and videos

_No response_

### Version

1.17.14b

macOS - aarch64

Tabs

```shell

```

Anodev-sudo:
Removes the zen-newtab guard that prevents owner tab assignment when a new tab is created via hotkey or url bar. Closing a new tab will now return to the tab it was opened from.  

Previously, closing these tabs would jump to the first regular tab instead of returning to the opener.

Ownership will not be assigned if multiple background tabs are created, but this matches current behavior. Single background tabs or newly created tabs will have ownership.

My testing hasn’t shown any other changes in behavior.

Fixes #11775

### DIFF
--- a/prefs/firefox/browser.yaml
+++ b/prefs/firefox/browser.yaml
@@ -72,3 +72,6 @@
 
 - name: browser.tabs.notes.enabled
   value: false
+
+- name: browser.tabs.dragDrop.moveOverThresholdPercent
+  value: 50 # Percentage of tab height to trigger move over on drag-and-drop
--- a/prefs/zen/zen.yaml
+++ b/prefs/zen/zen.yaml
@@ -20,6 +20,9 @@
 - name: zen.tabs.close-window-with-empty
   value: true
 
+- name: zen.tabs.select-recently-used-on-close
+  value: true
+
 - name: zen.tabs.use-legacy-drag-and-drop
   value: false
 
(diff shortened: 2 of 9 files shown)


mr-cheffy (changes requested):
I believe this is a hacky solution since you are simply allowing ALL tabs to have an owner (meaning you can close them, for example, with the back button, which is not what tab.owner is supposed to do). 

Also, tab.owner might get cleared in some cases such as when opening a new background tab from the recently opened tab. 

I think the fix rests around the `_findtabtoblur` (https://searchfox.org/firefox-main/source/browser/components/tabbrowser/content/tabbrowser.js#5859) function, where the next tab should be decided based on the last accessed time

Anodev-sudo:
I removed the previous changes. Now, lastAccessed will be used to determine the last active tab, which will be switched to on tab close.

mr-cheffy (approved):

PR #11942, opened 2026-01-18
feat: new history/library component for zen browser for a more modern age.

MiraiDevv:
Just like every browser has a different menu, I thought I'd do the same for Zen, since the one it has is still from its older sibling, Firefox. Here's a video of it working (I was showing the functionalities but for some reason it stopped recording, maybe because it's a secondary window, but as you can see at the beginning, it opens and works!).

Any additions to this PR are welcome, and if anyone wants to offer any criticism about the design, feel free to do so.

Also, if anyone has any ideas for new features I could add, I'd be happy to know; I haven't been able to think of any useful ones!


MiraiDevv:
Here's the video:

https://github.com/user-attachments/assets/614818dc-8342-4d5f-88e1-bdb8720c71f6

mr-cheffy:
Sorry, but we are trying to not mantain firefox components that dont really are "zen features" (extra maintenance cost). Instead, we are working on our own library implementation at the sidebar in https://github.com/zen-browser/desktop/pull/11889.

PR #11943, opened 2026-01-18
Importal Week1: add per-window windowType scaffold

nimlik:


mr-cheffy:
Whats this exactly for or what does it fix?

PR #11956, opened 2026-01-20
Fix: Show container indicator for non-default containers in workspace

### LINKED ISSUE #11951: Non-default container indicator not showing in workspace
### Preliminary Checks

### What happened?

I have a workspace with a default container set, and I have the "Hide the default container indicator in the tab bar" setting enabled. When I open a tab in a different container in that workspace, the container indicator does not show up on the tab (the container name and icon is still shown in the URL bar, however). This makes it difficult to distinguish which container a tab is assigned to without clicking into it. If I disable "Hide the default container indicator in the tab bar," then all tabs show container indicators.

### Expected behavior

A tab opened in a different container than the workspace's default container should show the container indicator on the tab.

### Actual behavior

No container indicator is shown on a tab opened in a different container than the workspace's default container.

### Steps to reproduce

Enable "Hide the default container indicator in the tab bar". Assign a default container to a workspace. Open a tab in that workspace, but in a different container. No container indicator is shown on the tab.

### Screenshots and videos

_No response_

### Version

1.17.15b (Firefox 146.0.1) (aarch64)

macOS - aarch64

Workspaces

```shell

```

ZeusCraft10:
Hey there! 👋

This fixes #11951 where the container indicator wasn't showing up when opening tabs in a different container than the workspace's default.

## The Problem

When you have a workspace with a default container set (say, "Personal") and the "Hide the default container indicator" setting enabled, opening a tab in a *different* container (like "Work") would still hide the indicator. This makes it hard to tell which container a tab belongs to without clicking into it.

## The Fix

One-line change in `ZenWorkspaces.mjs` - when the code detects a tab should go to a different workspace based on its container, it was incorrectly marking the tab as having the 'default' container context. Changed that from `true` to `false` so the indicator shows correctly.

## Testing

- Have workspace A with container X as default
- Have workspace B with container Y as default
- Enable "Hide the default container indicator in the tab bar"
- Open a tab in container Y while in workspace A
- The container indicator should now show (before this fix, it was hidden)

### DIFF
--- a/src/zen/workspaces/ZenWorkspaces.mjs
+++ b/src/zen/workspaces/ZenWorkspaces.mjs
@@ -2801,7 +2801,7 @@ class nsZenWorkspaces {
       if (matchingWorkspaces.length === 1) {
         const workspace = matchingWorkspaces[0];
         if (workspace.uuid !== this.getActiveWorkspaceFromCache().uuid) {
-          return [userContextId, true, workspace.uuid];
+          return [userContextId, false, workspace.uuid];
         }
       }
     }


src/zen/workspaces/ZenWorkspaces.mjs
```diff
@@ -2801,7 +2801,7 @@ class nsZenWorkspaces {
       if (matchingWorkspaces.length === 1) {
         const workspace = matchingWorkspaces[0];
         if (workspace.uuid !== this.getActiveWorkspaceFromCache().uuid) {
-          return [userContextId, true, workspace.uuid];
+          return [userContextId, false, workspace.uuid];
```
mr-cheffy, line 2804: How does this fix the issue exactly? This shouldn't even run based on the description from the issue

mr-cheffy (changes requested):

PR #11998, opened 2026-01-24
feat: Add search option for URL-like input

chelokot:
One of the inconveniences I personally often face when using Zen in my day to day life is occasional lack of search option when opening new tab.

Here is example from my real life practice today. I encountered some description of `nvcr.io/nvidia/tritonserver` during my work. I wanted to find relevant information about it. So I clicked "New Tab" with intention to search for it. And here is what I received:

image

There is no option to search. Zen sees that it's a url and proposes me to visit it. That's nice, but in this case, I clearly don't need to visit this url, it won't give me anything, I want to search it!

There are lot of such cases that you can name: getting link to some website and wanting to find some info about, finding info about some IP address or maybe even IP:port combination for something like finding info about minecraft server, etc. 

Basically, my point is: there is *always* a possibility that user wants to search what they entered. And they don't really have other options how to do it conveniently. The only workaround that I personally know is moving cursor into the beginning, adding some character and then adding space. Some browsers work with just adding space, but Zen doesn't. This feels bad each time I experience it.

I believe that correct and most convenient behavior is to **always give search option**. It shouldn't be main option for URLs of course, but it should be the second option.

This is how it works in this PR:

image

Simple, intuitive, consistent, easy to use, accessible for both keyboard (↓ + Enter) and mouse (single click) users. Code change is also pretty simple. It ensures that there is always search option added.

---

The code is modified to provide search option in different cases too, not just URLs (as those are managed by different functions in the code)

For example, do you think nobody ever searches for `127.0.0.1`, not to visit it:

image

... but to search what is it?

image

image

I think there are such people.

And do you think there are no people who want to not open `about:config` but search info about it?

image

image

I think there are such people. There are people that search for everything. Search option is one of the most generally useful options.

---

To keep this PR as minimal as possible, those behaviors are not made configurable in settings in any way. You you believe it's better to have some setting for this, I am open to propositions on it.

---

My reasoning to add separate search option and not, for example, make it search when adding space to the end, is that I personally had times in life when I had some URLs copied with spaces before or after. I think if I enter such URL, click Enter following my muscle memory and end up in Google, I will be pretty annoyed. Moreover, after such mistake happening, it's not quick to paste URL again and strip spaces from it. So, while it is arguable, adding spaces will break some behavior that people generally count on.

Moreover, adding space to the end of the url isn't even that intuitive. Like, why do I have to figure out to add space to URL to search for it, what does space have to do with search? It's essentially some kind of trick that lot of us know, but some people might not even think about.

My change, on other hand, doesn't break general behaviors but instead adds another option - and this option is clear and visible to everyone, without knowledge of any tricks

---

This is my first contribution to Zen Browser, so I am open to any feedback on style of contributing that you prefer


ExposedCat:
I'd love to see this is Zen, I struggle with this so often

src/browser/components/urlbar/UrlbarMuxerStandard-sys-mjs.patch
```diff
@@ -1,16 +1,32 @@
...
+index c12d875172650dddfe7de623a776149517c83302..1fc68a0b02a9a7f15605d171b12606452198a9f6 100644
 --- a/browser/components/urlbar/UrlbarMuxerStandard.sys.mjs
 +++ b/browser/components/urlbar/UrlbarMuxerStandard.sys.mjs
-@@ -854,6 +854,7 @@ class MuxerUnifiedComplete extends UrlbarMuxer {
+@@ -843,17 +843,22 @@ class MuxerUnifiedComplete extends UrlbarMuxer {
+     }
+ 
+     // UrlbarProviderHeuristicFallback may add non-heuristic results in some cases,
+-    // but those should be retained only if the heuristic result comes from it.
++    // but those should be retained only if the heuristic result comes from it or is a URL.
```
mr-cheffy: Don't edit this comment, we are trying to make patches as small as possible (even if the comments stop making sense)

src/browser/components/urlbar/UrlbarMuxerStandard-sys-mjs.patch
```diff
@@ -1,16 +1,32 @@
...
++      !(
++        result.type == UrlbarUtils.RESULT_TYPE.SEARCH &&
++        state.context.heuristicResult?.type == UrlbarUtils.RESULT_TYPE.URL
++      )
+     ) {
+       return false;
      }
  
      if (result.providerName == "UrlbarProviderTabToSearch") {
-+      return false;
```
mr-cheffy, line 9: Why did you remove this?

src/browser/components/urlbar/UrlbarMuxerStandard-sys-mjs.patch
```diff
@@ -1,16 +1,32 @@
...
+     // UrlbarProviderHeuristicFallback may add non-heuristic results in some cases,
+-    // but those should be retained only if the heuristic result comes from it.
++    // but those should be retained only if the heuristic result comes from it or is a URL.
+     if (
+       !result.heuristic &&
+       result.providerName == "UrlbarProviderHeuristicFallback" &&
+       state.context.heuristicResult?.providerName !=
+-        "UrlbarProviderHeuristicFallback"
++        "UrlbarProviderHeuristicFallback" &&
++      !(
```
mr-cheffy: Could move that `&&` to the line below to avoid that extra line of patch

src/browser/components/urlbar/UrlbarProviderHeuristicFallback-sys-mjs.patch
```diff
@@ -0,0 +1,52 @@
...
++        let [searchCandidate] = UrlbarUtils.stripPrefixAndTrim(
++          trimmedSearchString,
++          {
++            trimSlash: true,
+           }
++        );
++        if (
++          lazy.UrlbarPrefs.get("keyword.enabled") &&
++          (lazy.UrlUtils.looksLikeUrl(searchCandidate) ||
++            lazy.UrlUtils.looksLikeOrigin(searchCandidate, {
```
mr-cheffy, line 29: This code looks duplicated from the removed lines above, what actually did you change here?

mr-cheffy (changes requested):

mr-cheffy:
@chelokot how's it going? need any help?

mr-cheffy (approved):

PR #12033, opened 2026-01-26
fix: Update notification popup icon styles to account for showing state

### LINKED ISSUE #11870: When on the login page, the extension icon in the address bar cannot be clicked
### Preliminary Checks

### What happened?

https://github.com/user-attachments/assets/ff5b8077-47aa-41b9-b757-b4c23398a93b

### Expected behavior

the extension icon in the address bar can be clicked

### Actual behavior

When on the login page, the extension icon in the address bar cannot be clicked, and the option to remember password always pops up

### Steps to reproduce

1. open some login page
2. input username and password, then click login btn,The backend returned some errors,still stay login page 
3. click the extension icon in the address bar 

### Screenshots and videos

_No response_

### Version

1.17.15b

Windows - x64

URL Bar

```shell

```

busybox11:
Fixes #11870 

Icons in the omnibox indicate pending actions (e.g., password save modal) and should remain visible until the user completes or rejects the action.

When a modal is dismissed, JS keeps showing=true on the icon, but Zen wasn't accounting for this and hid the icon with opacity: 0 while keeping it clickable.
This also blocked clicks to icons positioned underneath.

https://github.com/user-attachments/assets/545fc42b-363f-4457-bcee-bec4957849a5

### DIFF
--- a/src/zen/common/styles/zen-omnibox.css
+++ b/src/zen/common/styles/zen-omnibox.css
@@ -309,7 +309,7 @@
     order: 2;
   }
 
-  #notification-popup-box:not([open]) {
+  #notification-popup-box:not([open]):not(:has(> [showing="true"])) {
     margin-inline-start: calc(-10px - 2 * var(--urlbar-icon-padding));
     opacity: 0;
     transition: all 0.2s;


src/zen/common/styles/zen-omnibox.css
```diff
@@ -309,7 +309,7 @@
     order: 2;
   }
 
-  #notification-popup-box:not([open]) {
+  #notification-popup-box:not([open]):not(:has(> [showing="true"])) {
```
mr-cheffy, line 312: I don't really like the use of :has, is there no other way this could be done? Maybe we should apply these styles to the icons inside, instead of the entire `notification-popup-box`. So, for example, `#notification-popup-box:not([open]) <element>:not([showing="true"]) {}`

mr-cheffy (changes requested):

src/zen/common/styles/zen-omnibox.css
busybox11: Yeah I don't either. It was the only approach that I could think of that would allow now writing rules for every single element.

I'm not familiar enough with Firefox's urlbar to know what icons are supposed to have this behavior other than the password one, I don't think I have encountered any other yet. Selecting only the password element sounds like it'd work for me

busybox11: @mr-cheffy quick update - i feel like this is the overall best solution anyway

im generally not a fan of `:has`, but this specific instance makes sense, with virtually [no performance impact](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/:has#performance_considerations) as the observed DOM tree is shallow and often not event present. it's essentially less code and requied maintenance for the same result

It's definitely the simplest "universal" fix, used a couple times in newer FF UI for similar stuff too.

lmk if you'd still prefer doing it per element though. more granularity could be useful, you'll know what would be best. that's just my opinion!

mr-cheffy: Ok, i'll allow it this time as it indeed doesn't seem like theres much perf impact as its just checking the children underneath.

mr-cheffy (approved):

PR #12085, opened 2026-01-27
fix(urlbar): guard against null messages in l10n formatting

TogiFerretFerret:
Fixes a TypeError where `l10n.formatMessages` returns an array containing `null` or `undefined` entries, causing console spam and logic failures.

The previous check `!messages?.length` was insufficient because the array can exist but contain invalid entries (e.g., `[null]`). This patch adds a guard to ensure the message object exists before accessing `.value`.

Fixes console error:
`TypeError: can't access property "value", messages[0] is null`


mr-cheffy:
Do you know what IDs are trying to be fetched? And could I get a screenshot of those errors? Maybe it would be more benefitial to fix why its not finding any translations rather than adding a guard

TogiFerretFerret:
Nevermind, the errors seem to have disappeared naturally. Might have been a locale issue.

PR #12086, opened 2026-01-27
Fix #12074: Preserve container when opening View Page Source

### LINKED ISSUE #12074: View Page Source opens in No Container Tab
### Preliminary Checks

### What happened?

Image

I am signed in to different accounts in different container tabs for developer purposes. Whenever I want to view the page source from said account, it makes it in No Container instead, which means without the logged-in user, and the source code is therefore different.

Consequently, ctrl+T also creates No Container tab, which is annoying as well. But it is reported already:
 - #8332

### Expected behavior

Opens in container tab

### Actual behavior

Opens in no container tab

### Steps to reproduce

1. Do "View page source" from a container tab
2. Opens outside the container tab

### Screenshots and videos

Image

### Version

1.18.1b

Windows - x64

Tabs

```shell

```

R0h1tAnand:
## Description

Fixes #12074 

This PR addresses the issue where "View Page Source" was opening in the default container (No Container) instead of preserving the container from the originating tab.

## Changes Made

Modified `src/browser/components/tabbrowser/content/tabbrowser-js.patch` to detect when a view-source URL is being opened without an explicit `userContextId` and inherit it from the currently selected tab.

The fix adds a check before `getContextIdIfNeeded()` to ensure proper container propagation:

```
js
// Fix for issue #12074: Inherit container from current tab for view-source
// If no userContextId is provided and we're opening a view-source URL,
// inherit the container from the currently selected tab
if (userContextId === undefined && aURI?.spec?.startsWith("view-source:") && this.selectedTab) {
  const selectedUserContextId = this.selectedTab.getAttribute("usercontextid");
  if (selectedUserContextId) {
    userContextId = parseInt(selectedUserContextId, 10);
  }
}

```
---
### Testing
Note: I was unable to build and test this locally due to setup constraints (30GB+ disk space and long compile time).

Suggested test steps:

Open a tab in any container (Personal, Work, etc.)
Navigate to a website
Press Ctrl+U / Cmd+U, or right-click → View Page Source
Verify the source tab shows the same container as the original tab
### Additional testing:

Test with different container types
Verify normal tab creation is unaffected
Confirm tabs without containers still work correctly
### Checklist
 

- [x] Followed contributing guidelines
- [x]  Kept changes minimal and targeted
- [x]  Added clear comments explaining the fix
- [ ]  Tested locally (unable due to build constraints - requesting maintainer testing)

Would appreciate if someone with a local build could verify this works as expected. Happy to make any adjustments if needed!

### DIFF
--- a/src/browser/components/tabbrowser/content/tabbrowser-js.patch
+++ b/src/browser/components/tabbrowser/content/tabbrowser-js.patch
@@ -266,6 +266,17 @@ index 0eaca7a58e0026237b71b2ad515efe84d9e8c779..2ab8633d0378e5a25cf75c50b29d4689
          );
        }
  
++      if (
++        userContextId === undefined &&
++        aURI?.spec?.startsWith("view-source:") &&
++        this.selectedTab
++      ) {
++        const selectedUserContextId = this.selectedTab.getAttribute("usercontextid");
++        if (selectedUserContextId) {
++          userContextId = parseInt(selectedUserContextId, 10);
++        }
++      }
++
 +      let hasZenDefaultUserContextId = false;
 +      let zenForcedWorkspaceId = undefined;
 +      if (typeof gZenWorkspaces !== "undefined" && !_forZenEmptyTab) {


Brambor:
I apreciate you taking on my issue. However, your solution seems to be using AI. Which is fine as long as you check it yourself.

But you put the burden of testing on maintainers. Such behaviour is closing down open source.

You can see this rant of a YouTube video, for example: https://youtu.be/uZy1M_s-8zs

R0h1tAnand:
> I apreciate you taking on my issue. However, your solution seems to be to use AI. Which is fine as long as you check it yourself.
> 
> But you put the burden of testing on maintainers. Such behaviour is closing down open source.
> 
> You can see this rant of a YouTube video, for example: https://youtu.be/uZy1M_s-8zs

You are right about it but when I said about the testing issue on me that did not mean I did not try when building the entire project it was taking a lot of time only then I asked it anyone else who has the firefox engine already downloaded can test or sometimes the test is automated via the gh actions only then I pushed it. I am still on process building it.

I am very aware about your concern, will keep this is mind. Apologies for it.

Brambor:
There is no rush. Take your time. Download, compile, test, fix etc.

codeCraft-Ritik (commented):
Solid improvement — inheriting userContextId from the selected tab seems like the expected UX here.
Thanks for including detailed manual test steps as well.

PR #12160, opened 2026-01-30
fix(split-view): correct backwards navigation in grid layout

### LINKED ISSUE #12070: Grid Split navigations is backwards
### Preliminary Checks

### What happened?

When I open 3 or 4 tabs and use CTRL+G to split into the Grid view and navigate via keyboard CTRL-Page UP or Down or CTRL-TAB or CTRL-SHIFT-TAB instead of the tabs going from the left to the right they go in reverse order from right to left. It is very confusing on other split views it goes from left to right or up to down while the grid view goes in reverse order. When your TAB is at the very top you expect to go to the right side the grid navigates backwards.

### Expected behavior

When you open 3 or 4 tabs you expect it to navigate from the LEFT side UP to DOWN then RIGHT TOP and then DOWN for navigation

### Actual behavior

It goes in reverse order it goes from the RIGHT side then it goes UP TO DOWN then it goes to LEFT side then TOP and DOWN. 

### Steps to reproduce

Open 4 tabs, select the with CTRL left click, press CTRL-G for grid split, press CTRL-Page Up and Down or CTRL-TAB and CTRL-SHIFT-TAB. They will navigate from the RIGHT side to the LEFT not from the LEFT to the RIGHT.

### Screenshots and videos

Please reverse the order from RIGHT to LEFT

https://github.com/user-attachments/assets/a6df6cd7-8a3d-45ee-b15c-3060301ce11b

### Version

1.17.15b, 1.18.1b

Windows - x64, Linux (Flatpak)

Split View

```shell

```

TogiFerretFerret:
The grid layout was being constructed using `addChild`'s default behavior, which prepends nodes. This caused columns to be added in reverse order (Right-to-Left). This change explicitly sets `prepend` to `false` to ensure columns are appended Left-to-Right, matching the logical tab order.

Closes #12070

### DIFF
--- a/src/zen/split-view/ZenViewSplitter.mjs
+++ b/src/zen/split-view/ZenViewSplitter.mjs
@@ -1386,10 +1386,10 @@ class nsZenViewSplitter extends nsZenDOMOperatedFeature {
           new nsSplitLeafNode(tabs[i], 50),
           new nsSplitLeafNode(tabs[i + 1], 50),
         ];
-        rootNode.addChild(columnNode);
+        rootNode.addChild(columnNode, false);
       }
       if (tabs.length % 2 !== 0) {
-        rootNode.addChild(new nsSplitLeafNode(tabs[tabs.length - 1], rowWidth));
+        rootNode.addChild(new nsSplitLeafNode(tabs[tabs.length - 1], rowWidth), false);
       }
     }
 


TogiFerretFerret:
I'm fixing the linting...

TogiFerretFerret:
Linting fixed

mr-cheffy (approved):

PR #12180, opened 2026-01-31
feat(ui): Add pref to force native xdg popups instead of subsurfaces

DADA30000:
Just a small patch that changes all type hints related to some kind of popups to POPUP_MENU, so that they would use xdg_popup instead of subsurfaces, this allows compositors to apply some effects to them like blur.
Also kept under a pref to keep it opt-in, to not cause any sudden problems to every user.

### DIFF
--- a/src/widget/gtk/nsWindow-cpp.patch
+++ b/src/widget/gtk/nsWindow-cpp.patch
@@ -1,0 +1,88 @@
+diff --git a/widget/gtk/nsWindow.cpp b/widget/gtk/nsWindow.cpp
+index e8cef42106..1eff90f9ee 100644
+--- a/widget/gtk/nsWindow.cpp
++++ b/widget/gtk/nsWindow.cpp
+@@ -794,2 +794,11 @@ bool nsWindow::WidgetTypeSupportsNativeCompositing() {
+ 
++// Small wrapper for hints that changes them to xdg_popups based on pref
++static GdkWindowTypeHint GetGdkTypeHint(GdkWindowTypeHint aOriginalHint) {
++  if (Preferences::GetBool("zen.widget.wayland.force_native_menus", false)) {
++    return GDK_WINDOW_TYPE_HINT_POPUP_MENU;
++  }
++
++  return aOriginalHint;
++}
++
+ static bool IsPenEvent(GdkEvent* aEvent, bool* isEraser) {
+@@ -1793,3 +1802,3 @@ bool nsWindow::WaylandPopupConfigure() {
+     case PopupType::Tooltip:
+-      gtkTypeHint = GDK_WINDOW_TYPE_HINT_TOOLTIP;
++      gtkTypeHint = GetGdkTypeHint(GDK_WINDOW_TYPE_HINT_TOOLTIP);
+       LOG("  popup type Tooltip");
+@@ -1797,3 +1806,3 @@ bool nsWindow::WaylandPopupConfigure() {
+     default:
+-      gtkTypeHint = GDK_WINDOW_TYPE_HINT_UTILITY;
++      gtkTypeHint = GetGdkTypeHint(GDK_WINDOW_TYPE_HINT_UTILITY);
+       LOG("  popup type Utility");
+@@ -1806,3 +1815,3 @@ bool nsWindow::WaylandPopupConfigure() {
+     LOG("  not tracked in popup hierarchy, switch to Utility");
+-    gtkTypeHint = GDK_WINDOW_TYPE_HINT_UTILITY;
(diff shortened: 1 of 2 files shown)


mr-cheffy (approved):
Looks good, but it would be interesting to send this patch to upstream firefox so we can off-load these patches.

DADA30000:
> Looks good, but it would be interesting to send this patch to upstream firefox so we can off-load these patches.

It's just that the process of sending a patch to upstream firefox is not that easy, also would take a looot of time, and I kinda feel like it could be rejected, but I will try
And in meantime, I would like this patch to be merged for now :)

DADA30000:
> Looks good, but it would be interesting to send this patch to upstream firefox so we can off-load these patches.

hey, don't want to be annoying, but when this will be merged? Like do I need to do something else? I just really-really want this to be merged, as compiling zen from source on my poor FX 8350 is, well, you know

mr-cheffy (approved):
I still recommend you send this patch to firefox upstream

src/widget/gtk/nsWindow-cpp.patch
```diff
@@ -0,0 +1,87 @@
...
++  }
++
++  return aOriginalHint;
++}
++
+ static bool IsPenEvent(GdkEvent* aEvent, bool* isEraser) {
+@@ -1793,3 +1802,3 @@ bool nsWindow::WaylandPopupConfigure() {
+     case PopupType::Tooltip:
+-      gtkTypeHint = GDK_WINDOW_TYPE_HINT_TOOLTIP;
++      gtkTypeHint = GetGdkTypeHint(GDK_WINDOW_TYPE_HINT_TOOLTIP);
```
mr-cheffy, line 20: I dont understand why make this change? you are giving GTK the wrong hint now

mr-cheffy (changes requested):

src/widget/gtk/nsWindow-cpp.patch
DADA30000: to also make tooltips use xdg_popups, as they also use subsurfaces when possible by default, so can't apply blur to them
I personally didn't notice any problems after this patch with tooltips, like they still behave absolutely the same

mr-cheffy: Im not too sure... There must be a reason why firefox added these specific hints, probably rendering related.

DADA30000: > Im not too sure... There must be a reason why firefox added these specific hints, probably rendering related.

it's locked behind a pref that is disabled by default anyway. it's like also called "force" native popups, as it like forces things that are not really supposed to be xdg popups, to be xdg popups, so some kind of breakage is expected
like there are far more broken things locked behind prefs (like gtk widgets pref, that completely breaks right click menu), it's nice to have some kind of option

although maybe it's better to change name to force xdg popups

mr-cheffy:
Im not too in favor of accepting this PR mainly because of maintenance concerns. What I recommend is submitting this patch to Firefox and then we can pull this patch instantly into zen so you dont have to wait until it reaches to stable. 

You can see how to submit patches to them at https://firefox-source-docs.mozilla.org/contributing/contribution_quickref.html#to-write-a-patch.

Sorry...

DADA30000:
> Im not too in favor of accepting this PR mainly because of maintenance concerns. What I recommend is submitting this patch to Firefox and then we can pull this patch instantly into zen so you dont have to wait until it reaches to stable.
> 
> You can see how to submit patches to them at https://firefox-source-docs.mozilla.org/contributing/contribution_quickref.html#to-write-a-patch.
> 
> Sorry...

well let's hope they merge it
https://bugzilla.mozilla.org/show_bug.cgi?id=2015283

mr-cheffy:
I'll CC to it and add it into zen if it gets merged

DADA30000:
> I'll CC to it and add it into zen if it gets merged

hey it got accepted and soon will be merged https://phabricator.services.mozilla.com/D282283
I didn't think it would be that fast actually, I guess thanks for pushing me :P

mr-cheffy:
Seems to have some regressions: https://bugzilla.mozilla.org/show_bug.cgi?id=2016172

DADA30000:
> Seems to have some regressions: https://bugzilla.mozilla.org/show_bug.cgi?id=2016172

yeah, just noticed, fixed it now
now we wait for another build check

DADA30000:
> Seems to have some regressions: https://bugzilla.mozilla.org/show_bug.cgi?id=2016172

https://bugzilla.mozilla.org/show_bug.cgi?id=2015283
https://bugzilla.mozilla.org/show_bug.cgi?id=2016172

it's merged now :)

PR #11935, opened 2026-01-17
Address potential memory leaks in Zen Browser

JorySeverijnse:
## Summary
Address potential memory leaks on Arch Linux by improving event listener cleanup and RefPtr handling, which may help reduce ghost window accumulation and shared memory growth.
## Changes
- **ZenStyleSheetCache.cpp**: Correct RefPtr assignment to use `result.unwrap()` instead of `unwrapOr(nullptr)`, ensuring proper refcounting for stylesheet loading.
- **ZenWorkspaces.mjs**: Add event listener cleanup on `pagehide` (bfcache-compatible), store bound handlers as instance properties, and make cleanup idempotent to prevent dangling refs.
## Root Cause
Ghost windows and memory growth could be contributed to by ZenWorkspaces adding window event listeners without removal, plus incorrect RefPtr handling in stylesheet cache. These fixes target identified Zen-specific issues.
## Testing
- Code compiles and passes syntax checks
- Follows Mozilla patterns for event listener management
- No regressions in existing functionality
- Memory reports should be monitored post-fix to confirm impact
Addresses #11721

PR #12285, opened 2026-02-07
Fix memory leak in ZenWorkspaces by adding event listener cleanup

JorySeverijnse:
- Store bound event listeners as instance properties for later removal
- Add cleanup() method with (_cleanupDone flag)
- Remove all window listeners (TabOpen/Close/Select, resize, AppCommand, etc.)
- Remove workspace-bookmarks-updated observer
- Uses unload event
- Also fixes RefPtr assignment in ZenStyleSheetCache

### DIFF
--- a/src/zen/mods/ZenStyleSheetCache.cpp
+++ b/src/zen/mods/ZenStyleSheetCache.cpp
@@ -64,9 +64,9 @@ auto ZenStyleSheetCache::LoadSheetFile(nsIFile* aFile,
   if (MOZ_UNLIKELY(result.isErr())) {
     return;
   }
-  mModsSheet = result.unwrapOr(nullptr);
+  mModsSheet = result.unwrap();
 }
-  
+
 /* static */
 auto ZenStyleSheetCache::Singleton() -> ZenStyleSheetCache* {
   MOZ_ASSERT(NS_IsMainThread());
@@ -77,7 +77,7 @@ auto ZenStyleSheetCache::Singleton() -> ZenStyleSheetCache* {
 }
 
 nsresult ZenStyleSheetCache::RebuildModsStylesheets(const nsACString& aContents) {
-  // Re-parse the mods stylesheet. By doing so, we read 
+  // Re-parse the mods stylesheet. By doing so, we read
   // Once we have the data as a nsACString, we call ReparseSheet from the
   // StyleSheet class to re-parse the stylesheet.
   auto sheet = GetModsSheet();
@@ -93,6 +93,6 @@ nsresult ZenStyleSheetCache::RebuildModsStylesheets(const nsACString& aContents)
 }
 
 
-mozilla::StaticRefPtr<ZenStyleSheetCache> ZenStyleSheetCache::gZenModsCache; 
+mozilla::StaticRefPtr<ZenStyleSheetCache> ZenStyleSheetCache::gZenModsCache;
 
 } // namespace: zen
--- a/src/zen/mods/ZenStyleSheetCache.cpp
+++ b/src/zen/mods/ZenStyleSheetCache.cpp
@@ -64,7 +64,7 @@ auto ZenStyleSheetCache::LoadSheetFile(nsIFile* aFile,
   if (MOZ_UNLIKELY(result.isErr())) {
     return;
   }
-  mModsSheet = result.unwrap();
+  mModsSheet = result.unwrapOr(nullptr);
 }
 
 /* static */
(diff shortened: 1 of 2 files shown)


src/zen/mods/ZenStyleSheetCache.cpp
```diff
@@ -64,9 +64,9 @@ auto ZenStyleSheetCache::LoadSheetFile(nsIFile* aFile,
   if (MOZ_UNLIKELY(result.isErr())) {
     return;
   }
-  mModsSheet = result.unwrapOr(nullptr);
+  mModsSheet = result.unwrap();
```
mr-cheffy: Why make this change? We want it to be null

src/zen/workspaces/ZenWorkspaces.mjs
```diff
@@ -130,27 +130,29 @@ class nsZenWorkspaces {
...
-      window.addEventListener("unload", () => {
-        Services.obs.removeObserver(observerFunction, "workspace-bookmarks-updated");
-      });
+
+      Services.obs.addObserver(
+        this.workspaceBookmarksObserver,
+        "workspace-bookmarks-updated"
+      );
     }
-  }
```
mr-cheffy, line 153: Syntax error?

mr-cheffy (changes requested):

src/zen/workspaces/ZenWorkspaces.mjs
JorySeverijnse: Thanks for pointing that out

src/zen/mods/ZenStyleSheetCache.cpp
JorySeverijnse: Sorry, i just edited the code to the changes i made last time in https://github.com/zen-browser/desktop/pull/11935. I'll revert

JorySeverijnse:
Should be good now? Have also built the browser myself (from dev branch with my changes) without any issues

src/zen/workspaces/ZenWorkspaces.mjs
```diff
@@ -130,25 +130,28 @@ class nsZenWorkspaces {
...
 
     if (!this.privateWindowOrDisabled) {
-      const observerFunction = async () => {
+      this.workspaceBookmarksObserver = async () => {
         delete this._workspaceBookmarksCache;
         await this.workspaceBookmarks();
         this._invalidateBookmarkContainers();
       };
-      Services.obs.addObserver(observerFunction, "workspace-bookmarks-updated");
-      window.addEventListener("unload", () => {
```
mr-cheffy, line 149: I would keep this

src/zen/workspaces/ZenWorkspaces.mjs
```diff
@@ -951,16 +955,18 @@ class nsZenWorkspaces {
...
+      window.addEventListener("TabClose", this.tabUpdateListener);
+      window.addEventListener("TabAddedToEssentials", this.tabUpdateListener);
+      window.addEventListener("TabRemovedFromEssentials", this.tabUpdateListener);
+      window.addEventListener("TabPinned", this.tabUpdateListener);
+      window.addEventListener("TabUnpinned", this.tabUpdateListener);
+      window.addEventListener("aftercustomization", this.tabUpdateListener);
+      this.onLocationChangeBound = this.onLocationChange.bind(this);
+      window.addEventListener("TabSelect", this.onLocationChangeBound);
+      this.onTabBrowserInsertedBound = this.onTabBrowserInserted.bind(this);
+      window.addEventListener("TabBrowserInserted", this.onTabBrowserInsertedBound);
```
mr-cheffy: Instead of doing this, why not just add the unload method here? Instead of creating new properties

src/zen/workspaces/ZenWorkspaces.mjs
```diff
@@ -3166,6 +3172,40 @@ class nsZenWorkspaces {
       document.getElementById("cmd_closeWindow").doCommand();
     }
   }
+
+  cleanup() {
+    // Defensive: ensure cleanup runs only once
+    if (this._cleanupDone) {
```
mr-cheffy: On what cases would cleanup be called more than once? And why would it matter if it would be a no-op anyways?

mr-cheffy (changes requested):
Please run formatting

mr-cheffy:
Looks good! Please re-open with the git conflicts merged if you are still interested in contributing! Sorry for the delay

PR #12295, opened 2026-02-07
Add zen identification info (closes #12198)

### LINKED ISSUE #12198: Zen doesn't identify itself in the extensions API
### Preliminary Checks

### What happened?

Extensions need to be able to detect the type of browser they are running in, because different browsers work in different ways and need per-browser code to implement some features or work around quirks.  Zen Browser is more quirky than most, which makes this extra important... but it does not seem to identify itself.  This means extensions will break and won't be able to apply browser-specific fixes.

Almost anything would work, as long as it is consistent, relatively easy to discover, and provides sufficient info.  For example...

- Brave creates an object called `navigator.brave`, so detection is as easy as `isBrave = (typeof navigator.brave !== 'undefined');`.

- Maxthon creates a top-level object called `maxthon`.

- MS Edge registers itself as a brand in `navigator.userAgentData` and identifies itself in the user agent:
```js
let isEdge = / Edg\//.test(navigator.userAgent);
if (navigator.userAgentData?.brands) {
  isEdge = navigator.userAgentData.brands
    .some(b => b.brand === "Microsoft Edge");
}
```

- Firefox (and derivatives) are trivial to detect because they define `browser` instead of using `chrome` as the top-level API object... but if a derivative doesn't distinguish itself further, it may be mistaken for Firefox.

Simply setting `navigator.zenBrowser = true` in the API would help quite a bit, though to make it more robust, it should ideally provide more info than that... like `navigator.zenBrowser.version = [1, 18, 3, 2];` or `navigator.zenBrowser.version = '1.18.3b';`, to allow extensions to apply version-specific fixes in addition to browser-specific fixes.  Or perhaps something like `browser.zen.version`, and expose user config options so extensions can check on config settings they might need to support... `if (browser.zen.getConfig('zen.window-sync.enabled')) { // handle "window sync" mode quirks }`.

Zen's quirks break extensions, but that's not a bug... it's just a side effect of being different.  The bug is that Zen doesn't provide the info extensions need to be able to function correctly.

### Expected behavior

Browser identifies itself in the extension API.

### Actual behavior

Zen seems to be indistinguishable from Firefox, so extensions can't apply Zen-specific fixes.

### Version

1.18.3b

Linux (AppImage)

Extensions

TogiFerretFerret:
This pull request (attempts) to close #12198 properly. 
In addition, it adds branding to from-source builds to prevent issues when testing certain features on development builds.


surfer.json
```diff
@@ -45,6 +45,16 @@
           "repo": "zen-browser/desktop"
         }
       }
+    },
+    "unofficial": {
```
mr-cheffy: Revert this change

src/toolkit/components/extensions/parent/ext-runtime-js.patch
```diff
@@ -1,13 +1,13 @@
...
+index 0d7a3e505b6bd30548c6dda1504dd343a517b083..bf16027dba3022a9bf53935fb27377e3295e2ce1 100644
 --- a/toolkit/components/extensions/parent/ext-runtime.js
 +++ b/toolkit/components/extensions/parent/ext-runtime.js
 @@ -333,7 +333,7 @@ this.runtime = class extends ExtensionAPIPersistent {
  
          getBrowserInfo: function () {
            const { name, vendor, version, appBuildID } = Services.appinfo;
 -          const info = { name, vendor, version, buildID: appBuildID };
-+          const info = { name: 'Firefox', vendor, version: AppConstants.ZEN_FIREFOX_VERSION, buildID: appBuildID };
++          const info = { name: 'Firefox', vendor, version: AppConstants.ZEN_FIREFOX_VERSION, buildID: appBuildID, zen: {version: AppConstants.MOZ_APP_VERSION_DISPLAY, platform: AppConstants.platform} };
```
mr-cheffy: I wouldn't include the platform here

mr-cheffy (changes requested):

mr-cheffy (approved):

PR #12300, opened 2026-02-08
Use optional chaining and nullish coalescing in resetButton

rmdeuce:


mr-cheffy:
Why exactly? In what scenario is `parentElement` null?

mr-cheffy (approved):

rmdeuce:
> Why exactly? In what scenario is `parentElement` null?
hi, the parentElement can be null in several edge cases:
during DOM manipulation, when the element is temporarily detached 
and if the getter is called before connectedCallback completes or after disconnectedCallback

mr-cheffy:
And where do we temporarily detach the element and *use* resetButton's element at the same time?

PR #12322, opened 2026-02-10
fix: Restore current page URL when urlbar is emptied and blurred

franklegolasyoung:
## Summary
- When users click the address bar, clear all text, and click away to blur, the urlbar stays empty instead of showing the current page URL
- Added a `handleRevert()` call in `endLayoutExtend()` for the non-newtab path when `_untrimmedValue` is empty, restoring the current page URL on blur
- Partial input is still preserved on blur as before — only a fully empty urlbar triggers the restore

## Test plan
- [ ] Open any webpage, click the address bar, type some text, click away → text should be preserved (unchanged behavior)
- [ ] Open any webpage, click the address bar, delete all text so it's empty, click away → address bar should restore the current page URL (fixed)
- [ ] Open a new tab via the new tab button, type in the urlbar, click away → should behave as before (newtab mode unaffected)
- [ ] Click the address bar without typing anything, click away → should show URL as normal (unchanged behavior)

### DIFF
--- a/src/browser/components/urlbar/content/UrlbarInput-mjs.patch
+++ b/src/browser/components/urlbar/content/UrlbarInput-mjs.patch
@@ -1,5 +1,5 @@
 diff --git a/browser/components/urlbar/content/UrlbarInput.mjs b/browser/components/urlbar/content/UrlbarInput.mjs
-index edb0482f0bfb22c70a585b0770e5b0437983779e..545f65db4fe8b807b388d5552842cd9e56a72eb3 100644
+index edb0482f0bfb22c70a585b0770e5b0437983779e..109e9eb18b90c650caff1287caef04dce857e122 100644
 --- a/browser/components/urlbar/content/UrlbarInput.mjs
 +++ b/browser/components/urlbar/content/UrlbarInput.mjs
 @@ -68,6 +68,13 @@ const lazy = XPCOMUtils.declareLazy({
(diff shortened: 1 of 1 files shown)


mr-cheffy (approved):

franklegolasyoung:
Hi @mr-cheffy , I have just committed a new update. The previous CI failed because the patch file had incorrect hunk headers. When I added 2 new lines to the endLayoutExtend() hunk (changing it from +2627,24 to +2627,26), I didn't update the + side start line numbers in all 18 subsequent hunk headers. They were each off by 2. git apply strictly validates these offsets, so it reported corrupt patch at line 153.

The fix was straightforward: bump the + side start line number by 2 for every hunk after the modified one (e.g. +2675 → +2677, +2929 → +2931, etc.). The actual code change itself remains unchanged, with only the patch metadata corrected. Thanks.

mr-cheffy:
Im sorry, but is this patch AI generated? Where are these random numbers coming from? How are you creating the patch file?

franklegolasyoung:
Yes, AI did help in this. Actually the code change is just 3 lines in endLayoutExtend(), adding an else if branch to call handleRevert() when the urlbar is empty on blur. But since this project works with .patch files rather than source files directly, I edited the patch file manually, which caused the hunk header line numbers to be off and broke CI. Seems like it is still causing it. Should I try with run the project's import/apply command to apply patches to the actual Firefox source and then make the code change in the UrlbarInput.mjs code for fixes?

mr-cheffy:
`npm run export <file>` to create patch files

franklegolasyoung:
Hi @mr-cheffy , I have just force committed the new update. I've downloaded the firefox engine, ran surfer import, made the code change directly in UrlbarInput.mjs, and used surfer export to regenerate the patch file. All line numbers are now auto-generated by the tooling.

I've also built and tested locally, and the fix works as expected:
- Typing in the urlbar and clicking away → input is preserved (unchanged behavior)
- Clearing the urlbar completely and clicking away → current page URL is restored (the fix)
- New tab mode → unaffected

franklegolasyoung:
Hi @mr-cheffy  still pending approval

PR #12341, opened 2026-02-12
fix: Restore tab position within workspace on cross-workspace undo close tab

7ito:
When a tab is closed in one workspace and restored (Ctrl+Shift+T) from a different workspace, the tab was always appended to the bottom of the workspace's tab list instead of its original position/index.

This implements a fix that saves the workspace tab index at close time and uses it to insert the tab at the correct position during restore.

Changes:
- SessionStore.sys.mjs — compute and save zenWorkspaceTabIndex in maybeSaveClosedTab, pass it through in undoCloseTab
- tabbrowser.js — accept zenWorkspaceTabIndex in addTab, set it as a transient attribute on the tab element
- ZenWorkspaces.mjs — read the attribute in onTabBrowserInserted, use it in moveTabToWorkspace to insert at the correct position (excluding the empty tab from index calculations)

### DIFF
--- a/src/zen/workspaces/ZenWorkspaces.mjs
+++ b/src/zen/workspaces/ZenWorkspaces.mjs
@@ -1531,10 +1531,12 @@ class nsZenWorkspaces {
         ? workspaceContainer?.pinnedTabsContainer
         : workspaceContainer?.tabsContainer;
       if (container?.contains(tab)) {
+        delete tab._zenRestoreTPos;
         continue;
       }
 
       if (tab.hasAttribute("zen-essential")) {
+        delete tab._zenRestoreTPos;
         continue;
       }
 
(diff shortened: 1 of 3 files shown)


mr-cheffy:
I would rather figure out why the tab isn't positioned at the correct position in the first time. Maybe we should use the tab's "current" position before switching its space and tanslate it over?

src/zen/workspaces/ZenWorkspaces.mjs
```diff
@@ -1520,11 +1520,11 @@ class nsZenWorkspaces {
...
   }
 
-  moveTabToWorkspace(tab, workspaceID) {
-    return this.moveTabsToWorkspace([tab], workspaceID);
+  moveTabToWorkspace(tab, workspaceID, tabIndex) {
+    return this.moveTabsToWorkspace([tab], workspaceID, tabIndex);
   }
 
-  moveTabsToWorkspace(tabs, workspaceID) {
+  moveTabsToWorkspace(tabs, workspaceID, tabIndex) {
```
mr-cheffy: What I meant was that, we dont need this as a parameter. We can just use the same location the tab would be in before `gBrowser.zenHandleTabMove`. You could use, for example, `tab._tPos`

mr-cheffy (changes requested):

7ito:
To avoid adding a parameter to moveTabsToWorkspace, the restore flow now records a tab’s pre‑close position as a short‑lived _zenRestoreTPos, which is used to pick the insertion point in the workspace (normal tabs only). About the >= condition: once a tab closes, everything after it shifts up by one, so using >= lines up the pre‑close position with the right slot and avoids the off‑by‑one jump. I think this approach keeps the original order across workspaces without persisting extra state or changing method signatures now?

Thanks for the reviews!

PR #12349, opened 2026-02-12
Update UrlbarMuxerStandard-sys-mjs.patch

dontmindaditya:


mr-cheffy:
Why this change exactly? This makes the patch file become corrupted

dontmindaditya:
Hey i am really sorry about thats i was juss checking thats how its working
but in future i ll do valid contribution

PR #12394, opened 2026-02-15
fix: Replace hardcoded Firefox with Zen in error pages

### LINKED ISSUE #10527: Firefox-Zen browser name inconsistency in error page
### Preliminary Checks

### What happened?

In the error page, the browser is addressed both as Firefox and Zen.

It happens regardless of the language in which the browser is displayed.

### Expected behavior

Change the first line to say "Zen can't ..."

### Actual behavior

The first line says "Firefox can't..."

### Steps to reproduce

Open any non-existent URL to view the error page.

### Screenshots and videos

Image

### Version

1.16b

Windows - x64

Localization

```shell

```

abdushsk:
## Summary
- Adds a new patch for `browser/locales/en-US/chrome/overrides/appstrings.properties` that replaces all hardcoded "Firefox" references with "Zen" in browser error pages (e.g. connection failure, file not found, offline mode, proxy errors, etc.)
- Fixes 15 error message strings that incorrectly displayed "Firefox" instead of "Zen"

Closes #10527

## Test plan
- [ ] Open a non-existent URL (e.g. `https://thisdoesnotexist.invalid`) and verify the error page says "Zen" instead of "Firefox"
- [ ] Trigger an offline error (`about:preferences` -> uncheck network, then navigate) and verify it says "Zen is currently in offline mode"
- [ ] Check proxy error messages display "Zen" branding

### DIFF
--- a/src/browser/locales/en-US/chrome/overrides/appstrings-properties.patch
+++ b/src/browser/locales/en-US/chrome/overrides/appstrings-properties.patch
@@ -1,0 +1,55 @@
+diff --git a/browser/locales/en-US/chrome/overrides/appstrings.properties b/browser/locales/en-US/chrome/overrides/appstrings.properties
+--- a/browser/locales/en-US/chrome/overrides/appstrings.properties
++++ b/browser/locales/en-US/chrome/overrides/appstrings.properties
+@@ -4,10 +4,10 @@
+
+ malformedURI2=Please check that the URL is correct and try again.
+-fileNotFound=Firefox can't find the file at %S.
++fileNotFound=Zen can't find the file at %S.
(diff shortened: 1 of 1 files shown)


mr-cheffy:
This has been discussed previously. And this fix is just some sort of way to avoid the core issue and that is that firefox doesn't take the brand name for these strings. These only work for en-US anyways so it also doesn't fix the issue for other languages. 

Thanks for attempting a fix though! But we are not planning to fix this as its mostly an issue with how firefox does translations. They are currently trying to move from these strings to fluent, so at some point it might be fixed by simply syncing from upstream.

abdushsk:
Thanks for the context! I'll look into other issues to contribute to.

PR #12395, opened 2026-02-15
fix: Merge duplicate fullscreen shortcuts into single toggle

### LINKED ISSUE #12237: Duplicated Toggle Full Screen and View Full Screen for keyboard shortcut
### Preliminary Checks

### What happened?

The two shortcut is separate, one is for entering and another exiting fullscreen

### Expected behavior

One shortcut for both entering and exiting Fullscreen

### Actual behavior

When changing shortcut for Toggle Fullscreen, i can only use that shortcut for either entering OR exiting Fullscreen. I am also unable to set both shortcut to the same key.

### Steps to reproduce

Settings page at shortcut keyboard

### Screenshots and videos

Image

### Version

1.18.4b

macOS - aarch64

Keyboard Shortcuts

```shell

```

abdushsk:
## Summary
- Removes the separate "Exit Full Screen" shortcut entry from keyboard settings, since `key_enterFullScreen` already toggles fullscreen (both enter and exit via `View:FullScreen` command)
- Renames "Enter Full Screen" label to "Toggle Full Screen" to accurately reflect its behavior
- Changes across 3 files:
  - `src/zen/kbs/ZenKeyboardShortcuts.mjs` — removed `zen-key-exit-full-screen` from `windowAndTabManagement` group
  - `src/browser/components/preferences/zen-settings.js` — removed `key_exitFullScreen` from l10n map
  - `locales/en-US/.../zen-preferences.ftl` — renamed label and removed exit string

Closes #12237

## Test plan
- [ ] Open Settings > Keyboard Shortcuts > Window & Tab Management
- [ ] Verify only one fullscreen shortcut appears ("Toggle Full Screen") instead of two
- [ ] Verify the shortcut correctly toggles fullscreen both in and out
- [ ] Verify editing the shortcut works as expected

### DIFF
--- a/src/browser/components/preferences/zen-settings.js
+++ b/src/browser/components/preferences/zen-settings.js
@@ -818,6 +818,13 @@ const zenMissingKeyboardShortcutL10n = {
   key_accessibility: "zen-devtools-toggle-accessibility-shortcut",
 };
 
+var zenIgnoreKeyboardShortcutIDs = [
+  "key_enterFullScreen_old",
+  "key_enterFullScreen_compat",
+  "key_exitFullScreen_old",
+  "key_exitFullScreen_compat",
+];
+
 var zenIgnoreKeyboardShortcutL10n = [
   "zen-full-zoom-reduce-shortcut-alt-b",
   "zen-full-zoom-reduce-shortcut-alt-a",
@@ -888,7 +895,11 @@ var gZenCKSSettings = {
 
       const labelValue = zenMissingKeyboardShortcutL10n[keyID] ?? l10nID;
 
-      if (zenIgnoreKeyboardShortcutL10n.includes(labelValue) || shortcut.shouldBeEmpty) {
+      if (
+        zenIgnoreKeyboardShortcutIDs.includes(keyID) ||
+        zenIgnoreKeyboardShortcutL10n.includes(labelValue) ||
+        shortcut.shouldBeEmpty
+      ) {
         continue;
       }
 


mr-cheffy:
Is changing for the "enter fullscreen" shortcut also changing the "exit fullscreen" shortcut?

abdushsk:
No — only the primary `key_enterFullScreen` shortcut is editable. The exit variant (`key_exitFullScreen`) is disabled by Firefox by default and exists only as a reserved key. The `View:FullScreen` command that `key_enterFullScreen` maps to already handles both entering and exiting fullscreen, so a single editable entry is sufficient.

Also updated the PR to handle all 6 macOS fullscreen key variants (enter/exit + _old + _compat) — they're now all hidden except the primary one, which shows as "Toggle Full Screen".

abdushsk:
Actually, to clarify my earlier response — changing the "Toggle Full Screen" shortcut only updates `key_enterFullScreen`. It won't automatically sync to `key_exitFullScreen`.

That said, this shouldn't cause issues in practice since `View:FullScreen` is a toggle command, so the enter shortcut handles both entering and exiting. The exit key variants exist as a reserved (`reserved="true"`) safety net that Firefox enables during fullscreen to guarantee an exit even if a web page blocks keyboard input.

The only edge case would be: if a user changes the toggle shortcut, the reserved fallback still uses the original keybinding. But that's also how it worked before this change — the two were always independent entries. This PR just cleans up the UI to avoid showing redundant/confusing duplicates.

If syncing them is something you'd want, happy to look into that — it'd need changes in the shortcut manager to link the two keys together.

mr-cheffy:
> No — only the primary key_enterFullScreen shortcut is editable. The exit variant (key_exitFullScreen) is disabled by Firefox by default and exists only as a reserved key. The View:FullScreen command that key_enterFullScreen maps to already handles both entering and exiting fullscreen, so a single editable entry is sufficient.

True, wouldn't it make more sense to ALSO simply not display any `reserved` keys on the list instead of filtering out by individual IDs? `isUserEditable` doesnt check for reserved, im not sure what "reserved" would mean in this context

abdushsk:
That makes sense — I'll add a `this.#reserved` check to `isUserEditable()` to filter those out generically.

One thing though: on macOS there are also `key_enterFullScreen_old` (⌘⇧F) and `key_enterFullScreen_compat` (F11) which are NOT reserved, just duplicate enter bindings. Those still need the ID-based filtering to hide. Want me to handle both in this PR, or just the reserved approach for now?

mr-cheffy:
Sure, you should filter those out as well, thanks!

abdushsk:
Updated — added a `this.#reserved` check in `isUserEditable()` to generically filter out reserved keys (covers all exit fullscreen variants). The enter `_old` and `_compat` duplicates are still handled via the ignore list since they're not reserved. Should be good now!

mr-cheffy:
I was just checking firefox's source code, and it seems like exit fullscreen IS being used, https://searchfox.org/firefox-main/source/browser/base/content/browser-fullScreenAndPointerLock.js#637-654.

So maybe we shouldn't continue with this change. I agree on removing the compat ones though. Also, could you please run `npm run lint:fix`? Thanks

abdushsk:
Ah my bad, should've checked Firefox's source more carefully. Reverted the exit fullscreen changes — the updated commit now only filters out the `_old` and `_compat` duplicate variants (both enter and exit), and keeps Enter/Exit Full Screen visible as separate entries. Also fixed the prettier issue, should be clean now.

src/browser/components/preferences/zen-settings.js
```diff
@@ -821,6 +825,10 @@ const zenMissingKeyboardShortcutL10n = {
 var zenIgnoreKeyboardShortcutL10n = [
```
mr-cheffy, line 828: Maybe it would be better to add a `zenIgnoreKeyboardShortcutIDs` variable

mr-cheffy (changes requested):

src/browser/components/preferences/zen-settings.js
abdushsk: Done — added a `zenIgnoreKeyboardShortcutIDs` array that filters by element ID directly. Cleaner than going through the l10n mapping.

abdushsk:
Updated — let me know if this is what you had in mind!

mr-cheffy (approved):

PR #12383, opened 2026-02-15
feat: [UI] Modern Ctrl Tab Preview

gholts:
Related history: [zen-browser/theme-store/pull/1912](https://github.com/zen-browser/theme-store/pull/1912) and the previous [Reddit thread](https://www.reddit.com/r/zen_browser/comments/1r4nzc1/comment/o5fg998/).

I have optimized the Ctrl+Tab Preview UI for Zen Browser. This update improves the display of the favicon background and includes full support for both light and dark modes.

image
image

PR #12397, opened 2026-02-15
feat [UI]: Refined ctrl+tab panel

gordonn:
This PR aims to implement an Arc-like ctrl+tab panel with ability to cycle through entire tab list.

https://github.com/user-attachments/assets/01158994-cf12-4392-bfe1-85adfdbde72b

image

- [x] Ability to cycle through tab list from top to bottom like standard Ctrl+Tab behavior
- [x] Add about:config pref to cycle through tabs by recently used order like Arc and alt-tabbing
- [ ] Remove default Firefox Ctrl+Tab panel and migrate preferences to settings


12th-devs:
@gordonn I figured I would share this pr also being developed. https://github.com/zen-browser/desktop/pull/12383

src/zen/tabs/zen-tab-switcher.css
```diff
@@ -0,0 +1,152 @@
...
+  padding-right: 1px;
+  padding-bottom: 1px;
+  scroll-behavior: smooth;
+  scrollbar-width: none; 
+  position: relative;
+  box-sizing: border-box;
+  flex-shrink: 0;
+}
+
+#zen-tab-switcher-tabs::-webkit-scrollbar {
```
FlorianButz: Why `-webkit-scrollbar`? It's not supported in Firefox, is it?

src/zen/tabs/ZenTabSwitcher.mjs
```diff
@@ -0,0 +1,525 @@
...
+    if (!tab || tab.closing || tab.hidden) return;
+
+    const index = this.#recentlyUsedTabs.indexOf(tab);
+    if (index !== -1) {
+      this.#recentlyUsedTabs.splice(index, 1);
+    }
+
+    this.#recentlyUsedTabs.unshift(tab);
+
+    if (this.#recentlyUsedTabs.length > 50) {
```
FlorianButz: Why 50? Could this be a constant instead, or explained with a comment?

src/zen/tabs/ZenTabSwitcher.mjs
```diff
@@ -0,0 +1,525 @@
...
+    } else {
+      const currentTabIndex = this.#tabList.indexOf(gBrowser.selectedTab);
+      this.#currentIndex = currentTabIndex >= 0 ? currentTabIndex : 0;
+    }
+
+    await this.#preCacheThumbnailsForVisible();
+
+    this.#renderTabs();
+
+    // Center panel on screen
```
FlorianButz: Shouldn't it be centered in the browser?

src/zen/tabs/ZenTabSwitcher.mjs
```diff
@@ -0,0 +1,525 @@
...
+    if (this.#thumbnailCache.has(tabId)) return;
+    const browser = tab.linkedBrowser;
+    if (!browser) return;
+
+    try {
+      const canvas = document.createElementNS("http://www.w3.org/1999/xhtml", "canvas");
+      canvas.width = 320;
+      canvas.height = 180;
+
+      const { PageThumbs } = ChromeUtils.importESModule(
```
FlorianButz: This could be lazily imported at the top, no?
The function is called quite often I assume

src/zen/tabs/ZenTabSwitcher.mjs
```diff
@@ -0,0 +1,525 @@
...
+      }
+
+      this.tabsContainer.appendChild(tabCard);
+    });
+
+    this.#applySelectionTextStyles();
+  }
+
+  // Creates a single tab card element with thumbnail, favicon, and title
+  // Returns: XUL <vbox> element representing the tab card
```
FlorianButz: You should be using block comments for function documentation and the proper `@returns` attribute

FlorianButz (changes requested):

src/zen/tabs/ZenTabSwitcher.mjs
gordonn: I just picked a arbitrary value but yeah your right. I set it as a constant variable.

src/zen/tabs/ZenTabSwitcher.mjs
gordonn: Cheff wanted it to be centered on the screen and I think that's the better approach since it doesn't rely on the browser window being a certain width to be able to fit the panel.

CosmoCreeper:
A **whole lot** of commits for something that could be as simple as a few tweaks to the existing ctrl+tab panel in my opinion. Not sure why there's a whole separate ctrl+tab panel component for these changes but something tells me it's AI's fault (the markdown file in the first commit).

gordonn:
> A **whole lot** of commits for something that could be as simple as a few tweaks to the existing ctrl+tab panel in my opinion. Not sure why there's a whole separate ctrl+tab panel component for these changes but something tells me it's AI's fault (the markdown file in the first commit).

It's quite different from Firefox's panel implementation and has extra functionality such as scrolling throught entire tab list in order, seeing unloaded tabs, better thumbnail updating and utilizing native panel styling for each OS. Maybe review the most recent files instead of the initial commits which were vibe coded because I just was just prototyping. Since them I have tested it thoroughly and optimized it by hand coding it myself to remove any mistakes done by AI.

CosmoCreeper:
Okay, no problem, but I just wasn't sure why these couldn't just be improvements to the initial Firefox ctrl+tab panel. Like generic tweaks to the panel that would be beneficial regardless of the browser could be packaged in a merge request to the initial Firefox code, while the styling for Zen could just be applied here.

Only reason I say this is cause the improvements you're mentioning sound like they could at least partially be pushed into Firefox's source.

PR #12437, opened 2026-02-20
fix: reset the search mode after closing the urlbar

### LINKED ISSUE #8704: Cmd + K (Focus Search) still active after closing it
### Preliminary Checks

### What happened?

After pressing Cmd + K and then closing it, it seems Focus Search is still active. This can easily be seen in the URL bar in the sidebar. The user then has to manually deactivate it in the URL bar, if not, Focus Search will open again when the user attempts to open a new tab (Cmd + T)! But interestingly, pressing Cmd + T twice deactivates Focus Search.

### Expected behavior

After pressing Cmd + K and then closing it, Focus Search should be closed for good. Should not be present in URL bar, and not should not open again when pressing Cmd + T

### Actual behavior

Focus Search is still active after closing it, which tampers with the URL bar and creation of new tabs

### Steps to reproduce

1. Press Cmd + K (Focus Search)
2. Either press Escape or click outside to close Focus Search
3. Open Sidebar and check URL bar. Focus Search is still active, even though it was closed
4. Press Cmd + T to open new tab
5. Observe that Focus Search was opened again
6. Do step 4 again, but this time, observe that Focus Search is no longer active

### Screenshots and videos

If not able to reproduce, I can provide a screen recording. But steps should be easy to follow

### Version

1.12.8b

macOS - aarch64

Keyboard Shortcuts

```shell

```

kaedeng:
Search mode was persisting the state between separate uses of the urlbar (ex. using `cmd+k`'s and typing anything persisted the search mode onto `cmd+t`\\`cmd+l`'s urlbar).

This fix resets the search mode every time the urlbar closes 

Fixes #8704


mr-cheffy:
I think it would be better to modify https://github.com/zen-browser/desktop/blob/a8e245b28a01edf352619b25278224211ddca567/src/browser/components/urlbar/content/UrlbarInput-mjs.patch#L133 to check for the search mode?

kaedeng:
> I think it would be better to modify
> 
> https://github.com/zen-browser/desktop/blob/a8e245b28a01edf352619b25278224211ddca567/src/browser/components/urlbar/content/UrlbarInput-mjs.patch#L133
> to check for the search mode?

Done, lmk if I misunderstood the comment

mr-cheffy:
Great! Thanks

mr-cheffy:
Maybe on a next extent, we could start reverting even when there's no search mode?

Issue #12439, opened 2026-02-20
Sidebar should not hide when mouse pointer is still at the edge of the screen


gauravkdne:
### Preliminary Checks

- [x] I have read and understood the important section above.
- [x] I have searched existing issues and avoided creating duplicates.
- [x] I am not filing an enhancement request.
- [x] I have checked that this issue cannot be reproduced on Mozilla Firefox.
- [x] I have checked that this issue can be reproduced once I removed all my Mods and Custom CSS.

### What happened?

Sidebar hides even if the mouse pointer is still at the edge of the screen.

This does not happen in fullscreen mode.

### Expected behavior

When the mouse pointer is kept at the edge of the screen, the sidebar should stay open as long as the pointer is kept at the edge.

### Actual behavior

Sidebar hides even if the pointer is still kept at the edge of the screen.

### Steps to reproduce

1. Keep Zen in the regular windowed mode. (NOT Fullscreen)
2. Move the mouse pointer to the left edge of the screen until you cannot move further.
3. Now, the sidebar will show up. — This is expected behavior.
4. **Problem**: Now, the sidebar will automatically hide even if the pointer is still at the edge of the screen.
5. Sidebar should stay open as long as the pointer is at the left edge of the screen.

### Version

1.18.9b

### What platform are you seeing the problem on?

macOS - aarch64

### What component is this issue related to?

Other

### Relevant log output if applicable

```shell

```

mr-cheffy:
This has been discussed before, not sure at what point it becomes annoying

gauravkdne:
I keep Sidebar hidden all the time. To launch the sidebar, I flick the mouse pointer all the way to the edge of the screen. Now, I keep the pointer parked there and expect the sidebar in a visible state so that I can read the tabs. If the sidebar is hidden then I have to again precisely align the pointer in a presumed area of sidebar so that it stays visible. 
So, two actions are required when one action is assumed by muscle memory — 
1. Flick the pointer to the edge. (Sidebar shows briefly but hides again) After this -> 2. Re-align the pointer carefully so the sidebar is visible as long as the pointer is within the sidebar's area.

This issue seems trivial when it's consciously tested few times. But, over the time it gets annoying for the user because the muscle memory expects the sidebar being visible when pointer is at the edge of screen. The repeated adjustment of pointer makes one think 'what happened, what's wrong, where's the sidebar'.

[...]

PR #12444, opened 2026-02-20
Fix compact sidebar auto-hide when cursor is at the screen edge

elijahfriedman:
Fixes issue #12439

This fixes a compact mode regression where the sidebar hides even while the cursor remains pinned to the screen edge in non-fullscreen windows. 

In windowed mode, moving the cursor to the left edge correctly reveals the sidebar, but it auto-hides after the hover timeout even if the cursor has not left the edge. 

The mouseleave flow always started a timed hide (flashElement) and relied on a subsequent mousemove to keep/clear state. At a physical screen edge, no additional mousemove may fire, so the sidebar collapsed incorrectly.

Updated ZenCompactMode.mjs to:
- Detect when the pointer is still at the relevant physical screen edge.
- Keep zen-has-hover active while pinned to that edge.
- Collapse only after pointer movement indicates it left that edge.
- Add listener cleanup to avoid stale edge-tracking handlers.

### DIFF
--- a/src/zen/compact-mode/ZenCompactMode.mjs
+++ b/src/zen/compact-mode/ZenCompactMode.mjs
@@ -42,6 +42,7 @@ window.gZenCompactModeManager = {
   _flashTimeouts: {},
   _eventListeners: [],
   _removeHoverFrames: {},
+  _edgeHoverMoveListeners: {},
 
   // Delay to avoid flickering when hovering over the sidebar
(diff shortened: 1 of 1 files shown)


mr-cheffy:
Sorry, but have you tested this? How is it supposed to work? It doesn't seem to fix the issue to me

mr-cheffy:
Sorry, but this doesn't seem to be working at all, let alone tested. Almost like if it where AI generated. Please do let me know if im wrong so I can re-open. Thanks!

PR #12465, opened 2026-02-22
feat: Smart tab hibernation for inactive background tabs

AndreNijman:
This adds an MVP smart tab hibernation feature that proactively unloads inactive background tabs using Zen’s existing unload path. Tabs are tracked for inactivity and the oldest eligible background tab is discarded on a periodic timer when the feature is enabled.

The implementation reuses gBrowser.explicitUnloadTabs to preserve Firefox session behavior and avoids introducing a new tab termination path. Eligibility checks exclude selected tabs, pinned tabs, audio-playing tabs, and PiP tabs. The feature is controlled via about:config using zen.tabs.smart_hibernate.enabled and zen.tabs.smart_hibernate.timeout_minutes.

Manual testing confirms that inactive background tabs unload after the configured timeout, excluded tabs remain active, and discarded tabs restore normally when selected.

### DIFF
--- a/locales/en-US/browser/browser/preferences/zen-preferences.ftl
+++ b/locales/en-US/browser/browser/preferences/zen-preferences.ftl
@@ -56,6 +56,20 @@ pane-settings-workspaces-title = Workspaces
 zen-tabs-select-recently-used-on-close =
     .label = When closing a tab, switch to the most recently used tab instead of the next tab
 
+zen-tabs-smart-hibernate-enabled =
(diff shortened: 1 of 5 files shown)


mr-cheffy:
Sorry for not accepting the patch, but we are currently using Firefox's existing tab unload method intentionally as it works based on the system pressure instead of "monkey-unloading" tabs based on a timer. Thanks for trying to contribute though!

PR #12469, opened 2026-02-22
Fix macOS split-view unsplit shortcut conflict with View Source

### LINKED ISSUE #11568: [macOS] Default Unsplit shortcut (Opt+Cmd+U) conflicts with native 'View Source' binding
### Preliminary Checks

### What happened?

On macOS, when using the default keyboard shortcut `Opt + Cmd + U` to "Unsplit" a tab from Split View, the browser ignores the `Opt` modifier and instead triggers the underlying Firefox engine's default command for `Cmd + U` (View Page Source). 

Instead of returning the tab to a single view, a new tab opens displaying the HTML source code of the active page. This effectively makes the "Unsplit" action inaccessible via the keyboard for macOS users unless the keybinding is manually changed to avoid the letter 'U'.

### Expected behavior

Pressing Opt+Cmd+U should trigger the "Unsplit Tab" function, removing the split view overlay and expanding the current tab to full width, without opening the "View Source" window.

### Actual behavior

The "View Source" window opens (behavior consistent with Cmd+U), and the split view remains active. The `Opt` modifier appears to be masked or ignored by the event listener.

### Steps to reproduce

1. Open Zen Browser on macOS.
2. Enter Split View on any tab (e.g., press `Opt + Cmd + V`).
3. Ensure the split view is active/focused.
4. Press the default shortcut for Unsplit: `Opt + Cmd + U`.
5. Observe that a new tab opens with the Page Source, and the original tab remains split.

### Screenshots and videos

_No response_

### Version

1.17.8b 

macOS - aarch64

Split View

```shell

```

AndreNijman:
This patch:

- Changes the macOS default unsplit binding to `Cmd+Opt+Shift+U`
- Adds keyboard-shortcut migration v17 so existing profiles with the old default are updated
- Preserves user-customized shortcuts (only migrates unchanged old default)

Closes zen-browser/desktop#11568

### DIFF
--- a/src/zen/kbs/ZenKeyboardShortcuts.mjs
+++ b/src/zen/kbs/ZenKeyboardShortcuts.mjs
@@ -772,7 +772,11 @@ class nsZenKeyboardShortcutsLoader {
         "U",
         "",
         ZEN_SPLIT_VIEW_SHORTCUTS_GROUP,
-        nsKeyShortcutModifiers.fromObject({ accel: true, alt: true }),
+        nsKeyShortcutModifiers.fromObject(
+          AppConstants.platform == "macosx"
+            ? { accel: true, alt: true, shift: true }
+            : { accel: true, alt: true }
+        ),
         "cmd_zenSplitViewUnsplit",
         "zen-split-view-shortcut-unsplit"
       )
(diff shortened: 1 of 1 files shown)


mr-cheffy (changes requested):
I think it would be better to change view source's keyboard shortcut as it is less used maybe?

AndreNijman:
I reverted the unsplit shortcut change and instead moved the macOS View Source alias shortcut (`key_viewSourceSafari`) from Cmd+Opt+U to Cmd+Opt+Shift+U. `zen-split-view-unsplit` now stays on Cmd+Opt+U, and the v17 migration was adjusted to migrate unchanged `key_viewSourceSafari` defaults accordingly.

PR #12476, opened 2026-02-22
fix: resolve copy URL delay during pending navigations

### LINKED ISSUE #11835: Copy URL shortcut does not copy URL before page has loaded
### Preliminary Checks

### What happened?

When you use the copy URL shortcut, if the latest page has not loaded, it will copy the previous URL or about:blank (if its a new tab)

Note I selected Windows, but this issue is present on MacOS as well

### Expected behavior

Shortcut copies what is in the URL bar

### Actual behavior

Shortcut copies the page from the page before

### Steps to reproduce

Option 1:  
1. Open a slow loading page in a new tab
2. Hit the copy URL shortcut before the page has loaded
3. Paste the URL somewhere else or inspect clipboard
4. The value will be about:blank

Option 2:
1. Open a slow loading page in the same tab
2. Hit the copy URL shortcut before the new page has loaded
3. Paste the URL somewhere else or inspect clipboard
4. The value will be the page you were on before, not the new page

### Screenshots and videos

https://github.com/user-attachments/assets/479d807d-71d9-47d5-8987-a536073f8d05

### Version

1.17.15b

Windows - x64

Keyboard Shortcuts

```shell

```

phannawich:
Closed #11835 

# Summary

Fixes an issue where the "Copy URL" (`Ctrl+Shift+C`) and "Copy URL as Markdown" shortcuts copy the outdated/previous URL if pressed while a new page is still loading.

# Problem

When a user clicks a link that takes a while to load, pressing the Copy URL shortcut historically grabbed the URL of the currently fully loaded website or redirect to about:blank rather than the pending navigation. As a result, users expecting to quickly copy the link they just clicked would end up with the wrong URL in their clipboard.

# Solution
Updated the logic in `ZenCommonUtils.mjs` to prioritize the user's immediate intent over the resolved document state.

The new logic determines the correct URL in the following order:

1. **SPA & Click Interception:** Added a `passive: true` capture listener on `tabbrowser-tabpanels` that briefly caches `XULBrowserWindow.overLink` upon a mouse click. If the user copies within 4 seconds of clicking, it instantly grabs this link (perfectly bypassing SPA background-loading delays).
2. **Pending Engine State:** Checks `gBrowser.selectedBrowser.currentURI` to capture active browser navigations that haven't fully resolved into a loaded document yet. 
3. **Address Bar UI State:** Checks `gURLBar.value` (and automatically re-appends `https://` if visually trimmed) to guarantee the copied URL matches exactly what the user currently sees in their address bar. 
4. **Original Fallback:** If all the above are empty, it gracefully falls back to the original `currentUrl.displaySpec`.

Both `copyCurrentURLToClipboard` and `copyCurrentURLAsMarkdownToClipboard` have been updated to use this new logic.

### DIFF
--- a/src/zen/common/modules/ZenCommonUtils.mjs
+++ b/src/zen/common/modules/ZenCommonUtils.mjs
@@ -85,7 +85,29 @@ export class nsZenPreloadedFeature {
 window.gZenCommonActions = {
   copyCurrentURLToClipboard() {
     const [currentUrl, ClipboardHelper] = gURLBar.zenStrippedURI;
-    const displaySpec = currentUrl.displaySpec;
+
+    let clickedUrl = "";
+    if (window._zenPendingClick && Date.now() - window._zenPendingClick.time < 4000) {
+      clickedUrl = window._zenPendingClick.url;
+      window._zenPendingClick = null;
+    }
+
+    const activeBrowser = gBrowser.selectedBrowser;
+    const engineUrl = activeBrowser.currentURI ? activeBrowser.currentURI.displaySpec : "";
+
+    let visibleUrl = gURLBar.value;
+    if (
+      visibleUrl &&
+      !visibleUrl.startsWith("http") &&
+      !visibleUrl.startsWith("about:") &&
+      !visibleUrl.startsWith("file:") &&
+      !visibleUrl.startsWith("zen:")
+    ) {
+      visibleUrl = "https://" + visibleUrl;
+    }
+
+    const displaySpec = clickedUrl || engineUrl || visibleUrl || currentUrl.displaySpec;
+
     ClipboardHelper.copyString(displaySpec);
     let button;
     /* eslint-disable mozilla/valid-services */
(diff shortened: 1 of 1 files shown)


mr-cheffy (changes requested):
Not sure if this is the best way to do it. Wouldn't it make more sense to check where does the URLbar get the value from?

phannawich:
@mr-cheffy 

**The Problem**
In Single Page Applications (SPAs) like GitHub, link clicks are intercepted by JavaScript via preventDefault(). The browser’s native navigation engine remains unaware of the destination until the web app calls history.pushState(), which typically happens only after the background data fetch is complete.

**Current Behavior**
If a user triggers copyCurrentURLToClipboard during this loading state:
- `gURLBar.zenStrippedURI` reflects the originating page.
- `gBrowser.selectedBrowser.currentURI` reflects the originating page.
- `gURLBar.untrimmedValue` reflects the originating page.

Consequently, the clipboard receives the "old" URL, creating a perceived lag in the "Copy URL" shortcut.

I am looking for guidance on how should I approach this issue.
The idea that I have is intercept the `pushState` call itself so the clipboard updates the instant the website decides to change the URL, rather than waiting for the UI to catch up. Is this the way to go?

Please let me know how should I proceed with this or if there is a preferred Zen-specific pattern for handling these types of UI/Engine state mismatches.

mr-cheffy:
Unfortunately, there's virtually no way we can know what state we are going towards from this stage AFAIK, but this is not what the original issue is talking about. The original issue is talking about how it copies about:blank when copying a page and it hasn't finished loading

phannawich:
Oh OK, I think I missed interpret the point of the original issue.

From my understanding currently based on the original issue #11835 

- Option 1: return `about:blank` is a bug and need to be fixed
- Option 2: Show the original URL while the page still loading is intended behavior

Am I understand this correctly? if yes, I will investigate more on how to fix the Option 1 problem.

mr-cheffy:
Yes, option 1 is the underlying issue. Option 2 doesn't really matter because the page didn't even load in the first place

PR #12508, opened 2026-02-25
feat: Open URL lists from the urlbar in multiple tabs

salmonumbrella:
## Summary

- Detect when a user pastes or types a list of URLs/domains into the urlbar and open each in a separate tab
- Only allow `http` and `https` schemes — `file://`, `ftp://`, `javascript:`, `data:`, and `about:` are all blocked
- Preserve URLs with balanced parentheses (e.g. Wikipedia links)
- Cap at 20 tabs per paste with a confirmation dialog via `OpenInTabsUtils`
- Recognise common list formats: whitespace, commas, semicolons, numbered lists, lettered lists, bullet points

I am looking to solve accidentally pasting more than one URL, without adding bloat to Zen.


mr-cheffy:
Interesting! I wonder how annoying more than useful it could be though, i'll have a discussion about this internally

PR #12511, opened 2026-02-25
fix: Add guard to prevent invalid split drops

octaviusz:


src/zen/drag-and-drop/ZenDragAndDrop.js
```diff
@@ -683,6 +683,10 @@
 
       const dt = event.dataTransfer;
       const draggedTab = dt.mozGetDataAt(TAB_DROP_TYPE, 0);
+      if (!draggedTab) {
```
mr-cheffy: `!isTab(draggedTab)`

mr-cheffy (changes requested):

src/zen/drag-and-drop/ZenDragAndDrop.js
octaviusz: true

mr-cheffy (approved):

PR #12519, opened 2026-02-26
Folders for unpinned tabs

azizkoumenji:
Features added:
- New toggle in settings under tab management called "Automatically pin folders" so you can disable or enable having folders pinned by default when creating a new folder
- You can now create folders with unpinned tabs
- New option in the context menu of folders where you pin or unpin a folder


mr-cheffy:
Please rebase the branch as you seem to have picked some commits that have issues with the dev branch

PR #12521, opened 2026-02-26
gh-11835: resolve copy URL return about:blank during pending navigations

### LINKED ISSUE #11835: Copy URL shortcut does not copy URL before page has loaded

phannawich:
Close #11835 

# Summary

Fixes an issue where the "Copy URL" (Ctrl+Shift+C) and "Copy URL as Markdown" shortcuts return about:blank, if triggered while a new page or tab is still initializing.

# Problem
When a user navigates to a new page especially a link in a Single Page Application (SPA), there is a delay before the URL bar visually updates or the document fully resolves.

**The SPA Navigation Gap**

In SPAs like GitHub or YouTube, JavaScript intercepts clicks via `preventDefault()`. The browser's native `onLocationChange` event (which updates the URL bar) often doesn't fire until the background data fetch is nearly complete. During this 2–4 second window, the internal `currentURI` and the visual `gURLBar.value` still point to the previous page. While copying the "old" URL during an in-tab SPA transition is often considered intended behavior (based on discussion with mr-cheffy in #12476), the issue is critical when it results in about:blank.

**The about:blank Race Condition**

When opening links in new tabs, the background browser starts at `about:blank`. If the user triggers the "Copy URL" shortcut immediately, the shortcut grabs the resolved `currentUrl.displaySpec` before the network request has connected, resulting in a clipboard containing only `about:blank`.

**Discovery: Sidebar vs. URL Bar**

The resolution for this was found by observing Zen's sidebar. Even when the URL bar is empty and the page is "loading," the sidebar instantly displays the destination domain (e.g., youtube.com/). This confirms that the Tab (the UI layer) knows the destination immediately, even if the Browser Engine (the document layer) hasn't resolved it yet.

**Steps to Reproduce** (This is easier to reproduce `about:blank` problem than what shown in the original issue)
1. Go to YouTube or GitHub with opening new tab (e.g., `Ctrl + T`).
2. Ensure that the search actually point to Youtube or Github website and NOT `Search with Google`
3. Enter than website
4. Observation: The sidebar already shows the destination host such as `youtube.com/`, but the URL bar is still empty/loading. (Later, the sidebar will show YouTube instead)
5. Press `Ctrl + Shift + C`.
6. Result: The clipboard contains `about:blank` instead of the YouTube/GitHub link.

# Solution
The shortcut logic in `ZenCommonUtils.mjs` has been updated to use a strict native fallback cascade that prioritizes User Intent and Tab State over the document state:
1. `activeBrowser.userTypedValue`: Captures the exact URL for newly initialized tabs (preventing the `about:blank` issue) and explicitly typed navigations that haven't loaded yet.
2. `gURLBar.untrimmedValue`: Captures the native, unformatted string currently held by the URL bar.
3. `activeBrowser.currentURI.spec`: Captures the engine's immediate network destination. This updates the instant a link is clicked, matching the state the sidebar uses to display the pending host.
4. `currentUrl.displaySpec`: The original fallback for fully loaded, resolved documents.
5. Tab Label Failsafe: If the engine is queried so early that it returns `about:blank` or an empty string, it safely falls back to the `gBrowser.selectedTab.label` (provided it isn't "New Tab").


src/zen/common/modules/ZenCommonUtils.mjs
```diff
@@ -84,8 +84,22 @@ export class nsZenPreloadedFeature {
...
+    const activeBrowser = gBrowser.selectedBrowser;
     const [currentUrl, ClipboardHelper] = gURLBar.zenStrippedURI;
-    const displaySpec = currentUrl.displaySpec;
+    
+    let displaySpec = activeBrowser.userTypedValue || 
+                      gURLBar.untrimmedValue || 
+                      (activeBrowser.currentURI ? activeBrowser.currentURI.spec : "") || 
+                      currentUrl.displaySpec;
+
+    if (displaySpec === "about:blank" || displaySpec === "") {
```
mr-cheffy: Why would you copy the tab's label tough? This seems like a hacky solution to the real issue

mr-cheffy (changes requested):

src/zen/common/modules/ZenCommonUtils.mjs
phannawich: I found out the the tab's label get the label even before the URL bar and website finish loading.
And it can be used to put in the URL bar to search for the website

What do you recommend me to approach this problem?
1. Delete the if check block entirely and let it fall back to `about:blank` when the browser cannot grab any URL from all the case above?
2. Any ideas on your side?

mr-cheffy: Honestly, I have no idea yet... Let me investigate a bit and ill come back to you because this is a complex issue, even if it doesn't look like one

mr-cheffy: Might be worth to checkout what firefox is doing here:

https://searchfox.org/firefox-main/source/browser/components/customizableui/CustomizableWidgets.sys.mjs#524

phannawich:
Convert to draft, since it is still in the process of development and would like a help on patch file

phannawich:
Currently investigate [`UrlbarInput.mjs`](https://searchfox.org/firefox-main/source/browser/components/urlbar/content/UrlbarInput.mjs)

I put this code to help debug in `engine/browser/components/urlbar/content/UrlbarInput.mjs`
```javascript
get zenStrippedURI() {
    let activeBrowser = this.window.gBrowser.selectedBrowser;
    let tab = this.window.gBrowser.selectedTab;

    console.error("=== ZEN PROVISIONAL URL DEBUG ===");
    console.error("1. userTypedValue:", activeBrowser.userTypedValue);
    console.error("2. currentURI:", activeBrowser.currentURI ? activeBrowser.currentURI.spec : "null");
    console.error("3. documentURI:", activeBrowser.documentURI ? activeBrowser.documentURI.spec : "null");
    console.error("4. browser src attribute:", activeBrowser.getAttribute("src"));
    console.error("5. tab pendingurl:", tab.getAttribute("pendingurl"));
    console.error("6. tab label:", tab.label);
    console.error("7. gURLBar untrimmed:", this._untrimmedValue);
    console.error("=================================");

    // Fallback to the default behavior so the browser doesn't crash during the test
    let uri = activeBrowser.currentURI || Services.io.newURI("about:blank");
    return [uri, lazy.ClipboardHelper];
  }
```

This is the result I see when hit shortcut `Ctrl + C`
```bash
console.error: "=== ZEN PROVISIONAL URL DEBUG ==="
console.error: "1. userTypedValue:" null
console.error: "2. currentURI:" "about:blank"
console.error: "3. documentURI:" "null"
console.error: "4. browser src attribute:" null
console.error: "5. tab pendingurl:" null
console.error: "6. tab label:" "github.com/zen-browser/desktop"
console.error: "7. gURLBar untrimmed:" ""
console.error: "================================="
```

src/browser/components/urlbar/content/UrlbarInput-mjs.patch
```diff
@@ -92,9 +92,20 @@ index 2e6e2be9d7e28c3f189131ec19a26d552d13af99..101b5a3a70c24f28a755f2ca6630a6bc
 +
 +  get zenStrippedURI() {
 +    let strippedURI = null;
++    let activeBrowser = this.window.gBrowser.selectedBrowser;
++    let uriString = activeBrowser.userTypedValue || 
++                    (activeBrowser.currentURI ? activeBrowser.currentURI.spec : "");
++
++    // Convert the string back into a URI object so Firefox's strippers can read it
```
mr-cheffy: Please remove the comments and use `npm run export` so the patch files arent corrupted

mr-cheffy (changes requested):

src/browser/components/urlbar/content/UrlbarInput-mjs.patch
phannawich: I am struggling on try to make the patch file work. When I use `npm run export`, how do you fix it
```
>  npm run export

> zen-desktop@1.0.0 export
> surfer export

00:00:00 Dynamic config 'brand not set, defaulting to 'unofficial'
error: missing required argument 'file'
```

mr-cheffy: `npm run export <file>`

phannawich: From my understanding, 
1. I edit the file in Engine directory
2. Test if it works by running `npm run build` and `npm run start`
3. Edit the patch file by running `npm run export -- browser/components/urlbar/content/UrlbarInput.mjs`
4. Commit the patch file to the PR

mr-cheffy: yes

mr-cheffy (approved):

PR #12530, opened 2026-02-26
Folders for unpinned tabs feature

azizkoumenji:
Features added:
- New toggle in settings under tab management called "Automatically pin folders" so you can disable or enable having folders pinned by default when creating a new folder
- You can now create folders with unpinned tabs
- New option in the context menu of folders where you pin or unpin a folder

Features to be added: 
- "move tab/tabs to folder" context menu option

Bugs:

- Unpinned folders:
-- When the last tab of a folder is focused, and the user wants to duplicate the tab, the duplicated tab is created outside the folder
-- When selecting a tab inside a folder and positioned just before a subfolder (focused), when you use "new tab below" it creates the new tab at the end of the folder and not below the original tab
-- When selecting a tab inside a folder and positioned just before a subfolder (unfocused), both "new tab below" and "duplicate tab" create the tab at the end of the folder

- Pinned folders:
-- Using duplicate tab on a focused tab creates the new duplicated tab outside the folder and not below the original tab
-- When selecting a tab inside a folder and positioned just before a subfolder (unfocused or focused), "new tab below" will create a tab at the end of the folder and "duplicate tab" will create a tab outside the folder (exception: when the tab is unfocused, the duplicated tab is created at the end of the folder)

- Pinned tabs:
-- "new tab below" and "duplicate tab" on pinned tabs (outside folders) (focused or unfocused) create the new tab in the unpinned tabs section
-- Opening links inside pinned tabs (outside or inside a folder) opens the tab in the unpinned section (same thing for opening glanced links)

Some bugs might not be considered bugs (behaviour that some users expect) so let's have a discussion about that.

### DIFF
--- a/locales/en-US/browser/browser/preferences/zen-preferences.ftl
+++ b/locales/en-US/browser/browser/preferences/zen-preferences.ftl
@@ -66,6 +66,8 @@ zen-tabs-cycle-by-attribute =
   .label = Ctrl+Tab cycles within Essential or Workspace tabs only
 zen-tabs-cycle-ignore-pending-tabs =
   .label = Ignore Pending tabs when cycling with Ctrl+Tab
+zen-tabs-folders-auto-pin =
+  .label = Automatically pin folders
 zen-tabs-cycle-by-attribute-warning = Ctrl+Tab will cycle by recently used order, as it is enabled
 
 zen-look-and-feel-compact-toolbar-themed =
--- a/locales/en-US/browser/browser/zen-folders.ftl
+++ b/locales/en-US/browser/browser/zen-folders.ftl
@@ -8,6 +8,12 @@ zen-folders-search-placeholder =
 zen-folders-panel-rename-folder =
     .label = Rename Folder
 
+zen-folders-panel-pin-folder =
+    .label = Pin Folder
+
+zen-folders-panel-unpin-folder =
+    .label = Unpin Folder
+
 zen-folders-panel-unpack-folder =
     .label = Unpack Folder
 
--- a/prefs/zen/folders.yaml
+++ b/prefs/zen/folders.yaml
@@ -11,5 +11,8 @@
 - name: zen.folders.max-subfolders
   value: 5
 
+- name: zen.folders.auto-pin
+  value: true
+
 - name: zen.folders.owned-tabs-in-folder
   value: false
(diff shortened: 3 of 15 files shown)


mr-cheffy:
I dont think we should have unpinned folders, they shouldn't serve that purpose at least.

azizkoumenji:
I added the option of automatically pinning folders in settings, so the people who want to keep the old workflow (inspired by Arc Browser) can do that. I've seen many people suggest some kind of tab grouping for unpinned tabs, including myself. I personally think it's good to have the option to do this. Zen Browser doesn't have to be a one-to-one copy of Arc since different people have different workflows. Idk, maybe a poll will be  great to see if this is in demand or not

mr-cheffy:
Yes I agree, but it shouldn't be folders. Maybe a secondary type of groups that are meant for temporary actions, unlike folders which are meant for persistent organization.

azizkoumenji:
I can give it a try if you want. 

So, from what I understand, is that folders should stay as they are implemented now (basically tab groups in the pinned section of the sidebar only), and unpinned tabs should have a separate grouping mechanism with a different name (and maybe UI, even though I like how folders look and work, maybe we can keep the same look but different code).

Alternatively, without the need for creating "tab groups" for unpinned tabs, the functioning of pinned tabs (including the ones inside folders) could be modified in way that users won't need to have "unpinned folders" and by that I mean when selecting a tab/tabs in the pinned section (including inside folders) and choosing "new tab below" or "duplicate tab" or when opening links from the pinned tab or using glance the new tab/tabs will be created below the original tab (and inside the folder for tabs inside folders) and automatically be pinned instead of the current default behavior which creates the new tab in the unpinned section (a new toggle in settings should be enabled to get this behavior and by default it will be disabled to keep folders working as they are implemented now) what do you think? @mr-cheffy

mr-cheffy:
Yes, but I dont understand why you would want a toggle? Adding un-pinned groups would be a huge rework on zen's tab management system and I would have to make designs and all of that, which I simply don't have time to do that right now unfortunately.

But its certainly on the ToDo, but other things require more my attention. I rather unpinned groups are done by me, if thats ok. But thanks for the contribution though! its just that this is a pretty big change, not just in code, but for philosophies and it takes lots of time to plan out and I don't want you to keep waiting with this PR open

azizkoumenji:
I understand, and it makes sense for the main maintainer to have this feature added due to the big changes. I look forward to when the feature will be implemented! thanks

mr-cheffy:
Thank you for understanding! There are more ways to contribute though! Just ping me on github, reddit, discord, etc before doing it so this doesn't happen again 😅

Issue #12516, opened 2026-02-25
Repeated SIGSEGV in libxul.so on Linux (1.18.10b, Fedora 43, AMD)


georgehanlon:
### Preliminary Checks

- [x] I have read and understood the important section above.
- [x] I have searched existing issues and avoided creating duplicates.
- [x] I am not filing an enhancement request.
- [x] I have checked that this issue cannot be reproduced on Mozilla Firefox.
- [x] I have checked that this issue can be reproduced once I removed all my Mods and Custom CSS.

## Describe the bug

  Zen Browser 1.18.10b crashes repeatedly with SIGSEGV (signal 11) on Fedora 43. 19 crashes recorded over 2 days,
  often in clusters (e.g. 5 crashes within 4 minutes). Crashes occur in both the main browser process and content
  processes ("Isolated Web Co"), all originating inside `libxul.so`.

  The crashes began after the 1.18.10b update (build 20260220112217).

  ## Environment

  - **Zen version:** 1.18.10b (build 20260220112217, installed to `/opt/zen/`)
  - **OS:** Fedora 43 (kernel 6.18.12-200.fc43.x86_64)
  - **GPU:** AMD Radeon RX 6800 XT (Navi 21)
  - **Mesa:** 25.3.5-1.fc43.x86_64
  - **Display:** Wayland (KDE)
  - **Install method:** Native (not Flatpak)

  ## Crash data

  ### Crash frequency

  19 SIGSEGV crashes total:
  - **Feb 23:** 8 crashes (16:04 - 18:59)
  - **Feb 25:** 11 crashes (09:54 - 16:50)

  ## What I've ruled out

  - **Not a GPU driver issue** — Mesa 25.3.5 is the latest for Fedora 43, and no GPU errors (like
  `VK_ERROR_OUT_OF_HOST_MEMORY`) appear in logs
  - **Not a system-wide instability** — no other applications crash with similar patterns
  - **Timing** — crashes began after installing 1.18.10b (built Feb 20, 2026)

### Expected behavior

Browser and tabs should not crash.

### Actual behavior

Somtimes just a tab that crashes. Sometimes whole browser.

## Steps to reproduce

  No specific trigger identified — crashes appear to happen during normal browsing. They tend to cluster (multiple
   crashes within minutes of each other).

### Version

1.18.10b (build 20260220112217, installed to `/opt/zen/`)

### What platform are you seeing the problem on?

Linux (AppImage)

### What component is this issue related to?

Other

[...]

PR #12538, opened 2026-02-27
fix: Add null safety checks to JXL decoder DMS colour management

georgehanlon:
## Summary

  - Add missing null checks in the JXL decoder's `JXL_DEC_COLOR_ENCODING`
  handler to prevent potential crashes
  - All early exits now reset `mChannels` and `mFormat` to 4-channel RGBA to
  prevent downstream buffer over-reads in `JXL_DEC_FULL_IMAGE`
  - Fixes a null dereference when `qcms_profile_from_memory` returns null
  (malformed ICC profile)
  - Fixes a pre-existing issue where incompatible profile color space skip
  didn't reset channel format
  - Handles `malloc` failure for `mCMSLine` and `qcms_transform_create` failure
   gracefully

  Found while investigating #12516. These are defensive fixes for real null-safety bugs — whether or not they are the specific cause of #12516, they are correctness issues worth fixing. Note that upstream Mozilla has since abandoned this C++ JXL decoder in favor of a Rust rewrite (D119700), partly due to these classes of memory safety concerns.

  ## Test plan

  - [ ] Verify JXL images with valid ICC profiles still decode with correct
  colors
  - [ ] Verify JXL images with malformed/missing ICC profiles decode gracefully
   (without color management) rather than crashing
  - [ ] Verify grayscale JXL images decode without crashes

### DIFF
--- a/src/image/decoders/nsJXLDecoder-cpp.patch
+++ b/src/image/decoders/nsJXLDecoder-cpp.patch
@@ -144,13 +144,22 @@ index ffb7f3cd51e1d0e480bf8ac5e936a90c11f0c93d..378ae56d7fa99fbe8bdea2e689949994
 +                                               icc_profile.data(), size))
 +
 +        mInProfile = qcms_profile_from_memory((char*)icc_profile.data(), size);
++        if (!mInProfile) {
++          // Profile parsing failed. Fall back to RGBA output without color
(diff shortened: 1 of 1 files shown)


mr-cheffy:
Sorry, but we are planning on removing JXL support once firefox adds their own rust implementation. To fix your issue, you can disable JXL in about:config though!

PR #12541, opened 2026-02-27
feat: use `sidebar-keep-hover` delay while hiding sidebar

black-dragon74:
This patch introduces a new preference as `zen.view.compact.sidebar-keep-hover.duration`.

The pref is used as a timeout to decide when to hide the sidebar if hover is no longer active.

### DIFF
--- a/prefs/zen/compact-mode.yaml
+++ b/prefs/zen/compact-mode.yaml
@@ -17,6 +17,9 @@
 - name: zen.view.compact.toolbar-hide-after-hover.duration
   value: 1000
 
+- name: zen.view.compact.sidebar-keep-hover.duration
+  value: 150
+
 - name: zen.view.compact.animate-sidebar
   value: true
 
--- a/src/zen/compact-mode/ZenCompactMode.mjs
+++ b/src/zen/compact-mode/ZenCompactMode.mjs
@@ -589,7 +589,7 @@ window.gZenCompactModeManager = {
       {
         element: this.sidebar,
         screenEdge: this.sidebarIsOnRight ? "right" : "left",
-        keepHoverDuration: 100,
+        keepHoverDuration: Services.prefs.getIntPref("zen.view.compact.sidebar-keep-hover.duration"),
       },
       {
         element: document.getElementById("zen-appcontent-navbar-wrapper"),
--- a/src/zen/compact-mode/ZenCompactMode.mjs
+++ b/src/zen/compact-mode/ZenCompactMode.mjs
@@ -589,7 +589,9 @@ window.gZenCompactModeManager = {
       {
         element: this.sidebar,
         screenEdge: this.sidebarIsOnRight ? "right" : "left",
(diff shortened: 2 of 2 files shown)


mr-cheffy (approved):

src/zen/compact-mode/ZenCompactMode.mjs
```diff
@@ -589,7 +589,7 @@ window.gZenCompactModeManager = {
       {
         element: this.sidebar,
         screenEdge: this.sidebarIsOnRight ? "right" : "left",
-        keepHoverDuration: 100,
+        keepHoverDuration: this.hideAfterHoverDuration,
```
mr-cheffy: The default value of this is 1000 though? This would make the sidebar really slow to close. This pref serves a different purpose.

mr-cheffy (changes requested):

src/zen/compact-mode/ZenCompactMode.mjs
black-dragon74: Apart from hiding the sidebar when focus is out of the window?

What do you suggest? A lower default value or a new preference key? 1s is not slow but it ain't snappy either. 

IMHO, having a shared preference would be ideal as both would be dealing with similar intents(unless I missed something).

mr-cheffy: Im not sure why you are trying to change this value though? It doesn't does what you think it does

black-dragon74: keepHoverDuration does keep the sidebar open additionally for specified duration. 

Could you share its purpose if it's not like that?

We could bump 100ms to 500ms if we don't wanna depend on the preferences or create a new one.

mr-cheffy: I'd make it a new pref

src/zen/compact-mode/ZenCompactMode.mjs
```diff
@@ -584,12 +584,16 @@ window.gZenCompactModeManager = {
...
+    return Services.prefs.getIntPref("zen.view.compact.sidebar-keep-hover.duration")
+  },
+
   get hoverableElements() {
     return [
       {
         element: this.sidebar,
         screenEdge: this.sidebarIsOnRight ? "right" : "left",
-        keepHoverDuration: 100,
+        keepHoverDuration: this.sidebarKeepHoverDuration,
```
mr-cheffy: No need to make it a getter, can just be inline

prefs/zen/compact-mode.yaml
```diff
@@ -17,6 +17,9 @@
 - name: zen.view.compact.toolbar-hide-after-hover.duration
   value: 1000
 
+- name: zen.view.compact.sidebar-keep-hover.duration
+  value: 500
```
mr-cheffy: Lower to 150

mr-cheffy (changes requested):

mr-cheffy (approved):

PR #12589, opened 2026-03-02
Refactor folders

octaviusz:
I'm just farming Additions and Deletions


mr-cheffy:
How's it going? Whats left?

octaviusz:
- [x] Refactor element position restoration. Current `prevTab.after(tab)` is too brittle and depends on order; throws an exception if `prevTab` is still in `groupActiveTabsContainer`.
- [x] Investigate why elements from other spaces are visible on startup. (I might need to set the `overflow`).
- [ ] Implement animations for appearing tabs in LiveFolders.
- [x] Add collapse/expand animations for folder movement (`animateGroupMove`).
*   Tests, tests, and more tests.
*   Anything else I’ve inevitably forgotten.

PR #12640, opened 2026-03-05
fix: remove duplicate patches, p=#12640

Hythera:
## Motivation and Context
Some patches include duplicate content. This doesn't cause issues when using Surfer but breaks the build when applying the patches manually. 
I patched the source with and without the removed patches. I compared both with the `diff` command and they are both completely identical so this shouldn't break anything.

## Changes
- `src/toolkit/components/asyncshutdown/AsyncShutdown-sys-mjs.patch` and `src/external-patches/firefox/fix_macos_crash_on_shutdown_firefox_149.patch` are not the exact same but the second one contains all of the changes of the first one so I dropped the non auto generated one
- `src/browser/app/macbuild/Contents/Info-plist-in.patch` and `src/external-patches/firefox/no_liquid_glass_icon.patch` contain the exact same content so I removed the non auto generated one


mr-cheffy (approved):

mr-cheffy:
Might be good to add a guard to https://github.com/zen-browser/desktop/blob/dev/scripts/recalculate-patches.sh so it ignores these files

PR #12689, opened 2026-03-08
feat: add mpris volume support to zen on linux

Aethar01:
This pr implements volume control for the mpris interface on linux.

MPRISServiceHandler-volume.patch patches MPRISServiceHandler to add Volume and SetVolume to the DBus interface which then outputs zen-mpris-volume-changed to be seen by ZenMediaController.mjs which then sets the volume.

I also added the MPL blurb at the top of new files but i dont know if that was the right thing to do.

### DIFF
--- a/src/widget/gtk/MPRISServiceHandler-volume.patch
+++ b/src/widget/gtk/MPRISServiceHandler-volume.patch
@@ -1,0 +1,111 @@
+diff --git a/widget/gtk/MPRISServiceHandler.h b/widget/gtk/MPRISServiceHandler.h
+--- a/widget/gtk/MPRISServiceHandler.h
++++ b/widget/gtk/MPRISServiceHandler.h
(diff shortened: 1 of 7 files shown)


mr-cheffy:
Not sure why we need this?

Aethar01:
Lots of people use top bars or programs that control their programs audio through the mpris dbus interface which is incompletely implemented in firefox. I use zen so I made this patch for my use. I have been using it for a while and thought other people might appreciate it.

mr-cheffy:
Please share this patch to Firefox (https://firefox-source-docs.mozilla.org/contributing/contribution_quickref.html) because this is getting a bit out of scope of zen and im primarily concerned about mantainance and seurity of this patch. Thanks for trying to help improve zen though! But as I said, im not comfortable with starting to add patches to the engine that doesn't directly affect one of our features... hope you understand.

PR #12700, opened 2026-03-09
feat: Add "Separate from pinned tab" when resetting pinned tab with CMD

bo2themax:
Pretty new to Firefox's code base, `duplicateTabIn` is the proper one, I think, I could find to duplicate a tab.

1. The first commit added a `onRestore` completion block parameter `duplicateTabIn` function, because we're resetting tab state when resetting a pinned tab, which would cause a duplicated tab to end up in the wrong index.

2. The second commit added event handlers for `keydown` and `keyup` to change the tab subtitle based on the combination. I didn't find a proper way to update the tool tip, but since Arc doesn't show a theme at all and I guess the current tool tip won't cause any confusion since the subtitle shows the action.

   This commit also added a new subtitle text, which I'm not sure whether it's the right way or not, please let me know if it's not.

https://github.com/user-attachments/assets/dba4d57b-c448-4d03-a099-76d34beae95e

### DIFF
--- a/src/browser/components/tabbrowser/content/tab-js.patch
+++ b/src/browser/components/tabbrowser/content/tab-js.patch
@@ -1,5 +1,5 @@
 diff --git a/browser/components/tabbrowser/content/tab.js b/browser/components/tabbrowser/content/tab.js
-index 836bee14d2b63604688ebe477a5d915a5e99b305..7e105a1ae07657b0a0e664a8e3d9d2eb894fa1d4 100644
+index 836bee14d2b63604688ebe477a5d915a5e99b305..922c1233fadab5bc0c0ea41ae86be63d8b4565ad 100644
 --- a/browser/components/tabbrowser/content/tab.js
 +++ b/browser/components/tabbrowser/content/tab.js
 @@ -21,6 +21,7 @@
@@ -110,7 +110,28 @@ index 836bee14d2b63604688ebe477a5d915a5e99b305..7e105a1ae07657b0a0e664a8e3d9d2eb
      }
  
      get splitview() {
-@@ -489,6 +515,8 @@
+@@ -444,6 +470,12 @@
+         : this;
+       gBrowser.warmupTab(tabToWarm);
+ 
++      if (event.target.classList.contains("tab-reset-pin-button")) {
++        gZenPinnedTabManager.updateTabWithResetPinButtonHovered(event.target.closest("tab"), event);
++      } else {
++        gZenPinnedTabManager.updateTabWithResetPinButtonHovered(null, event);
++      }
++      
+       // If the previous target wasn't part of this tab then this is a mouseenter event.
+       if (!this.contains(event.relatedTarget)) {
+         this._mouseenter();
+@@ -455,6 +487,7 @@
+       if (!this.contains(event.relatedTarget)) {
+         this._mouseleave();
+       }
++      gZenPinnedTabManager.updateTabWithResetPinButtonHovered(null, event);
+     }
+ 
+     on_dragstart(event) {
+@@ -489,6 +522,8 @@
          this.style.MozUserFocus = "ignore";
        } else if (
          event.target.classList.contains("tab-close-button") ||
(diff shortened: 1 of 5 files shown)


src/zen/tabs/ZenPinnedTabManager.mjs
```diff
@@ -110,8 +113,15 @@ class nsZenPinnedTabManager extends nsZenDOMOperatedFeature {
 
   _onTabResetPinButton(event, tab) {
     event.stopPropagation();
-    this._resetTabToStoredState(tab);
-    gBrowser.selectedTab = tab;
+    if (event.getModifierState("Accel")) {
+      window.duplicateTabIn(tab, "tabshifted", 0, () => {
```
mr-cheffy, line 117: Why not just listen for "SSTabRestored" event? It would reduce patch size significantly:

```js
let newTab = SessionStore.duplicateTab(window, tab, 0);
newTab.addEventListener("SSTabRestored", (event) => {
		this._resetTabToStoredState(tab);
  	gBrowser.selectedTab = tab;
}, {once:true});
```

src/zen/tabs/ZenPinnedTabManager.mjs
```diff
@@ -134,14 +144,26 @@ class nsZenPinnedTabManager extends nsZenDOMOperatedFeature {
     switch (action) {
       case "TabPinned":
         tab._zenClickEventListener = this._zenClickEventListener;
+        tab._zenKeyDownEventListener = this._zenKeyDownEventListener;
+        tab._zenKeyUpEventListener = this._zenKeyUpEventListener;
         tab.addEventListener("click", tab._zenClickEventListener);
+        window.addEventListener("keydown", tab._zenKeyDownEventListener);
+        window.addEventListener("keyup", tab._zenKeyUpEventListener);
```
mr-cheffy, line 151: No need to always listen for keyups, this can be bad for performance. You can just register `window.addEventListener("keyup", () => ...), {once:true})` once keydown happens (defining this inside _zenKeyDownEventListener).

src/zen/tabs/ZenPinnedTabManager.mjs
```diff
@@ -134,14 +144,26 @@ class nsZenPinnedTabManager extends nsZenDOMOperatedFeature {
     switch (action) {
       case "TabPinned":
         tab._zenClickEventListener = this._zenClickEventListener;
+        tab._zenKeyDownEventListener = this._zenKeyDownEventListener;
+        tab._zenKeyUpEventListener = this._zenKeyUpEventListener;
         tab.addEventListener("click", tab._zenClickEventListener);
+        window.addEventListener("keydown", tab._zenKeyDownEventListener);
```
mr-cheffy, line 150: Wouldn't it be better to instead, check for "mouseenter" and check if the cmd modifier is being used?

src/browser/components/tabbrowser/content/tab-js.patch
```diff
@@ -110,7 +110,28 @@ index 836bee14d2b63604688ebe477a5d915a5e99b305..7e105a1ae07657b0a0e664a8e3d9d2eb
      }
  
      get splitview() {
-@@ -489,6 +515,8 @@
+@@ -444,6 +470,12 @@
+         : this;
+       gBrowser.warmupTab(tabToWarm);
+ 
++      if (event.target.classList.contains("tab-reset-pin-button")) {
++        gZenPinnedTabManager.updateTabWithResetPinButtonHovered(event.target.closest("tab"), event);
```
mr-cheffy, line 118: Here you can check cmd modifier

src/zen/tabs/ZenPinnedTabManager.mjs
```diff
@@ -163,6 +185,39 @@ class nsZenPinnedTabManager extends nsZenDOMOperatedFeature {
...
+      });
+    }
+  }
+
+  updateTabWithResetPinButtonHovered(tab, event) {
+    if (!tab) {
+      // reset subtitle first
+      this._updateZenTabSublabel(event);
+    }
+    this._tabWithResetPinButtonHovered = tab
```
mr-cheffy, line 214: No need to keep this as reference, might just want to add a temporal event listener

mr-cheffy (changes requested):

src/zen/tabs/ZenPinnedTabManager.mjs
bo2themax: Checking modifier with `mouseenter` only works if you moving the mouse with cmd pressed, it won't be updated if you release the key inside the reset button or press after the mouse is entered

mr-cheffy: I see, still, it would be beneficial to only register this even *once* the mouse is ontop of the tab

PR #12655, opened 2026-03-06
feat: Start implementing Zen Library

mr-cheffy:

PR #12701, opened 2026-03-09
feat: Implement Zen Library data backend

JustAdumbPrsn:
Extends #12655

Implements the data backend for the history and downloads sections of the Zen Library

1. history uses `PlacesQuery` with live updates through `observeHistory`
2. downloads using `DownloadHistory.getList` for Places-persisted downloads
3. Added empty state for both downloads and history sections
4. Uses `DownloadUtils` and `DownloadsCommon.strings` for size formatting
5. Add empty state handling for both sections because why not


CosmoCreeper:
AI??

Minemetero:
> AI??

shhhhh

PR #12710, opened 2026-03-10
feat: Add "Separate from pinned tab" when resetting pinned tab with CMD

bo2themax:
Force pushing to my origin seems broken, so I had to create another one, sorry about this.

Based on previous comments for #12700, I asked Claude to do some cleanup and refactoring.

https://github.com/user-attachments/assets/a6a6fefa-9b59-4a6d-855b-b523ced3e6a9

### DIFF
--- a/src/browser/components/tabbrowser/content/tab-js.patch
+++ b/src/browser/components/tabbrowser/content/tab-js.patch
@@ -1,5 +1,5 @@
 diff --git a/browser/components/tabbrowser/content/tab.js b/browser/components/tabbrowser/content/tab.js
-index 836bee14d2b63604688ebe477a5d915a5e99b305..7e105a1ae07657b0a0e664a8e3d9d2eb894fa1d4 100644
+index 836bee14d2b63604688ebe477a5d915a5e99b305..922c1233fadab5bc0c0ea41ae86be63d8b4565ad 100644
 --- a/browser/components/tabbrowser/content/tab.js
 +++ b/browser/components/tabbrowser/content/tab.js
 @@ -21,6 +21,7 @@
@@ -110,7 +110,26 @@ index 836bee14d2b63604688ebe477a5d915a5e99b305..7e105a1ae07657b0a0e664a8e3d9d2eb
      }
  
      get splitview() {
-@@ -489,6 +515,8 @@
+@@ -444,6 +470,10 @@
+         : this;
+       gBrowser.warmupTab(tabToWarm);
+
++      if (event.target.classList.contains("tab-reset-pin-button")) {
++        gZenPinnedTabManager.onResetPinButtonMouseOver(this, event);
++      }
++
+       // If the previous target wasn't part of this tab then this is a mouseenter event.
+       if (!this.contains(event.relatedTarget)) {
+         this._mouseenter();
+@@ -455,6 +491,7 @@
+       if (!this.contains(event.relatedTarget)) {
+         this._mouseleave();
+       }
++      gZenPinnedTabManager.onResetPinButtonMouseOut(this);
+     }
+ 
+     on_dragstart(event) {
+@@ -489,6 +522,8 @@
          this.style.MozUserFocus = "ignore";
        } else if (
          event.target.classList.contains("tab-close-button") ||
@@ -119,7 +138,7 @@ index 836bee14d2b63604688ebe477a5d915a5e99b305..7e105a1ae07657b0a0e664a8e3d9d2eb
          event.target.classList.contains("tab-icon-overlay") ||
          event.target.classList.contains("tab-audio-button")
        ) {
-@@ -543,6 +571,10 @@
+@@ -543,16 +578,17 @@
        this.style.MozUserFocus = "";
      }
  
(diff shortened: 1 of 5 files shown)


src/zen/tabs/ZenPinnedTabManager.mjs
```diff
@@ -163,6 +173,42 @@ class nsZenPinnedTabManager extends nsZenDOMOperatedFeature {
...
+    if (tab) {
+      let accelHeld = e.getModifierState("Accel") || (e.metaKey && e.type == "keydown");
+      this._setResetPinSublabel(tab, accelHeld);
+    }
+  }
+
+  _setResetPinSublabel(tab, accelHeld) {
+    let label = tab.querySelector(".zen-tab-sublabel");
+    if (label) {
+      window.document.l10n.setArgs(label, {
```
mr-cheffy: no need the `window` part

src/zen/tabs/ZenPinnedTabManager.mjs
```diff
@@ -163,6 +173,42 @@ class nsZenPinnedTabManager extends nsZenDOMOperatedFeature {
...
+        tabSubtitle: accelHeld ? "zen-default-pinned-cmd" : "zen-default-pinned",
+      });
+    }
+  }
+
+  onResetPinButtonMouseOver(tab, event) {
+    if (!this._resetPinButtonHovered) {
+      this._resetPinButtonHovered = true;
+      window.addEventListener("keydown", this._onAccelKeyChangeBound);
```
mr-cheffy: Should be `{once:true}` and keyup should be defined inside the change bound. Remove `_resetPinButtonHovered` as it would not be needed then

mr-cheffy (changes requested):

src/zen/tabs/ZenPinnedTabManager.mjs
```diff
@@ -110,8 +111,16 @@ class nsZenPinnedTabManager extends nsZenDOMOperatedFeature {
 
   _onTabResetPinButton(event, tab) {
     event.stopPropagation();
-    this._resetTabToStoredState(tab);
-    gBrowser.selectedTab = tab;
+    if (event.getModifierState("Accel")) {
+      let newTab = gBrowser.duplicateTab(tab, true);
+      newTab.addEventListener("SSTabRestored", () => {
+        this._resetTabToStoredState(tab);
+        gBrowser.selectedTab = tab;
```
mr-cheffy: Now that I think about it, why do we need the `selectTab` to happen after the tab has been restored? It can be determinedly

src/zen/tabs/ZenPinnedTabManager.mjs
```diff
@@ -163,6 +172,36 @@ class nsZenPinnedTabManager extends nsZenDOMOperatedFeature {
...
+      this._setResetPinSublabel(tab, accelHeld);
+      // Up <-> down events until the mouse leaves the button.
+      let nextEvent = e.type == "keydown" ? "keyup" : "keydown";
+      window.addEventListener(nextEvent, this._onAccelKeyChangeBound, { once: true });
+    }
+  }
+
+  _setResetPinSublabel(tab, accelHeld) {
+    let label = tab.querySelector(".zen-tab-sublabel");
+    if (label) {
```
mr-cheffy: Why would it be null?

src/zen/tabs/ZenPinnedTabManager.mjs
```diff
@@ -163,6 +172,36 @@ class nsZenPinnedTabManager extends nsZenDOMOperatedFeature {
...
+      document.l10n.setArgs(label, {
+        tabSubtitle: accelHeld ? "zen-default-pinned-cmd" : "zen-default-pinned",
+      });
+    }
+  }
+
+  onResetPinButtonMouseOver(tab, event) {
+    window.addEventListener("keydown", this._onAccelKeyChangeBound, { once: true });
```
mr-cheffy: Rather call `_onAccelKeyChange` here, no? What if im hovering with the key down already?

src/zen/tabs/ZenPinnedTabManager.mjs
```diff
@@ -163,6 +172,36 @@ class nsZenPinnedTabManager extends nsZenDOMOperatedFeature {
     }
   }
 
+  _onAccelKeyChange(e) {
+    let hoveredButton = document.querySelector(".tab-reset-pin-button:hover");
```
mr-cheffy: This is very expensive to do, maybe keep a reference on `onResetPinButtonMouseOver` and clear it on `onResetPinButtonMouseOut`.

e.g.

```
this._hoveringPinButton = ...
...

delete this._hoveringPinButton;
```

src/browser/components/tabbrowser/content/tab-js.patch
```diff
@@ -130,22 +149,63 @@ index 836bee14d2b63604688ebe477a5d915a5e99b305..7e105a1ae07657b0a0e664a8e3d9d2eb
...
-+      if (event.target.classList.contains("tab-reset-pin-button")) {
++      if (
++        !event.shiftKey && 
++        event.target.classList.contains("tab-reset-pin-button")
++      ) {
 +        gZenPinnedTabManager._onTabResetPinButton(event, this, 'reset');
 +        gBrowser.tabContainer._blockDblClick = true;
-+      } else if (event.target.classList.contains("tab-reset-button")) {
++      } else if (
++        !(event.getModifierState("Accel") || event.shiftKey) &&
```
mr-cheffy: Why wouldn't you want to prevent double click when not pressing ctrl? Also, this button isn't the same as `tab-reset-pin-button`

mr-cheffy (changes requested):

src/zen/tabs/ZenPinnedTabManager.mjs
bo2themax: Because the reset tab can also happen to non-selected pinned tabs. This was mentioned in the earlier pr

src/browser/components/tabbrowser/content/tab-js.patch
bo2themax: Here I was just thinking to keep the same behavior as before, since previously there was a guard on top all of these branches

src/zen/tabs/ZenPinnedTabManager.mjs
mr-cheffy: Yes, but why does it have to be inside the restore event callback?

mr-cheffy (changes requested):

src/zen/tabs/ZenPinnedTabManager.mjs
```diff
@@ -163,13 +167,40 @@ class nsZenPinnedTabManager extends nsZenDOMOperatedFeature {
...
-          "zen.pinned-tab-manager.wheel-close-if-pending"
-        ),
+        closeIfPending: Services.prefs.getBoolPref("zen.pinned-tab-manager.wheel-close-if-pending"),
       });
     }
   }
 
+  _onAccelKeyChange(e) {
+    let tab = this._tabWithResetPinButtonHovered;
+    if (tab) {
```
mr-cheffy: you could do `!tab` and do an early return

src/zen/tabs/ZenPinnedTabManager.mjs
```diff
@@ -163,13 +167,40 @@ class nsZenPinnedTabManager extends nsZenDOMOperatedFeature {
...
+      let accelHeld = e.getModifierState("Accel") || (e.metaKey && e.type == "keydown");
+      this._setResetPinSublabel(tab, accelHeld);
+      // Up <-> down events until the mouse leaves the button.
+      // When hovered with accelHeld, we should listen to the next keyup event
+      let nextEvent = accelHeld ? "keyup" : "keydown";
+      window.addEventListener(nextEvent, this._onAccelKeyChangeBound, { once: true });
```
mr-cheffy: Instead of binding the function, why not have it as a local variable and make it call its self?

mr-cheffy (changes requested):
I would like a quick unit test as well if you dont mind, thank you so much for the PR!

src/browser/components/tabbrowser/content/tab-js.patch
```diff
@@ -130,22 +149,63 @@ index 836bee14d2b63604688ebe477a5d915a5e99b305..7e105a1ae07657b0a0e664a8e3d9d2eb
      on_click(event) {
        if (event.button != 0) {
          return;
-@@ -603,6 +635,14 @@
+       }
+ 
+-      if (event.getModifierState("Accel") || event.shiftKey) {
```
mr-cheffy, line 154: Maybe just shift click then?

src/browser/components/tabbrowser/content/tab-js.patch
```diff
@@ -130,22 +149,63 @@ index 836bee14d2b63604688ebe477a5d915a5e99b305..7e105a1ae07657b0a0e664a8e3d9d2eb
...
+         if (this.activeMediaBlocked) {
+           if (this.multiselected) {
+@@ -583,7 +620,10 @@
+         return;
+       }
+ 
+-      if (event.target.classList.contains("tab-close-button")) {
++      if (
++        !(event.getModifierState("Accel") || event.shiftKey) &&
++        event.target.classList.contains("tab-close-button")
```
mr-cheffy: No need to format this

mr-cheffy (changes requested):
Almost done!

mr-cheffy (approved):

PR #12733, opened 2026-03-11
Fix video fullscreen rendering when split view is active

### LINKED ISSUE #11559: Unable to go full screen when in split view
### Preliminary Checks

### What happened?

When I split 2 tabs for example and one of them has a video that I want to open in full screen it won't let me. You can't use full screen mode for videos in split view can you add this option if you have 2, 3 or 4 split tabs to be able to go full screen for videos like Youtube for example.

### Expected behavior

Open 2 tabs on one of them you play a video.

### Actual behavior

When you try to click for opening a video in full screen it doesn't work cuz the split tab prevents it from opening

### Steps to reproduce

Open any site that has a video go to split view with any other site. Try to open the video in full screen when it's played it won't let you it will display both tabs that are split instead.

### Screenshots and videos

https://github.com/user-attachments/assets/c1f87fd8-24d9-4457-b015-954c259c105a

### Version

1.17.12b

Windows - x64, Linux (Flatpak)

Tabs

```shell

```

S31ZUR3:
Fixes #11559

## Problem
Fullscreen videos do not work correctly when split view is active.
Firefox expects only one active browser docshell for fullscreen,
but Zen split view keeps multiple browsers active, preventing the
fullscreen browser from being promoted correctly.

## Solution
When entering fullscreen:
- Only the fullscreen browser remains active
- Other split browsers are deactivated
- Non-fullscreen split containers are hidden
- The fullscreen container expands to fill the window

Split layout is restored automatically when exiting fullscreen.

## Changes
- Added fullscreenchange handler in ZenViewSplitter
- Toggle zen-split-fullscreen and zen-fullscreen attributes
- Deactivate other split browsers using docShellIsActive
- CSS override to hide non-fullscreen split containers

## Testing
Tested with:
- 2–4 split tabs
- YouTube fullscreen
- Fullscreen exit restoring split layout

### DIFF
--- a/src/zen/split-view/zen-decks.css
+++ b/src/zen/split-view/zen-decks.css
@@ -7,6 +7,7 @@
 /** Zen Decks - ONLY USED FOR SPLIT VIEWS, DO NOT USE THIS CLASS FOR ANY OTHER PURPOSE **/
 
 %include zen-split-group.inc.css
+
 #tabbrowser-tabpanels[zen-split-view="true"] {
   display: flex;
   flex-direction: row;
@@ -264,6 +265,3 @@ zen-split-fake-tab {
   flex: 1;
 }
 
-:root[inDOMFullscreen="true"] #tabbrowser-tabpanels[zen-split-view="true"] {
-  display: block !important;
-}
--- a/src/zen/split-view/zen-decks.css
+++ b/src/zen/split-view/zen-decks.css
@@ -7,7 +7,6 @@
 /** Zen Decks - ONLY USED FOR SPLIT VIEWS, DO NOT USE THIS CLASS FOR ANY OTHER PURPOSE **/
 
 %include zen-split-group.inc.css
-
 #tabbrowser-tabpanels[zen-split-view="true"] {
   display: flex;
   flex-direction: row;
@@ -253,4 +252,20 @@ zen-split-fake-tab {
   background-color: color-mix(in srgb, var(--button-background-color-primary), transparent 40%);
   margin: var(--tab-block-margin);
   flex: 1;
+}
+
+:root[inDOMFullscreen="true"] #tabbrowser-tabpanels[zen-split-view="true"] {
+  display: block !important;
+}
+
+:root[zen-split-fullscreen] .browserSidebarContainer[is-zen-split="true"]:not([zen-fullscreen="true"]) {
+  display: none !important;
+}
+
+:root[zen-split-fullscreen] .browserSidebarContainer[is-zen-split="true"][zen-fullscreen="true"] {
+  inset: 0 !important;
+  width: 100% !important;
+  height: 100% !important;
+  margin: 0 !important;
+  z-index: 1000 !important;
 }
--- a/src/zen/split-view/zen-decks.css
+++ b/src/zen/split-view/zen-decks.css
@@ -37,6 +37,16 @@
     /* Fix for issue https://github.com/zen-browser/desktop/issues/7564 */
     position: absolute;
   }
+
+  :root[inDOMFullscreen="true"] & {
+    &:not(.deck-selected) {
+      -moz-subtree-hidden-only-visually: 1 !important;
+    }
+
+    &.deck-selected {
+      inset: 0 !important;
+    }
+  }
 }
 
 .browserSidebarContainer[is-zen-split="true"],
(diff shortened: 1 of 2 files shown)


S31ZUR3:
Tested with YouTube and other fullscreen video players.

src/zen/split-view/zen-decks.css
```diff
@@ -254,3 +253,19 @@ zen-split-fake-tab {
   margin: var(--tab-block-margin);
   flex: 1;
 }
+
+:root[inDOMFullscreen="true"] #tabbrowser-tabpanels[zen-split-view="true"] {
```
mr-cheffy: Why not simply just wrap all these styles under `:root:not([inDOMFullscreen="true"])`? This seems a bit over-kill

mr-cheffy (changes requested):

src/zen/split-view/zen-decks.css
S31ZUR3: I did try to use CSS-only but it resulted in a black screen. Zen already positions split browsers using inline styles:

browserContainer.style.inset = `${nodeRootPosition.top}% ${nodeRootPosition.right}% ${nodeRootPosition.bottom}% ${nodeRootPosition.left}%`;

Because of that, disabling the split CSS alone doesn't allow the fullscreen browser to expand properly. The small JS part just tells the engine which browser should remain active during fullscreen.

mr-cheffy: @S31ZUR3 you can just use this patch I made, it fixes the issue with much less complexity:

```patch
diff --git a/src/zen/split-view/zen-decks.css b/src/zen/split-view/zen-decks.css
index ae6f7b39e..2472af649 100644
--- a/src/zen/split-view/zen-decks.css
+++ b/src/zen/split-view/zen-decks.css
@@ -38,6 +38,16 @@
     /* Fix for issue https://github.com/zen-browser/desktop/issues/7564 */
     position: absolute;
   }
+
+  :root[inDOMFullscreen="true"] & {
+    &:not(.deck-selected) {
+      -moz-subtree-hidden-only-visually: 1 !important;
+    }
+
+    &.deck-selected {
+      inset: 0 !important;
+    }
+  }
 }
 
 .browserSidebarContainer[is-zen-split="true"],
@@ -67,7 +77,7 @@
   }
 }
 
-#tabbrowser-tabpanels[zen-split-view="true"] .browserSidebarContainer.deck-selected {
+:root:not([inDOMFullscreen="true"]) #tabbrowser-tabpanels[zen-split-view="true"] .browserSidebarContainer.deck-selected {
   &:not(.zen-glance-overlay) {
     outline: 2px solid var(--zen-active-split-outline-color) !important;
   }
```

S31ZUR3: Thanks! I've applied the suggested changes.

src/zen/split-view/zen-decks.css
```diff
@@ -7,7 +7,6 @@
 /** Zen Decks - ONLY USED FOR SPLIT VIEWS, DO NOT USE THIS CLASS FOR ANY OTHER PURPOSE **/
 
 %include zen-split-group.inc.css
-
 #tabbrowser-tabpanels[zen-split-view="true"] {
```
mr-cheffy, line 11: Add new line here please

src/zen/split-view/zen-decks.css
```diff
@@ -254,3 +263,7 @@ zen-split-fake-tab {
   margin: var(--tab-block-margin);
   flex: 1;
 }
+
+:root[inDOMFullscreen="true"] #tabbrowser-tabpanels[zen-split-view="true"] {
```
mr-cheffy: Remove this please

mr-cheffy (changes requested):

src/zen/split-view/zen-decks.css
S31ZUR3: Yep finished doing them. Check them out once!

src/zen/split-view/zen-decks.css
```diff
@@ -254,3 +264,4 @@ zen-split-fake-tab {
   margin: var(--tab-block-margin);
   flex: 1;
 }
+
```
mr-cheffy: Please remove the last empty line, its erroring out the linter

mr-cheffy (changes requested):

S31ZUR3:
My first contribution to something i use. Thanks for this opportunity dude!!

PR #12734, opened 2026-03-11
feat: close new tab popup on Cmd/Ctrl + T, p=#12734

Lecloow:
Add functionality to close the new tab popup by pressing the same shortcut to open the popup (cmd/ctrl + t)

### DIFF
--- a/src/zen/common/modules/ZenUIManager.mjs
+++ b/src/zen/common/modules/ZenUIManager.mjs
@@ -438,6 +438,12 @@ window.gZenUIManager = {
       return false;
     }
 
+    // Close the new tab popup on cmd/ctrl + t
+    if (gURLBar.hasAttribute("zen-newtab")) {
+      this.handleUrlbarClose(false, false);
+      return true;
+    }
+
     // Clear any existing timeout
     if (this._clearTimeout) {
       clearTimeout(this._clearTimeout);
--- a/src/zen/common/modules/ZenUIManager.mjs
+++ b/src/zen/common/modules/ZenUIManager.mjs
@@ -440,7 +440,7 @@ window.gZenUIManager = {
 
     // Close the new tab popup on cmd/ctrl + t
     if (gURLBar.hasAttribute("zen-newtab")) {
-      this.handleUrlbarClose(false, false);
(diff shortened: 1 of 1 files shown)


src/zen/common/modules/ZenUIManager.mjs
```diff
@@ -500,6 +438,12 @@ window.gZenUIManager = {
       return false;
     }
 
+    // Close the new tab popup on cmd/ctrl + t
+    if (gURLBar.hasAttribute("zen-newtab")) {
+      this.handleUrlbarClose(false, false);
```
mr-cheffy: Both arguments are false by default, please remove these

mr-cheffy (changes requested):

src/zen/common/modules/ZenUIManager.mjs
```diff
@@ -500,6 +438,12 @@ window.gZenUIManager = {
       return false;
     }
 
+    // Close the new tab popup on cmd/ctrl + t
+    if (gURLBar.hasAttribute("zen-newtab")) {
```
mr-cheffy: Should also be checking for `shouldOpenURLBar`?

mr-cheffy (changes requested):

src/zen/common/modules/ZenUIManager.mjs
Lecloow: The previous condition is

 if (!shouldOpenURLBar) {
      return false;
    }

So the only way to reach:

if (gURLBar.hasAttribute("zen-newtab")) {
  this.handleUrlbarClose(false, false);
  return true;
}

is when shouldOpenURLBar is true.
There’s no need to check shouldOpenURLBar again

mr-cheffy: True, thank you, I missed that.

mr-cheffy: In that case, we should at least check for `overridePreferance`?

mr-cheffy (approved):

mr-cheffy:
Thanks!

PR #12751, opened 2026-03-12
feat: Added 'unload all other spaces' option in spaces context menu

AfeefurR:
Screenshot 2026-03-12 at 11 16 32 PM

Clicking 'Unload All Other Spaces' will unload every space except the space on which the context menu was triggered. Basically the same thing as unloading every space (except one) individually

### Why?
When I know that I will work in one space exclusively for a while, I typically unload every other space to manually free memory. It gets tedious when doing it for 6-10 spaces, so I thought this might be useful.

### DIFF
--- a/src/browser/base/content/zen-panels/popups.inc
+++ b/src/browser/base/content/zen-panels/popups.inc
@@ -43,6 +43,7 @@
     <menupopup />
   </menu>
   <menuitem id="context_zenUnloadWorkspace" data-l10n-id="zen-workspaces-panel-unload" command="cmd_zenUnloadWorkspace"/>
+  <menuitem id="context_zenUnloadAllOtherWorkspace" data-l10n-id="zen-workspaces-panel-unload-others" command="cmd_zenUnloadAllOtherWorkspace"/>
   <menuseparator id="context_zenWorkspacesSeparator"/>
   <menuseparator/>
   <menuitem id="context_zenReorderWorkspaces" data-l10n-id="zen-workspaces-panel-context-reorder" command="cmd_zenReorderWorkspaces"/>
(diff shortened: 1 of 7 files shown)


src/browser/base/content/zen-panels/popups.inc
```diff
@@ -43,6 +43,7 @@
     <menupopup />
   </menu>
   <menuitem id="context_zenUnloadWorkspace" data-l10n-id="zen-workspaces-panel-unload" command="cmd_zenUnloadWorkspace"/>
+  <menuitem id="context_zenUnloadAllOtherWorkspace" data-l10n-id="zen-workspaces-panel-unload-others" command="cmd_zenUnloadAllOtherWorkspace"/>
```
mr-cheffy: I think this should be moved downwards with "reorder workspaces"? `context_zenUnloadWorkspace` as well while you are at it if you don't mind.

src/zen/spaces/ZenSpaceManager.mjs
```diff
@@ -1665,6 +1665,27 @@ class nsZenWorkspaces {
...
+        !tab.hasAttribute("pending")
+    );
+
+    if (tabsToUnload.length === 0) {
+      return;
+    }
+
+    this.log("Unloading all workspaces except", workspaceId);
+
+    await gBrowser.explicitUnloadTabs(tabsToUnload); // TODO: unit test this
```
mr-cheffy, line 1508: I'd like tests for this if you don't mind.

mr-cheffy (changes requested):

src/zen/spaces/ZenSpaceManager.mjs
AfeefurR: I'll get working on it!

mr-cheffy:
Could you also please fix these?
CleanShot 2026-03-13 at 17 36 29@2x

mr-cheffy (approved):

PR #12763, opened 2026-03-13
fix: keep bookmarks bar in fullscreen, b=closes #8163, p=#12763

### LINKED ISSUE #8163: When in fullscreen(F11) bookmarks bar disappear
### Captchas

### What happened?

Just like the title when the fullscreen mode is turned on, the bookmarks bar becomes empty and impossible to click any bookmark, as soon as the fullscreen mode is turned off it goes back to normal.

### Version

1.12.2b

Windows - x64

Bookmarks

```shell

```

djdillybdev:
Issue: [When in fullscreen(F11) bookmarks bar disappear](https://github.com/zen-browser/desktop/issues/8163)

Hi, I saw this issue and thought it could be a good first issue for me to start contributing to an open source project that I'm a fan of using. 

I found that just setting `fullscreentoolbar="true"` for the PersonalToolbar resolved this issue and made sure that the bookmark bar stays with all bookmarks shown in fullscreen mode. Can easily test by:
- Have top toolbar shown
- Have bookmark toolbar shown
- Enter fullscreen mode (F11)

### DIFF
--- a/src/browser/base/content/navigator-toolbox-inc-xhtml.patch
+++ b/src/browser/base/content/navigator-toolbox-inc-xhtml.patch
@@ -1,5 +1,5 @@
 diff --git a/browser/base/content/navigator-toolbox.inc.xhtml b/browser/base/content/navigator-toolbox.inc.xhtml
-index 4d4223c508560136aba220adb18528aac913a188..c29194057efe15e6144f5114d7977a0e11619092 100644
+index 4d4223c508560136aba220adb18528aac913a188..10d4d9cecbb0e7cec9191d78fb81a57376b37ff1 100644
 --- a/browser/base/content/navigator-toolbox.inc.xhtml
 +++ b/browser/base/content/navigator-toolbox.inc.xhtml
 @@ -2,7 +2,7 @@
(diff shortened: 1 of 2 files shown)


mr-cheffy:
I wonder if it should only be added when single toolbar is enabled

djdillybdev:
> I wonder if it should only be added when single toolbar is enabled

It fixes the issue for all 3 browser layout options so I think that condition might not be needed.

mr-cheffy:
Right, but then it might be deviating from firefox's behavior. They probably had a reason for it, not sure what to feel about that

djdillybdev:
Okay that makes sense. I just checked and firefox has the same issue where the bookmarks dont show up in fullscreen. There's a thread i found [here](https://connect.mozilla.org/t5/ideas/show-bookmarks-bar-in-full-screen-mode/idi-p/56220) about it, but doesn't seem like they're gonna add it and people just use workarounds or the bookmark sidebar instead.

So I can make this change only apply for the single toolbar and the other layouts keep the same behavior as firefox (only difference in zen i noticed is that with the top toolbar there is still an empty space where the bookmarks would go in fullscreen mode)

Let me know what you prefer

mr-cheffy:
It's fine, lets see how it goes

mr-cheffy (changes requested):
To avoid extra patches, could you please maybe add the attribute at `gZenUIManager.init()`? At the end of the function, for example

mr-cheffy (approved):

Issue #10686, opened 2025-10-04
macOS Altcodes / Modifier Keys Cause Keyboard Shortcut Issues


Coops0:
### Preliminary Checks

- [x] I have read and understood the important section above.
- [x] I have searched existing issues and avoided creating duplicates.
- [x] I am not filing an enhancement request.
- [x] I have checked that this issue cannot be reproduced on Mozilla Firefox.
- [x] I have checked that this issue can be reproduced once I removed all my Mods and Custom CSS.

### What happened?

When trying to assign keyboard shortcuts on macOS, some shortcuts just do not work. Others bind incorrectly to other unicode characters which still doesn't work.

### Expected behavior

The keyboard shortcut to properly display the correct binded key then actually function.

### Actual behavior

It either doesn't work or displays a different unicode symbol and doesn't work.

### Steps to reproduce

1. On macOS, navigate to keyboard shortcuts in settings
2. Try to assign "Switch to Workspace 1" (or some other workspace) as "Cmd+Shift+1" (then press escape so it's binded)
3. This will successfully bind.
4. Go to another tab and try the shortcut to switch between workspaces. It just doesn't work.

Other example:
1. Same first steps, go and try to assign to a workspace button.
2. Try to assign it to "Cmd+Option+2" (and press escape)
3. It will be binded as "Cmd+Option+™" (and not work)

I am unsure if these are exactly correlated but they both don't work.

I'm just using workspaces as an example but these should work with other shortcuts too.

Also unrelated, but in the documentation and in these settings, spaces are still referred to as workspaces. Should that be changed?

### Screenshots and videos

<img width="95" height="35" alt="Image" src="https://github.com/user-attachments/assets/908018d7-a82e-4e3a-8f04-c5140b56adc0" />

### Version

1.16.3b

### What platform are you seeing the problem on?

macOS - aarch64

### What component is this issue related to?

Keyboard Shortcuts

marchfra:
I'm having the same problem but I have more context I can add: the problem seems to be input source-related.

I assigned "Ctrl+Option+Q" and "Ctrl+Option+E" to Backward and Forward Workspace respectively. The shortcuts bind as expected, and also work as expected when I'm using the Italian input source. However, if I then switch to the US input source, the Forward Workspace shortcut stops working.

I know Zen sees the shortcut correctly, because when I try to assign the same shortcut with the US input source it correctly binds to "Ctrl+Option+E", but then it just doesn't work until I switch back to the Italian input source.

[...]

PR #12776, opened 2026-03-15
fix: Keyboard shortcut assignment issue caused by modifier, b=closes https://github.com/zen-browser/desktop/issues/10686, p=#12776

gitmichaelqiu:
As described in #10686, when option/shift is included in the shortcut, Zen's `event.key` will return the result of the alt-code (such as `™`) instead of the physical key (such as `2`). This can break the shortkey listener.

Switching to `event.code` for digits and adding a mapping for symbol keys (such as comma) help translate the alt-code back to physical key. Since Windows does not have such issue, the fix is only applied to macOS.


mr-cheffy:
Why not for the rest of the OSs as well?

gitmichaelqiu:
Sure thing. I added the macOS check because the issue seems only occur on macOS. Now, I have removed the macOS system check. It should apply for all OSs.

mr-cheffy (approved):

PR #12780, opened 2026-03-15
feat: add .deb and .rpm package builds for Linux releases

mbergo:
Zen ships Linux builds as .tar.xz and AppImage but lacks native system packages. This creates friction for users expecting apt/dnf integration, breaks enterprise environments that block AppImage execution, and makes password manager native messaging harder to configure.

This PR adds:

build/linux/deb/         — Debian package structure
  DEBIAN/control         — package metadata (version, arch, deps)
  DEBIAN/postinst        — registers update-alternatives, MIME types,
                           and creates /usr/lib/zen/native-messaging-hosts/
                           so system-installed apps (1Password, Bitwarden,
                           KeePassXC) can place manifests there automatically
  DEBIAN/prerm           — clean removal of alternatives

build/linux/rpm/zen-browser.spec — RPM spec file with equivalent %post/%preun

scripts/package-linux-deb.sh  — repackages zen.linux-*.tar.xz into .deb scripts/package-linux-rpm.sh  — repackages zen.linux-*.tar.xz into .rpm
  (no source rebuild required — wraps existing Linux build artifacts)

.github/workflows/linux-package-build.yml
  — workflow_call job that runs after the Linux build, produces
    zen-browser_VERSION_amd64.deb, arm64.deb, x86_64.rpm, aarch64.rpm

.github/workflows/build.yml
  — wires linux-packages into the release pipeline and adds .deb/.rpm
    to release artifact upload for both twilight and stable releases

The native-messaging-hosts directory created in postinst/%post directly complements the NativeManifests path fix in the sibling PR, making the full integration seamless for end users on system-packaged Zen installs.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -360,6 +360,15 @@ jobs:
       build-version: ${{ needs.build-data.outputs.version }}
       release-branch: ${{ inputs.update_branch }}
 
+  linux-packages:
+    name: Linux packages (.deb / .rpm)
+    uses: ./.github/workflows/linux-package-build.yml
+    needs: [linux, build-data]
+    secrets: inherit
+    with:
+      build-version: ${{ needs.build-data.outputs.version }}
+      release-branch: ${{ inputs.update_branch }}
+
   appimage:
     name: AppImage build - Linux ${{ matrix.arch }}
     permissions:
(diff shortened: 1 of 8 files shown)


mr-cheffy (commented):
Has this been done with AI? There seems to be a lot of random values being added?

PR #12794, opened 2026-03-17
gh-12740: remove deleted workspace tabs from switch-to-tab

### LINKED ISSUE #10345: Tabs are lost then a workspace deletion is synced
### Preliminary Checks

### What happened?

When a workspace is deleted on an other computer, tabs from it are lost, when the deletion is synced.

### Expected behavior

Tabs should not disappear. Moving them to an other space (first one?) is a better solution.

### Actual behavior

Tabs are lost.

### Steps to reproduce

1) create a new workspace
2) use it (create tabs in it...)
3) after some time this workspace is synced to other computers
4) delete this workspace at an other computer
5) when deletion is synced, all the tabs from the workspace disappear

### Screenshots and videos

_No response_

### Version

1.15.2b (Firefox 142.0.1) (64-bit)

Linux (Flatpak)

Workspaces

```shell

```

### LINKED ISSUE #12740: "Switch Tab" persists after deleting a space
### Preliminary Checks

### What happened?

When making a new space, filling them up with tabs, then deleting them, the tabs persist on the new tab menu telling to "switch tab" even though those tabs dont exist anymore.

### Expected behavior

No recall of the deleted tab

### Actual behavior

On the tab menu, it shows tabs from the previous, deleted space

### Steps to reproduce

> Make a new space
> Add pinned tabs, standard tabs
> Delete the new space
> Try to open a new tab in the default space, the tabs from the deleted space should persist

### Screenshots and videos

Image

Image

Image

### Version

1.19.2b

Windows - x64

Tabs

```shell

```

Vedaant-Rajoo:
## Summary

  Fixes #12740 by removing workspace-owned tabs through the normal tab close path when a
  workspace is deleted.

  Previously, deleting a workspace only removed the workspace metadata/UI. Tabs that
  still belonged to that workspace could remain alive in the browser tab model, which
  allowed stale `Switch Tab` results to continue appearing in the urlbar.

## What changed

  - close workspace-owned tabs before removing workspace metadata
  - include both regular and pinned workspace tabs
  - exclude essentials and empty tabs from workspace deletion
  - clear Zen's stored tab cache after deletion
  - add/update workspace tests around deletion and unload behavior

  ## Why this approach

  The stale urlbar result was a symptom of the tabs not actually being closed. The safest
  fix is to delete those tabs through the normal `gBrowser.removeTabs(...)` path so the
  usual browser and Zen cleanup runs, instead of trying to manually refresh urlbar state
  afterward.

  ## Notes

  - essentials are intentionally not deleted because they are not workspace-owned
  - the delete regression test passes
  - while working on this, some older unload tests also needed cleanup because they were
  creating unrealistic workspace state by mutating attributes directly

>   Short maintainer note:
>   The core reasoning here is that Switch Tab was not primarily a urlbar caching bug. The
>   deleted workspace tabs were surviving the workspace deletion flow, so the correct fix
>   point was the workspace delete path itself. Closing those tabs normally is lower risk
>   than adding special-case invalidation around urlbar providers.


src/zen/spaces/ZenSpaceManager.mjs
```diff
@@ -1261,12 +1261,14 @@ class nsZenWorkspaces {
...
 
   removeWorkspace(windowID) {
+    this.#deleteWorkspaceOwnedTabs(windowID);
     let workspacesData = this.getWorkspaces();
     // Remove the workspace from the cache
     workspacesData = workspacesData.filter(
       workspace => workspace.uuid !== windowID
     );
     this.#propagateWorkspaceData(workspacesData);
+    this._allStoredTabs = null;
```
mr-cheffy: Not needed?

src/zen/spaces/ZenSpaceManager.mjs
```diff
@@ -1261,12 +1261,14 @@ class nsZenWorkspaces {
   }
 
   removeWorkspace(windowID) {
+    this.#deleteWorkspaceOwnedTabs(windowID);
```
mr-cheffy, line 1265: I think it would be better to do this at https://github.com/zen-browser/desktop/blob/c929fea21d70766426813e7ed427b9bec4fe7f81/src/zen/spaces/ZenSpaceManager.mjs#L1334-L1337 ?

src/zen/spaces/ZenSpaceManager.mjs
```diff
@@ -1472,6 +1483,18 @@ class nsZenWorkspaces {
...
+  #deleteWorkspaceOwnedTabs(workspaceID) {
+    const tabs = this.#workspaceOwnedTabs(workspaceID);
+    if (!tabs.length) {
+      return;
+    }
+
+    gBrowser.removeTabs(tabs, {
+      closeWindowWithLastTab: false,
+    });
+    this._allStoredTabs = null;
```
mr-cheffy: Not needed

mr-cheffy (changes requested):

src/zen/spaces/ZenSpaceManager.mjs
Vedaant-Rajoo: yes, this is the most likely redundant one.

src/zen/spaces/ZenSpaceManager.mjs
Vedaant-Rajoo: `line 1271` is not redundant in all cases
the redundant one is more likely the clear inside `#deleteWorkspaceOwnedTabs()`, not the one in `removeWorkspace()`

src/zen/spaces/ZenSpaceManager.mjs
Vedaant-Rajoo: I tried moving deletion into the `propagateWorkspaces()` removal promise as suggested, but the stale `Switch Tab` result came back. That suggests the tab close needs to happen earlier than that async removal branch.
Even with window sync the deletion of the tabs propagate through windows right now.

src/zen/spaces/ZenSpaceManager.mjs
Vedaant-Rajoo: These are some cases where i found it useful:
- The workspace is deleted but it has no workspace-owned tabs left.
  In that case [`#deleteWorkspaceOwnedTabs()`](/home/vee/Coding/OSS/zen-browser/src/zen/spaces/ZenSpaceManager.mjs) may do nothing, so no tab-close-driven cache invalidation happens. But `allStoredTabs` is still stale because the workspace container set changed.

- The workspace only had tabs that are excluded from deletion logic.
  Example: no normal workspace-owned tabs, only essentials/empty-tab edge cases. The workspace still disappears from the workspace UI/container model, so the cached tab walk can still be stale.

- You move tab deletion out of `removeWorkspace()` or it happens later/asynchronously.
  Then line 1271 becomes the only immediate invalidation at the time workspace data is changed.

Why this matters:
- [`allStoredTabs`](/home/vee/Coding/OSS/zen-browser/src/zen/spaces/ZenSpaceManager.mjs#L3068) is built by walking workspace containers, not just `gBrowser.tabs`.
- [`removeWorkspace()`](/home/vee/Coding/OSS/zen-browser/src/zen/spaces/ZenSpaceManager.mjs#L1263) changes workspace data and leads to workspace container removal.
- If no lower-level tab invalidation fires, the memoized cache can point at the old container-derived state.

So:
- if real tab removal definitely happens first, line 1271 may be redundant
- if workspace metadata/UI can change without guaranteed tab-removal invalidation, line 1271 is still useful

src/zen/spaces/ZenSpaceManager.mjs
Vedaant-Rajoo: P.S.


> While testing multi-window behavior for this PR, I noticed a separate sync/UI issue that may be relevant to workspace propagation: when I create a new space in one window, the other synced window sometimes renders that propagated space as a folder-like item alongside the new space instead of just showing a normal new workspace entry.
>
> I don’t think this is caused by the tab-deletion fix itself, but it does make multi-window verification noisy and may be relevant to the suggestion of moving more deletion logic into `propagateWorkspaces()`. I’m treating it as a separate workspace propagation/UI issue for now, not as evidence that the deletion fix is wrong.

mr-cheffy: >  Even with window sync the deletion of the tabs propagate through windows right now.

Very true, I missed that

src/zen/spaces/ZenSpaceManager.mjs
mr-cheffy: Please use `delete` then

src/zen/tests/workspaces/browser_workspace_unload.js
```diff
@@ -129,4 +160,51 @@ add_task(async function test_UnloadWorkspace_TargetedWorkspaceIsolation() {
 
   await gZenWorkspaces.removeWorkspace(inActiveWorkspaceId);
   await gZenWorkspaces.removeWorkspace(activeWorkspaceId);
+  removeNonWorkspaceTabs();
+});
+
+add_task(async function test_DeleteWorkspace_RemovesWorkspaceOwnedTabs() {
```
mr-cheffy: This test is a bit useless, i'd just check if TabRemove event is being fired for every tab and TabGroupRemoved

src/zen/spaces/ZenSpaceManager.mjs
```diff
@@ -1438,6 +1440,15 @@ class nsZenWorkspaces {
     );
   }
 
+  #workspaceOwnedTabs(workspaceID) {
+    return gBrowser.tabs.filter(
+      tab =>
+        tab.getAttribute("zen-workspace-id") === workspaceID &&
+        !tab.hasAttribute("zen-essential") &&
+        !tab.hasAttribute("zen-empty-tab")
```
mr-cheffy: `!tab.hasAttribute("zen-empty-tab")` not needed here AFAIK. At least i'd do `(!tab.hasAttribute("zen-empty-tab") && !tab.group)` so it also removes empty tabs of the folders and therefor, remove the folder its self

mr-cheffy (changes requested):

src/zen/spaces/ZenSpaceManager.mjs
Vedaant-Rajoo: done.

src/zen/tests/workspaces/browser_workspace_unload.js
Vedaant-Rajoo: Updated the test to cover the actual removal lifecycle instead of only checking the final tab state. It now waits for the relevant `TabClose` events and `TabGroupRemoved`, then asserts the workspace-owned tabs are gone. The updated coverage is in [browser_workspace_unload.js](/home/vee/Coding/OSS/zen-browser/src/zen/tests/workspaces/browser_workspace_unload.js).

src/zen/spaces/ZenSpaceManager.mjs
Vedaant-Rajoo: Fixed this by handling workspace folders explicitly during workspace deletion instead of trying to fold `zen-empty-tab` handling into the normal tab filter. `removeWorkspace()` now deletes workspace folders first, which also clears the folder empty-tab path, and then removes the remaining workspace-owned tabs. I also extended the test to cover a folder case in [browser_workspace_unload.js](/home/vee/Coding/OSS/zen-browser/src/zen/tests/workspaces/browser_workspace_unload.js). The deletion change is in [ZenSpaceManager.mjs](/home/vee/Coding/OSS/zen-browser/src/zen/spaces/ZenSpaceManager.mjs).

Vedaant-Rajoo: > [!IMPORTANT]
> I verified the workspace-deletion change in isolation and the updated deletion test passes.
>
> I still couldn’t get a clean full-file run, but this does not look introduced by this PR: the same file already fails on a full `origin/dev` checkout, and it fails more broadly there. The main issue is unload tests not fully unloading tabs (`pending` stays false / `linkedPanel` still exists), which then leaves unexpected tabs behind at the end of the run.
>
> Because of that, I used a temporary local `.only()` only to confirm the new deletion behavior independently, then removed it afterward.

src/zen/spaces/ZenSpaceManager.mjs
```diff
@@ -1449,6 +1450,18 @@ class nsZenWorkspaces {
     );
   }
 
+  #workspaceFolders(workspaceID) {
+    const workspaceElement = this.workspaceElement(workspaceID);
+    if (!workspaceElement) {
+      return [];
+    }
+
+    return [
```
mr-cheffy: This doesn't take into account sub folders I think? Just use `gBrowser.tabGroups`

mr-cheffy (changes requested):

src/zen/spaces/ZenSpaceManager.mjs
mr-cheffy: Sorry, better to use `gBrowser.tabContainer._invalidateCachedVisibleTabs();`

mr-cheffy (changes requested):

src/zen/spaces/ZenSpaceManager.mjs
Vedaant-Rajoo: I tracked down why `this._allStoredTabs = null` originally looked like the fix.

What made it confusing is that there are actually two delete paths with different behavior:
- deleting the currently active workspace
- deleting a different workspace while staying in the current one

Deleting the active workspace worked because the removal code was using `gBrowser.tabs`, and in Zen that effectively reflects the active workspace strip. So the tabs from the workspace being deleted were present in that enumeration and got removed.

Deleting a different workspace failed because its tabs were not part of that same `gBrowser.tabs` view, so `#deleteWorkspaceOwnedTabs()` could miss them entirely. That is why stale `Switch to Tab` entries remained.

I initially thought `this._allStoredTabs = null` was the right fix because clearing that cache made the stale results disappear in practice. After tracing it further, the real issue was not the open-tabs/urlbar unregister path, and not `moz_openpages_temp` either; those were being cleaned up correctly on workspace deletion. The actual problem was that the delete path was enumerating tabs from an active-workspace-scoped source.

So the fix here is to use `allStoredTabs` for workspace-owned tab deletion, since that walks tabs across workspace containers instead of only what `gBrowser.tabs` currently exposes for the active space. I also updated the test to cover deleting a workspace from another workspace, since that was the failing branch.

src/zen/spaces/ZenSpaceManager.mjs
Vedaant-Rajoo: Switched folder lookup to `gBrowser.tabGroups` here.

The reason for that change is that the earlier container-child walk was only checking direct children of the workspace containers, which could miss nested folder structures. `gBrowser.tabGroups` in Zen already goes through `gZenWorkspaces.allTabGroups`, so it gives the cross-workspace group set instead of only the active workspace strip.

From there I filter to top-level Zen folders for the target workspace and let `folder.delete()` handle recursive cleanup through `allItemsRecursive`.

src/zen/spaces/ZenSpaceManager.mjs
Vedaant-Rajoo: See https://github.com/zen-browser/desktop/issues/12862

mr-cheffy:
You haven't ran the tests you created, have you?

mr-cheffy (approved):

Vedaant-Rajoo:
I did run the tests. But separately and indicidually.

PR #12795, opened 2026-03-17
fix(compact-mode): prevent sidebar flicker/jitter during snap-restore on Windows

### LINKED ISSUE #12299: Vertical Sidebar flickering/jittering during interaction
### Preliminary Checks

### What happened?

The vertical sidebar (workspace/tab switcher) exhibits rapid horizontal or vertical "shaking" (fluctuation) when interacting with the main page content or resizing the window. It appears the layout engine is repeatedly recalculating the sidebar's width or position, causing a jittery visual effect.

### Expected behavior

The sidebar Should not show up

### Actual behavior

It shows up and flickers 

### Steps to reproduce

Maximize and try to go to Windowed mode in Windows in Compact Mode while mouse being just a bit around where sidebar would be if it was active

### Screenshots and videos

https://github.com/user-attachments/assets/dac81ff5-d5d8-417e-8b30-372b2bdc22ff

### Version

1.18.5b (64-bit)

Windows - x64

Compact Mode

```shell

```

Ivorisnoob:
**Fixes #12299**

## What was happening

When you drag a maximized Zen window back to normal size on Windows, the compact mode sidebar would rapidly flicker in and out if your mouse was near the sidebar edge.

The reason is subtle: Windows sends mouse events to the browser not because your mouse moved, but because the *window* moved, shrinking under your stationary cursor. Firefox passes these along as normal `mouseenter`/`mouseleave` events, and Zen's compact mode hover detection has no way of knowing they're fake, so it kept showing and hiding the sidebar in a loop.

The obvious fix of setting a flag on `sizemodechange` or `resize` doesn't work because those events fire *after* the hover events are already queued.

## How it's fixed

We track the mouse's real screen position via a `mousemove` listener. In the `onEnter` handler, if the event's screen coordinates match the last recorded position (meaning the mouse hasn't actually moved), the hover is rejected as a false positive.

A `_mouseMovedSinceResize` flag adds a second layer: it resets on every window resize and only clears once a genuine `mousemove` is detected, blocking any hover events that sneak in before the mouse has actually moved after the resize.

## Testing

Windows x64, Zen 1.19.3b, compact mode on, sidebar hidden. Snap-dragging from maximized to windowed with the mouse near the sidebar edge no longer causes flickering.


src/zen/compact-mode/ZenCompactMode.mjs
```diff
@@ -1,4 +1,4 @@
-/* This Source Code Form is subject to the terms of the Mozilla Public
+﻿/* This Source Code Form is subject to the terms of the Mozilla Public
```
Ivorisnoob, line 1: I am not sure why this says changed, I think i Opened it in Zed and that made some formatting change

src/zen/compact-mode/ZenCompactMode.mjs
```diff
@@ -90,9 +90,24 @@ window.gZenCompactModeManager = {
     this.addHasPolyfillObserver();
 
     // Clear hover states when window state changes (minimize, maximize, etc.)
-    window.addEventListener("sizemodechange", () =>
-      this._clearAllHoverStates()
-    );
+    window.addEventListener("sizemodechange", () => {
+      this._clearAllHoverStates();
+      document.documentElement.setAttribute("zen-resizing", "true");
```
mr-cheffy: Why set an attribute?

src/zen/compact-mode/ZenCompactMode.mjs
```diff
@@ -90,9 +90,24 @@ window.gZenCompactModeManager = {
...
+      clearTimeout(this._zenResizingTimeout);
+      this._zenResizingTimeout = setTimeout(() => {
+        document.documentElement.removeAttribute("zen-resizing");
+      }, 400);
```
mr-cheffy: Why 400? Seems redundant

mr-cheffy (changes requested):

src/zen/compact-mode/ZenCompactMode.mjs
Ivorisnoob: Good catch, that's a leftover from an earlier approach and can be removed.

src/zen/compact-mode/ZenCompactMode.mjs
Ivorisnoob: You're right, this whole block is redundant and can be removed along with the attribute. let me fix it, thanks

src/zen/compact-mode/ZenCompactMode.mjs
Minemetero: bro speaks like an AI agent

Ivorisnoob: > bro speaks like an AI agent


Lmao nah it Just happens, I tapped on some Tab autocomplete in the middle

src/zen/compact-mode/ZenCompactMode.mjs
```diff
@@ -90,9 +90,22 @@ window.gZenCompactModeManager = {
     this.addHasPolyfillObserver();
 
     // Clear hover states when window state changes (minimize, maximize, etc.)
-    window.addEventListener("sizemodechange", () =>
-      this._clearAllHoverStates()
-    );
+    const onWindowResize = () => {
+      // Fix #12299: block hover events until the mouse genuinely moves after a resize.
```
mr-cheffy: nit: use gh- format

```suggestion
      // Fix gh-12299: block hover events until the mouse genuinely moves after a resize.
```

src/zen/compact-mode/ZenCompactMode.mjs
```diff
@@ -731,6 +744,13 @@ window.gZenCompactModeManager = {
   },
 
   addMouseActions() {
+    this._mouseMovedSinceResize = true;
+    this._onMouseMove = () => { this._mouseMovedSinceResize = true; };
```
mr-cheffy: Why are you assigning it to `this`?

src/zen/compact-mode/ZenCompactMode.mjs
```diff
@@ -731,6 +744,13 @@ window.gZenCompactModeManager = {
...
+    window.addEventListener("mousemove", this._onMouseMove, true);
```
mr-cheffy: I would allow the mouse move listener ONLY inside onWindowResize and called with `{ once: true }`

mr-cheffy (changes requested):

src/zen/compact-mode/ZenCompactMode.mjs
Ivorisnoob: Done, moved the mousemove listener inside onWindowResize with { once: true } so it self-removes after the first real mouse movement. Removed the this assignment as well

src/zen/compact-mode/ZenCompactMode.mjs
```diff
@@ -91,11 +91,14 @@ window.gZenCompactModeManager = {
...
     // Clear hover states when window state changes (minimize, maximize, etc.)
     const onWindowResize = () => {
-      // Fix #12299: block hover events until the mouse genuinely moves after a resize.
+      // Fix gh-12299: block hover events until the mouse genuinely moves after a resize.
       // During snap-restore on Windows, the window moves under the cursor firing fake
       // mouseenter events. We suppress them until a real mousemove is detected.
       this._mouseMovedSinceResize = false;
       this._clearAllHoverStates();
+      window.addEventListener("mousemove", () => {
+        this._mouseMovedSinceResize = true;
```
mr-cheffy: Would be nice to add some logs here and on top, by using `this.log`

mr-cheffy (changes requested):

src/zen/compact-mode/ZenCompactMode.mjs
Ivorisnoob: surely, in a bit

src/zen/compact-mode/ZenCompactMode.mjs
```diff
@@ -731,6 +749,9 @@ window.gZenCompactModeManager = {
   },
 
   addMouseActions() {
+    this._mouseMovedSinceResize = true;
+    this.log("Compact mode mouse actions initialized");
```
mr-cheffy: No need for this

src/zen/compact-mode/ZenCompactMode.mjs
```diff
@@ -731,6 +749,9 @@ window.gZenCompactModeManager = {
   },
 
   addMouseActions() {
+    this._mouseMovedSinceResize = true;
```
mr-cheffy: Can just be declared at the top of the class

src/zen/compact-mode/ZenCompactMode.mjs
```diff
@@ -543,9 +561,9 @@ window.gZenCompactModeManager = {
               this.sidebar,
               this.sidebarIsOnRight
                 ? {
-                    marginRight: [`-${sidebarWidth}px`, 0],
-                    transform: ["translateX(100%)", "translateX(0)"],
-                  }
+                  marginRight: [`-${sidebarWidth}px`, 0],
```
mr-cheffy: Remember to run npm run lint:fix

src/zen/compact-mode/ZenCompactMode.mjs
```diff
@@ -90,9 +90,27 @@ window.gZenCompactModeManager = {
...
+        this._mouseMovedSinceResize = true;
+      }, { once: true });
+    };
+
+    window.addEventListener("sizemodechange", () => {
+      onWindowResize();
+    });
+
+    window.addEventListener("resize", event => {
+      if (!event.isTrusted) return;
```
mr-cheffy: isTrusted?

mr-cheffy (changes requested):

src/zen/compact-mode/ZenCompactMode.mjs
Ivorisnoob: @mr-cheffy  getAndApplySidebarWidth dispatches new window.Event("resize") internally to trigger layout recalc, and that's script-initiated so isTrusted is false on it. Without the check that would also hit onWindowResize and reset _mouseMovedSinceResize on every sidebar width calc, if i am getting the js correct this which would be bad

Ivorisnoob:
Looks like the Lint command changed a lot of formatting in the file.

mr-cheffy:
Not sure what you ran, but that unformatted the file even more

src/zen/compact-mode/ZenCompactMode.mjs
```diff
@@ -90,9 +91,33 @@ window.gZenCompactModeManager = {
...
+        { once: true }
+      );
+    };
+
+    window.addEventListener("sizemodechange", () => {
+      onWindowResize();
+    });
+
+    window.addEventListener("resize", event => {
+      if (!event.isTrusted) {
```
mr-cheffy, line 116: So on every resize, there's always going to be a new move listener?

mr-cheffy (changes requested):

mr-cheffy:
Sorry, but im sick of having to review AI generated PRs

PR #12804, opened 2026-03-17
Feature/tab switching

Skxxtz:
### Description

This PR introduces customizable shortcuts for sequential tab navigation (Next Tab/Previous Tab).

Currently, Zen provides a "previous tab" shortcut based on tab usage history. However, for users who need to navigate through their tab strip linearly and quickly, the default Ctrl+Tab behavior isn't always optimal. This PR adds the necessary infrastructure to allow users to bind their own keys for sequential tab switching.

### Changes

**Locales:** Added `en-US` Fluent strings for both "Forward Tab" and "Backward Tab" shortcuts. Also added `de` locale as native German speaker (also stripped empty white spaces at the end of lines)

**Migration:** Updated `KBS_VERSION` to `17` and implemented a migration block in `shortcuts.mjs` to inject these new entries into existing profiles.

**Defaults:** The shortcuts are left empty by default. This ensures no interference with existing user configurations while giving users the power to opt-in via settings.

**Categorization:** Added the new shortcuts to the `windowAndTabManagement` group for visibility in the Shortcuts Settings UI.

### Checklist

- [x] `en-US` locales provided in shortcuts.ftl.
- [x] Shortcuts added to migration logic and `KBS_VERSION` incremented.
- [x] Shortcuts left empty by default to prevent keybind conflicts.

https://github.com/user-attachments/assets/c9e84d15-6e2a-4b03-8fec-2d2fa313b6b4


mr-cheffy:
Hi, sorry, but im closing this as its not planned: you can see the reason for this at the comment inside https://github.com/zen-browser/desktop/pull/11258. We are waiting for firefox to do it natively

PR #12832, opened 2026-03-19
gh-12782: Fix app menu items cut off on small screens (macOS)

### LINKED ISSUE #12782: macOS: Context Menu Items Cut Off at Bottom
### Preliminary Checks

### What happened?

The context menu is using the available space on screen, if the height of the screen is too short, then the items at the bottom can not be seen.

### Expected behavior

Show all the avaiable items when scrolling to the bottom

### Actual behavior

Items at the bottom are cut off.

### Steps to reproduce

1. Change display settings to larger text(1147x745)
2. Open Twilight/Zen
3. Open application menu by click the top right button

### Screenshots and videos

Image

### Version

1.20t (Firefox 148.0.2) (2026-03-15) (aarch64)

macOS - aarch64

Settings

```shell
I think this is pure UI issues, I couldn't find the responsible component just yet.
```

bo2themax:
image

### AI Disclosure

I asked Claude to debug and find the rc of #12782. The reasoning makes sense to me, and I ran the unit test before and after the fix, also manually tested myself. 

Since I don't know whether this is an issue for other platforms as well, there is a check for the platform.


src/zen/common/modules/ZenUIManager.mjs
```diff
@@ -331,7 +331,36 @@ window.gZenUIManager = {
     this._popupTrackingElements.remove(element);
   },
 
+  // On macOS, the app menu panel is displayed as a native NSPopover which
+  // silently clips content beyond the screen without informing Firefox's
+  // layout engine. This makes bottom menu items unreachable by scrolling.
+  // Setting max-height based on available screen space lets Firefox's layout
+  // handle the constraint, enabling proper overflow scrolling.
+  // See https://github.com/zen-browser/desktop/issues/12782
```
mr-cheffy: nit:


```suggestion
  // See gh-12782
```

src/zen/common/modules/ZenUIManager.mjs
```diff
@@ -331,7 +331,36 @@ window.gZenUIManager = {
...
+    const spaceBelow = screenBottom - anchorScreenBottom;
+    const spaceAbove = anchorScreenTop - window.screen.availTop;
+    const maxHeight = Math.max(spaceBelow, spaceAbove);
+    if (maxHeight > 0 && maxHeight < window.screen.availHeight) {
+      panel.style.maxHeight = `${maxHeight}px`;
+    }
+  },
+
   onPopupShowing(showEvent) {
+    if (AppConstants.platform === "macosx") {
```
mr-cheffy: i'd also check if the preference is enabled?

src/zen/tests/popover/browser.toml
```diff
@@ -0,0 +1,7 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+[DEFAULT]
+
+["browser_popover_height_constraint.js"]
```
mr-cheffy, line 8: It should only run on macos?

mr-cheffy (changes requested):

src/zen/common/modules/ZenUIManager.mjs
bo2themax: Strangely, I tried native popup Boolean settings; it doesn't seem to work as expected: the background is just removed, and the style didn't change to what Firefox has.

bo2themax: Claude added it once during the process, I can add it back

src/zen/common/modules/ZenUIManager.mjs
```diff
@@ -331,7 +331,39 @@ window.gZenUIManager = {
...
+    }
+    const anchorRect = anchor.getBoundingClientRect();
+    const screenBottom = window.screen.availTop + window.screen.availHeight;
+    const anchorScreenBottom = window.screenY + anchorRect.bottom;
+    const anchorScreenTop = window.screenY + anchorRect.top;
+    const spaceBelow = screenBottom - anchorScreenBottom;
+    const spaceAbove = anchorScreenTop - window.screen.availTop;
+    const maxHeight = Math.max(spaceBelow, spaceAbove);
+    if (maxHeight > 0 && maxHeight < window.screen.availHeight) {
+      panel.style.maxHeight = `${maxHeight}px`;
```
mr-cheffy: Does this ever get reset?

mr-cheffy (changes requested):

src/zen/common/modules/ZenUIManager.mjs
bo2themax: > Does this ever get reset?

No, I don't think it needs to be reset, since this constraint should be on every `appMenu-popup`, but I did find some regression after reading this. 

The previous fix will cause the popover to be too small if you move the browser window, and the anchor button is in the centre of the screen. I changed to another approach -- always set the `maxHeight` to the available usable space on this screen.

image

image

mr-cheffy (approved):

Issue #12241, opened 2026-02-05
Bookmark temporarily removes


TheGlorySaint:
### Preliminary Checks

- [x] I have read and understood the important section above.
- [x] I have searched existing issues and avoided creating duplicates.
- [x] I am not filing an enhancement request.
- [x] I have checked that this issue cannot be reproduced on Mozilla Firefox.
- [x] I have checked that this issue can be reproduced once I removed all my Mods and Custom CSS.

### What happened?

In my bookmarks-bar i have multiple bookmarks saved. The first bookmark is also the Homepage in about:preferences#home and opens with every new window.

When i open the first Zen-Window everything loads normal but with opening every new window the first bookmark, that is also the Homepage, is temporarily removed.

### Expected behavior

The first bookmark is displayed

### Actual behavior

The first bookmark is not displayed that is also the homepage

### Steps to reproduce

1. Set a Page as Homepage under about:preferences#home
2. Add that Page that is Homepage as the first bookmark in bookmarksbar 
3. Add a second bookmark to bookmarksbar
4. Leave the Homepage/first bookmarked Page open on the first window
5. Open a second window
6. Then the homepage/first bookmark is removed from the bookmarksbar

### Version

1.18.4 b

### What platform are you seeing the problem on?

macOS - aarch64

### What component is this issue related to?

Bookmarks

### Relevant log output if applicable

```shell

```

TheGlorySaint:
Is there an update on this issue?

TheGlorySaint:
Tested again with v1.19.3b. Issue still exists

TheGlorySaint:
Issue still exists with 1.19.8

mr-cheffy:
Fixed with https://github.com/zen-browser/desktop/pull/13168

mr-cheffy:
Could you please verify this has been in twilight if you don't mind?

TheGlorySaint:
Yesterday, i build me twilight locally and there it worked. I used [this](https://docs.zen-browser.app/contribute/desktop/building) Documentation for that.

Now downloaded twilight with brew, Version 1.20t and there the fix is not available.

mr-cheffy:
Seems to have failed, please check after https://github.com/zen-browser/desktop/actions/runs/24238467493

TheGlorySaint:
> Seems to have failed, please check after https://github.com/zen-browser/desktop/actions/runs/24238467493

The links does not work

PR #12838, opened 2026-03-19
Fix workspace bookmark cache lookup to prevent temporary bookmark disappearance

TheGlorySaint:
# Issue: https://github.com/zen-browser/desktop/issues/12241
## What this fixes
This change resolves a workspace bookmark edge case where bookmarks could be treated as missing temporarily (notably when the homepage bookmark is involved across multiple windows/workspaces).

## Changes made
- Fixed `workspaceBookmarks()` to return the bookmark cache object (`_workspaceBookmarksCache`) instead of the workspace cache.
- Hardened `isBookmarkInAnotherWorkspace(...)` by:
  - validating bookmark/cache state before processing,
  - resolving the active workspace safely from cache,
  - ignoring stale/invalid workspace entries,
  - requiring valid bookmark-guid arrays before membership checks.

## Why
The previous cache return path and loose validation could lead to inconsistent workspace-bookmark detection, which caused UI state to incorrectly hide/show bookmarks in some multi-window scenarios.

## Scope
Only `src/zen/spaces/ZenSpaceManager.mjs` was changed.

### DIFF
--- a/src/zen/spaces/ZenSpaceManager.mjs
+++ b/src/zen/spaces/ZenSpaceManager.mjs
@@ -153,7 +153,7 @@ class nsZenWorkspaces {
     if (!this.privateWindowOrDisabled) {
       const observerFunction = async () => {
         delete this._workspaceBookmarksCache;
-        await this.workspaceBookmarks();
+        await this.#ensureWorkspaceBookmarksCache();
         this._invalidateBookmarkContainers();
       };
       Services.obs.addObserver(observerFunction, "workspace-bookmarks-updated");
@@ -746,17 +746,17 @@ class nsZenWorkspaces {
     return spacesForSS;
   }
 
-  async workspaceBookmarks() {
+  async #ensureWorkspaceBookmarksCache() {
     if (this.privateWindowOrDisabled) {
       this._workspaceBookmarksCache = {
         bookmarks: [],
         lastChangeTimestamp: 0,
       };
-      return this._workspaceBookmarksCache;
+      return;
     }
 
     if (this._workspaceBookmarksCache) {
-      return this._workspaceBookmarksCache;
+      return;
     }
 
     const [bookmarks, lastChangeTimestamp] = await Promise.all([
@@ -765,8 +765,6 @@ class nsZenWorkspaces {
     ]);
 
     this._workspaceBookmarksCache = { bookmarks, lastChangeTimestamp };
-
-    return this._workspaceBookmarksCache;
   }
 
   restoreWorkspacesFromSessionStore(aWinData = {}) {
@@ -826,7 +824,7 @@ class nsZenWorkspaces {
     return (async () => {
       await this.#waitForPromises();
       this.#afterLoadInit();
-      await this.workspaceBookmarks();
+      await this.#ensureWorkspaceBookmarksCache();
       await this.changeWorkspace(activeWorkspace, { onInit: true });
       this.#fixTabPositions();
       this.onWindowResize();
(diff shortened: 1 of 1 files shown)


src/zen/spaces/ZenSpaceManager.mjs
```diff
@@ -766,7 +766,7 @@ class nsZenWorkspaces {
 
     this._workspaceBookmarksCache = { bookmarks, lastChangeTimestamp };
 
-    return this._workspaceCache;
+    return this._workspaceBookmarksCache;
```
mr-cheffy: I dont think this function's value is read anywhere, could you please remove this and rename the function?

src/zen/spaces/ZenSpaceManager.mjs
```diff
@@ -3017,23 +3017,43 @@ class nsZenWorkspaces {
   }
 
   isBookmarkInAnotherWorkspace(bookmark) {
-    if (!this._workspaceBookmarksCache?.bookmarks) {
+    const workspaceBookmarks = this._workspaceBookmarksCache?.bookmarks;
+    if (!workspaceBookmarks || typeof workspaceBookmarks !== "object") {
```
mr-cheffy: second condition already takes it into account?
```suggestion
    if (typeof workspaceBookmarks !== "object") {
```

mr-cheffy (changes requested):

src/zen/spaces/ZenSpaceManager.mjs
TheGlorySaint: Thanks for the suggestion. Hopefully i did unterstand the changes you wanted correct

src/zen/spaces/ZenSpaceManager.mjs
TheGlorySaint: Thanks for the suggestion. Implemented these changes ✅

src/zen/spaces/ZenSpaceManager.mjs
```diff
@@ -3017,23 +3015,43 @@ class nsZenWorkspaces {
...
   isBookmarkInAnotherWorkspace(bookmark) {
-    if (!this._workspaceBookmarksCache?.bookmarks) {
+    const workspaceBookmarks = this._workspaceBookmarksCache?.bookmarks;
+    if (typeof workspaceBookmarks !== "object") {
       return false;
     }
+
     const bookmarkGuid = bookmark.bookmarkGuid;
-    const activeWorkspaceUuid = this.activeWorkspace;
+    if (!bookmarkGuid) {
```
mr-cheffy, line 3024: In what scenario is bookmarkguid null?

src/zen/spaces/ZenSpaceManager.mjs
```diff
@@ -3017,23 +3015,43 @@ class nsZenWorkspaces {
...
+      return false;
+    }
+
+    const activeWorkspaceUuid = this.getActiveWorkspaceFromCache()?.uuid;
+    if (!activeWorkspaceUuid) {
```
mr-cheffy, line 3029: Why wouldn't there be no id?

src/zen/spaces/ZenSpaceManager.mjs
```diff
@@ -3017,23 +3015,43 @@ class nsZenWorkspaces {
...
+      return false;
+    }
+
+    const validWorkspaceIds = new Set(
```
mr-cheffy, line 3033: Why's this needed?

mr-cheffy (changes requested):
Sorry, is this AI generated? some of these changes dont make lots of sense

TheGlorySaint:
> Sorry, is this AI generated? some of these changes dont make lots of sense

Yes, I AI-Generated with Claude Code Opus. Don't know JS or MJS but the Bug i reported over a month ago is really annoying or hinders me from using Zen

mr-cheffy:
Have you at least tested it?

mr-cheffy:
Im sorry but this PR clearly doesn't fix anything and its really annoying because it takes time and effort to review these where as I could've spent time actually fixing bugs

PR #12842, opened 2026-03-20
fix: persist activeZenSpace in sidebar session data

### LINKED ISSUE #12651: Zen always opens "Default" space
### Preliminary Checks

### What happened?

When opening Zen, it always opens the "Default" space. In this "Default" space I cannot edit the theme or change to other spaces from the bottom of the tab section.
Once I do open my desired space, on closing Zen, it will again open into this "Default" space, not the desired one. 

### Expected behavior

Zen should open the desired, last active space

### Actual behavior

Zen opens a "Default" workspace on launch, in which I cannot edit the theme or switch to other spaces easily

### Steps to reproduce

1. Open Zen with a non-"Default" workspace
2. Close Zen 
3. Reopen 

### Screenshots and videos

Image

### Version

1.19.1b

Other

Workspaces

```shell

```

wcork:
Persist activeZenSpace in the sidebar session object (zen-sessions.jsonlz4) so the last active workspace is restored on startup.

Fixes #12651

### DIFF
--- a/src/zen/sessionstore/ZenSessionManager.sys.mjs
+++ b/src/zen/sessionstore/ZenSessionManager.sys.mjs
@@ -755,6 +755,7 @@ export class nsZenSessionManager {
     sidebarData.splitViewData = firstWindow.splitViewData;
     sidebarData.groups = firstWindow.groups;
     sidebarData.spaces = firstWindow.spaces;
+    sidebarData.activeZenSpace = firstWindow.activeZenSpace;
   }
 
   /**
@@ -801,6 +802,9 @@ export class nsZenSessionManager {
     // Folders are always pinned, so we dont need to check for the pinned state here.
     aWindowData.folders = sidebar.folders;
     aWindowData.spaces = sidebar.spaces;
+    if (!aWindowData.activeZenSpace && sidebar.activeZenSpace) {
+      aWindowData.activeZenSpace = sidebar.activeZenSpace;
+    }
     this.log("Restored sidebar data into window", {
       tabs: aWindowData.tabs?.length || 0,
       groups: aWindowData.groups?.length || 0,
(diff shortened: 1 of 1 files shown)


mr-cheffy:
This is already handled here though?

https://github.com/zen-browser/desktop/blob/9bd76d3ad0b3390129544a3de286d5f97ff732af/src/browser/components/sessionstore/SessionStore-sys-mjs.patch#L191

This isnt related to the issue anyways, OP is reporting how it always opens a blank window, not that the space being restored is wrong

wcork:
> This is already handled here though?
> 
> https://github.com/zen-browser/desktop/blob/9bd76d3ad0b3390129544a3de286d5f97ff732af/src/browser/components/sessionstore/SessionStore-sys-mjs.patch#L191
> 
> This isnt related to the issue anyways, OP is reporting how it always opens a blank window, not that the space being restored is wrong

This appears to assign the active workspace to individual tabs when saved. This does not persist which workspace was active so that it may be restored.

In the case of the issue, "Default" workspace is a blank one when the session cannot be restored.

Default is the fallback name:

https://github.com/zen-browser/desktop/blob/9bd76d3ad0b3390129544a3de286d5f97ff732af/src/zen/spaces/ZenSpaceManager.mjs#L2531

So when the active workspace UUID doesn't survive the restore, this fallback path can create a brand new workspace named "Default" which is what I and OP of the issue see.

https://github.com/zen-browser/desktop/blob/9bd76d3ad0b3390129544a3de286d5f97ff732af/src/zen/spaces/ZenSpaceManager.mjs#L659

mr-cheffy:
No, default workspace only appears for blank windows, we need to figure out why that window has been marked as blank to begin with

PR #12876, opened 2026-03-22
gh-12897: prevent live folder dismissed tabs from reappearing

### LINKED ISSUE #12897: Live folder dismissed tabs reappear on startup
### Preliminary Checks

### What happened?

When using live folders (e.g. GitHub pull requests), dismissed tabs reappear after restarting Zen. The user closes a PR tab, but on a subsequent startup, all previously dismissed PRs show up again.

## Fix 

Guard the dismissed-items cleanup with itemIds.size > 0 so that an empty fetch result does not wipe all dismissals. See #12876 

### Expected behavior

When a PRs is dismissed, it shouldn't re-appear again

### Actual behavior

dismissed tabs periodically re-appear and they have to be dismissed over and over again

### Steps to reproduce

1. Create a GitHub pull requests live folder
2. Wait for PRs to load
3. Dismiss (close) several PR tabs
4. Wait for a background fetch that returns empty results (e.g. GitHub auth expires between fetches)
5. Restart Zen — all previously dismissed PR tabs reappear

### Screenshots and videos

_No response_

### Version

1.19.3b

macOS - aarch64

Tab Folders

```shell

```

genu:
Dismissed live folder tabs (e.g. GitHub PRs) would reappear on startup. Two issues caused this:

1. **Empty fetch wipes dismissals**: `onLiveFolderFetch` cleaned up dismissed items by removing any not in the current fetch results. If a fetch returned an empty list (auth expired, HTML structure changed), all dismissals were cleared — causing every PR to reappear on the next successful fetch.

2. **`saveState` early return**: If any single folder's DOM element wasn't found, the entire `saveState` method returned early, preventing dismissed items for all folders from being persisted.

Fix: guard the dismissed-items cleanup with `itemIds.size > 0`, and change `return` to `continue` in `saveState`.

### DIFF
--- a/src/zen/live-folders/ZenLiveFoldersManager.sys.mjs
+++ b/src/zen/live-folders/ZenLiveFoldersManager.sys.mjs
@@ -384,13 +384,18 @@ class nsZenLiveFoldersManager {
       animate: !folder.collapsed,
     });
 
-    // Remove the dismissed items that are no longer in the given list
-    for (const dismissedItemId of this.dismissedItems) {
-      if (
-        dismissedItemId.startsWith(`${liveFolder.id}:`) &&
-        !itemIds.has(dismissedItemId)
-      ) {
-        this.dismissedItems.delete(dismissedItemId);
+    // Remove the dismissed items that are no longer in the given list.
+    // Only do this when the fetch returned results — an empty list may
+    // indicate a transient failure (e.g. auth expired, HTML changed)
+    // and we must not wipe all dismissals in that case.
+    if (itemIds.size > 0) {
+      for (const dismissedItemId of this.dismissedItems) {
+        if (
+          dismissedItemId.startsWith(`${liveFolder.id}:`) &&
+          !itemIds.has(dismissedItemId)
+        ) {
+          this.dismissedItems.delete(dismissedItemId);
+        }
       }
     }
 


src/zen/live-folders/ZenLiveFoldersManager.sys.mjs
```diff
@@ -543,8 +548,10 @@ class nsZenLiveFoldersManager {
 
       const folder = this.getFolderForLiveFolder(liveFolder);
       if (!folder) {
-        // Assume browser is quitting.
-        return;
+        // Skip this folder but keep saving the rest. Using `return`
+        // here would abort the entire write, preventing dismissed
+        // items for *all* folders from being persisted to disk.
+        continue;
```
Slowlife01: What about on browser quit? But I still don't see any reasons why it should continue since this should never happen if it does it might be other bugs

genu: Reverted the return → continue change per feedback. You're right — during normal operation the folder should always be found, and during quit we don't want to write partial data. The previous save on disk is already correct. The fix now only guards the dismissed-items cleanup against empty fetch results.

mr-cheffy (approved):
Does this even run if there has been an error while fetching? Also, is there an existing issue for this? If not, please create one if you dont mind

genu:
To answer your question — no, this cleanup code does not run when `fetchItems()` returns an error. Error results (strings like `"zen-live-folder-github-no-auth"`) cause `onLiveFolderFetch` to return early at the top of the method, before reaching the cleanup loop.

The guard is needed for a different case: when the fetch *succeeds* (HTTP 200) but the HTML does not contain the expected structure — e.g. GitHub auth expired with a redirect instead of a 404, or a rate-limiting page. In that case `fetchItems()` returns an empty array (not an error string), and without the guard, the cleanup loop would wipe all dismissals.

Created #12894 for tracking.

mr-cheffy:
Your explanation doesn't make much sense and it just looks like something AI would spit out to justify the change (`GitHub auth expired with a redirect instead of a 404`. In that case, the response code wouldnt be 200 anyways), but it doesn't seem to make much harm so its fine.

PR #12893, opened 2026-03-23
gh-12892 Fix a split-vew tab shrinks when pressing on close button

### LINKED ISSUE #12892: Split-view tab shrinks when pressing close button
### Preliminary Checks

### What happened?

For an ordinary tab, pressing close button cause the button to shrink while the tab per se keeps in same size. 
But for a split-view tab, tab shrinks anyway whether you press close button or tab itself.

### Expected behavior

A split-view tab should not shrink when press its close button

### Actual behavior

 A split-view tab shrinks anyway whether you press close button or tab itself.

### Steps to reproduce

1. Open two tabs, and make them into split-view.
2. Press the close button of split-view tab.

### Screenshots and videos

_No response_

### Version

1.19.3b (Firefox 148.0.2)

Windows - x64

Split View

```shell

```

sporocyst:
This shoud fix #12892

Implementing a condition in CSS rule to prevent split-view tab scaling down when close button is pressed.

### DIFF
--- a/src/zen/split-view/zen-split-group.inc.css
+++ b/src/zen/split-view/zen-split-group.inc.css
@@ -120,7 +120,7 @@ tab-group[split-view-group] {
       }
     }
 
-    &:active {
+    &:active:not(:has(.tab-content > image:active)) {
       scale: var(--zen-active-tab-scale);
     }
 


mr-cheffy (approved):
Not very in favor of using `:has` but my guess is that its not very harmful here. Thanks!

PR #12910, opened 2026-03-24
refactor: Switch to box for pinned tabs

CosmoCreeper:
Switch to box for pinned tabs, allowing CSS mods to modify flex-direction (typically locked by XUL styling).
This change is specifically to allow SuperPins to be less buggy when it comes to the auto-grow feature for pinned tabs.


src/zen/spaces/ZenSpace.mjs
```diff
@@ -61,7 +61,7 @@ export class nsZenWorkspace extends MozXULElement {
           <toolbarbutton class="toolbarbutton-1 chromeclass-toolbar-additional zen-workspaces-actions" context="zenWorkspaceMoreActions" />
         </vbox>
         <arrowscrollbox orient="vertical" class="workspace-arrowscrollbox">
-          <vbox class="zen-workspace-tabs-section zen-workspace-pinned-tabs-section" hide-separator="true">
+          <box class="zen-workspace-tabs-section zen-workspace-pinned-tabs-section" hide-separator="true">
```
mr-cheffy, line 64: Please use `html:div`

mr-cheffy (changes requested):

mr-cheffy:
Please properly test this

CosmoCreeper:
I did. Not sure what the problem is, it works on my build.

EDIT: Oh. Why is there a change i don't remember making. 😭 Must've been an accidental selection.

Issue #10409, opened 2025-09-17
Opening links from external apps creates a new window but doesn’t redirect to the target URL


ahmedsomaa:
### Preliminary Checks

- [x] I have read and understood the important section above.
- [x] I have searched existing issues and avoided creating duplicates.
- [x] I am not filing an enhancement request.
- [x] I have checked that this issue cannot be reproduced on Mozilla Firefox.
- [x] I have checked that this issue can be reproduced once I removed all my Mods and Custom CSS.

### What happened?

Whenever I try to open a link from an external desktop app (e.g., Slack, Telegram), Zen Browser opens a new window but does not navigate to the clicked link. The window stays on the default homepage/start page instead of redirecting to the intended URL.

On Firefox

https://github.com/user-attachments/assets/9f840251-fe1f-466a-8d75-c298c0471b31

On Zen

https://github.com/user-attachments/assets/998fa31c-d5d0-434e-9e4a-f85c6369e80d

### Expected behavior

When clicking a link from an external desktop app, Zen Browser should open a new window (or tab, if configured) and automatically navigate to the target URL of the clicked link.

### Actual behavior

Instead of opening the clicked link, Zen Browser only opens a new window that stays on the default start/homepage. The requested URL is never loaded.

### Steps to reproduce

1. Open any desktop app that can launch external links (e.g., Slack, Telegram).
2. Click on a hyperlink inside the app.
3. Zen Browser launches a new window.
4. Observe that the window opens but does not redirect to the clicked link — it stays on the default start/homepage.

Video
https://github.com/user-attachments/assets/998fa31c-d5d0-434e-9e4a-f85c6369e80d

### Screenshots and videos

https://github.com/user-attachments/assets/998fa31c-d5d0-434e-9e4a-f85c6369e80d

### Version

1.15.5b

### What platform are you seeing the problem on?

Linux (Tarball)

### What component is this issue related to?

Other

### Relevant log output if applicable

```shell

```

garderobier:
I can confirm this bug on 1.17.3b on macOS 26.0.1 (25A362). So its not just an issue on Linux.

cameronjlarsen:
Experiencing a variation of this issue on windows using 1.17.3b. External links open in a new window but will load the url. I've tried forcing opening external links via browser.link.open_newwindow.override.external = 3

OS Info
Edition	Windows 11 Pro
Version	24H2
Installed on	‎9/‎23/‎2024
OS build	26100.6899
[...]

Issue #9292, opened 2025-07-02
Zen not opening links from another application in Ubuntu


ghost:
### Preliminary Checks

- [x] I have read and understood the important section above.
- [x] I have searched existing issues and avoided creating duplicates.
- [x] I am not filing an enhancement request.
- [x] I have checked that this issue cannot be reproduced on Mozilla Firefox.
- [x] I have checked that this issue can be reproduced once I removed all my Mods and Custom CSS.

### What happened?

When another application (VS Code or Zed for example) tries to open a link for authentication with Github or something, Zen opens a new window and do nothing else. Just a black window with a floating URL  bar.

### Expected behavior

Zen opens exact link form the given application

### Actual behavior

Zen catn't opensexact link form the given application and opens a new black different window.

### Steps to reproduce

- Try using Zed Editor to sign in with Github.
- Or try any other application that offers sign-in with the browser.

### Version

1.14.1b

### What platform are you seeing the problem on?

Linux (Tarball)

### What component is this issue related to?

Other

### Relevant log output if applicable

```shell

```

NilsMoller:
Happening to me on Linux Mint as well.

raphraph:
Same here on Ubuntu 24.04.3 LTS and Zen 1.15.3b (Firefox 142.0.1).

I checked and it's a second Zen window, not a second process.
The window has same spaces, pinned tabs/folders, but each space has no "current" tabs.

lethib:
I have the same issue on macOS 15.5. Zen browser version: 1.15.5b (Firefox 142.0.1)

calt-laboratory:
I have the same issue on Ubuntu 24.04.3 LTS.
For example, when I want to open a link from Thunderbird, Telegram, or Signal, a second Zen window opens.

skcb64:
Adding another testimony to this issue continuing to exist; on Mint 22.2 (Ubuntu 24.04).

raphraph:
Maybe someone from the dev team (@mr-cheffy ?) can tell us if we can do anything to narrow down the problem ?

This problem is a no-go for many users... I've been right-clicking and copying links outside Zen several times a day for 3 months now and it's a pain. I'll be glad to help in any way, but unfortunately I'm not a developer.

mr-cheffy:
@Propheticus any ideas?

Propheticus:
As a first check, because the same happens between Zen windows when opening a (new) profile editing page.
[...]

PR #12963, opened 2026-03-28
fix external app links not working on linux

savashn:
Related to #9292

This PR adds %u to the Exec line in zen.desktop file if missing, according to [#10409 (comment)](https://github.com/zen-browser/desktop/issues/10409#issuecomment-3592503811)

I faced with this problem and adding %u fixed it. So I wanted to contribute with a code block that checks zen.desktop file once on startup and sets a pref to avoid rewriting the file on every launch. Changes are tested on Fedora 43 KDE.

Additionally: Tested also the Flatpak version and didn't face with the same issue, so the issue is related to tarball version I think. I don't know how to fix for Windows and MacOS, so this PR for Linux only.

Thank you for this great project.

### DIFF
--- a/src/zen/common/modules/ZenStartup.mjs
+++ b/src/zen/common/modules/ZenStartup.mjs
@@ -59,6 +59,7 @@ class ZenStartup {
       setTimeout(() => {
         gZenUIManager.init();
         this.#checkForWelcomePage();
+        this.#editZenDesktopLinux();
       }, 0);
     } catch (e) {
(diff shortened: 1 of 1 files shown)


mr-cheffy:
Why not just edit the .desktop file directly? This seems very hacky and it assume the file is named zen.desktop anyway. Sorry, but I dont think this is the way of fixing it. Open to discuss about it though!

savashn:
I didn't know the filename could be anything different from zen.desktop, thought it's a standard name. Sorry for the noise.

PR #12965, opened 2026-03-28
gh-12104: Fix notification tabs overlap

### LINKED ISSUE #12104: Notification can't be closed when tabs are on the right side (mailto notify)
### Preliminary Checks

### What happened?

Notification: add "name of the application" as an application for mailto links

Hello,
I receive this notification every time I open Outlook. It cannot be closed because the tabs are located on the right side of the screen. When I hover over the “X” on the notification, the tabs respond to hover and clicks instead (I accidentally closed a few tabs this way).

### Expected behavior

When the tabs are on the right side, the notification should not cover them, or it should be prioritized for user interaction.

### Actual behavior

Tabs respond to hover and clicks instead of the pop-up

### Steps to reproduce

Open web-based e-mail client.

### Screenshots and videos

Image

### Version

1.18.2b (Firefox 147.0.2) (aarch64)

macOS - aarch64

Tabs

```shell

```

Rugved-dev18:
## Problem

When tabs are positioned vertically (e.g., on the right side), notifications overlap the tab area but fail to capture user interactions. Hover and click events are instead handled by the tabs, making the notification difficult or impossible to interact with.

## Cause

The notification deck (tab-notification-deck-template) is added to zen-appcontent-wrapper, which can put it in a layer that clashes with the tab/sidebar UI, especially when vertical tabs are on.

## Solution

1. Changed where the notification deck is put in during startup.
2. When the deck is available, it is now added to the top of the browser container, making sure it is above the tab/sidebar layer.
3. If necessary, it went back to zen-appcontent-wrapper to keep the original behavior.

## Changes

Update logic in ZenStartup.mjs to place the notification deck in a more appropriate container.

## Result

Notifications are now fully interactive, and there are no problems with vertical tabs. The fix is cleaner and stronger because it addresses the root cause instead of relying on styling overrides.

## Notes

Thanks @mr-cheffy  for pointing me to ZenStartup.mjs , that helped identify the root cause.

This change is based on analyzing the notification deck placement and how it interacts with the tab layout. Please let me know if you’d like me to test or adjust further.

### DIFF
--- a/src/zen/common/modules/ZenStartup.mjs
+++ b/src/zen/common/modules/ZenStartup.mjs
@@ -46,17 +46,14 @@ class ZenStartup {
         }
         newContainer.appendChild(node);
       }
-     // Fix notification deck
-const deckTemplate = document.getElementById(
-  "tab-notification-deck-template"
-);
+      // Fix notification deck
+      const deckTemplate =
+        document.getElementById("tab-notification-deck-template") ||
+        document.getElementById("tab-notification-deck");
 
-if (deckTemplate) {
-  // overlap and interaction issues with vertical tabs
-  const browser = document.getElementById("browser");
-  browser.prepend(deckTemplate);
-}
-      
+      // overlap and interaction issues with vertical tabs
+      document.getElementById("browser").prepend(deckTemplate);
+
       gZenWorkspaces.init();
       setTimeout(() => {
         gZenUIManager.init();
--- a/src/zen/common/modules/ZenStartup.mjs
+++ b/src/zen/common/modules/ZenStartup.mjs
@@ -46,23 +46,17 @@ class ZenStartup {
         }
         newContainer.appendChild(node);
       }
-      // Fix notification deck
-      const deckTemplate = document.getElementById(
-        "tab-notification-deck-template"
-      );
+     // Fix notification deck
+const deckTemplate = document.getElementById(
+  "tab-notification-deck-template"
+);
 
-      if (deckTemplate) {
-        // overlap and interaction issues with vertical tabs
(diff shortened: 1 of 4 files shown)


src/zen/common/modules/ZenStartup.mjs
```diff
@@ -46,15 +46,23 @@ class ZenStartup {
...
         "tab-notification-deck-template"
       );
+
       if (deckTemplate) {
-        document.getElementById("zen-appcontent-wrapper").prepend(deckTemplate);
+        // overlap and interaction issues with vertical tabs
+        const browser = document.getElementById("browser");
+        const wrapper = document.getElementById("zen-appcontent-wrapper");
+
+        if (browser) {
```
mr-cheffy: Why would browser not exist?

mr-cheffy (changes requested):

src/zen/common/modules/ZenStartup.mjs
Rugved-dev18: I looked into the notification deck logic and changed the insertion point so that it goes in the browser container instead of the zen-appcontent-wrapper.  It will still work the way it did before. This stops the problems with vertical tabs overlapping and interacting with each other without needing to change CSS.

mr-cheffy: In what scenario will #browser not exist though?

Rugved-dev18: ok, I will remove the fallback and simplify the logic to always use #browser for inserting the notification deck.

mr-cheffy (approved):

PR #13044, opened 2026-04-01
Live Folders: support GitHub Enterprise instances

gcsecsey:
## Related issues

Fixes https://github.com/zen-browser/desktop/discussions/12768

## Proposed Changes

The GitHub Live Folder provider currently hardcodes `https://github.com` as its only source, preventing users with GitHub Enterprise (GHE) instances from using the feature.

This PR adds full GitHub Enterprise support with two authentication strategies:

**1. Configurable GitHub host per folder**
- Each GitHub live folder targets a configurable instance (defaults to `github.com`)
- First folder creation skips the prompt for a frictionless default experience; subsequent folders prompt for the host URL
- "Instance: hostname" context menu option allows changing the host after creation
- Folder label includes the hostname for non-github.com instances (e.g., "Pull Requests (github.corp.com)")

**2. Personal Access Token (PAT) authentication for GHE**
- For GHE instances, cookie-based auth won't work, so PAT authentication is supported via the GitHub REST API (`/search/issues`)
- Tokens are stored securely in Firefox's Login Manager (encrypted, supports Primary Password) — never in the JSON state file
- "Set Access Token…" / "Remove Access Token" context menu options for token management
- Clicking the error triangle on a GHE folder opens the PAT creation page with pre-selected `repo` scope
- Token expiry is handled gracefully: 401/403 clears the stored token and shows an actionable error
- When a PAT is available, fetching switches from HTML scraping to the REST API (more reliable and future-proof)

**3. Security hardening**
- HTTPS-only for host URLs (HTTP rejected to prevent cleartext PAT transmission)
- `_hasToken` runtime flag excluded from state file serialization
- Stored PAT cleaned up from Login Manager when folder is deleted (if no other folder shares the host)
- API queries parallelized with `Promise.all` for better performance

**4. Broader auth error detection**
- Any non-2xx HTTP response treated as auth failure (not just 404)
- GHE instances without a PAT skip HTML scraping entirely and show the auth error directly
- Empty API results (authenticated but no matching items) no longer incorrectly treated as auth failures

### Files changed

| File | Purpose |
|------|---------|
| `providers/GithubAuth.sys.mjs` | **New** — Secure PAT storage CRUD via Firefox Login Manager |
| `providers/GithubLiveFolder.sys.mjs` | Host state, REST API fetch path, PAT UI options, GHE-specific error handling |
| `ZenLiveFolder.sys.mjs` | Added `headers` parameter to base `fetch()` method |
| `ZenLiveFoldersManager.sys.mjs` | Conditional host prompt, PAT cleanup on folder delete |
| `moz.build` | Register new `GithubAuth.sys.mjs` module |
| `zen-live-folders.ftl` | Localization strings for instance prompt, PAT options, error messages |
| `browser_github_live_folder.js` | Tests for custom host, GHE auth flow, error detection |

## Testing Instructions

1. **Default github.com flow** — Create a GitHub PR live folder via right-click sidebar → Create Live Folder → Pull Requests. Should work as before with no prompt (defaults to github.com).
2. **Add a GHE instance** — Create a second PR folder. Enter your GHE URL in the prompt (e.g., `https://github.corp.com`). The PAT creation page should open on your GHE instance.
3. **Set the PAT** — Right-click the GHE folder → Live Folder Options → "Set Access Token…" → paste the token. PRs should appear after refresh.
4. **Verify error triangle** — For a GHE folder without a PAT, the triangle should appear. Clicking it opens the PAT creation page (no prompt window).
5. **Change instance** — Right-click → Live Folder Options → "Instance: hostname" → enter a different URL.
6. **Delete folder** — Delete the GHE folder. Verify the PAT is cleaned up (check `about:logins`).
7. **Run tests** — `npm run test live-folders`


mr-cheffy:
Not too sure about this, it certainly seems to add quite the level of complexity from the user's side...

mirai-toto:
> Not too sure about this, it certainly seems to add quite the level of complexity from the user's side...

What makes you think that? Is it the concept or the usage? I think if this feature is greenlit, we could have settings on a Live Folder when right-clicking it, where we could set a token or URL. 

Or is it the customization of Live Folders that feels too advanced?

gcsecsey:
@mr-cheffy thanks for the feedback. Yes, I was experimenting with using only cookie-based auth to make the UX just as simple as for github.com, but GHE instances are typically behind SSO/SAML, so the cookie-based approach is not reliable.

This requires us to manage the entry and storage of the PAT in some way, I'm not sure if  the current approach is the best UX. This is how the current flow looks like:

1. User creates a second GitHub live folder, and gets prompted for the host URL
CleanShot 2026-04-02 at 09 41 23@2x
2. Clicking the error triangle opens the GHE's "Create new token" page with pre-selected scopes
CleanShot 2026-04-02 at 09 42 48@2x
3. User then right-clicks the folder → "Set Access Token…" to paste the token
CleanShot 2026-04-02 at 09 46 54@2x
CleanShot 2026-04-02 at 09 47 00@2x
4. Once the PAT is set, the user can remove it by right-clicking the folder → "Remove Access Token" then set a new one
CleanShot 2026-04-02 at 09 53 34@2x

Could you advise how we can make this flow easier to use?

mr-cheffy:
> Or is it the customization of Live Folders that feels too advanced?

Yes, I feel like part of the magic behind live folders is that everything feels seeming-less... Give me some time to think about this, maybe there's a nicer solution

src/zen/live-folders/providers/GithubLiveFolder.sys.mjs
```diff
@@ -11,24 +19,44 @@ export class nsGithubLiveFolderProvider extends nsZenLiveFolderProvider {
...
-        (this.state.options.assignedMe ?? true) ||
-        (this.state.options.reviewRequested ?? false);
+      const token = await GithubTokenManager.getToken(this.state.host);
+      this.state._hasToken = !!token;
+      if (token) {
+        return this.#fetchItemsViaApi(token);
+      }
 
-      if (!hasAnyFilterEnabled) {
+      // GHE instances require a PAT — HTML scraping won't work without cookies
```
mr-cheffy: We are sending cookies though?

mr-cheffy (changes requested):

src/zen/live-folders/providers/GithubLiveFolder.sys.mjs
gcsecsey: Not at the moment, I started working on this by testing my own use-case (SAML). However, GHE support [these 4 auth methods](https://docs.github.com/en/enterprise-server@3.4/admin/identity-and-access-management/managing-iam-for-your-enterprise/about-authentication-for-your-enterprise#built-in-authentication), and for instances using the built in auth, the same flow should work that already works with github.com.

I'll update the PR to try to use cookies first, and only fall back to requiring a PAT if that approach fails. Unfortunately, AFAIK there's no way to detect the auth method from the client side.

gcsecsey: Sorry for the delay, I updated the PR. I could actually test that this works with the GHE instance I have access to, without setting a PAT:

CleanShot 2026-04-28 at 12 28 34@2x

However, for GHE instances where different auth methods are used, the PAT approach is still required.

gcsecsey: @mr-cheffy could you let me know if sending the cookies and falling back to PAT if fetching through cookies didn't work is a good UX in your opinion? If not, what can we do to make it better?

gcsecsey: @mr-cheffy I'd like to progress this, but for that I need to know what you think the UX blockers are. Could you please check the most recent commit on this PR and let me know? Thanks!

gcsecsey:
@mr-cheffy I'd still like to push this forward, but for that I need to know your feedback on the UX. Could you please check the most recent commit on this PR and let me know? Thanks!

PR #13080, opened 2026-04-04
gh-12112: auto-focus URL bar when `zen.urlbar.replace-newtab` is disabled

### LINKED ISSUE #12112: New window does not auto focus url bar with zen.urlbar.replace-newtab set to false
### Preliminary Checks

### What happened?

New windows, both standard and private do not autofocus the URL bar when zen.urlbar.replace-newtab is set to false. Possible regression or related to #9412.  Opening a regular new tab in the current window does not have this problem. I do not believe I have any other relevant config settings that would interfere with URL bar.  May have been present in both 1.18b but confirmed in 1.18.2b

### Expected behavior

Open a new window or private window -> immediately begin typing new URL

### Actual behavior

Open a new window or private window -> must hit Ctrl+L to focus -> begin typing new URL

### Steps to reproduce

Set `zen.urlbar.replace-newtab` to `false` in `about:config`.  Hit `Ctrl + N` or `Ctrl + Shift + P`, attempt to type in a new URL.

### Screenshots and videos

_No response_

### Version

1.18.2b

Windows - x64

URL Bar

```shell

```

qBitnaut:
Fixes #12112

When zen.urlbar.replace-newtab=false, opening a new window or new tab no longer auto-focused the URL bar.

- New window (Ctrl+N): selectStartPage() never called openLocation() because the `showed` flag is only set on the _tabToRemoveForEmpty path, which requires replace-newtab=true. Fix: track initialTabWasEmpty separately and use it to call openLocation().

- New tab (Ctrl+T): PR #13002 added && !_hasSetSingleToolbar to _adjustFocusAfterTabSwitch to prevent focus-stealing on tab switches, but this also blocked focus for new blank tabs. Fix: allow focus when the tab URL is a blank page (isBlankPageURL).

### DIFF
--- a/src/browser/components/tabbrowser/content/tabbrowser-js.patch
+++ b/src/browser/components/tabbrowser/content/tabbrowser-js.patch
@@ -191,7 +191,7 @@ index d88bc0e5570c8fd428a84fdf5af0f6bab1e2a636..be4bedce98f404325e547dd8a4e73e89
        // In full screen mode, only bother making the location bar visible
        // if the tab is a blank one.
 -      if (gURLBar.getBrowserState(newBrowser).urlbarFocused) {
-+      if (gURLBar.getBrowserState(newBrowser).urlbarFocused && (!gZenVerticalTabsManager._hasSetSingleToolbar || isBlankPageURL(newBrowser.currentURI?.spec))) {
++      if (gURLBar.getBrowserState(newBrowser).urlbarFocused && !gZenVerticalTabsManager._hasSetSingleToolbar) {
          let selectURL = () => {
            if (this._asyncTabSwitching) {
              // Set _awaitingSetURI flag to suppress popup notification
--- a/src/browser/components/tabbrowser/content/tabbrowser-js.patch
+++ b/src/browser/components/tabbrowser/content/tabbrowser-js.patch
@@ -191,7 +191,7 @@ index d88bc0e5570c8fd428a84fdf5af0f6bab1e2a636..be4bedce98f404325e547dd8a4e73e89
        // In full screen mode, only bother making the location bar visible
        // if the tab is a blank one.
 -      if (gURLBar.getBrowserState(newBrowser).urlbarFocused) {
-+      if (gURLBar.getBrowserState(newBrowser).urlbarFocused && !gZenVerticalTabsManager._hasSetSingleToolbar) {
++      if (gURLBar.getBrowserState(newBrowser).urlbarFocused && (!gZenVerticalTabsManager._hasSetSingleToolbar || isBlankPageURL(newBrowser.currentURI?.spec))) {
          let selectURL = () => {
            if (this._asyncTabSwitching) {
              // Set _awaitingSetURI flag to suppress popup notification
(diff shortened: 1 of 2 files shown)


src/browser/components/tabbrowser/content/tabbrowser-js.patch
```diff
@@ -191,7 +191,7 @@ index d88bc0e5570c8fd428a84fdf5af0f6bab1e2a636..be4bedce98f404325e547dd8a4e73e89
        // In full screen mode, only bother making the location bar visible
        // if the tab is a blank one.
 -      if (gURLBar.getBrowserState(newBrowser).urlbarFocused) {
-+      if (gURLBar.getBrowserState(newBrowser).urlbarFocused && !gZenVerticalTabsManager._hasSetSingleToolbar) {
++      if (gURLBar.getBrowserState(newBrowser).urlbarFocused && (!gZenVerticalTabsManager._hasSetSingleToolbar || isBlankPageURL(newBrowser.currentURI?.spec))) {
```
mr-cheffy: Why's this needed?

mr-cheffy (changes requested):

src/browser/components/tabbrowser/content/tabbrowser-js.patch
qBitnaut: This could be entirely wrong, but it did resolve the issue in my local build. _adjustFocusAfterTabSwitch is called for both tab switches and new blank tab opens; it doesn't distinguish between them.
                                                                                                         
PR #13002 added && !_hasSetSingleToolbar to fix focus-stealing when switching tabs in single-toolbar mode (correct fix). But when replace-newtab=false and the user opens a new tab (Ctrl+T), that new blank tab goes through the same code path with urlbarFocused=true. The !_hasSetSingleToolbar guard then blocks URL bar focus for it too, even though focusing the URL bar on a brand new blank tab is the expected behavior.

isBlankPageURL(newBrowser.currentURI?.spec) restores that case: for a regular tab switch to an existing page, isBlankPageURL is false so the single-toolbar guard still prevents focus-stealing; for a new blank tab, it's true so the guard is bypassed and the URL bar gets focused as expected.

mr-cheffy: The change is unrelated to this PR though, it shouldn't be affecting this. Specially since the PR got merged way after the OG issue appeared.

qBitnaut: You're right. That is separate from this issue. I've removed it from this PR and will file a separate issue for it.

src/zen/spaces/ZenSpaceManager.mjs
```diff
@@ -944,14 +946,17 @@ class nsZenWorkspaces {
     }
 
     showed &&= Services.prefs.getBoolPref("zen.urlbar.open-on-startup", true);
+    initialTabWasEmpty &&= Services.prefs.getBoolPref("zen.urlbar.open-on-startup", true);
```
mr-cheffy: We seem to be fetching `zen.urlbar.open-on-startup` twice here, could you please assign it to a variable and use that?

mr-cheffy (changes requested):

mr-cheffy (approved):

PR #13082, opened 2026-04-04
gh-13081: focus URL bar on new blank tab in single-toolbar mode

### LINKED ISSUE #13081: New tab (Ctrl+T) doesn't focus URL bar when replace-newtab=false (Possible regression from gh-13002)
### Preliminary Checks

### What happened?

After #13002 merged, opening a new tab with zen.urlbar.replace-newtab=false no longer auto-focuses the URL bar when Zen's single-toolbar mode is active.

### Expected behavior

Ctrl+T, New tab with New Tab Window enabled (urlbar-replace-newtab=false), URL bar is auto-focused when typing

### Actual behavior

URL bar is not focused

### Steps to reproduce

1. Set zen.urlbar.replace-newtab=false in about:config
2. Ensure single-toolbar mode is enabled
3. Open a new tab (Ctrl+T)

### Screenshots and videos

_No response_

### Version

1.19.6b

Windows - x64

Other

```shell

```

qBitnaut:
gh-13002 added `&& !_hasSetSingleToolbar` to _adjustFocusAfterTabSwitch to prevent URL bar focus-stealing on tab switches in single-toolbar mode. However, new blank tabs (Ctrl+T with replace-newtab=false) go through the same code path with urlbarFocused=true, so the guard incorrectly blocks URL bar focus for them too.

Allow focus when the tab's URL is a blank page so the fix only applies to actual tab switches, not new tab opens.


mr-cheffy (approved):
Looks good! But I would like a test about this if you don't mind. Thank you for the PR!

PR #13106, opened 2026-04-05
feat(tabs): Improve the UX of tab context menu  

Panemiko:
# Rework tab context menu with quick actions

Replaces the current tab context menu with a redesigned version focused on speed and intuition.

[zen.webm](https://github.com/user-attachments/assets/79fd4429-4213-4dfd-961f-fbd740cc875a)

## Why?

The current menu is a wall of text that forces users to read through every option to find what they want, even for the most frequent actions.

The new menu surfaces the most common actions as an icon bar at the top, so users can act before they even finish reading.

> **This is a draft PR.** The quick actions bar is working, but there a full context menu redesign is coming.

## Tasks

- [x] Create the quick actions section.
- [x] Remove redundant items from the menu.
- [x] Add icons to some of the other remarkable actions
- [ ] Make quick actions work with multiple tabs
- [ ] Reorder some of the items from the menu
- [ ] Apply icons to all variations of items
- [ ] Update localization

### Edge context menus
image

### Zen and Firefox page context menu
image


sharmaanuj334:
can you also add the tooltip while making these icons

Panemiko:
> can you also add the tooltip while making these icons

Already have it working. Will just need to update the localization at the end of the work

image

Panemiko:
Update

image

mr-cheffy:
Thank you for your efforts but I don't think we'll be going forward with this as icons in the context menu doesn't follow our design language (For some reason you are using filled icons in some places, outline icons in others and "emoji" icons as well?), it breaks user workflow/habits, and seems to be AI generated(?). Not sure how it will work out with macos and GTK as well, as we are using native context menus for those OSs.

Thanks again for your work on this. We'll pause here and won't be moving forward with this direction.

PR #13142, opened 2026-04-08
gh-11881: Fix orphaned PiP window when source tab closes

### LINKED ISSUE #11881: PiP window stays open empty and buttons don’t work
### Preliminary Checks

### What happened?

Sometimes when I close a Picture-in-Picture video, an empty PiP window stays stuck on top of the screen. The window has the PiP controls (close, minimize, etc.) but none of the buttons work and no video is shown.

- Happens: A Few times per week 
- Seems more likely when: Multiple tabs open  / multiple videos playing

### Expected behavior

The PiP window should close completely and not leave any empty or unresponsive window on screen.

### Actual behavior

An empty PiP window remains stuck on top of other windows. No controls respond (close, minimize, back-to-tab).

### Steps to reproduce

The bug is intermittent, but this is roughly what I’m doing when it occurs

1. Enable Picture-in-Picture.
2. Open YouTube and start a video.
3. Switch other tabs, PIP starts automatically 
5. Eventually, closing the PiP player leaves an empty floating window that can’t be closed.

### Screenshots and videos

Image

### Version

1.17.15b

macOS - aarch64

Other

```shell
TypeError: can't access property "setAttribute", tab is undefined PictureInPicture.sys.mjs:880:5
```

SomeoneWithOptions:
## Summary

Fixes #11881.

This fixes a Picture-in-Picture race where Zen can leave behind a blank, unresponsive, always-on-top PiP window if the originating tab disappears while the PiP window is still being initialized.

The failure path is:

1. `handlePictureInPictureRequest(...)` opens the PiP chrome window.
2. The source tab is closed or no longer resolves through `gBrowser.getTabForBrowser(browser)`.
3. `tab.setAttribute("pictureinpicture", true)` throws because `tab` is undefined.
4. `win.setupPlayer(...)` never runs, so the PiP shell remains open without a working player/controller.

## Changes

- Added `getPipTabForBrowser(browser)` to validate the source browser still maps to a live tab.
- Validate the source tab before opening the PiP window.
- Re-validate the source tab after `openPipWindow(...)`; if it disappeared, close the newly opened PiP shell before player setup.
- Keep PiP player count telemetry updates after the second validation so aborted shells are not counted as initialized players.
- Harden `focusTabAndClosePip(...)` so missing source tab/browser mappings close the PiP window instead of dereferencing `tab`.
- Harden `removePiPBrowserFromWeakMap(...)` for missing browser mappings.
- Preserve Zen workspace behavior by using `gZenWorkspaces.switchIfNeeded(browser)` for valid source tabs.

## Validation

- Verified the patch applies cleanly against Firefox `FIREFOX_149_0_2_RELEASE`.
- Ran `node --check` against the patched `PictureInPicture.sys.mjs`.

### DIFF
--- a/src/toolkit/components/pictureinpicture/PictureInPicture-sys-mjs.patch
+++ b/src/toolkit/components/pictureinpicture/PictureInPicture-sys-mjs.patch
@@ -2,7 +2,7 @@ diff --git a/toolkit/components/pictureinpicture/PictureInPicture.sys.mjs b/tool
 index 098742100858b266aebc8f764f918c85815f3c5f..71b8f8d71d48229fc96d6c84c635d09012a82250 100644
 --- a/toolkit/components/pictureinpicture/PictureInPicture.sys.mjs
 +++ b/toolkit/components/pictureinpicture/PictureInPicture.sys.mjs
-@@ -122,6 +122,9 @@ export class PictureInPictureToggleParent extends JSWindowActorParent {
+@@ -122,6 +122,9 @@
          if (browser.ownerGlobal.gBrowser.selectedBrowser == browser) {
            break;
          }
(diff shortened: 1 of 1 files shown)


mr-cheffy:
Thank you for your patch! However though, I feel like it would be better to send it to firefox so they can take a look at it and so they also have it fixed from their side :)

PR #13160, opened 2026-04-09
no-bug: Remove non-existent hooks property

CosmoCreeper:
There is no hooks property in package.json configurations. Husky will automatically find the pre-commit file from the .husky configuration folder.


mr-cheffy:
Shouldn't you also run the uninstall command for husky?

mr-cheffy (approved):

Issue #12721, opened 2026-03-10
Multiple Spaces (Workspaces) for Bookmarks and Bookmark folders no longer work


PEC-IT2026:
### Preliminary Checks

- [x] I have read and understood the important section above.
- [x] I have searched existing issues and avoided creating duplicates.
- [x] I am not filing an enhancement request.
- [x] I have checked that this issue cannot be reproduced on Mozilla Firefox.
- [x] I have checked that this issue can be reproduced once I removed all my Mods and Custom CSS.

### What happened?

On an older version of Zen, I was able to connect bookmark folders and bookmarks to multiple spaces.
On the newest version of Zen on a new computer, I sync'd my Bookmarks and then started setting up my spaces again. 
When editing a folder or bookmark, I am allowed to select more than one space.  **Clicking Save doesn't show any errors. Then, reopening the folder/bookmark settings show no spaces are mapped.**
By contrast, if I only select one space and save, the settings save correctly and persist. I am unable to connect another space after connecting to only one; it keeps my original selection and doesn't add the additional space I selected. 

From what I experienced it seems like the multiple space configuration for bookmarks stopped working. If there is something odd about this process to set up Zen on a new computer let me know: 

1. Install latest version of Zen
2. Sign in to get Bookmarks to sync
3. Set up Tab Containers
4. Set up spaces and connect them to Tab Containers
5. Customize Bookmarks to associate different folders and bookmarks with different spaces

I like this method of setup and it worked really well when it was working. For example I could have a main "Expenses" folder linked to two spaces, and then only show specific bookmarks to different spaces so I don't log into the incorrect account. 

### Expected behavior

Map a bookmark folder or single bookmark to more than one space

### Actual behavior

Multiple space selection results in no changes saved. Only one space selection is saving successfully. 

### Steps to reproduce

1. Set up at least 2 container tabs
2. Set up a space for every container tab and make that profile associate with each container tab
3. Edit a bookmark (or bookmark folder) and attempt to map that to more than one space and Save
4. Open bookmark settings to see the changes were not saved

### Version

1.19.1b

### What platform are you seeing the problem on?

Windows - x64

### What component is this issue related to?

Bookmarks

### Relevant log output if applicable

```shell

```

Dcfunkster:
[...]

PR #13184, opened 2026-04-09
fix: restore workspace bookmark scoping via compatibility layer 

### LINKED ISSUE #12134: Workspace bookmark
### Preliminary Checks

### What happened?

Every single day when I update the Zen, all bookmarks change.

I divide two workspaces: Daily and Work. Each workspace has its own bookmarks. 
However, when I update Zen, their workspace change. Their workspace features are gone and both bookmarks appear on both workspaces. I don't know why this happens, and also why there is no simple feature to change multiple bookmarks into the specific workspace so that I can recover my bookmark setting. I need to change each boorkmark's workspace "individually". 

I am currently using Zen in linux, windows, and mac.
This bug should be fixed as soon as possible in my opinion. Thank you.

### Expected behavior

Bookmark features change

### Actual behavior

Just after update Zen

### Steps to reproduce

Sometimes, after updates, it happens.

### Screenshots and videos

_No response_

### Version

1.17.15b

Windows - x64

Bookmarks

```shell

```

skpaz:
Fixes/should fix: https://github.com/zen-browser/desktop/issues/12134, https://github.com/zen-browser/desktop/issues/12721

Fixes an issue where bookmarks and bookmark folder assignments to newly-created workspaces do not persist.

Workspaces are now primarily stored in session state (`zen-sessions.jsonlz4`), while bookmark-to-workspace associations continue to rely on the `zen_workspaces` table in `places.sqlite`. Since new workspaces are no longer written to this table, bookmark mappings for them are not saved, causing folders to appear across all workspaces.

This change introduces a small compatibility layer that keeps the `zen_workspaces` table in sync with the current session state workspaces by:

- Adding `syncWorkspacesToPlaces()` to upsert active workspaces into `places.sqlite`
- Invoking this sync during workspace initialization and on workspace updates
- Avoiding unnecessary writes via a simple change detection guard

This preserves existing bookmark scoping behavior without altering the current session-based workspace model.

Steps to reproduce:

1. Create a new workspace.
2. Create a bookmark folder.
3. Assign folder to workspace.
4. Switch workspaces; folder appears globally.

After this change, I tested with a local build and confirmed:

- New workspaces are persisted to `zen_workspaces`.
- Bookmark assignments create rows in `zen_bookmarks_workspaces`.
- Folder visibility is correctly scoped.

### DIFF
--- a/src/zen/sessionstore/ZenSessionManager.sys.mjs
+++ b/src/zen/sessionstore/ZenSessionManager.sys.mjs
@@ -296,7 +296,7 @@ export class nsZenSessionManager {
     }
     if (
       Services.prefs.getBoolPref("zen.session-store.log-tab-entries", false)
-    ) {z
+    ) {
       for (const tab of this.#sidebar.tabs || []) {
         this.log("Tab entry in session file:", tab);
       }
--- a/src/zen/sessionstore/ZenSessionManager.sys.mjs
+++ b/src/zen/sessionstore/ZenSessionManager.sys.mjs
@@ -16,6 +16,7 @@ ChromeUtils.defineESModuleGetters(lazy, {
   gWindowSyncEnabled: "resource:///modules/zen/ZenWindowSync.sys.mjs",
   gSyncOnlyPinnedTabs: "resource:///modules/zen/ZenWindowSync.sys.mjs",
   DeferredTask: "resource://gre/modules/DeferredTask.sys.mjs",
+  PlacesUtils: "resource://gre/modules/PlacesUtils.sys.mjs",
 });
 
 XPCOMUtils.defineLazyPreferenceGetter(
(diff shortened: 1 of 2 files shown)


skpaz:
First PR for this repo. I checked some other PRs and didn't see a specific format, so I went with the basics. Please let me know if any additional information is needed.

mr-cheffy:
Hi, thank you for your contribution! Unfortunately, `zen_workspaces` has been deprecated long time ago and only exists as a potential backup when user upgrade their versions. I haven't looked deep into the issue yet but I don't feel comfortable with going back to storing the spaces in the SQL database (mainly due to development overhead). Sorry...

skpaz:
@mr-cheffy To be clear, it doesn't regress to storing workspaces in `zen_workspace`. It simply synchronizes existing session-based data with the old location so that bookmark assignment to workspaces functions like it should. It's a backwards-compatibility layer. 

This has been broken for some time and has several open issues. It's completely broken my own workflow.

mr-cheffy:
I understand, but im not exactly sure how storing them on the database fixes the issue? We shouldn't make the workspace_id in bookmarks be reliant on zen_workspaces table

skpaz:
When Zen switched to session-based workspaces, the functionality that allowed users to assign bookmarks and bookmark folders to specific workspaces was left out.

While the DB remained and worked as a backup as you describe, any new bookmarks or bookmark folders created after the switch could no longer be assigned to new or existing workspaces. Previously, this data was stored in a couple of tables in `places.sqlite`. 

This fix introduces a compatibility layer that takes the session data and synchronizes it with the old database so that a) the current session-based architecture remains unaffected and b) the old workspace/bookmark associations still work.

It's my understanding that fixing this exclusively using the new session-based model might require extensive refactoring and involves some core Firefox bookmark code. This layer or shim seemed like a more pragmatic approach until a real fix can be implemented.

Issue #7150, opened 2025-03-31
Blocking "Override keyboard shortcuts" permission disables delete and backspace keys


somnambulist247:
### Captchas

- [x] I have read the instructions.
- [x] I have searched existing issues and avoided creating duplicates.
- [x] I am not filing an enhancement request.
- [x] I have checked that this issue cannot be reproduced on Mozilla Firefox.
- [x] I have checked that this issue can be reproduced once I removed all my Mods and Custom CSS.

### History:
As i have come from Librewolf i tried setting "permissions.default.shortcuts" to 2. This default blocks websites from deciding what to do with my key-presses.

While websites like google docs need to override keyboard shortcuts like "Ctrl+S" to save the doc not the webpage some websites mess with my custom shortcuts and work flow so i enable and disable "override keyboard shortcuts" permission for individual websites.

### Problematic Behavior:
In zen browser blocking "override keyboard shortcuts" disables the use of backspace and delete keys in text fields. You can't use backspace nor delete keys to edit text. However only if u select the text you can use delete key to delete it.

### Expected Behavior:
Blocking "override keyboard shortcuts" permission does not hinder your ability to edit text. Text fields and the keys "delete" and "backspace" work as expected.

### Steps To Reproduce:
1. Open any website with a text field. (Example: [google](https://www.google.com/))
2. Press "Ctrl+i" to open page info window. (You can alternatively find it from security icon in address bar, clicking on "Connection Secure", clicking on more information)
3. Navigate to "Permissions" tab.
4. Scroll down and find the setting "Override keyboard shortcuts".
5. Uncheck "Use Default".
6. Select "Block".
7. Now select a text field in the website you opened in step 1.
8. Type anything.
9. Try to use keys "delete" and "backspace" to edit text.

### What could have caused this:
Zen browser comes with customizable shortcuts which includes "delete" key. According to a [bugzilla entry](https://bugzilla.mozilla.org/show_bug.cgi?id=1445942) non-reserved keys might be blocked by shortcut permissions. My uneducated guess is while making zen keyboard shortcuts "delete" and "backspace" keys were made non-reserved.

### Version

1.10.3b

### What platform are you seeing the problem on?

Windows - x64

### What component is this issue related to?

Keyboard Shortcuts

### Relevant log output if applicable

```shell

```

TV2-Lmor:
+1 Windows 11; Zen Browser 1.11.5b (Firefox 137.0.2) (64-bit)
[...]

Issue #6647, opened 2025-03-20
Shortcut Override Disables Backspace Input


darshvatsa:
### Captchas

- [x] I have read the instructions.
- [x] I have searched existing issues and avoided creating duplicates.
- [x] I am not filing an enhancement request.

### What happened?

In `about:config`, I set `permissions.default.shortcuts` to 2 in order to have my browser shortcuts override any website-specific shortcuts. While it did the job, it created a new issue where **pressing backspace won't actually delete any text on a lot of websites.**

I was initially unsure if this has to do with the `permissions.default.shortcuts` value change but when I changed it to 0, 1 or 3, the backspace was back to its normal functioning, although the browser-wide shortcut override won't work anymore.

Not sure if this might have something to do with the old Firefox feature where backspace functioned as a shortcut to go back. Just to be clear though, when my backspace was not working with `permissions.default.shortcuts` set to 2, it didn't lead to the old "go back" feature either. The backspace was completely unresponsive. 

### Reproducible?

- [ ] I have checked that this issue cannot be reproduced on Mozilla Firefox.

### Version

1.10b

### What platform are you seeing the problem on?

Windows

### Relevant log output if applicable

```shell

```

zen-browser-auto:
## Closing this issue as part of a repo cleanup

Due to technical-debt we had in the past, we are currently clearing out old, duplicate, stale, and non-actionable issues to improve issue tracking and make it easier to focus on fixing real bugs.

If this issue is still relevant, please open a new issue using our updated issue template, which will help us categorize and address it more efficiently.

Thanks for your understanding and for helping us improve the project! 🚀

This issue can be reopened at: https://github.com/zen-browser/desktop/issues/new?template=bug_report.yml&title=Shortcut%20Override%20Disables%20Backspace%20Input%0A

PR #13188, opened 2026-04-09
feat: add about:config pref to reserve specific keyboard shortcuts from page override

ben-tanen:
## Problem

This addresses #8894 — web pages (GitHub, TikTok, Google Docs, etc.) can intercept Zen's keyboard shortcuts before the browser handles them. The most common example: GitHub's "Copy raw file" button registers `data-hotkey="Meta+Shift+C"`, which steals Zen's Copy URL shortcut on file-view pages.

Firefox already has a mechanism for this — the `reserved` attribute on `<key>` elements. Shortcuts like Cmd+Q, Cmd+W, and Cmd+T are marked `reserved="true"` so pages can't override them. But none of Zen's custom shortcuts use this attribute, so they're all vulnerable to page-level interception.

## Approach

Rather than hardcoding `reserved="true"` on specific shortcuts (which is what #10422 attempted before being closed), this PR introduces a new `about:config` pref — `zen.keyboard.shortcuts.reserved` — that accepts a comma-separated list of shortcut IDs. Any shortcut ID in this list gets `reserved="true"` set on its `<key>` element at build time.

For example, setting the pref to `zen-copy-url,zen-copy-url-markdown` reserves just the Copy URL shortcuts from page override.

**Why a pref instead of a hardcoded change:**
- Default value is `""` (empty string), so this changes nothing for existing users
- Power users who are affected by #8894 can selectively reserve the shortcuts that conflict with their most-used sites
- It's granular — you can reserve just the shortcuts that are causing problems, rather than an all-or-nothing approach
- It sidesteps the web compatibility concern (Firefox intentionally leaves most shortcuts unprotected so pages can use them) by making it opt-in

## Changes

- **`prefs/firefox/browser.yaml`** — added `zen.keyboard.shortcuts.reserved` pref with empty string default
- **`src/zen/kbs/ZenKeyboardShortcuts.mjs`** — added `#isReservedByPref()` method to `KeyShortcut` class, called during `<key>` element creation. If the shortcut's ID is in the pref's comma-separated list, `reserved="true"` is set on the element.

## Testing

Tested locally on macOS with a dev build:
- Set `zen.keyboard.shortcuts.reserved` to `zen-copy-url,zen-copy-url-markdown` in `about:config`
- Restarted the browser
- Navigated to a GitHub file-view page (which registers `Meta+Shift+C` via `data-hotkey`)
- Cmd+Shift+C correctly copied the page URL (toast notification appeared, clipboard had the URL — not the raw file content)
- Verified Cmd+Shift+C still works on pages without conflicting hotkeys
- Verified Cmd+Shift+Alt+C (markdown copy) also works correctly
- With the pref set to `""`, behavior is unchanged from current — GitHub's hotkey takes priority as before

## Future considerations

- Currently requires a browser restart for pref changes to take effect (the pref is read when `<key>` elements are built during startup). A pref observer that triggers `_applyShortcuts()` on change could be added to make it live — happy to add that if it would help, but wanted to keep the diff small for now.
- Could also be surfaced in the Keyboard Shortcuts settings UI as a per-shortcut toggle, though that's a larger change.

### DIFF
--- a/prefs/firefox/browser.yaml
+++ b/prefs/firefox/browser.yaml
@@ -63,6 +63,9 @@
   locked: true
   value: false
 
+- name: zen.keyboard.shortcuts.reserved
+  value: ""
+
 - name: browser.tabs.closeWindowWithLastTab
   value: false
 
--- a/src/zen/kbs/ZenKeyboardShortcuts.mjs
+++ b/src/zen/kbs/ZenKeyboardShortcuts.mjs
@@ -452,8 +452,8 @@ class KeyShortcut {
     if (this.#disabled) {
       key.setAttribute("disabled", this.#disabled);
     }
-    if (this.#reserved) {
-      key.setAttribute("reserved", this.#reserved);
+    if (this.#reserved || this.#isReservedByPref()) {
+      key.setAttribute("reserved", "true");
     }
     if (this.#internal) {
       key.setAttribute("internal", this.#internal);
@@ -521,6 +521,14 @@ class KeyShortcut {
 
   isReserved() {
     return this.#reserved;
+  }
+
+  #isReservedByPref() {
+    const reserved = Services.prefs.getStringPref("zen.keyboard.shortcuts.reserved", "");
+    if (!reserved) {
+      return false;
+    }
+    return reserved.split(",").map(s => s.trim()).includes(this.#id);
   }
 
   isInternal() {


mr-cheffy:
There's already a config for this, its called `permissions.default.shortcuts`

ben-tanen:
yep, i'm aware of `permissions.default.shortcuts` — i looked into it before going this route. the issue is that setting it to `2` treats *all* unset-reserved shortcuts as reserved, which breaks backspace and delete keys in text fields (see #7150 and #6647). some folks have reported that `3` works a bit better, but it's not well-documented and still has edge cases.

the idea behind this pref is that it's per-shortcut rather than all-or-nothing. you can reserve just the specific shortcuts that are causing problems (e.g., `zen-copy-url`) without affecting text input or other page-level keyboard behavior. it sets `reserved="true"` directly on the `<key>` element for only the shortcuts you list, which is the same mechanism Firefox uses for Cmd+Q/W/T.

mr-cheffy:
I think we should rather ask for a confirmation to trigger the same shortcut again, just like what arc does

milch:
IMO it depends on the shortcut. I don't ever want some of my core shortcuts to be overridden at all, like cmd + S

ben-tanen:
I always found the Arc confirmation to trigger a reasonable way to inform you of when something was going to be overridden (or could be), especially for stuff like CMD+D or CMD+R in Google Sheets, but as @milch points out I do think there are certain core shortcuts that you never want overridden (CMD+S; for me, CMD+Shift+C; etc.)

My implementation here was meant to tackle the latter case in what seems to be a relatively simple opt-in way for power users (those discussing it/finding it in #8894) while leaving the option of doing the Arc style confirmation open for the future, since I think these can be complementary/exist in the same space.

mr-cheffy:
The issue with your implementation is that we then don't have a way of knowing which shortcut was originally reserved

ben-tanen:
do you mean that the implementation and application of `reserved` is based on the IDs `zen-copy-url` and `zen-copy-url-markdown` as opposed to the actual shortcuts that the user has them enabled to? By default `zen-copy-url` is mapped to CMD+Shift+C, but they could have remapped it to CMD+Shift+L, so the override over that would be unclear?

or do you mean something else?

i had originally thought about having the pref enable a sort of column of check boxes to the right of the Keyboard Shortcuts menu so you could go in and click/toggle the reserved application, but I thought that was elevating it too much for a standard user and it might confuse them–also more of an implementation change. But if you think that might work better, happy to try tackling something like that! 

image

mr-cheffy:
As you said, this is just a "workaround" once the confirmation toast is implemented, so why not just work on the final implementation than just workarounds? Let me think about this, i'll come back to you

milch:
Circling back here, any updates? IMO there are two competing requirements here, and this PR solves the second:

- A website overrides a some behavior that they think is a better fit than the default, e.g. cmd-r just reloading some part of the site rather than a full browser reload. In this case a confirmation toast is fine since you may be fine with the site's adjusted behavior if it matches your intent. 
- A website overrides some behavior to something completely different than the core browser shortcuts that I have muscle memory for and always want to retain. A confirmation toast would very much be an incomplete workaround for this. I don't want to have to press the shortcut twice, or see a confirmation. 

Based on the discussion in the linked issue, I believe people are asking for the second, not the first.

PR #13202, opened 2026-04-10
feat: Add ZenDevUrlDetector for localhost/dev URL detection

DannyAmzq:
## Summary

Added a dev URL indicator to the Zen URL bar, similar to what Arc Browser does when you're on a local dev server.

- Added a new `ZenDevUrlDetector` module that detects local/dev URLs (localhost, 127.0.0.1, `*.local`, `*.localhost`, `*.internal`, `*.test`, `file://`) and sets a `zen-dev-url` attribute on `<html>`
- Added CSS in `zen-omnibox.css` to show an orange outline and a slight background tint on the URL bar when the attribute is set
- Listens to tab switches and navigation in real time using `gBrowser.addTabsProgressListener`
- Can be turned off with the `zen.urlbar.show-dev-indicator` pref (default: `true`), changes apply right away without restarting

## Test plan

- [ ] Go to `localhost` or `127.0.0.1` and check that the URL bar gets an orange outline
- [ ] Go to `google.com` and check that the orange outline goes away
- [ ] Switch between a localhost tab and a regular tab and make sure the indicator updates correctly
- [ ] Set `zen.urlbar.show-dev-indicator = false` in `about:config` and check that the indicator disappears right away
- [ ] Check that `file://` URLs also trigger the indicator

### DIFF
--- a/src/zen/common/ZenPreloadedScripts.js
+++ b/src/zen/common/ZenPreloadedScripts.js
@@ -11,6 +11,7 @@
   ChromeUtils.importESModule("resource:///modules/zen/ZenSpaceManager.mjs", { global: "current" });
   ChromeUtils.importESModule("chrome://browser/content/zen-components/ZenCompactMode.mjs", { global: "current" });
   ChromeUtils.importESModule("chrome://browser/content/ZenUIManager.mjs", { global: "current" });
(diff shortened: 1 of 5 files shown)


mr-cheffy:
I feel like this is too much overhead for such simply a cosmetic change. I think we should have a more clear idea on how dev mode should be and work once we have some more functionality into it. Thanks for trying to contribute!

PR #13209, opened 2026-04-11
feat(folders): allow unpinned tabs in folders behind pref

doyled-it:
## Summary

Adds `zen.folders.allow-unpinned` preference (default `false`). When enabled, folders may contain unpinned tabs in the regular tab area, not only pinned tabs.

- Right-click a **pinned** tab → pinned folder (unchanged behavior)
- Right-click an **unpinned** tab → folder created in regular tab area, tabs stay unpinned
- Existing folders stay pinned — the pref only affects newly created folders
- Toggling the pref off does not retroactively pin existing unpinned folders

Default is `false`, so existing users see zero behavior change.

## Changes

1. **`prefs/zen/folders.yaml`** — new `zen.folders.allow-unpinned` pref (default `false`)
2. **`src/zen/folders/ZenFolder.mjs`** — `pinned` getter/setter backed by `zen-unpinned-folder` attribute + `_isUnpinnedFolder` instance flag initialized in `connectedCallback`; label element pinned state conditioned
3. **`src/zen/folders/ZenFolders.mjs` (`createFolder`)** — compute `makeUnpinned` from first tab's state; skip `gBrowser.pinTab()` for unpinned; select unpinned workspace strip as insertion container; create sentinel empty tab with matching pinned state; pass `pinned: !makeUnpinned` to `_createFolderNode`
4. **`src/zen/folders/ZenFolders.mjs` (`_createFolderNode`)** — honor `options.pinned === false` by setting `zen-unpinned-folder` attribute before DOM attachment
5. **`src/zen/folders/ZenFolders.mjs` (`on_TabOpen`)** — short-circuit for `_isUnpinnedFolder` groups so unpinned tabs aren't force-pinned
6. **`src/zen/drag-and-drop/ZenDragAndDrop.js`** — condition `#applyDragoverIndicator` guard on `draggedTab.group?.pinned` so unpinned folders can be drag-reordered within the regular tab strip

## Session restore

No changes needed to `restoreDataFromSessionStore`. The method already passes `pinned: folderData.pinned` to `_createFolderNode`; with the attribute-based approach, unpinned folders restore naturally:
- `storeDataForSessionStore` serializes the correct `pinned: false` via the updated getter
- The sentinel empty tab is created unpinned, so Firefox's standard restore places it in the regular tab area
- `_createFolderNode` sets the `zen-unpinned-folder` attribute from the serialized data

## Design decisions

- **Only new folders can be unpinned** — no "convert pinned folder to unpinned" UI. The pinned/unpinned state is decided at creation time based on the right-clicked tab.
- **Pref governs creation only** — flipping the pref off doesn't retroactively pin existing unpinned folders, making it safe to toggle.
- **Attribute-based state** — `zen-unpinned-folder` element attribute is the source of truth, read in `connectedCallback`. This unifies both creation and restore paths through `_createFolderNode`.

### DIFF
--- a/prefs/zen/folders.yaml
+++ b/prefs/zen/folders.yaml
@@ -13,3 +13,13 @@
 
 - name: zen.folders.owned-tabs-in-folder
   value: false
+
+- name: zen.folders.allow-unpinned
+  value: false
+  # Allow creating folders that contain unpinned tabs in the regular
+  # tab area. When false (default), folders may only be created in the
+  # pinned area and any tab added to a folder is force-pinned. When
+  # true, right-clicking an unpinned tab creates a folder in the
+  # regular tab area whose members stay unpinned. This pref only
+  # governs creation — toggling it off does not retroactively pin
+  # existing unpinned folders.
--- a/src/zen/drag-and-drop/ZenDragAndDrop.js
+++ b/src/zen/drag-and-drop/ZenDragAndDrop.js
@@ -1309,6 +1309,7 @@
       if (
         isTabGroupLabel(draggedTab) &&
         draggedTab.group?.isZenFolder &&
+        draggedTab.group?.pinned &&
         (((isTab(dropElement) ||
           dropElement.hasAttribute("split-view-group")) &&
           (!dropElement.pinned || dropElement.hasAttribute("zen-essential"))) ||
--- a/src/zen/folders/ZenFolder.mjs
+++ b/src/zen/folders/ZenFolder.mjs
@@ -68,7 +68,8 @@ export class nsZenFolder extends MozTabbrowserTabGroup {
 
   connectedCallback() {
     super.connectedCallback();
-    this.labelElement.pinned = true;
+    this._isUnpinnedFolder = this.hasAttribute("zen-unpinned-folder");
+    this.labelElement.pinned = !this._isUnpinnedFolder;
     if (this.#initialized) {
       return;
     }
@@ -205,7 +206,7 @@ export class nsZenFolder extends MozTabbrowserTabGroup {
   }
 
   get pinned() {
-    return this.isZenFolder;
+    return this.isZenFolder && !this._isUnpinnedFolder;
   }
 
   /**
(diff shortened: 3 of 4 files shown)


mr-cheffy:
Hi, thank you for your contribution! Unfortunately, we aren't planning on adding unpinned folders as they offer a level of organization that you wouldn't need for your unpinned tabs. We are planning on adding groups though!

doyled-it:
Hey, thanks for the quick response! I appreciate you taking a look.

I'd push back a little on "you wouldn't need" though. I built this specifically because I do need it. I have a lot of tabs open at any given time and folders would help me keep them organized, the same way they do in the pinned area.

A couple things worth considering:

- **It's behind a pref** (`zen.folders.allow-unpinned`, default `false`). It's completely invisible to anyone who doesn't opt in. Zero impact on the default experience.
- **Folders and groups solve different problems.** Groups (like Chrome's tab groups) are typically visual separators, good for short-lived groupings. Folders are persistent containers you can collapse and manage over time. They're complementary, not redundant.

I'm happy to adapt the implementation if you'd like it shaped differently, or to wait and see how groups land and revisit this afterward. But I think there's a real use case here that groups alone won't cover.

Either way, no hard feelings. Just wanted to make the case.

doyled-it:
@mr-cheffy ☝️ Any response to the above comment?

mr-cheffy:
> It's behind a pref (zen.folders.allow-unpinned, default false). It's completely invisible to anyone who doesn't opt in. Zero impact on the default experience.

Only a small sub-set of people would want this and an even smaller one would go out of their way to find and change advanced prefs, not sure if its worth the maintenance overhead.

> Folders and groups solve different problems. Groups (like Chrome's tab groups) are typically visual separators, good for short-lived groupings. Folders are persistent containers you can collapse and manage over time. They're complementary, not redundant.

Thats exactly why they shouldn't be unpinned. Because unpinned tabs are meant to be temporary, they shouldn't have a more advanced organizational mechanism other than basic grouping. Think of pinned tabs like bookmarks, normal tabs shouldn't need such level of organization.

> I'm happy to adapt the implementation if you'd like it shaped differently, or to wait and see how groups land and revisit this afterward. But I think there's a real use case here that groups alone won't cover.

as I said, its all very early still. Im figuring out designs and mechanisms for unpinned tab grouping, so there's not much I can provide currently unfortunately. Sorry

doyled-it:
@mr-cheffy:

Okay, fair enough I guess. This comes down to difference in philosophy and how we use tabs, because I still would like them given your points... 

> Think of pinned tabs like bookmarks, normal tabs shouldn't need such level of organization.

That is how I think about it: pinned tabs are like bookmarks. But sometimes I want to organize my ephemeral tabs into a task I'm working on (like github or gitlab issues) or a set of articles I need to read. I don't think they're important enough to bookmark but I don't want them in a formless sea of tabs. 

---

I did try to create a mod instead that would allow me to do this, but there's no way for me to do this with only CSS 😞 

> Only a small sub-set of people would want this and an even smaller one would go out of their way to find and change advanced prefs, not sure if its worth the maintenance overhead.

You can @doyled-it and I'll maintain it! 😅 (half joking)

PR #13220, opened 2026-04-12
fix: persist custom tab icons across sessions for favicon-less tabs

### LINKED ISSUE #13020: Custom icons for tabs without favicons not saving
### Preliminary Checks

### What happened?

After closing the browser, custom icons for tabs without favicons will occasionally be discarded.

### Expected behavior

Preserved icons

### Actual behavior

Discarded icons

### Steps to reproduce

Add custom icon to a tab without a favicon and close browser a few times.

### Screenshots and videos

https://github.com/user-attachments/assets/bb388908-92aa-4c79-bfcb-cdc4864d5e46

### Version

1.19.5

Windows - x64

Workspaces

```shell

```

galsakuri:
Fixes #13020

## Problem

Custom icons set on tabs without favicons were occasionally discarded after restarting the browser. The root cause was a save/restore mismatch:

- When a custom icon was set, `TabStateCache.update({ image: null })` explicitly cleared the image field in the session cache
- Both restore paths checked `zenHasStaticIcon && tabData.image` — this condition always evaluated to false since `image` was null
- Tabs with real favicons were unaffected because Firefox's async favicon service would later overwrite the null; favicon-less tabs never received such a write

## Fix

- Added a dedicated `zenStaticIconURL` field to the session state so the icon URL is saved independently of `tabData.image`
- Updated both restore paths to prefer `zenStaticIconURL` with fallback to the old `image`-based path for backward compatibility with existing sessions
- Replaced the `image: null` cache update with explicit `zenStaticIconURL` and `zenHasStaticIcon` fields to keep the cache consistent

## Files changed

- `src/browser/components/sessionstore/TabState-sys-mjs.patch`
- `src/browser/components/sessionstore/SessionStore-sys-mjs.patch`
- `src/zen/common/modules/ZenSessionStore.mjs`
- `src/zen/tabs/ZenPinnedTabManager.mjs`

### DIFF
--- a/src/browser/components/sessionstore/SessionStore-sys-mjs.patch
+++ b/src/browser/components/sessionstore/SessionStore-sys-mjs.patch
@@ -310,8 +310,9 @@ index 3bebeb70e1a88aefb3ef4bc2114bd85b0d8d1e51..5414adeca242f00bff98c5e41de74f33
 +    if (typeof tabData.zenStaticLabel === "string") {
 +      tab.zenStaticLabel = tabData.zenStaticLabel;
 +    }
-+    if (tabData.zenHasStaticIcon && tabData.image) {
-+      tab.zenStaticIcon = tabData.image;
++    const restoredIcon = tabData.zenStaticIconURL || (tabData.zenHasStaticIcon && tabData.image);
++    if (restoredIcon) {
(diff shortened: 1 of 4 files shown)


mr-cheffy:
Sorry, but this looks clearly done with AI

Issue #11155, opened 2025-11-05
Multiple inconsistancies with private window theme


RockClapps:
### Preliminary Checks

- [x] I have read and understood the important section above.
- [x] I have searched existing issues and avoided creating duplicates.
- [x] I am not filing an enhancement request.
- [x] I have checked that this issue cannot be reproduced on Mozilla Firefox.
- [x] I have checked that this issue can be reproduced once I removed all my Mods and Custom CSS.

### What happened?

There are a couple distinct issues surrounding private window theming that do not exist within vanilla firefox

1. Setting the "browser.theme.dark-private-windows" flag to "false" does nothing. Private browsing windows are permanently given a dark theme. 
2. "Website Appearance" setting "Automatic" has pages and extensions inherit the window theme, not the system theme. This causes all pages in private windows to be forced to dark mode regardless of system theme.  

Additionally, opening a new page in a private window will cause extensions like "Dark Reader" (when set to inherit system theme) to automatically convert all tabs in all windows to dark mode. This is not a separate issue rather just an additional annoyance caused by the above issues

### Expected behavior

"browser.theme.dark-private-windows" set to false should cause private browsing windows to have a light or automatic theme. Website appearance should inherit from the system theme, not the window theme, so tabs in private windows on a system set to light theme should also be light themed.

### Actual behavior

Setting the "browser.theme.dark-private-windows" flag to "false" does nothing. Private browsing windows are permanently given a dark theme. "Website Appearance" setting "Automatic" has pages and extensions inherit the window theme, not the system theme. This causes all pages in private windows to be forced to dark mode regardless of system theme.

### Steps to reproduce

1. Set system theme to light 
2. Set the "browser.theme.dark-private-windows" flag to "false"
3. Open a private browsing window
4. Observe that the window is still dark

1. Set system theme to light
2. Go to "Settings > General > Language and Appearance > Website apperance" and set it to "Automatic"
3. Open a private browsing window
4. Open a tab that has automatic theming 
5. Observe that the tab follows a dark theme

### Version

1.17.5b

### What platform are you seeing the problem on?

Linux (Tarball)

### What component is this issue related to?

Mods / Themes

### Relevant log output if applicable

```shell

```

RockClapps:
Yes this is still a relevant issue. It is particularly annoying when using an extension like _Dark Reader_, which I mentioned in my initial submission.

koaledu:
[...]

PR #13222, opened 2026-04-12
fix: #11155 multiple inconsistancies with private window theme

gitmichaelqiu:
Targeting #11155: websites in private browing are always in dark theme when `browser.theme.dark-private-windows` is set to false.

- In `ZenGradientGenerator.mjs`, `isDarkMode` was hardcoded to return true whenever a window was private
    - changed to `return Services.prefs.getBoolPref("browser.theme.dark-private-windows", true);`
    - `shouldBeDarkMode` has a special case handling for private + `dark-private-windows` flag
- In `zen-theme.css`, the rules for private windows were hardcoded to always include `dark.inc.css`, and the css was not aware of the `browser.theme.dark-private-windows` flag
    - wrapped the code with `@media -moz-pref("browser.theme.dark-private-windows")`
    - added `color-scheme: light dark !important` for private windows

The default behavior (`browser.theme.dark-private-windows` is true) is not changed. When the flag is false, both the browser UI and the web content should follow system theme. The light theme comes from `schemes/light.inc.css`.

### DIFF
--- a/src/zen/common/styles/zen-theme.css
+++ b/src/zen/common/styles/zen-theme.css
@@ -227,17 +227,21 @@
 
   --toolbar-field-color: var(--toolbox-textcolor) !important;
 
-  &[zen-private-window='true'][zen-dark-private-windows='true'] {
-    --zen-main-browser-background: color-mix(in srgb, rgb(11, 10, 11) 90%, var(--zen-themed-toolbar-bg-transparent));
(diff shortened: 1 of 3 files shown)


mr-cheffy:
We purposefully ignore this pref in order to make a clear distinction between dark mode and normal windows, sorry.

PR #13244, opened 2026-04-14
Linux: add .deb packaging and desktop integration helper

RaymonDev:
Adds a user-local desktop entry installer for unpacked Linux builds, plus a CI-built .deb package flow for Ubuntu.\n\nTesting:\n- Downloaded the official zen.linux-x86_64.tar.xz release artifact\n- Built dist/zen-browser_1.19.8b-test_amd64.deb\n- Installed it with apt\n- Launched zen successfully and confirmed it was running\n\nNotes:\n- The .deb repackages the existing Linux tarball into a standard Debian layout\n- This is intended to make Zen behave like a normal installed app on Ubuntu

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -438,6 +438,48 @@ jobs:
           name: zen-${{ matrix.arch }}.AppImage.zsync
           path: ./dist/zen-${{ matrix.arch }}.AppImage.zsync
 
+  deb:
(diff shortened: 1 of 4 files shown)


mr-cheffy:
Hi, thank you for your contribution! Unfortunately, we don't support deb packages on purpose as they can't auto update. Sorry

PR #13292, opened 2026-04-17
gh-13095: restore multi-select highlight for split-view tabs

### LINKED ISSUE #13095: Multi selection looks inconsistent on split views
### Preliminary Checks

### What happened?

When selecting multiple tabs, the appearance of split views will be inconsistent with how it looks when its regularly selected.

### Expected behavior

Same apperance on multi selection as on single selection.

### Actual behavior

Inconsistent apperance.

### Steps to reproduce

Select multiple tabs without initially selecting the split tabs.

### Screenshots and videos

https://github.com/user-attachments/assets/1d906dd8-73b2-47c2-a136-93d4abf681e2

### Version

1.19.6b

Windows - x64

Split View

```shell

```

KelTech-Services:
Fixes #13095.

The selector change in fe19b44 replaced the :has(multiselected) match with [hasactivetab], which only fires when the active tab is inside the split group. Multi-selecting tabs without the split tab in the initial selection dropped the highlight.

Added the multiselected path back as a union so both cases get the selected styling.

### DIFF
--- a/src/zen/split-view/zen-split-group.inc.css
+++ b/src/zen/split-view/zen-split-group.inc.css
@@ -104,7 +104,8 @@ tab-group[split-view-group] {
       }
     }
 
-    tab-group[split-view-group]:where([hasactivetab]) & {
+    tab-group[split-view-group]:where([hasactivetab]) &,
+    &:has(> tab:is([multiselected])) {
       background-color: var(--tab-selected-bgcolor);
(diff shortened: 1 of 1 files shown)


mr-cheffy (approved):

mr-cheffy:
Not too fond of using `:has` but it shouldn't impact too much performance, lets see. Thanks!

PR #13340, opened 2026-04-21
fix: prevent detached popups from converting to NSPopover on macOS (f…

JacuXx:
…ixes gh-XXXXX)

This fix adds an additional validation to the ShouldShowAsNSPopover() function to check that the anchor rect is valid and non-empty. This prevents detached popups (like Shopify theme editor windows) from being incorrectly converted to NSPopover, which would fail to display due to NSPopover's requirement that the anchor must be within the parent view.

- Add anchor rect validation in ShouldShowAsNSPopover()
- Prevents 'empty anchor' popups from using NSPopover
- Maintains NSPopover benefits for properly-anchored UI popups
- Fixes Shopify theme editor not opening on macOS


mr-cheffy:
AI didn't even include the issue and didnt export the patch correctly

JacuXx:
Well, at least I tried. c:

PR #13368, opened 2026-04-23
gh-8643: Fix stuck sidebar in compact mode after DnD exception

### LINKED ISSUE #8643: Sidebar stuck open, does not collapse when compact mode turned on.
### Captchas

### What happened?

With compact mode turned on, the side bar does not collapse when the mouse hovers away from it. After booting the browser this is not an issue, but randomly, roughly an hour into usage it happens and the side bar stops collapsing automatically. The sidebar is still functional, so is workspace switching when this happens. It also does not happen when the laptop awakes from sleep, it happens randomly when I move my mouse out of the sidebar, and it just doesn't go back in and stays like that.

This has been happening for the last few versions so I don't think it was something introduced just now. 

The only solution I have is to restart the browser entirely which is annoying because it forces me to lose progress, or to disable compact mode which is also not ideal as it reduces the page size too much for me.

### Version

1.12.8b (Firefox 138.0.4) (aarch64)

macOS - aarch64

Compact Mode

```shell

```

cfarvidson:
## Why

Closes #8643.

In compact mode the sidebar randomly stops auto-collapsing on mouse leave after ~1h of use and stays pinned open until restart. Reported reliably on macOS, intermittently elsewhere.

Root cause: `gZenCompactModeManager._isTabBeingDragged` is the latch that keeps the sidebar expanded during a drag. It's set in `startTabDrag` and cleared at the very *end* of `handle_dragend`. If any step inside `handle_dragend` throws — most commonly `super.handle_dragend` after an upstream drop failure like `adoptTab` hitting a null `linkedBrowser` — the delete never runs. The mouseleave path in `ZenCompactMode.mjs:830` then early-returns forever because the flag is still truthy, and the sidebar stays pinned open until the process restarts.

## What

One-line fix: move the `delete gZenCompactModeManager._isTabBeingDragged` to the top of `handle_dragend` (with optional chaining to survive a null owner global). Now the latch is released before anything that could throw runs, so the sidebar can always collapse on the next mouseleave — regardless of whether later cleanup succeeds.

## Test plan

- [ ] Launch Zen, enable compact mode with auto-hide sidebar.
- [ ] Drag a tab within the sidebar, drop it — sidebar collapses on mouseleave (no regression).
- [ ] Drag a tab across workspaces — workspace still switches, sidebar still collapses.
- [ ] Drag a tab out to create a new window — adoption still works, sidebar still collapses.
- [ ] Synthetic repro: throw from `super.handle_dragend` (e.g. monkey-patch to throw, or reproduce the `linkedBrowser is null` / `children.at(...)` condition from #8643) — confirm the sidebar still collapses on mouseleave after the failed drag.
- [ ] Long-session check: after a known bad drag error fires, hover out of the sidebar — it collapses.

### DIFF
--- a/src/zen/drag-and-drop/ZenDragAndDrop.js
+++ b/src/zen/drag-and-drop/ZenDragAndDrop.js
@@ -1139,6 +1139,12 @@
       const dt = event.dataTransfer;
       const draggedTab = dt.mozGetDataAt(TAB_DROP_TYPE, 0);
       let ownerGlobal = draggedTab?.ownerGlobal;
+      // Release the compact-mode drag latch up front so the sidebar can
+      // collapse on mouseleave again even if any step below throws (e.g.
+      // super.handle_dragend hitting a stale tab after an upstream drop
+      // failure). Otherwise _isTabBeingDragged stays true until restart and
+      // the sidebar pins open. See gh-8643.
+      delete ownerGlobal?.gZenCompactModeManager?._isTabBeingDragged;
       draggedTab.style.visibility = "";
       let thisFromGlobal = ownerGlobal?.gBrowser.tabContainer.tabDragAndDrop;
       let currentEssenialContainer =
@@ -1172,7 +1178,6 @@
         thisFromGlobal._tempDragImageParent.remove();
         delete thisFromGlobal._tempDragImageParent;
       }
-      delete ownerGlobal.gZenCompactModeManager._isTabBeingDragged;
       if (dt.dropEffect !== "move") {
         ownerGlobal.gZenCompactModeManager._clearAllHoverStates();
       }


src/zen/drag-and-drop/ZenDragAndDrop.js
```diff
@@ -644,7 +644,17 @@
...
+      try {
+        super.handle_dragover(event);
+      } catch (e) {
+        // Firefox's handle_dragover can throw transiently (e.g.
+        // `children.at(...) is undefined` during workspace switches or on
+        // empty tab groups). Swallow so the drag lifecycle keeps producing
+        // dragend events and compact mode can un-pin the sidebar again.
+        // See gh-8643.
+        console.error(e);
+        return;
```
mr-cheffy: What's the point of this if you are going to return anyways...?

src/zen/drag-and-drop/ZenDragAndDrop.js
```diff
@@ -1138,43 +1155,59 @@
...
-      thisFromGlobal._clearDragOverSplit();
-      this.#maybeClearVerticalPinnedGridDragOver();
-      thisFromGlobal.originalDragImageArgs = [];
-      window.removeEventListener(
-        "dragenter",
-        thisFromGlobal.handle_windowDragEnter,
-        {
-          capture: true,
+      let ownerGlobal = draggedTab?.ownerGlobal ?? window;
+      try {
```
mr-cheffy: Why wrap this in a catch block?

mr-cheffy (changes requested):

src/zen/drag-and-drop/ZenDragAndDrop.js
cfarvidson: Good catch — dropped it. The only thing the try/catch gave us was swallowing the console noise; the downstream behavior on throw was already identical to the `return`. The real fix lives in `handle_dragend`, which I've simplified to just move the `delete _isTabBeingDragged` to the top of the method.

src/zen/drag-and-drop/ZenDragAndDrop.js
cfarvidson: Fair — the try/finally was overkill. I've dropped it and the `handle_drop` try/catch too. The whole fix is now one moved line: the `delete gZenCompactModeManager._isTabBeingDragged` now runs at the top of `handle_dragend` instead of at the bottom, so if `super.handle_dragend` (or anything in between) throws after an upstream drop failure, the latch is already released and the sidebar can collapse on mouseleave again.

sporocyst:
As a layman on coding, I am just curious shouldn't we figure out first how to address those failures happening before `delete gZenCompactModeManager._isTabBeingDragged` ?

cfarvidson:
> As a layman on coding, I am just curious shouldn't we figure out first how to address those failures happening before delete gZenCompactModeManager._isTabBeingDragged ?

Sure, but I think that "perfect is the enemy of good". I cannot use the browser now with the tabs-menu are always opened.. 😊

mr-cheffy (changes requested):
Wouldn't it make more sense to actually handle why its erroring out rather than simply moving the delete instruction upwards?

cfarvidson:
I mean, you could do both. 

But perfect it is...Then I'm out.

mr-cheffy:
Not sure what you mean with "perfect"?

PR #13401, opened 2026-04-25
no-bug: `contain` of `.zen-glance-sidebar-container` too strict

Ivaon:
The confirm button was cut off due to

https://github.com/zen-browser/desktop/blob/0199e256d68d62bc6b7926792601f5fa5cc405b1/src/zen/glance/zen-glance.css#L13

image


mr-cheffy (approved):
Thanks!

PR #13454, opened 2026-04-28
fix: Hide appcontent navbar strip in compact + single-toolbar mode

### LINKED ISSUE #7094: Top toolbar is shown when hovering the top side of the screen in "single toolbar" layout with "compact mode"
### Captchas

### What happened?

Image

### Version

1.10.1b

Linux (Flatpak)

Customizable UI / Toolbars

```shell

```

AndreNijman:
## Summary

- In compact mode with single-toolbar enabled, `#zen-appcontent-navbar-wrapper` retains its default height of `--zen-element-separation` (8px), showing a thin visible strip at the top of the content area when the sidebar is hidden.
- In single-toolbar mode all navigation lives in the sidebar, so the content navbar wrapper is always empty. The fix collapses it to zero height in compact mode when no hover/popup/active state is requesting it to expand.

## Steps to reproduce

1. Enable compact mode with `zen.view.use-single-toolbar = true`
2. Move mouse away from sidebar so it collapses
3. Observe a thin strip above the content area (height equals `--zen-element-separation`)

## Fix

CSS rule in `sidebar.inc.css` (compiled into the compact mode media query):

```css
&[zen-single-toolbar='true'] {
  #zen-appcontent-navbar-wrapper:not([zen-has-hover])
    :not([has-popup-menu]):not([zen-compact-mode-active]):not([flash-popup]) {
    height: 0;
    min-height: 0;
    overflow: hidden;
    pointer-events: none;
  }
}
```

The exclusions preserve expansion when a popup or hover interaction needs the wrapper visible.

Fixes #7094


mr-cheffy:
Same as the other PRs

PR #13455, opened 2026-04-28
fix: Clear pending attribute on synced tabs when source tab is loaded

### LINKED ISSUE #12142: Loaded tabs are faded out in new windows
### Preliminary Checks

### What happened?

When I create a new window, UI elements appear as though the tabs in that window are unloaded, even though they are loaded (just not in that window)

### Expected behavior

Tabs that are loaded in any window will appear loaded in any window

### Actual behavior

Tabs only appear as loaded in windows where they've been visited

### Steps to reproduce

1. Enable browser.tabs.fadeOutExplicitlyUnloadedTabs and browser.tabs.fadeOutUnloadedTabs
2. Open more than one tab with colorful icons
3. Create a new window
4. Notice that in the newly created window, the icons on the tabs you're not currently in are grayed out

### Screenshots and videos

Notice that in the image below, the second Google tab in the right window is slightly transparent
Image

The issue is far more obvious when using a mod that makes unloaded tabs more noticeable:
Image

### Version

1.18.1b

Linux (Tarball)

Tabs

```shell

```

AndreNijman:
## Summary

- When `ZenWindowSync` opens placeholder tabs in a new window via `addTrustedTab("about:blank", { createLazyBrowser: true })`, each placeholder inherits `pending="true"` from the lazy browser creation.
- This causes synced tabs to appear faded/dimmed even when the corresponding source tab in the original window is fully loaded — the `pending` attribute is never cleared.

## Steps to reproduce

1. Open Zen with multiple windows synced (zen window sync enabled)
2. Open a second window
3. Observe: all tabs in the new window appear faded even though the source tabs are loaded

## Fix

In `ZenWindowSync.sys.mjs`, `on_TabOpen`, after `newTab.id = tab.id`:

```js
if (!tab.hasAttribute("pending")) {
  newTab.removeAttribute("pending");
}
```

If the source tab is not pending (i.e. it is loaded), the `pending` attribute is removed from the placeholder so the visual state matches the source.

Fixes #12142


mr-cheffy:
Same as the other PRs

PR #13458, opened 2026-04-28
fix: Use CanvasText for tab preview panel text color

### LINKED ISSUE #12410: Tab preview text illegible in front of light background
### Preliminary Checks

### What happened?

When hovering over a tab to see its preview and title, if the open tab has a white/light background, the tab preview text becomes completely illegible due to the light background

### Expected behavior

I would expect the text in the tab preview be dark enough to read in front of a white backgroun

### Actual behavior

The text in the tab preview is not dark enough to read in front of a white background

### Steps to reproduce

1. Open a tab with a white page
2. Hover over a closed tab to see the tab preview

### Screenshots and videos

Image

Image

### Version

1.18.8b

macOS - aarch64

Tabs

```shell

```

AndreNijman:
## Summary

- Tab hover preview text (title + URL) renders invisible on light color schemes because `--text-color-deemphasized` inherits `currentColor` from the panel's dark-themed paint context, producing white text against a light panel background.
- Adds `color: CanvasText` on `#tab-preview-panel` so text color always matches the system foreground color regardless of inherited paint context.

## Steps to reproduce

1. Enable a light color scheme / light system theme
2. Hover over any tab in the sidebar
3. Tab preview popup appears — title/URL text is white-on-white (invisible)

## Fix

One rule in `zen-popup.css`:

```css
#tab-preview-panel {
  color: CanvasText;
}
```

`CanvasText` is a CSS system color that resolves to the system's foreground text color, adapting correctly to both light and dark themes without hardcoding a value.

Fixes #12410

### DIFF
--- a/src/zen/common/styles/zen-popup.css
+++ b/src/zen/common/styles/zen-popup.css
@@ -428,3 +428,7 @@ menuseparator {
 #window-modal-dialog {
   inset: 0;
 }
+
+#tab-preview-panel {
+  color: CanvasText;
+}


mr-cheffy:
Have you tested this? How does it look?

PR #13461, opened 2026-04-28
fix: Intercept DOMWindowClose to prevent closing pinned Places tab on Ctrl+W

### LINKED ISSUE #12353: "Close Pinned Tab Shortcut Behavior" ignored for "Places" tab
### Preliminary Checks

### What happened?

Hi!
This is a rare case, I believe. I'm using the `chrome://browser/content/places/places.xhtml` tab to have a dedicated full-window history tab, because I find Sidebar History hard to use and navigate.
I accidentally closed this tab, and found it out way later - so I had to reopen it again (and as I don't recall this entire URI, I had to find it again). 
I found that there's a "Close Tab Shortcut Behavior" setting, which I thought would help me prevent this from happening again. I tinkered with it, with no result. Just when I wanted to forget about this option, I tried it on the 2nd essential tab, with "regular page" opened there, and the tab didn't close, as per "Switch to next tab" behavior, set in the Settings.
It acts the same way when the tab is pinned, not added to the Essentials.

**OS:** Windows 10

### Expected behavior

On the essential/pinned tab closing, the configured behavior is executed, like for the "regular page" tabs

### Actual behavior

The page is closed regardless of the "Close Tab Shortcut Behavior" configuration

### Steps to reproduce

Steps:
1. Set "Close Tab Shortcut Behavior" to "Switch To next tab".
2. Open `chrome://browser/content/places/places.xhtml`.
3. "Pin Tab" or "Add to Essentials".
4. Open this tab.
5. Press `Ctrl+W`.

### Screenshots and videos

Image

### Version

1.18.5b

Windows - x64

Tabs

AndreNijman:
## Summary

- The Places sidebar tab (`about:privatebrowsing` / bookmark manager) can be closed with Ctrl+W even when it's pinned, because `window.close()` dispatches a `DOMWindowClose` event that bypasses `cmd_close` (which already checks for pinned tabs).
- Adds a `DOMWindowClose` capture-phase listener on `gBrowser` to intercept and cancel the close event when the target tab is pinned.
- Also defers listener registration to `window load` to avoid a `gBrowser is undefined` crash (the parent `init()` runs at `DOMContentLoaded` before `gBrowser` is available).

## Steps to reproduce

1. Pin the Places/bookmarks sidebar tab
2. Focus it and press Ctrl+W
3. The tab closes despite being pinned

## Fix

In `ZenPinnedTabManager.mjs`, `_initClosePinnedTabShortcut()`:
- Wrap `gBrowser.addEventListener(...)` in `window.addEventListener("load", ..., { once: true })`
- Add `DOMWindowClose` handler that resolves the browser element for remote tabs and cancels the event if the tab is pinned

Fixes #12353

### DIFF
--- a/src/zen/tabs/ZenPinnedTabManager.mjs
+++ b/src/zen/tabs/ZenPinnedTabManager.mjs
@@ -253,22 +253,28 @@ class nsZenPinnedTabManager extends nsZenDOMOperatedFeature {
       cmdClose.addEventListener("command", this.onCloseTabShortcut.bind(this));
     }
 
-    gBrowser.addEventListener(
-      "DOMWindowClose",
-      (event) => {
-        let browser = event.target;
-        if (!browser.isRemoteBrowser) {
-          browser = event.target.docShell?.chromeEventHandler ?? browser;
-        }
-        const tab = gBrowser.getTabForBrowser(browser);
-        if (!tab?.pinned) {
-          return;
-        }
-        event.stopImmediatePropagation();
-        event.preventDefault();
-        this.onCloseTabShortcut(event, tab);
+    window.addEventListener(
+      "load",
+      () => {
+        gBrowser.addEventListener(
+          "DOMWindowClose",
+          (event) => {
+            let browser = event.target;
+            if (!browser.isRemoteBrowser) {
+              browser = event.target.docShell?.chromeEventHandler ?? browser;
+            }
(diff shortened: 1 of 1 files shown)


mr-cheffy:
Seems a bit overkill, I think the issue will be eventually be marked as WONTFIX if a better solution isn't found... Thanks for trying though!

PR #13468, opened 2026-04-28
gh-7094: Fixed topbar dropping down when titlebar is enabled

### LINKED ISSUE #7094: Top toolbar is shown when hovering the top side of the screen in "single toolbar" layout with "compact mode"

sporocyst:
From #7094 , we need to prevent topbar pulling down when hovering if following criterias are all meet:
1. Titlebar is enabled
2. Bookmark bar is disabled
3. In compact mode
4. Only sidebar layout

Thanks to previous effort, we already have a similar mechanism when `zen.view.experimental-no-window-controls`  is `true`. And firefox also has a pref reflecting the status of titlebar (`browser.tabs.inTitlebar`)

So, to tap in that existing mechanism, this PR puts a conditonal judgement that either `zen.view.experimental-no-window-controls` or `browser.tabs.inTitlebar` can achieve same effect.

However, since I only tested the changes on Windows, I don't know if it will lead to same result in MAC OS or linux.

### DIFF
--- a/src/zen/common/styles/zen-browser-container.css
+++ b/src/zen/common/styles/zen-browser-container.css
@@ -43,7 +43,7 @@
   }
 
   /* stylelint-disable-next-line media-query-no-invalid */
-  @media (not -moz-pref("zen.view.shift-down-site-on-hover")) and (not ((-moz-pref("zen.view.experimental-no-window-controls") or (not -moz-pref("zen.view.hide-window-controls"))) and -moz-pref("zen.view.use-single-toolbar"))) {
+  @media (not -moz-pref("zen.view.shift-down-site-on-hover")) and (not ((-moz-pref("zen.view.experimental-no-window-controls") or (not -moz-pref("zen.view.hide-window-controls")) or (not -moz-pref("browser.tabs.inTitlebar"))) and -moz-pref("zen.view.use-single-toolbar"))) {
     .browserSidebarContainer:is(.deck-selected, [zen-split="true"]) .browserContainer {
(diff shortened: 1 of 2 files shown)


mr-cheffy (approved):
I have a weird feeling that there's a check we are missing... Does the webcontent get pushed down when the bookmarks bar is enabled AND native titlebar?

sporocyst:
> I have a weird feeling that there's a check we are missing... Does the webcontent get pushed down when the bookmarks bar is enabled AND native titlebar?

You are right on this. I did see that happen during testing. That said, since it is a glitch that has been there for a while triggered merely by enabling both `zen.view.experimental-no-window-controls` and bookmark bar, and no one ever reported an inconvenience about it, I just assume users don't really care that much.

CcMenta:
Is it possible to disable this feature? I have a setup where I have a panel on top and when I maximize the program the titlebar gets "integrated" in to the panel (the titlebar buttons are in the panel). Without the topbar dropping down it's really annoying to move the window.

sporocyst:
>  when I maximize the program the titlebar gets "integrated" in to the panel.

How did you do that? Depend on your apprach, maybe we can find a workaround

CcMenta:
> How did you do that?

I use kde plasma and there are widgets that can hide/disable the titlebar.

sporocyst:
@CcMenta Press `ctrl+shift+B`, that will put a bookmark bar into topbar and allow topbar to drop down even when titlebar is activated

PR #13479, opened 2026-04-29
no-bug: Make macOS app icon changes available from settings

yordis:
- Makes the running macOS app icon switchable from about:preferences#zenLooks, so icon variants can be exercised from settings instead of requiring separate branded builds.
- Shows the selected icon variant directly in settings and packages the alternate asset, so the macOS-specific behavior is visible before it is applied.
- Keeps this first implementation scoped to macOS because the current runtime icon hook only exists for AppKit, and Windows and Linux remain unchanged for now.
- Syncs the Zen en-US overlay strings during local import and build, so the new settings UI renders correctly in local development builds.


yordis:
https://github.com/user-attachments/assets/8c0800ff-3d9c-4ef4-97dd-527cc8c23871

src/zen/toolkit/common/cocoa/ZenAppIcon.mm
```diff
@@ -0,0 +1,55 @@
...
+  if (!imageData) {
+    return NS_ERROR_INVALID_ARG;
+  }
+
+  NSImage* iconImage = [[[NSImage alloc] initWithData:imageData] autorelease];
+  if (!iconImage) {
+    return NS_ERROR_FAILURE;
+  }
+
+  [NSApp setApplicationIconImage:iconImage];
```
mr-cheffy: I think it would be better to use NSWorkspace setIcon?

src/zen/common/modules/ZenStartup.mjs
```diff
@@ -69,15 +76,84 @@ class ZenStartup {
...
+  observe(aSubject, aTopic, aData) {
     // This nsIObserver method allows us to defer initialization until after
     // this window has finished painting and starting up.
     if (aTopic == "browser-delayed-startup-finished" && aSubject == window) {
       Services.obs.removeObserver(this, "browser-delayed-startup-finished");
       this.delayedStartupFinished();
+      return;
+    }
+
+    if (aTopic == "nsPref:changed") {
```
mr-cheffy: This would be called on every instance of the window, please make it a service and hook the observers there via a manifest file. For example: https://github.com/zen-browser/desktop/blob/dev/src/zen/sessionstore/SessionComponents.manifest

src/zen/images/app-icons/alternate-preview.png
mr-cheffy, line 1: We should be setting .car files as icons so they work best with tahoe

mr-cheffy (changes requested):

LemonPotion:
Is there a way to implement this on Windows and Linux? Would be great 🙏

yordis:
@LemonPotion I never done it for Win and Linux, but I can try to figure it out

mr-cheffy:
Hey yordis! I'll close this in favor of https://www.askvg.com/firefox-now-allows-users-to-change-or-customize-app-icon-natively/

PR #13483, opened 2026-04-29
gh-7094: Fix webcontent shift when bookmark bar pulled down in the presence of title bar

### LINKED ISSUE #7094: Top toolbar is shown when hovering the top side of the screen in "single toolbar" layout with "compact mode"

sporocyst:
Per https://github.com/zen-browser/desktop/pull/13468#pullrequestreview-4191276729 , handling the timing of webcontent shift with bookmark bar in mind

### DIFF
--- a/src/zen/common/styles/zen-browser-container.css
+++ b/src/zen/common/styles/zen-browser-container.css
@@ -43,7 +43,7 @@
   }
 
   /* stylelint-disable-next-line media-query-no-invalid */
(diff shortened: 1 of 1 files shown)


mr-cheffy (approved):
Thanks!

Issue #11626, opened 2025-12-15
Split tab divider can't be moved to resize with bookmarks sidebar open


alvanx:
### Preliminary Checks

- [x] I have read and understood the important section above.
- [x] I have searched existing issues and avoided creating duplicates.
- [x] I am not filing an enhancement request.
- [x] I have checked that this issue cannot be reproduced on Mozilla Firefox.
- [x] I have checked that this issue can be reproduced once I removed all my Mods and Custom CSS.

### What happened?

The vertical divider in split view can't be dragged sideways to resize whenever the bookmarks sidebar is open.

### Expected behavior

I should be able to drag-to-resize the central divider in a split view. 

### Actual behavior

When the bookmarks sidebar is open, the divider can't be moved to resize. When the sidebar is closed, it starts working.

### Steps to reproduce

1. Tested on macOS 15.6, observed for several weeks now.
2. Open the bookmarks sidebar (might work with any FF sidebar view).
3. Select two tabs, right-click and select to split them.
4. Attempt to drag the vertical divider to resize the tabs. 

### Version

1.17.13b

### What platform are you seeing the problem on?

macOS - aarch64

### What component is this issue related to?

Split View

### Relevant log output if applicable

```shell

```

MagicCheese1:
Can reproduce on Windows - x64, version 1.17.14b

willrossignol:
The issue is that the drag-and-drop part of the divider is still in the position where it would be if the Sidebar wasn't open. (See the small vertical line to the left of the divider in the picture below)

<img width="1403" height="843" alt="Image" src="https://github.com/user-attachments/assets/39d93358-da27-469c-bfb5-2caa110e7237" />

sai-educ:
Can reproduce this issue when the AI chat sidebar is active.

willrossignol:
I can take this up, I've got a good idea of how to go about fixing this. I think the `zen-splitview-overlay-wrapper` needs to be moved to the `tabbrowser-tabpanels` element instead of being in `tabbrowser-tabbox`.

Propheticus:
@mr-cheffy  Clear bug, easy to repro. Want to let @willrossignol  take this one?

Fsyrl:
+1 …

alvanx:
Issue still present on 1.19.13b (Firefox 150.0.3) (aarch64).

PR #13511, opened 2026-04-30
Fix splitter bad position when sidebar open

willrossignol:
- Move `zen-splitview-overlay.inc.xhtml` into `tabpanels` element to handle correct drag-and-drop logic when sidebar is open
- Insert `zen-split-view-splitter` directly into `tabpanels` element for correct placement when sidebar is open
- Adjust inset calculation for `zen-split-view-splitter` to center in `tabpanels` element
- Add `translateX: calc(-50%)` to `::before` styles in order to properly center splitter

image

### DIFF
--- a/src/browser/base/content/browser-box-inc-xhtml.patch
+++ b/src/browser/base/content/browser-box-inc-xhtml.patch
@@ -1,5 +1,5 @@
 diff --git a/browser/base/content/browser-box.inc.xhtml b/browser/base/content/browser-box.inc.xhtml
-index d58fcdf99843d110b708f3fbf9fb317787fadfcf..cfc2aad902641609c3804e615c4cb66ce65299b7 100644
+index d58fcdf99843d110b708f3fbf9fb317787fadfcf..45ad1e28b31a4993e56dfce0f88f49f48f25f2de 100644
 --- a/browser/base/content/browser-box.inc.xhtml
 +++ b/browser/base/content/browser-box.inc.xhtml
 @@ -3,12 +3,22 @@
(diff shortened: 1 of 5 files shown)


willrossignol:
Fix for #11626

sporocyst:
Looking forward to see this finalized. And, be sure to test the conditon when sidebar is open on the right

willrossignol:
> Looking forward to see this finalized. And, be sure to test the conditon when sidebar is open on the right

Can confirm the fixes work when the sidebar is open on the right. Thanks for the heads up!

image

willrossignol:
Splitter is properly centered after updates to `::before` styling

image

src/browser/base/content/browser-box-inc-xhtml.patch
```diff
@@ -25,16 +25,20 @@ index d58fcdf99843d110b708f3fbf9fb317787fadfcf..cfc2aad902641609c3804e615c4cb66c
...
-@@ -25,7 +35,7 @@
+@@ -25,8 +35,10 @@
      </stack>
    </vbox>
    <splitter id="sidebar-splitter" class="chromeclass-extrachrome sidebar-splitter" resizebefore="sibling" resizeafter="none" hidden="true"/>
 -  <tabbox id="tabbrowser-tabbox" flex="1" tabcontainer="tabbrowser-tabs">
+-    <tabpanels id="tabbrowser-tabpanels" flex="1" selectedIndex="0"/>
 +#include zen-tabbrowser-elements.inc.xhtml
-     <tabpanels id="tabbrowser-tabpanels" flex="1" selectedIndex="0"/>
++    <tabpanels id="tabbrowser-tabpanels" flex="1" selectedIndex="0">
```
mr-cheffy, line 35: Pretty sure `selectedIndex` is used to fetch childs within the tab panel. Adding elements inside might mess that up

mr-cheffy (changes requested):

src/browser/base/content/browser-box-inc-xhtml.patch
willrossignol: I looked into this a bit. Every tab keeps track of the index of its panel in the tabpanels element, so the index of the panels is adjusted to be +1 because of the overlay wrapper being inserted first into the tabpanels element. For the splitter, it is always appended to the list of children of tabpanels, so this also doesn't affect the indices of the panels in tabpanels.

I've been testing around creating new tabs, removing tabs, joining and splitting tabs, and using glance, and I haven't encountered any issues.

I can look into another solution though if you're not happy with this!

willrossignol: The logic that I was looking at is in `toolkit/content/widgets/tabbox.js`

sporocyst: But... even Firefox itself puts splitter into tab panel! It is probably not really a no-no.
  

Image

mr-cheffy: True, just not really comfortable with messing around firefox's internals. I guess it could be fine for now

mr-cheffy (changes requested):
Could you please run `npm run lint:fix`?

willrossignol:
> Could you please run `npm run lint:fix`?

Done. Corrected some linting errors in ZenViewSplitter.mjs

mr-cheffy:
Does drag and dropping split views from the top handle work for you? It doesnt show me the preview

willrossignol:
> Does drag and dropping split views from the top handle work for you? It doesnt show me the preview

Yeah it works but you're right that it doesn't show the preview. I'll look into this

src/zen/split-view/zen-split-view.css
```diff
@@ -116,6 +116,8 @@
   pointer-events: none;
   padding: inherit;
   inset: 0;
+  -moz-subtree-hidden-only-visually: 0;
+  z-index: 2;
```
willrossignol, line 120: Added these changes to ensure contents of `zen-splitview-overlay` is shown above the tabs. This fixes not seeing the `zen-splitview-dropzone`

PR #13564, opened 2026-05-05
feat: Open links in split view via shift+click

JordanTranchina:
## What

Adds `shift+click` on any page link as a shortcut to open that link in a new split-view tab alongside the current one.

## How

Implements a `ZenSplitView` JSWindowActor pair following the same pattern as `ZenGlance` and `ZenBoosts`:

- **`ZenSplitViewChild`** (content process) — listens for `click` events. Intercepts `shift+click` on `<a href>` elements and sends an async message to the parent to open the URL in split view. On `DOMContentLoaded` it also queries the Glance activation method so it can yield when Glance is configured to use `shift`.
- **`ZenSplitViewParent`** (chrome process) — responds to the query by reading `zen.glance.activation-method`, and handles the open action by calling `gZenViewSplitter.splitLinkFromURL(url)`.
- **`ZenViewSplitter`** — extracts a `splitLinkFromURL(url)` helper from `splitLinkInNewTab()` so the actor can invoke it directly without going through `gContextMenu`.
- **`split-view/moz.build`** — ships the actor files to `resource:///actors/`.
- **`zen/moz.build`** — adds `split-view` to `DIRS`.

## Conflict avoidance

Before intercepting a click, the child actor checks the Glance activation-method pref. If Glance is set to `shift`, the click is passed through unchanged so the two features don't conflict.

## Test plan

- [ ] `shift+click` a link → opens in a new split-view tab beside the current tab
- [ ] Plain click, `ctrl+click`, `alt+click`, `meta+click` are unaffected
- [ ] Glance activation = `shift`: `shift+click` still opens Glance; split view is not triggered
- [ ] Glance activation = `ctrl` (or other): `shift+click` opens split view; Glance opens on `ctrl+click`
- [ ] Context-menu "Open in Split View" still works unchanged


untitled500:
Question, @JordanTranchina : Will there be a setting to change open in split view behaviour between Shift+ / Cmd+ / Control+ Click? Of course, if so, then glance activation setting should not allow a user to set a conflicting configuration.

chirivelli:
isn't "Shift + Click" for Glance or Peek

can we do this with "Option + Click"

mr-cheffy:

PR #13565, opened 2026-05-05
no-bug: A few quality-of-life tweaks (build tooling)

iskunk:
Hello, I am using some of the Zen build tools in an alternate setting, a framework that "converts" a Debian source package of Firefox into the equivalent Zen one. To that end, having some additional flexibility in the invocation of the tooling is helpful.

----

`package.json`: Use the `$CARGO` environment variable if set, and pass arguments to `ffprefs` using the new convention (see below)

_(E.g. on Ubuntu, there is generally not a `cargo` executable in `PATH`, but there may be a `cargo-1.91`. This is normally specified via the `CARGO` variable, similar to `CC` and `CXX`)_

`update_service_dumps.py`: Allow specifying `DUMPS_FOLDER` and `ENGINE_DUMPS_FOLDER` on the command line

`ffprefs/src/main.rs`: Allow specifying the prefs and engine dirs on the command line instead of hard-coding their locations relative to a common root dir

### DIFF
--- a/tools/ffprefs/src/main.rs
+++ b/tools/ffprefs/src/main.rs
@@ -107,9 +107,9 @@ use std::env;
 use std::fs;
 use std::path::PathBuf;
 
-const STATIC_PREFS: &str = "../engine/modules/libpref/init/zen-static-prefs.inc";
-const FIREFOX_PREFS: &str = "../engine/browser/app/profile/firefox.js";
-const DYNAMIC_PREFS: &str = "../engine/browser/app/profile/zen.js";
+const STATIC_PREFS: &str = "modules/libpref/init/zen-static-prefs.inc";
+const FIREFOX_PREFS: &str = "browser/app/profile/firefox.js";
+const DYNAMIC_PREFS: &str = "browser/app/profile/zen.js";
 
 #[derive(Serialize, Deserialize, PartialEq, Debug)]
 struct Preference {
@@ -122,12 +122,6 @@ struct Preference {
     mirror: Option<String>,
     locked: Option<bool>,
     sticky: Option<bool>,
-}
-
-fn get_config_path() -> PathBuf {
-    let mut path = env::current_dir().expect("Failed to get current directory");
-    path.push("prefs");
-    path
 }
 
 fn ordered_prefs(mut prefs: Vec<Preference>) -> Vec<Preference> {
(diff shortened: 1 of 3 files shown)


mr-cheffy (approved):
Thanks!

tools/ffprefs/src/main.rs
mr-cheffy, line 309: Is this being used correctly? As the DIR might be different?

mr-cheffy (changes requested):

tools/ffprefs/src/main.rs
```diff
@@ -300,10 +288,10 @@ fn write_preferences(prefs: &[Preference]) {
     fs::write(&dynamic_prefs_path, dynamic_content).expect("Failed to write dynamic prefs");
 }
 
-fn prepare_zen_prefs() {
+fn prepare_zen_prefs(engine_path: &PathBuf) {
     // Add `#include zen.js` to the bottom of the firefox.js file if it doesn't exist
     let line = "#include zen.js";
-    let firefox_prefs_path = get_config_path().join(FIREFOX_PREFS);
+    let firefox_prefs_path = engine_path.join(FIREFOX_PREFS);
```
mr-cheffy, line 294: This can't be right? You are using the engine path here instead of the config one?

tools/ffprefs/src/main.rs
iskunk: Hm, I missed that relative-to-$CWD dir. It's effectively a third (if optional) input.

I've pushed an update to `package.json` that will run `ffprefs` in the top-level directory, instead of cd'ing down into `tools/ffprefs/`.

mr-cheffy (changes requested):

tools/ffprefs/src/main.rs
iskunk: Isn't `browser/app/profile/firefox.js` supposed to be in the engine (FF source)? There is no `firefox.js` in this Git repo.

mr-cheffy (approved):

tools/ffprefs/src/main.rs
```diff
@@ -351,15 +339,19 @@ fn expand_pref_values(prefs: &mut [Preference]) {
...
+    } else {
+        PathBuf::from("engine")
     };
-    env::set_current_dir(&root_path).expect("Failed to change directory");
 
-    prepare_zen_prefs();
-    let mut preferences = load_preferences();
+    prepare_zen_prefs(&engine_path);
+    let mut preferences = load_preferences(&prefs_path);
     expand_pref_values(&mut preferences);
```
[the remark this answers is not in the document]
mr-cheffy: @iskunk could you please check these out?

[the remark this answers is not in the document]
iskunk: I saw the Copilot commentary, but didn't find it relevant---Surfer already creates `.surfer/` in the working directory, so reading anything in there the same way is reasonable, IMO.

Would you rather this program have _three_ inputs? (base_dir, prefs, engine) I already considered setting `IS_TWILIGHT` outside, in the `package.json` invocation, but then figured you might want to pull more values from `dynamicConfig.brand.json` in the future.

mr-cheffy (approved):

iskunk:
Thanks! I put in a [companion PR](https://github.com/zen-browser/surfer/pull/50) for Surfer, along similar lines.

mr-cheffy:
fails to run `npm run import` on windows. https://github.com/zen-browser/desktop/pull/13605

iskunk:
Hm, I wasn't aware this needed to run on Windows. The `${CARGO:-cargo}` bit is likely at fault.

I see there are some [solutions](https://stackoverflow.com/questions/45082648/npm-package-json-os-specific-script) to running OS-specific scripts in a `package.json` file, so I'll review those and put together a reland.

PR #13607, opened 2026-05-07
fix: extension popup panel height on fractional scaling

### LINKED ISSUE #13518: Extensions spills out of the window, broken scaling on extension popup when opened by mouse clicks

shelldandy:
Closes #13518 

After:
image

Before:
Image

### DIFF
--- a/src/browser/components/extensions/ExtensionPopups-sys-mjs.patch
+++ b/src/browser/components/extensions/ExtensionPopups-sys-mjs.patch
@@ -1,12 +1,8 @@
-# https://github.com/zen-browser/desktop/issues/13518
-# Update `lastCalculatedInViewHeight` in the fixedWidth=false branch of
-# `resizeBrowser` so the panel shrinks to the actual popup height.
 diff --git a/browser/components/extensions/ExtensionPopups.sys.mjs b/browser/components/extensions/ExtensionPopups.sys.mjs
+index 1cd35175507bd442aae6857e65015cfed00058e3..376377629e9732b292085ba94309e47cc53ee2ef 100644
 --- a/browser/components/extensions/ExtensionPopups.sys.mjs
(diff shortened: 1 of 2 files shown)


mr-cheffy:
please use `npm run export` a file inside the engine dir to export a valid patch file. Also, it seems to be done on purpose to not set this variable in this branch: https://bugzilla.mozilla.org/show_bug.cgi?id=1403466

shelldandy:
@mr-cheffy  customRectGetter falls back to a stale viewHeight from initial attachment in the fixedWidth=false branch because lastCalculatedInViewHeight never gets set, so I've mirrored the fixedWidth=true branch exactly using Math.max(height, this.viewHeight).

mr-cheffy:
Wouldn't it make more sense to submit this patch to firefox and we pull it from upstream?

shelldandy:
Yeah makes sense. I added it here since I thought it might be faster. Let's keep this open and I'll open it as well on Firefox upstream

PR #13619, opened 2026-05-08
Respect newtab button top pref more consistently

ryan-griffin:


src/browser/components/tabbrowser/content/tabbrowser-js.patch
```diff
@@ -513,16 +513,24 @@ index 2643e1a2aa14ba5cb4a64a92e1c2dfa5f07d242f..3a21c40eb70621968b5fbfc4b9b6d841
              tabGroup = previousTab.group;
            }
            if (
-@@ -4823,7 +4969,7 @@
+@@ -4818,12 +4964,14 @@
+           ) {
+             elementIndex = Infinity;
+           } else if (previousTab.visible && previousTab.splitview) {
```
mr-cheffy, line 519: this is never true

src/zen/glance/ZenGlanceManager.mjs
mr-cheffy, line 1: Not sure about these, the pref is ignored on purpose here

mr-cheffy (changes requested):

mr-cheffy:
Seems unnecessary changes and some of them are done on purpose. Happy to discuss them and re-open if needed though!

ryan-griffin:
I just felt that its awkward that child tabs don't follow the bottom up approach like new tabs. It sorta messes with my brain but that might be just be me though.

PR #13630, opened 2026-05-09
gh-13629: fix tab renaming input field too narrow

### LINKED ISSUE #13629: Tab rename input only uses ~50% of the available tab width
### Preliminary Checks

### What happened?

When I rename a tab in the sidebar, the inline rename input is only about half as wide as the tab itself. The rest of the tab stays empty. For longer titles the text gets clipped on both sides while I'm typing, so I can't see the full name I'm editing.

### Expected behavior

The rename input should fill the same horizontal space the tab title uses in its normal state. Long names should stay readable while editing without horizontal scrolling inside the input.

### Actual behavior

The input is roughly 50% of the tab's width. Longer titles clip on both ends, even though there's clearly unused space to the right of the input. The normal (non-editing) state uses the full width before truncating; the editing state doesn't.

### Steps to reproduce

Open Zen with a sidebar wide enough to show full tab titles.
Have a tab with a longer title, e.g. `zen-browser/desktop: Welcome to a calmer internet`.
Double-click the tab to enter rename mode.
The input only takes up about half the tab width; the rest stays blank.

### Screenshots and videos

Normal state

Image

Rename state

Image

### Version

1.19.12b (Firefox 150.0.2) (64-bit)

Windows - x64

Tabs

```shell

```

sporocyst:
This PR expands renaming input field to full lenght of its container width.
Tested sucessefully on renaming tabs/folders/pinned tabs/spaces

Fix #13629


mr-cheffy (approved):

mr-cheffy:
Thanks!

PR #13637, opened 2026-05-10
no-bug: A few quality-of-life tweaks, take 2

iskunk:
This should support `npm run import` on Windows. I looked into [run-script-os](https://www.npmjs.com/package/run-script-os), and a couple other approaches, but then figured it would be simpler to just call a small Python script that looks up the environment variable and calls `cargo`/`$CARGO` as appropriate.

Please give this a try, and let me know if you see any regressions.

----

`package.json`: Invoke `cargo(1)` through a Python script that honors the `CARGO` environment variable if set, and pass arguments to `ffprefs` using the new convention (see below)

`run_cargo.py`: Script to invoke `cargo(1)`, as a workaround for reading an environment variable in a `package.json` script in a cross-platform manner

`update_service_dumps.py`: Allow specifying `DUMPS_FOLDER` and `ENGINE_DUMPS_FOLDER` on the command line

`ffprefs/src/main.rs`: Allow specifying the `prefs` and `engine` dirs on the command line instead of hard-coding their locations relative to a common root dir

### DIFF
--- a/configs/linux/mozconfig
+++ b/configs/linux/mozconfig
@@ -2,8 +2,11 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-# Setting the compiler based on the existence of clang bin directory
-if test -d "$HOME/.mozbuild/clang/bin"; then
+# Setting the compiler based on the environment or existence of the clang
+# bin directory
+if test "$CC" && test "$CXX"; then
+    true
+elif test -d "$HOME/.mozbuild/clang/bin"; then
     export CC="$HOME/.mozbuild/clang/bin/clang"
     export CXX="$HOME/.mozbuild/clang/bin/clang++"
 else
--- a/scripts/run_cargo.py
+++ b/scripts/run_cargo.py
@@ -1,0 +1,18 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import os
+import sys
+
+cargo = os.environ.get('CARGO', 'cargo')
+
+command = [cargo] + sys.argv[1:]
+
+try:
+  os.execvp(command[0], command)
+except OSError as e:
+  print(f"{sys.argv[0]}: '{command[0]}': {e}", file=sys.stderr)
+  if 'CARGO' in os.environ:
+    print('(CARGO environment variable is set)', file=sys.stderr)
+  sys.exit(1)
--- a/scripts/update_service_dumps.py
+++ b/scripts/update_service_dumps.py
@@ -3,6 +3,7 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 import os
+import sys
 import json
 from json_with_comments import JSONWithCommentsDecoder
 
@@ -60,4 +61,6 @@ def main():
 
 
 if __name__ == "__main__":
+  if len(sys.argv) == 3:
+    _, DUMPS_FOLDER, ENGINE_DUMPS_FOLDER = sys.argv
   main()
--- a/tools/ffprefs/src/main.rs
+++ b/tools/ffprefs/src/main.rs
@@ -107,9 +107,9 @@ use std::env;
 use std::fs;
 use std::path::PathBuf;
 
-const STATIC_PREFS: &str = "../engine/modules/libpref/init/zen-static-prefs.inc";
-const FIREFOX_PREFS: &str = "../engine/browser/app/profile/firefox.js";
-const DYNAMIC_PREFS: &str = "../engine/browser/app/profile/zen.js";
+const STATIC_PREFS: &str = "modules/libpref/init/zen-static-prefs.inc";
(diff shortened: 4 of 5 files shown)


scripts/update_service_dumps.py
```diff
@@ -60,4 +60,7 @@ def main():
 
 
 if __name__ == "__main__":
+  import sys
+  if len(sys.argv) == 3:
+    _, DUMPS_FOLDER, ENGINE_DUMPS_FOLDER = sys.argv
   main()
```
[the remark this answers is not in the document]
iskunk: `update_service_dumps.py` is not a general-purpose utility that needs extensive input validation and usage information. My goal here was a minimal change that allows the locations to be specified externally.

tools/ffprefs/src/main.rs
```diff
@@ -351,15 +339,19 @@ fn expand_pref_values(prefs: &mut [Preference]) {
...
     };
-    env::set_current_dir(&root_path).expect("Failed to change directory");
 
-    prepare_zen_prefs();
-    let mut preferences = load_preferences();
+    prepare_zen_prefs(&engine_path);
+    let mut preferences = load_preferences(&prefs_path);
     expand_pref_values(&mut preferences);
-    write_preferences(&preferences);
+    write_preferences(&engine_path, &preferences);
```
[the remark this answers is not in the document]
iskunk: Zen build tooling already works with `$CWD/.surfer/`.

scripts/run_cargo.py
```diff
@@ -0,0 +1,16 @@
...
+import sys
+
+cargo = os.environ.get('CARGO', 'cargo')
+
+command = [cargo] + sys.argv[1:]
+
+try:
+  os.execvp(command[0], command)
+except Exception:
+  print('Exec failed', file=sys.stderr)
```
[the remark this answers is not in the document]
iskunk: I've pushed an update that makes the error message more informative.

iskunk:
I've rebased and resolved the merge conflict, and also tweaked the Linux `mozconfig` to allow user-set values for `CC` and `CXX`.

configs/linux/mozconfig
```diff
@@ -2,8 +2,11 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-# Setting the compiler based on the existence of clang bin directory
-if test -d "$HOME/.mozbuild/clang/bin"; then
+# Setting the compiler based on the environment or existence of the clang
+# bin directory
+if test "$CC" && test "$CXX"; then
+    true
+elif test -d "$HOME/.mozbuild/clang/bin"; then
```
mr-cheffy, line 9: Not sure about this, maybe instead, we should just define CC and CXX if ZEN_RELEASE tests true?

scripts/update_service_dumps.py
```diff
@@ -60,4 +60,7 @@ def main():
 
 
 if __name__ == "__main__":
+  import sys
```
mr-cheffy: Imports at the top please

tools/ffprefs/src/main.rs
```diff
@@ -351,15 +339,19 @@ fn expand_pref_values(prefs: &mut [Preference]) {
...
 fn main() {
     let args: Vec<String> = env::args().collect();
-    let root_path = if args.len() > 1 {
+    let prefs_path = if args.len() > 1 {
         PathBuf::from(&args[1])
     } else {
-        env::current_dir().expect("Failed to get current directory")
+        PathBuf::from("prefs")
+    };
+    let engine_path = if args.len() > 2 {
```
mr-cheffy: First and second args should be required

mr-cheffy (changes requested):

configs/linux/mozconfig
iskunk: It's still good to set the variables generally, because the configuration might default to `gcc`/`g++` otherwise. And the user might want to do a `ZEN_RELEASE` build using their compilers instead of the ones in `.mozbuild/`.

scripts/update_service_dumps.py
iskunk: Got it

tools/ffprefs/src/main.rs
iskunk: Okay, no defaults for these.

(My Rust-fu isn't so good yet; let me know if you want this written differently)

configs/linux/mozconfig
mr-cheffy: In that case, you should just use your own mozconfig file at the root directory to override these

iskunk: That's non-obvious, however, for just overriding the compilers. It's customary across many build systems to use `CC`/`CXX` when set in the environment, and especially when the user isn't familiar with the ins and outs of the build system, supporting common conventions like that helps make it easier to deal with.

In any event, nothing changes here unless both variables are set, to avoid any ambiguity.

iskunk: I've rebased and resolved the new conflicts.

Note that a big part of the reason for using Surfer in the first place is to abstract the build away from the user, so that they don't have to be familiar with the Mozilla build system in order to build Zen. Requiring the user to mess with mozconfig files, for something as basic as CC/CXX, defeats that purpose.

PR #13645, opened 2026-05-10
no-bug: Improve flaky URLBar scrolling

jakubiakdev:
Yesterday when opening Zen after updating I checked out the changelog that mentioned fixing [CVE-2026-44659](https://github.com/zen-browser/desktop/security/advisories/GHSA-7p2r-fp29-9w69) and tried replicating the issue out of curiosity.

As I had `zen.urlbar.show-domain-only-in-sidebar` set to false, the new version didn't fix anything for me which prompted me to investigate further into this issue.

I found out that ce1ee92 tried to fix this by waiting for an animation frame to set the scroll position when trimming the URL, but with further testing it was flaky, inconsistent and would break when resizing the window/sidebar.

Turns out `browser/components/urlbar/UrlbarValueFormatter.sys.mjs` already manages the scroll position for the URL bar, and by moving the logic to this file we don't have to request the animation frame for the change to stick, and the function is executed more times instead of only on load. I also took the time to implement a fix for `zen.urlbar.show-domain-only-in-sidebar=false` case.

Before (stable .appImage):
ezgif-before

After (with `zen.urlbar.show-domain-only-in-sidebar=true`, the default)
ezgif-after-default

After (with `zen.urlbar.show-domain-only-in-sidebar=false`)
ezgif-after-nondefault


src/zen/common/modules/ZenUIManager.mjs
```diff
@@ -702,9 +702,7 @@ window.gZenUIManager = {
         const hoverAttr = "zen-has-implicit-hover";
         const hasHover = gNavToolbox.hasAttribute(hoverAttr);
         gNavToolbox.removeAttribute(hoverAttr);
```
mr-cheffy: Please remove the entire frame callback as its useless now.

src/browser/components/urlbar/UrlbarValueFormatter-sys-mjs.patch
```diff
@@ -11,7 +11,62 @@ index f28d277764158566bc9406ce0e6160d92d346a63..82dd5f8568ec1b12e87676e5c2b24382
      });
    }
  
-@@ -371,7 +371,7 @@ export class UrlbarValueFormatter {
+@@ -105,6 +105,41 @@ export class UrlbarValueFormatter {
+   }
+ 
+   #ensureFormattedHostVisible(urlMetaData) {
++    // Zen specific change to accommodate a trimmed url,
```
mr-cheffy: No need for this comment, lets try keep patches short :)

src/browser/components/urlbar/UrlbarValueFormatter-sys-mjs.patch
```diff
@@ -11,7 +11,62 @@ index f28d277764158566bc9406ce0e6160d92d346a63..82dd5f8568ec1b12e87676e5c2b24382
...
+   #ensureFormattedHostVisible(urlMetaData) {
++    // Zen specific change to accommodate a trimmed url, 
++    // making sure the domain is visible even with long subdomains 
++    // in a small URL input in the sidebar.
++    // In other cases we keep Firefox's default behavior
++    if (
++      this.#window.gZenVerticalTabsManager?._hasSetSingleToolbar
++      && Services.prefs.getBoolPref("zen.urlbar.show-domain-only-in-sidebar")
++    ) {
++      this.#inputField.scrollLeft = this.#inputField.scrollLeftMax;
```
mr-cheffy, line 27: Like what the animation frame did before, you should remove the attribute, and set it back again so the buttons don't interfere with the scroll

jakubiakdev:
I tried making my commit verified and force pushing but looks like it didn't go through, whoops

src/browser/components/urlbar/UrlbarValueFormatter-sys-mjs.patch
```diff
@@ -11,7 +11,62 @@ index f28d277764158566bc9406ce0e6160d92d346a63..82dd5f8568ec1b12e87676e5c2b24382
...
++    
++    urlMetaData = urlMetaData || this.#getUrlMetaData();
++    if (!urlMetaData) {
++      this.#urlbarInput.removeAttribute("domaindir");
++      return;
++    }
++    let { url, preDomain, domain } = urlMetaData;
++
++    if (
++      this.#window.gZenVerticalTabsManager?._hasSetSingleToolbar
```
mr-cheffy: I dont think we should be doing this, as I believe showing the protocol would be more helpful here. Regardless, this is a pref almost no one turns so it shouldn't be worth the hassle to maintain either.

mr-cheffy (changes requested):

src/browser/components/urlbar/UrlbarValueFormatter-sys-mjs.patch
jakubiakdev: TBH I'm using this perf daily and was actually sometimes noticing that I didn't see the full domain. I think it's a reasonable compromise to still be able to see it, especially since the way urlbarTrim works the protocol was trimmed either way.

src/browser/components/urlbar/UrlbarValueFormatter-sys-mjs.patch
jakubiakdev: I've re-implemented this inside UrlbarValueFormatter. It still quirky for some reason (doesn't always take the text back when buttons are shown) but the behavior is similar to stable. I will probably have to figure out how the buttons are being shown but I'm low on time for the next couple of days

jakubiakdev:
Force pushing because the first commit wasn't verified, the whole PR should be verified now

mr-cheffy (approved):
I've reverted some of the changes as I'd like to get the resize patch out ASAP, please feel free to create a new PR so we can properly test the new functionality :)

PR #13659, opened 2026-05-11
gh-13653: fix copy url button/shortcut to respect `decodeURLsOnCopy` spec

### LINKED ISSUE #13653: "Copy URL" button/shortcut does not respect `decodeURLsOnCopy` flag
### Preliminary Checks

### What happened?

By default, Firefox copies Unicode urls to clipboard using punycode. However, there is an `about:config` flag that changes this behavior, `browser.urlbar.decodeURLsOnCopy`. After setting it to `true`, if I select url in address bar and copy it from there, unicode urls are copied as-is. However, when I use Cmd+Shift+C shortcut or press Copy button to the left of the address bar, copied url still contains punycode.

Example url: `https://uk.wikipedia.org/wiki/Ефект_Мандели`
Cmd+C from address bar: `https://uk.wikipedia.org/wiki/Ефект_Мандели`
Cmd+Shift+C: `https://uk.wikipedia.org/wiki/%D0%95%D1%84%D0%B5%D0%BA%D1%82_%D0%9C%D0%B0%D0%BD%D0%B4%D0%B5%D0%BB%D0%B8`

### Expected behavior

Urls copied using Cmd+Shift+C shortcut don't contain punycode when decodeURLsOnCopy is enabled

### Actual behavior

Urls copied using Cmd+Shift+C shortcut still contain punycode when decodeURLsOnCopy is enabled

### Steps to reproduce

- Go to `about:config`
- Set `browser.urlbar.decodeURLsOnCopy` to `true`
- Go to `https://uk.wikipedia.org/wiki/Ефект_Мандели` (or any other url that contains non-ascii characters)
- Copy url using `Cmd+Shift+C` shortcut
- Paste copied url back into address bar
- It will look like `https://uk.wikipedia.org/wiki/%D0%95%D1%84%D0%B5%D0%BA%D1%82_%D0%9C%D0%B0%D0%BD%D0%B4%D0%B5%D0%BB%D0%B8` and not `https://uk.wikipedia.org/wiki/Ефект_Мандели`

### Screenshots and videos

https://github.com/user-attachments/assets/2fe72c8d-c0a0-4af6-8fef-6ae9adda4c2c

### Version

1.19.12b

macOS - aarch64

URL Bar

```shell

```

ashvwinn:
Decodes URL on copy by shortcut/button if `browser.urlbar.decodeURLsOnCopy` is true
fixes #13653


mr-cheffy (changes requested):
We should do it more like in https://searchfox.org/firefox-main/source/browser/components/urlbar/content/SmartbarInput.mjs#4144-4154

ashvwinn:
This uses the URL that the user selected which is not encoded. But getting `currentURI` directly with the API will return the encoded URI. Unless I am missing something.

mr-cheffy:
I meant the code structure instead of having 2 different calls to "copy to clipboard". Also, checking if the url is a data url

ashvwinn:
I have made the changes as requested. Do tell me if something is still not right.

src/zen/common/modules/ZenCommonUtils.mjs
```diff
@@ -118,8 +129,20 @@ window.gZenCommonActions = {
...
-    const markdownLink = `[${tabTitle}](${currentUrl.displaySpec})`;
+    let displaySpec = currentUrl.displaySpec;
+
+    // decode URI if `browser.urlbar.decodeURLsOnCopy` is enabled
+    // and is not a data URI
+    if (
+      Services.prefs.getBoolPref('browser.urlbar.decodeURLsOnCopy', false) &&
+      !currentUrl.schemeIs("data")
+    ) {
+      displaySpec = decodeURI(displaySpec);
```
mr-cheffy: Could decodeURI throw an error? Is there a possibility for that? Also, remove the useless comments please

mr-cheffy (changes requested):

src/zen/common/modules/ZenCommonUtils.mjs
ashvwinn: It can but only if the uri is malformed.

>[URIError](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/>Global_Objects/URIError)
> 
>    Thrown if encodedURI contains a % not followed by two hexadecimal digits, or if >the escape sequence does not encode a valid UTF-8 character.

This would only happen if the URI is not encoded properly. Passing a normal URI without any special characters still work.

ashvwinn: Now, it copies the encoded URI if decodeURI throws an error.

mr-cheffy (changes requested):
Please fix lint: https://github.com/zen-browser/desktop/actions/runs/25793929248/job/75766372262?pr=13659#step:11:56.

You can run `npm run lint:fix`

mr-cheffy (approved):

PR #13664, opened 2026-05-12
Fix: Preserve Essential tab status when moving to new window

### LINKED ISSUE #8309: Moving Essential Tab to New Window removes it from Essentials
### Captchas

### What happened?

Part of my workflow that I'm used to involves opening multiple windows at times, especially if I need to reference something from one site for another.

At the moment, I've found the easiest way is to use the "Move to a New Window" option in the context menu. But on essentials, that actually ends up removing it from essentials, and makes it become just a standard tab in the new window.

I think this could potentially be circumvented by adding a new option that allows you to "Duplicate" to a new window. But I do think the fact that this action by default is destructive to something that should be a bit harder to accidentally irreversibly alter is a big issue. Essentials should almost be set in stone unless explicitly changed by the user and not by some auxiliary action that doesn't imply such behaviour, so in my opinion I feel like the "Move" options should act like "Duplicate" when done on essentials.

### Version

1.12.4b

Windows - x64

Tabs

```shell

```

shalin-dev:
## Summary

Fixes #8309

When a user moves an Essential tab to a new window via "Move to a New Window" from the context menu, the tab loses its Essential status and becomes a regular tab. This is a destructive action that should not alter the Essential designation.

## Root Cause

The `adoptTab()` method in `tabbrowser.js` creates a new tab element and swaps the browser/docshell into it. DOM attributes like `zen-essential` are **not** transferred to the new element. The existing code only preserved `zenStaticLabel` but missed `zen-essential`, `_zenPinnedInitialState`, and `zenStaticIcon`.

This is the same pattern that `ZenPinnedTabManager.addToEssentials()` already handles for drag-and-drop operations (explicitly re-setting `zen-essential` after `adoptTab`), but the base `adoptTab` method itself was missing this preservation.

### `src/browser/components/tabbrowser/content/tabbrowser-js.patch`
Added preservation of Essential-specific properties during tab adoption:
- `zen-essential` attribute (the Essential status)
- `_zenPinnedInitialState` (needed for pinned tab reset functionality)
- `zenStaticIcon` (custom icon if set)

### `src/zen/tests/pinned/browser_issue_8309.js` (new)
Regression test that:
1. Creates a tab and marks it as Essential
2. Moves it to a new window via `replaceTabsWithWindow()`
3. Asserts the tab retains `zen-essential` attribute and appears in the essentials container

### `src/zen/tests/pinned/browser.toml`
Added test manifest entry for the new test file.

## Testing

- [x] Code follows existing patterns (matches `ZenPinnedTabManager.addToEssentials()` adoption pattern)
- [x] Regression test added
- [ ] CI build and test pass (awaiting CI)

## Acceptance Criteria (from issue)

- [x] Moving an Essential tab to a new window preserves its Essential status
- [x] The tab appears as Essential in the new window's sidebar (verified by test checking `.zen-essentials-container`)
- [x] Works on all platforms (fix is platform-independent JavaScript)
- [x] No regressions in standard tab move behavior (only adds conditional preservation, doesn't change existing behavior)
- [x] PR targets `dev` branch with issue reference

### DIFF
--- a/src/browser/components/tabbrowser/content/tabbrowser-js.patch
+++ b/src/browser/components/tabbrowser/content/tabbrowser-js.patch
@@ -1038,9 +1038,18 @@ index 2643e1a2aa14ba5cb4a64a92e1c2dfa5f07d242f..3a21c40eb70621968b5fbfc4b9b6d841
 @@ -7447,6 +7695,8 @@
        }
        params.skipLoad = true;
-       let newTab = this.addWebTab("about:blank", params);
+        let newTab = this.addWebTab("about:blank", params);
 +      newTab._zenContentsVisible = true;
 +      newTab.zenStaticLabel = aTab.zenStaticLabel;
++      if (aTab.hasAttribute("zen-essential")) {
++        newTab.setAttribute("zen-essential", "true");
++      }
++      if (aTab._zenPinnedInitialState) {
++        newTab._zenPinnedInitialState = aTab._zenPinnedInitialState;
++      }
++      if (aTab.zenStaticIcon) {
++        newTab.zenStaticIcon = aTab.zenStaticIcon;
++      }
  
        aTab.container.tabDragAndDrop.finishAnimateTabMove();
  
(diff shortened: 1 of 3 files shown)


mr-cheffy:
This is not what the issue meant

PR #13668, opened 2026-05-12
no-bug: Add keyboard shortcuts to move tabs between workspaces

kristof-nyr:
## Summary

Adds 12 new keyboard shortcuts to move the active (or multiselected) tabs to a specific workspace or to the next/previous workspace.

Addresses community requests in:
- https://github.com/zen-browser/desktop/discussions/4805
- https://github.com/zen-browser/desktop/discussions/4363
- https://github.com/zen-browser/desktop/discussions/1058

### What's included

- **Commands:** 12 new `<command>` elements in `zen-commands.inc.xhtml`
- **Move logic:** `shortcutMoveTabTo(index)` and `moveActiveTabShortcut(offset)`
  in `ZenSpaceManager.mjs`; empty tabs (`zen-empty-tab`) are filtered out to
  prevent moving placeholder tabs
- **Feedback:** Toast notification ("Moved tab to [name]" / "Moved N tabs to
  [name]") using existing `gZenUIManager.showToast`; the toast container's ARIA
  live region provides screenreader announcements automatically
- **Shortcuts:** Registered in `ZenKeyboardShortcuts.mjs` with migration to
  version 19; extracted shared `createMoveTabShortcuts()` factory so fresh-install
  and migration paths cannot diverge
- **Defaults:** Ctrl+Shift+1..0 on macOS (mirrors Ctrl+1..0 for switching);
  unbound on other platforms
- **Tests:** 10 mochitest tasks covering single move, multiselect, wrap/no-wrap,
  same-workspace guard, empty-tab fallback, and toast verification

### Design decisions

- User stays in the current workspace after moving a tab (matches Discussion
  #1058 consensus)
- Shortcuts unbound by default on non-macOS (mirrors existing workspace-switch
  pattern)
- Out-of-bounds and same-workspace moves are silent no-ops
- Empty/placeholder tabs are skipped (no-op if workspace only has an empty tab)

### Behavioral note for reviewers

In `zen-sets.js`, the `default` case was reordered: workspace-switch and
move-tab commands are now checked **before** `gZenGlanceManager.handleMainCommandSet()`.
Previously the glance manager saw all default-case commands including workspace
switches. The new order only passes truly unrecognized commands to the glance
manager. This is more correct (workspace commands aren't glance-related) but
reviewers should confirm the glance manager doesn't depend on seeing those events.


src/browser/base/content/zen-commands.inc.xhtml
```diff
@@ -30,6 +30,19 @@
...
+  <command id="cmd_zenMoveTabToWorkspace2" />
+  <command id="cmd_zenMoveTabToWorkspace3" />
+  <command id="cmd_zenMoveTabToWorkspace4" />
+  <command id="cmd_zenMoveTabToWorkspace5" />
+  <command id="cmd_zenMoveTabToWorkspace6" />
+  <command id="cmd_zenMoveTabToWorkspace7" />
+  <command id="cmd_zenMoveTabToWorkspace8" />
+  <command id="cmd_zenMoveTabToWorkspace9" />
+  <command id="cmd_zenMoveTabToWorkspace10" />
+  <command id="cmd_zenMoveTabForward" />
```
mr-cheffy, line 43: Between spaces right?

src/zen/kbs/ZenKeyboardShortcuts.mjs
```diff
@@ -628,6 +628,50 @@ class KeyShortcut {
   }
 }
 
+function createMoveTabShortcuts() {
+  const shortcuts = [];
+  for (let i = 10; i > 0; i--) {
+    shortcuts.push(
+      new KeyShortcut(
+        `zen-move-tab-to-workspace-${i}`,
```
mr-cheffy, line 636: To be honest, I dont think we should have shortcuts to move these to individual spaces, they would clutter up the keyboard shortcuts too much.

mr-cheffy (changes requested):

src/browser/base/content/zen-commands.inc.xhtml
kristof-nyr: Exactly. Missed putting it into the name

src/zen/kbs/ZenKeyboardShortcuts.mjs
kristof-nyr: I see your point. My original pain point was that most of the time I use shortcuts for switching between workspaces, and switching between tabs within a workspace (ctrl + [1-9], cmd + [1-9] respectively, on Mac). So when I clicked a link, and it opened the tab in a "wrong" workspace, I'd want to move it to the "correct" one. Since there is already a built-in "Move tab" option, when you right click the tab itself, I thought making a keybind for it would make sense. I also read the issues linked in the description, although those went kind of stale.
If you know any existing way to address my pain point, or propose alternatives, I'm open to collaborate.

Mr7F:
It would be nice to have 2 keybinds to move tabs within a workspace too (move the tab after/before the next/previous one)

andrico1234:
This would be an amazing feature to have. I'll be happy to help if anyone needs help getting this over the line

PR #13700, opened 2026-05-14
feat(folders): Add zen.folders.new-tabs-in-folder to open new tabs in current folder

mehdi-sy-h:
Adds a new `zen.folders.new-tabs-in-folder` preference. When enabled, new tabs via Ctrl + T will open under the currently open tab's parent folder. This saves having to continually move new tabs into the desired folder.

Relevant discussion: #13669

### DIFF
--- a/src/zen/folders/ZenFolders.mjs
+++ b/src/zen/folders/ZenFolders.mjs
@@ -397,6 +397,18 @@ class nsZenFolders extends nsZenDOMOperatedFeature {
   on_TabOpen(event) {
     const tab = event.target;
     const group = tab.group;
+    if (
+      Services.prefs.getBoolPref("zen.folders.new-tabs-in-folder")
+      && !group?.isZenFolder
+      && !tab.owner
+    ) {
+      const activeFolder = gBrowser.selectedTab?.group;
+      if (activeFolder?.isZenFolder) {
+        gBrowser.pinTab(tab);
+        activeFolder.addTabs([tab]);
+        return;
+      }
+    }
     if (!group?.isZenFolder || tab.pinned) {
       return;
     }
--- a/src/zen/folders/ZenFolders.mjs
+++ b/src/zen/folders/ZenFolders.mjs
@@ -398,26 +398,27 @@ class nsZenFolders extends nsZenDOMOperatedFeature {
     const tab = event.target;
     const group = tab.group;
     if (
-      Services.prefs.getBoolPref("zen.folders.new-tabs-in-folder")
-      && !group?.isZenFolder
-      && !tab.owner
+      Services.prefs.getBoolPref("zen.folders.owned-tabs-in-folder") &&
+      group?.isZenFolder &&
+      !tab.pinned
     ) {
-      const activeFolder = gBrowser.selectedTab?.group;
-      if (activeFolder?.isZenFolder) {
-        gBrowser.pinTab(tab);
-        activeFolder.addTabs([tab]);
-        return;
-      }
-    }
-    if (!group?.isZenFolder || tab.pinned) {
-      return;
-    }
-    // Edge case: In occations where we add a tab with an ownerTab
-    // inside a folder, the tab gets added into the folder in an
-    // unpinned state. We need to pin it and re-add it into the folder.
-    if (Services.prefs.getBoolPref("zen.folders.owned-tabs-in-folder")) {
+      // Edge case: In occations where we add a tab with an ownerTab
+      // inside a folder, the tab gets added into the folder in an
+      // unpinned state. We need to pin it and re-add it into the folder.
       gBrowser.pinTab(tab);
       group.addTabs([tab]);
+      return;
+    }
+    const activeFolder = gBrowser.selectedTab?.group;
+    if (
+      Services.prefs.getBoolPref("zen.folders.new-tabs-in-folder") &&
+      !group?.isZenFolder &&
+      !tab.owner &&
+      activeFolder?.isZenFolder &&
+      !activeFolder?.isLiveFolder &&
+      !activeFolder?.hasAttribute("split-view-group")
+    ) {
+      activeFolder.addTabs([tab]);
     }
   }
 
--- a/prefs/zen/folders.yaml
+++ b/prefs/zen/folders.yaml
@@ -13,3 +13,6 @@
 
 - name: zen.folders.owned-tabs-in-folder
   value: false
+
+- name: zen.folders.new-tabs-in-folder
+  value: false
--- a/src/zen/tests/folders/browser.toml
+++ b/src/zen/tests/folders/browser.toml
@@ -26,6 +26,8 @@ support-files = [
 
 ["browser_folder_multiselected.js"]
 
+["browser_folder_new_tabs.js"]
+
 ["browser_folder_owner_tabs.js"]
 
 ["browser_folder_reset_button.js"]
--- a/src/zen/tests/folders/browser_folder_new_tabs.js
+++ b/src/zen/tests/folders/browser_folder_new_tabs.js
@@ -1,0 +1,91 @@
+/* Any copyright is dedicated to the Public Domain.
+   https://creativecommons.org/publicdomain/zero/1.0/ */
+
+"use strict";
+
+add_task(async function test_New_Tab_Inside_Folder() {
+  await SpecialPowers.pushPrefEnv({
+    set: [["zen.folders.new-tabs-in-folder", true]],
+  });
+  const selectedTab = gBrowser.selectedTab;
+  const tab = BrowserTestUtils.addTab(gBrowser, "about:blank");
+  const folder = await gZenFolders.createFolder([tab], {
+    renameFolder: false,
+  });
+  gBrowser.selectedTab = tab;
+  const triggeringPrincipal = Services.scriptSecurityManager.getSystemPrincipal();
+
+  const newTab = gBrowser.addTab("https://example.com", {
+    inBackground: true,
+    triggeringPrincipal,
+  });
+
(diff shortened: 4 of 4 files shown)


mr-cheffy:
We don't typically add patches like these, specially if they are just going to be hidden advanced preferences and without any tests. What would be the use case of this?

mehdi-sy-h:
Sorry yes I can add some tests  I was just wondering if this is a feature that might get merged. There's already the `zen.folders.owned-tabs-in-folder` pref which does something similar but for tabs that were middle clicked.

As for use case, generally when you're browsing/researching about a specific topic you might want to keep all those tabs under one folder, especially if you're someone who has a lot of tabs already and want things structured.  Middle click is useful but not complete because you might read something for example and then search for that specific thing with Ctrl+T and have that tab be under the same folder. Without this feature there's a lot of friction in moving the relevant new tab(s) into the folder again and again and your tabs end up being cluttered again.

Imo this is a slightly separate use case than Spaces, because you might want the same containers, theme settings, etc but you still want grouping by subtopic within that space. Say you have a space for "Work"  with its own container, and you're working on a bunch of different projects at work, so you group them by folder. Switching to different projects at a given time is rare, but searching for things related to the same topic is more common (you're reading docs, you see a concept you're not aware of, you open a new tab searching for it, etc).

src/zen/folders/ZenFolders.mjs
```diff
@@ -397,6 +397,18 @@ class nsZenFolders extends nsZenDOMOperatedFeature {
...
     const group = tab.group;
+    if (
+      Services.prefs.getBoolPref("zen.folders.new-tabs-in-folder")
+      && !group?.isZenFolder
+      && !tab.owner
+    ) {
+      const activeFolder = gBrowser.selectedTab?.group;
+      if (activeFolder?.isZenFolder) {
+        gBrowser.pinTab(tab);
+        activeFolder.addTabs([tab]);
```
mr-cheffy: Doesn't `addTabs` already pin the tab itself?

src/zen/folders/ZenFolders.mjs
```diff
@@ -397,6 +397,18 @@ class nsZenFolders extends nsZenDOMOperatedFeature {
   on_TabOpen(event) {
     const tab = event.target;
     const group = tab.group;
+    if (
+      Services.prefs.getBoolPref("zen.folders.new-tabs-in-folder")
+      && !group?.isZenFolder
+      && !tab.owner
+    ) {
+      const activeFolder = gBrowser.selectedTab?.group;
+      if (activeFolder?.isZenFolder) {
```
mr-cheffy: We should ignore livefolders as well maybe?

src/zen/folders/ZenFolders.mjs
```diff
@@ -397,6 +397,18 @@ class nsZenFolders extends nsZenDOMOperatedFeature {
   on_TabOpen(event) {
     const tab = event.target;
     const group = tab.group;
+    if (
+      Services.prefs.getBoolPref("zen.folders.new-tabs-in-folder")
+      && !group?.isZenFolder
```
mr-cheffy: Unsure how this would work with split views.

mr-cheffy (changes requested):

src/zen/folders/ZenFolders.mjs
mehdi-sy-h: For split views should I see if all of the tabs are part of the same folder and then add them, otherwise not?

mehdi-sy-h: Actually it looks like trying to do a split view with a tab in a folder and a tab outside of a folder is buggy in upstream already. And for a split view of tabs across two different folders, it moves one tab into the other's folder?

mehdi-sy-h:
What kind of tests would be useful? I imagine one similar to the existing owner tabs test `test_Duplicate_Tab_Inside_Folder`.

gummms:
I'd really like to see this implemented. 
**Here's my analysis as a UX Designer and user:**

## **tl;dr**
The lack of this feature goes against Zen's principles and is a potential "exit door" to some users. It's implementation is a very minor development change that would output a great usability performance.

---

## To reinforce @mehdi-sy-h answer on what would be the use case
First of all, I'm aware that folders are intended to be used more as bookmarks instead of for navigation, but unless I'm mistaken, there's no other functionality that allows to create "sub-spaces" (a smaller group of tabs nested inside a major group of tabs (spaces)). While the Container feature can work as a tabs "categorizer", it's goals are completely different from what is being proposed here, which is simply an UX improvement.

>Say you have a space for "Work" with its own container, and you're working on a bunch of different projects at work, so you group them by folder.

This is almost exactly my case, but within my "Work" space, I have folders not for different projects, but for different jobs. Let's say "Hiring Company" (HC), "My Company" (MC) and "Freelances" (F). If I'm doing research related to MC, within it's folder, and come across a term that I'm curious about, I'll hit `Ctrl + T` to research on that subject and suddenly I'm kicked out of the folder, doing a specific research among generalists tabs. Then I have to drag this tab back to the folder I was using. **Using different spaces for HC, MC and F would quickly overcrowd my sidebar with buttons and I don't want to change environments while changing folders, since they can (and they do) have related tabs and subjects.**

## The negative impacts from a UX perspective
This seems like a very minor issue (and at the great development scope, most likely is), but being kicked out of the environment I was plus having to manually drag a component is a major bummer and workflow breaker. Specially since this doesn't happen once or two times per session - I hit a lot of `Ctrl + T`'s while browsing. **At this point I am heavily invested in Zen, but this made me consider changing browsers, so it might not be that small of an issue.**

## Still, why bother?
>We don't typically add patches like these, specially if they are just going to be hidden advanced preferences

- The main selling point of Zen is cleaness, productivity and anti-stress navigation. The main page features phrases such as *"we care about your experience"* and *"Zen [...] help you stay productive and focused. Browsers should be tools that help you get things done, not distractions that keep you from your work."*. **This issue directly attacks Zen's philosophy. When I experience this, it actually stresses me and breaks my focus.**

- This is a **very** minor change on a development point of view that outputs a **great** usability improvement. Besides the work already being made here at this PR, the functionality `zen.folders.new-tabs-in-folder` is very similar to what is being proposed here. **Either implementing this feature to this already existing config or using it as a base to a new one should be relatively simple.**

- Zen's another selling points are it's beauty, simplicity and customization capabilities. Arguing that *"we usually don't add features that will stay hidden at advanced preferences"* goes against this. Zen's avarage user **will** care about customization and **probably** will come across advanced settings while using it (specially since the general settings don't change much). The user is even encouraged to try mods, community themes, addons and plugins. **This is actually the main reason I haven't changed browsers. I already experienced how easy it is to customize Zen and have done it heavily, and I don't want to do it again. Having one more customization option wouldn't be a bother, but a welcomed update.**

MrDananas:
Hello! I also want to pile on onto this topic. I'd definitely consider myself a power user but I think this issue is as basic as it gets. I see a lot of pushback on different forums stating it's supposed to be this way "because of bookmarks", which, fair enough if you have a vision for it. But why then have so many fantastic productivity features to organize and be productive within the pinned tabs section? If you're supposed to use it like bookmarks, why have so many advanced features that bookmarks usually don't have? Just the fact that this section is at the forefront of the browser where bookmarks typically don't are. If it's supposed to be bookmarks, call it that and not something more general and misleading like "pinned tabs". And also, what's the point of having essentials, which are also a kind of bookmark/starred system, if pinned tabs are supposed to be that? It does not make any sense at all. 

Either let us be able to use folders outside of pinned tabs or just do this incredibly easy change to optionally be able to open a new tab below sister tab and open links below sister tab.

I recently found zen browser after constantly hopping between other browsers, never finding the perfect productivity browser, and I really thought this was it. I even recommended it to lot of my folks saying it was the be-all-end-all browser but now I think I might need to move on, again...

mehdi-sy-h:
Hi, I'm still willing to work on this or hand it off to anyone else who's interested. Given how small this change is, and assuming I haven't added a bug in the few lines of code I changed, an alternative implementation would likely converge to something similar anyway.

Please let me know what changes you'd like me to make, or if this is not a feature you would consider merging. Regarding your comment on split views, I simply decided not to add them in that situation.

washingdone:
I'd like to bump this PR as it seems like an incredibly easy thing to implement and honestly feels like an oversight to not have included already. I've only been using Zen for a couple of days and it's already been a game-changing experience, but this functionality is the one thing i find myself missing from other browsers. My personal use case for folders is to segment out my tabs int groups for easier navigation - the best example I can provide is when I'm working on coding projects. I'm currently working on a script that makes use of multiple 3rd party APIs as well as several libraries that I am not familiar with, leading me to have a lot of reference tabs open, and using folders to make it easier to navigate between different groups of documentation has been a significant improvement over traditional tab groups. However, every time I open a new piece of reference I'm required to pause my flow for a moment to manually drag the tab into the correct folder. This gets old very quickly, especially if i need to open the references for several functions at once. This one tweak would be all it takes for this browser to go from "the browser that made me realize how much I hate browsers" to "the browser that's literally perfect for me", so I hope it can be implemented.

walefisch:
Would also love to have this added.

DarqSh:
Any progress on this? As @gummms has pointed out, there is no way to make "subspaces" for now, so not allowing to make new tabs inside folders seems like an overlooked design flaw. 

I'd like two separate spaces for my "Home" usage in which I'd like to have folders e.g. "Cooking recipes", "Comparing a price of xyz product between different stores", "Movie subscription services" and then I'd like to have a space for "University" stuff where I'll be able to sort tabs based on which course I need them for. Instead, a space has to be done for each of those topics which are logically separated but have into the same "Spaces" space.

PR #13724, opened 2026-05-17
no-bug: Move boost import/export buttons

FlorianButz:


mr-cheffy (approved):

src/zen/boosts/zen-boost-editor.inc.xhtml
```diff
@@ -100,10 +100,6 @@
       </button>
 
     </vbox>
-    <hbox flex="1" class="footer" id="zen-boost-toolbar-wrapper">
```
mr-cheffy, line 103: Should we now remove the `zen-boost-toolbar-wrapper` styles for footer?

PR #13728, opened 2026-05-17
fix(kbs): restore missing pin tab shortcut

### LINKED ISSUE #4504: "Pin Tab" missing in keyboard shortcut settings
### Captchas

### What happened?

I have set it up to CMD+E sometime ago and it worked fine. Today I noticed it doesn't work, so I wanted to make sure it is set up correctly and it looks like it disappeared from the settings (or at least the label).

Image

### Reproducible?

### Version

1.7.1b

macOS - aarch64

```shell

```

puneetdixit200:
## Summary

  - Bump the Zen keyboard shortcut schema version from 18 to 19.
  - Add a guarded migration that restores the `zen-toggle-pin-tab` shortcut when it is missing from an existing profile.
  - Reuse the existing `cmd_zenTogglePinTab` command and `zen-toggle-pin-tab-shortcut` label so the shortcut appears again
  in Settings > Keyboard Shortcuts.

  ## Root Cause

  The Toggle Pin Tab shortcut was added in an earlier migration, but profiles that reached the latest shortcut version while
  missing that entry had no later repair path. `fillDefaultIfNotPresent()` only backfills shortcuts from
  `zenGetDefaultShortcuts()`, and the pin-tab shortcut is migration-added rather than part of that default list.

  ## Testing

  - `node --check src/zen/kbs/ZenKeyboardShortcuts.mjs`
  - `git diff --check`
  - Static verification that the latest keyboard shortcut version is 19 and the migration is guarded against duplicate `zen-
  toggle-pin-tab` entries.

  `npm run lint -- src/zen/kbs/ZenKeyboardShortcuts.mjs` could not run in this shallow checkout because the generated/
  downloaded Firefox `engine` directory is not present.

  Fixes #4504

### DIFF
--- a/src/zen/kbs/ZenKeyboardShortcuts.mjs
+++ b/src/zen/kbs/ZenKeyboardShortcuts.mjs
@@ -832,7 +832,7 @@ class nsZenKeyboardShortcutsLoader {
 }
 
 class nsZenKeyboardShortcutsVersioner {
-  static LATEST_KBS_VERSION = 18;
+  static LATEST_KBS_VERSION = 19;
 
   constructor() {}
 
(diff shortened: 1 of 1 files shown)


sporocyst:
can't this shortcut automaticllly be added back by setting `zen.keyboard.shortcuts.version` to 0? Sorry if I said something ignorant.

PR #13768, opened 2026-05-20
gh-13767: fix special key shortcuts displaying incorrectly until page reload

### LINKED ISSUE #13767: special key shortcuts display incorrectly until page reload
### Preliminary Checks

### What happened?

While setting a shortcut with a special key like Arrow Keys, Escape etc. The page needs a reload to display them correctly.
Like the "ArrowLeft" key will be shown as "ArrowLeft" and after reload it will become "←".

### Expected behavior

The keys should be displayed correctly without reload.

### Actual behavior

The page needs a reload for it to display correctly.

### Steps to reproduce

1. Set a keyboard shortcut that has arrow keys like "Ctrl+ArrowLeft" or just the arrow key "ArrowLeft".
2. Reload the page.
3. Notice the keybind change.

### Screenshots and videos

https://github.com/user-attachments/assets/65ee664b-c0bc-436c-937c-d66f57b136c0

### Version

v1.19.13b

Linux (Tarball)

Keyboard Shortcuts

```shell

```

ashvwinn:
The problem was that `input.value` was getting the raw shortcut and not really converting it to the normalized displayable string.
There was no method to just get the `displayString` for a shortcut without creating a new one, so I put that logic into `gZenZenKeyboardShortcutsManager`. The static function in `KeyShortcut` class is just to reduce code duplication.

fixes: #13767


src/browser/components/preferences/zen-settings.js
```diff
@@ -1124,7 +1124,7 @@ var gZenCKSSettings = {
     this._hasSafed = false;
     input.classList.remove(`${ZEN_CKS_INPUT_FIELD_CLASS}-invalid`);
     input.classList.remove(`${ZEN_CKS_INPUT_FIELD_CLASS}-not-set`);
-    input.value = modifiers.toDisplayString() + shortcut;
+    input.value = modifiers.toDisplayString() + gZenKeyboardShortcutsManager.getKeyDisplay(shortcut);
```
mr-cheffy, line 1127: Why not just use the original shortcut `toDisplayString` without adding it to `modifiers.toDisplayString() +`?


```suggestion
    input.value = shortcut.toDisplayString();
```

mr-cheffy (changes requested):

src/browser/components/preferences/zen-settings.js
ashvwinn: The `shortcut` here is just a string not an object of `KeyShortcut`. And to use the `toDisplayString()` method, a new keybind has to be set which cannot be done here.

The previous implementation was this:
```js
input.value = modifiers.toDisplayString() + shortcut;
this._latestValidKey = shortcut;
```

src/zen/kbs/ZenKeyboardShortcuts.mjs
```diff
@@ -1541,4 +1548,22 @@ window.gZenKeyboardShortcutsManager = {
...
+    }
+
+    let key = shortcut;
+    let keycode = "";
+    for (let kc of Object.keys(KEYCODE_MAP)) {
+      if (kc == shortcut.toUpperCase()) {
+        keycode = KEYCODE_MAP[kc];
+        key = "";
+        break;
+      }
```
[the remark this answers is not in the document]
ashvwinn: It uses the same pattern as in `KeyShortcut.setNewBinding()`
```js
  setNewBinding(shortcut) {
    for (let keycode of Object.keys(KEYCODE_MAP)) {
      if (keycode == shortcut.toUpperCase()) {
        this.#keycode = KEYCODE_MAP[keycode];
        this.#key = "";
        return;
      }
    }

    this.#keycode = ""; // Clear the keycode
    this.#key = shortcut;
  }
```

Should I also change this? Or just keep it the same in both.

src/browser/components/preferences/zen-settings.js
mr-cheffy: Oh I see, thanks!

src/zen/kbs/ZenKeyboardShortcuts.mjs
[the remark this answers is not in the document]
mr-cheffy: Keep it, its fine.

mr-cheffy (approved):

PR #13777, opened 2026-05-20
gh-13776: fix shortcut conflict detection failing with special keys

### LINKED ISSUE #13776: shortcut conflict detection failing with special keys
### Preliminary Checks

### What happened?

Conflict detection fails when creating a shortcut with special keys like Arrow keys, enter, backspace, escape etc.

### Expected behavior

Conflict should be detected if there is one

### Actual behavior

Conflict is not detected even if there is one

### Steps to reproduce

1. Create a shortcut like Ctrl+Enter
2. Create another shortcut with the same keybind
3. Notice conflict is not detected

### Screenshots and videos

Image

### Version

v1.19.13b

Linux (Tarball)

Keyboard Shortcuts

```shell

```

ashvwinn:
`getKeyNameOrCode()` returns the keycode as defined in the `KEYCODE_MAP` which will never match to the `realShortcut`.
Doing a reverse lookup in `KEYCODE_MAP` fixes the issue.
I created a `REVERSE_KEYCODE_MAP` to avoid an extra loop when doing a reverse lookup.

fixes: #13776

### DIFF
--- a/src/zen/kbs/ZenKeyboardShortcuts.mjs
+++ b/src/zen/kbs/ZenKeyboardShortcuts.mjs
@@ -44,6 +44,10 @@ const KEYCODE_MAP = {
   SCROLL_LOCK: "VK_SCROLL",
 };
 
+const REVERSE_KEYCODE_MAP = Object.fromEntries(
+  Object.entries(KEYCODE_MAP).map(([k, v]) => [v, k])
+);
+
 const defaultKeyboardGroups = {
   windowAndTabManagement: [
     "zen-window-new-shortcut",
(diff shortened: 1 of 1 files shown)


src/zen/kbs/ZenKeyboardShortcuts.mjs
```diff
@@ -1507,9 +1511,11 @@ window.gZenKeyboardShortcutsManager = {
         continue;
       }
 
+      const keyNameOrCode = targetShortcut.getKeyNameOrCode();
+      const key = REVERSE_KEYCODE_MAP[keyNameOrCode] ?? keyNameOrCode;
```
mr-cheffy, line 1515: Why reverse it?

mr-cheffy: Never mind, I understand now.

PR #13779, opened 2026-05-20
feat: auto-close inactive tabs

lcflight:
# feat: auto-close inactive tabs

Addresses #2326.

Opt-in periodic sweep that closes tabs older than a configurable threshold. New settings group in Tab Management, default prefs, Fluent strings, per-window sweep module, and an undo toast.

> **Status:** built and exercised locally. 

## Files changed

- `prefs/zen/zen.yaml` — default prefs
- `src/zen/zen.globals.mjs` — register `gZenAutoCloseTabs`
- `src/zen/tabs/jar.inc.mn` — chrome:// mapping
- `src/zen/common/ZenPreloadedScripts.js` — preload at startup
- `src/zen/tabs/ZenAutoCloseTabs.mjs` — new sweep module
- `src/browser/components/preferences/zenTabsManagement.inc.xhtml` — new groupbox
- `src/browser/components/preferences/zen-settings.js` — pref bindings
- `locales/en-US/browser/browser/preferences/zen-preferences.ftl` — pane strings
- `locales/en-US/browser/browser/zen-workspaces.ftl` — toast string

## Prefs

| Pref | Type | Default | Surfaced in UI? |
| --- | --- | --- | --- |
| `zen.tabs.auto-close.enabled` | bool | `false` | ✓ checkbox |
| `zen.tabs.auto-close.threshold-value` | int | `7` | ✓ number input |
| `zen.tabs.auto-close.threshold-unit` | string | `"days"` | ✓ menulist (`hours` / `days`) |
| `zen.tabs.auto-close.skip-audible` | bool | `true` | ✓ checkbox |
| `zen.tabs.auto-close.check-interval-minutes` | int | `15` | about:config only |

## Settings UI

New `Auto-close Inactive Tabs` group in *Settings → Tab Management*, sibling to `zenWorkspacesGroup`. Threshold is `[number] [hours/days ▾]`:

screenshot_000

## Sweep behavior

Per-window singleton extending `nsZenDOMOperatedFeature`. `setTimeout` re-armed after each pass; flipping `enabled` off via the pref observer stops the loop cleanly.

The sweep walks `gZenWorkspaces.allStoredTabs` and judges each candidate against the eligibility filter:

- `tab.visible`, `!tab.pinned`, `!tab.selected`, `!tab.busy`
- `!tab.hasAttribute("zen-empty-tab")` and `!tab.hasAttribute("zen-glance-tab")`
- `!tab.soundPlaying` when `skip-audible` is set
- `tab.lastAccessed < cutoff`

The currently-selected tab is the only structural guarantee — workspaces can be drained to zero tabs by design (matches discussion #2326 feedback).

**Split-view tabs are judged as a unit.** Both halves close together when the most-recently-interacted half passes the threshold, and the whole group is skipped if any tab in it is protected (selected, audible, busy, etc.).

**Tabs with unsaved state are skipped, not prompted.** Each candidate is pre-checked via `linkedBrowser.asyncPermitUnload("dontUnload")`. Tabs that would prompt (beforeunload returns false) are dropped from the victim list. The actual `removeTabs` call uses `skipPermitUnload: true` since the check is already done. Discarded tabs are assumed safe (no JS running = no dirty state).

## Edge cases

- **Essentials** — `tab.pinned === true`, covered by the eligibility filter.
- **Last tab in a Space** — eligible. Workspaces can become empty; opening one shows the new-tab UI.
- **Unsaved form data** — skipped (no prompt).
- **Glance / split-view tabs** — glance is skipped outright; split-view is treated as a unit (see above).
- **Folder tabs** — eligible. Folders aren't pinned in Zen, just containers.
- **Multi-window** — sweep runs per window; each window has its own `gZenAutoCloseTabs` instance.
- **Clock change / sleep** — `tab.lastAccessed` and `Date.now()` move together, so a sleep doesn't artificially age anything.

## Open questions (also raised in #2326)

1. Default off — opt-in to match Zen's calm defaults. Easy to flip.
2. `check-interval-minutes` kept out of the pane (`about:config` only) to keep the UI tight. Push back if you'd rather surface it.

## Out of scope (this PR)

- Per-workspace override of the threshold
- Per-domain whitelist (TabWrangler-style)
- "Snooze" / archive instead of close
- Pre-close notification

## Test plan

- [x] `enabled=false` → no sweep timer scheduled
- [x] `enabled=true`, 1h threshold → tab older than 1h closes within `check-interval-minutes`
- [x] Pinned + Essential tabs never close
- [x] Audible tab skipped when `skip-audible=true`; closed when false
- [x] Active tab never closes
- [x] A workspace with one non-pinned tab CAN be emptied
- [x] Closed tab restorable via the restore-closed-tab shortcut
- [x] Tab with `beforeunload` (unsaved form) is **skipped**, no prompt
- [x] Split-view group: both halves close together when stale; if one half is audible/selected, neither closes
- [x] Multi-window: each window sweeps independently
- [x] Toggling prefs at runtime takes effect on the next scheduled sweep without restart


JPtheJP:
I agree on opt-in by default, as such it does not disrupt people who do not want it.
Agreed also with pre-close notification (particularly on the initial run, where a high number of tabs would go, even with a restore-closed-tab shortcut ).
Thank you for tackling this functionality.

lcflight:
> I agree on opt-in by default, as such it does not disrupt people who do not want it. Agreed also with pre-close notification (particularly on the initial run, where a high number of tabs would go, even with a restore-closed-tab shortcut ). Thank you for tackling this functionality.

So far i have been punting on the pre-close notification because it feels intrusive to the purpose of the auto-closing function. It may make since but I have not yet fully wrapped my head around what makes sense with that.
Maybe its a toggle in the settings for auto-closing tabs...  "Notify before closing". I just have a certain fear of it feeling spammy to the user.

mr-cheffy:
Sorry, we don't accept code done by claude

ccaum:
Why would you reject agent generated code? Code is code. If it's good code, merge it. If it's not, reject it. What's driving this policy?

willbonney:
> Why would you reject agent generated code? Code is code. If it's good code, merge it. If it's not, reject it. What's driving this policy?

Because then every layperson will hop on the vibecoding train and there won't be enough time to review all the vanity feature PRs with bad code.

mr-cheffy:
There's also a license issue with AI code, as its not technically yours

lcflight:
@mr-cheffy 

I'm not going to try to push back and say the code should be accepted. I think its a reasonable position to take for a project under MPL-2.0. I do wonder were the line is. Is there something I can do to make this PRs work (or similar work) acceptable? 

I'd want to honor the spirit of the policy in future contributions, but I also use AI as part of my dev workflow, and I'd rather be upfront about that than have it be ambiguous. If "no AI assistance at all" is the bar, I can respect that and work that way for Zen contributions. Zen has become a browser I really love and I'd love to find a way that I can still contribute.

mr-cheffy:
AI assist is fine, though letting an agent do all the stuff isn't. Regardless, this isn't a feature we want to have *yet* until we have library going on, so there's little confusion as to whats happening.

lcflight:
Understood. I appreciate it, and I will make sure to follow that guideline for any future PRs.

PR #13797, opened 2026-05-22
gh-13796: Tab order reversed when dragging multiple tabs

### LINKED ISSUE #13796: Tab order is reversed when dragging multiple tabs
### Preliminary Checks

### What happened?

I was moving multiple tabs from one folder to another folder and saw that the order was reversed. The order is reversed when dragging the tabs in a normal folder, a nested folder and from a folder to outside any folder (the normal tab space).

I am using compact mode but I tested it with compact mode disabled and the issue persists. I searched for similar bugs and found #12159 which could maybe be a duplicate but I am not sure. The reverse order bug seems to exist for every drag action.

### Expected behavior

The tabs should be moved to the folder in the same order they had outside that folder.

### Actual behavior

The tabs are moved to the folder in the reverse order from what they had outside that folder.

### Steps to reproduce

# Scenarios

## Outside to folder

1. Create a folder
2. Open multiple tabs
3. Select multiple tabs
4. Drag the selected tabs to the folder

## Folder to outside

1. Have multiple tabs inside a folder
2. Select multiple tabs inside the folder
3. Drag the selected tabs outside the folder

## Folder to folder

1. Create two folders
2. Place multiple tabs in one folder
2. Select multiple tabs inside the folder
3. Drag the selected tabs to the other folder

## Original order

Image

## Order after dragging to new folder

Image

### Version

1.19.13b

Windows - x64

Tab Folders

```shell

```

vanhci:
## Summary

When dragging multiple tabs from one folder to another (or outside a folder), the tabs were appearing in reversed order. 

## Root Cause

The bug was in `src/zen/tabs/ZenPinnedTabManager.mjs` in the `moveToAnotherTabContainerIfNecessary` function. The code was calling `.reverse()` on the `movingTabs` array:

```javascript
let ownedTabs = Array.from(movingTabs || draggedTab)
  .reverse()  // This was reversing the tab order
  .map(tab => {
```

This `.reverse()` was originally added to handle adopting tabs from a different window (to avoid multiple tab-switches), but it was also being applied to tabs being moved within the same window, causing the order reversal bug.

## Fix

Removed the unnecessary `.reverse()` call. The correct tab order is now preserved when dropping tabs.

**File changed:** `src/zen/tabs/ZenPinnedTabManager.mjs`

**Lines changed:** Line 670 - removed `.reverse()`

## Testing

The issue can be reproduced by:
1. Creating a folder
2. Opening multiple tabs
3. Selecting multiple tabs
4. Dragging the selected tabs to the folder
5. Observing that tabs appear in reversed order

After the fix, tabs maintain their original order when dropped.

Fixes: zen-browser/desktop#13796

### DIFF
--- a/src/zen/tabs/ZenPinnedTabManager.mjs
+++ b/src/zen/tabs/ZenPinnedTabManager.mjs
@@ -667,7 +667,6 @@ class nsZenPinnedTabManager extends nsZenDOMOperatedFeature {
     let newIndex = dropIndex;
     let fromDifferentWindow = false;
     let ownedTabs = Array.from(movingTabs || draggedTab)
-      .reverse()
       .map(tab => {
         if (!gBrowser.isTab(tab)) {
           return tab;


mr-cheffy (approved):

mr-cheffy:
What about the original issue though?

PR #13799, opened 2026-05-22
gh-13796: Fix tab order reversing on dnd & boosts window size / initial color state (gh-13799)

### LINKED ISSUE #13796: Tab order is reversed when dragging multiple tabs

FlorianButz:


src/zen/boosts/zen-boosts.css
```diff
@@ -11,6 +11,18 @@
   }
   appearance: none;
   border: none;
+
+  max-height: 582px;
+}
+
+#zenBoostWindow[mode="boost"] {
+  width: 192px;
```
mr-cheffy: Would be nice to have it as CSS variables, but that could be done in a followup

src/zen/boosts/ZenBoostsEditor.mjs
```diff
@@ -397,20 +400,22 @@ export class nsZenBoostEditor {
...
+        this.editorWindow.moveTo(
+          this.editorWindow.screenX + offset,
+          this.editorWindow.screenY
+        );
+      }
 
-    this.doc.getElementById("zen-boost-editor-root").style.display = "flex";
-    this.doc.getElementById("zen-boost-code-editor-root").style.display =
-      "none";
+      this.doc.getElementById("zen-boost-editor-root").style.display = "flex";
```
mr-cheffy, line 415: Shouldn't be needed now that we know what state the window should be in, but could be done in a follow-up

mr-cheffy (approved):

mr-cheffy (approved):

PR #13828, opened 2026-05-24
feat: Add gesture controls to PiP window

laurienicholas:
Adds two-finger trackpad pan to move the PiP window and pinch-to-zoom (touchscreen) / Ctrl+scroll (trackpad) to resize it, all clamped within screen bounds.

- Two-finger drag moves the window; capture-phase wheel listeners ensure events are received even when the cursor is over the OOP video frame
- Ctrl+scroll and two-finger pinch resize the window, zooming from the centre so the window grows/shrinks symmetrically
- A "minimize" button is added to the player controls that closes the PiP window (maps to the existing CloseButton reason)


mr-cheffy:
Sorry, we can't allow code generated by claude for legal reasons. However though, this seems like it can be patched into firefox, and it would end up in zen eventually!

laurienicholas:
> Sorry, we can't allow code generated by claude for legal reasons. However though, this seems like it can be patched into firefox, and it would end up in zen eventually!
> 
> https://firefox-source-docs.mozilla.org/contributing/contribution_quickref.html

Thanks Mr Cheffy.

It'd be great to have this added into Zen as it's the last little quality of life thing that I miss from Arc.

Now I've got a better feel for how to actually achieve this, would you open to a PR if I re-implement it without Claude?

Going through Mozilla seems like it might take a while!

PR #13903, opened 2026-05-28
Fix unbound shortcuts writing key="undefined" to the keyset

### LINKED ISSUE #13902: Unbound key_duplicateTab serializes as key="undefined", causing bare u to duplicate tab
## Summary

Unbound Firefox shortcut `key_duplicateTab` (`Browser:DuplicateTab`) is written to `zen-keyboard-shortcuts.json` without a `key` field. On reload, Zen applies it to the live `zenKeyset` as a XUL `<key>` with **`key="undefined"`** (literal string) and **no modifiers**. This appears to be the **only bare-key binding** in the browser, and pressing bare **`u`** triggers tab duplication.

There are two bugs here:

1. **Zen CKS** writes `key="undefined"` for unbound shortcuts (via `setAttribute("key", undefined)`).
2. **Gecko key matching** appears to treat `key="undefined"` with no modifiers as a bare **`u`** binding — almost certainly because **`u` is the first character** of the literal string `"undefined"`. This is incorrect behavior for an unbound/malformed key.

## Environment

- Zen Browser (Firefox 149-based), Linux
- Profile: `zen-keyboard-shortcuts.json` version 18
- Remote debugging used to inspect live DOM (devtools.debugger.remote-enabled)

## Steps to reproduce (profile)

1. Use a profile where `key_duplicateTab` has no `key` field in `zen-keyboard-shortcuts.json` (default after Zen CKS migration — field omitted because `undefined` is dropped by `JSON.stringify`).
2. Restart Zen.
3. Press bare **`u`**.

**Focus:** Page content focus is fine — browser-level `<key>` bindings are evaluated regardless of whether chrome or the page is focused, provided the page does not capture the key (e.g. an focused `<input>`, contenteditable, or a page/script keyboard handler). A normal page with no focused editable field reproduces the bug.

**Expected:** Nothing (shortcut is unbound by design; upstream Firefox also ships `key_duplicateTab` with no default key — see [Firefox bug 2031128](https://bugzilla.mozilla.org/show_bug.cgi?id=2031128)).

**Actual:** Current tab is duplicated via `Browser:DuplicateTab` (fresh load, session not preserved the same way as a normal duplicate).

## Minimal live-DOM repro (isolates Gecko matching)

This reproduces the trigger without needing a broken profile — useful to confirm the `key="undefined"` → bare `u` matching is a separate bug from the CKS serialization issue.

1. Open **Browser Toolbox** (chrome debugger).
2. In the **Browser Console** for `chrome://browser/content/browser.xhtml`, run:

```js
const el = document.getElementById("key_duplicateTab")
  || document.querySelector('#zenKeyset key[command="Browser:DuplicateTab"]');
el.setAttribute("key", "undefined");
el.setAttribute("modifiers", "");
el.setAttribute("command", "Browser:DuplicateTab");
el.outerHTML;
```

3. Press bare **`u`** (page content focus is fine if the page is not capturing the key — see above).

**Expected:** Nothing — `"undefined"` is not a valid single-key binding and this shortcut should be unbound.

**Actual:** Tab duplicates via `Browser:DuplicateTab`.

To verify the first-character theory, other leading characters in the `key` attribute value may also match (e.g. a key with `key="nonsense"` might fire on bare **`n`**). Worth testing.

To clean up after testing:

```js
el.removeAttribute("key");
// or restart Zen
```

## Live DOM evidence

After `_applyShortcuts()`, the element in `zenKeyset` is:

```xml
<key id="key_duplicateTab" key="undefined" group="other" modifiers=""
      command="Browser:DuplicateTab" zen-keybind="true"/>
```

Audit of all 140 `<key>` elements: **`key_duplicateTab` is the only binding with a key set and no modifiers.** Other `u` shortcuts correctly require modifiers (`accel+u` for View Source, `accel+alt+u` for split unsplit).

`CustomKeys.getDefaultKey("key_duplicateTab")` returns `null` — Firefox CustomKeys is not involved.

## Root cause (in `ZenKeyboardShortcuts.mjs`)

1. **`KeyShortcut.#parseFromJSON`**: missing `key` in JSON → `this.#key = undefined` (`key?.toLowerCase()`).
2. **`toJSONForm` / save**: `key: undefined` is omitted from JSON, so the field stays absent across restarts.
3. **`replaceWithChild`**: `key.setAttribute("key", this.#key)` with `undefined` coerces to the string **`"undefined"`** in XUL.
4. **`_applyShortcuts`**: does not skip `isEmpty()` shortcuts — unbound commands are still inserted into `zenKeyset`.

Zen creates the malformed element; Gecko then mis-handles it by matching bare **`u`**.

## Suggested fix

**Zen (CKS):**

1. Coerce missing/null keys to empty string on parse: `json.key ?? ""`.
2. In `replaceWithChild`, if key is empty, **`removeAttribute("key")`** instead of setting `"undefined"`.
3. In `_applyShortcuts`, **skip `key.isEmpty()`** shortcuts so unbound commands never become active `<key>` elements.
4. Optionally normalize on save so `"key": ""` is always present for unbound shortcuts.

**Gecko (may also warrant upstream report):**

5. Investigate why `<key key="undefined" modifiers="" command="...">` matches bare **`u`**. Multi-character `key` attribute values should not behave as bare single-key bindings based on the first character.

## Workaround

Manually add to `zen-keyboard-shortcuts.json` for `key_duplicateTab`:

```json
"key": ""
```

Then restart Zen. Confirmed working — bare `u` no longer duplicates after restart with `"key": ""`.

## Related

- Upstream XUL: `<key id="key_duplicateTab" command="Browser:DuplicateTab"/>` (no key attribute)
- Zen migration v17 adds separate `zen-duplicate-tab` → `cmd_zenDuplicateTab` (correctly unbound with `key: ""`)

margathon:
Fixes #13902

## Problem

When a shortcut has no `key` in `zen-keyboard-shortcuts.json`, Zen CKS leaves `#key` as `undefined`. On apply, `setAttribute("key", undefined)` stringifies to the literal **`key="undefined"`** in XUL. With no modifiers, Gecko mis-matches bare **`u`** (first character of `"undefined"`), firing commands like `Browser:DuplicateTab`.

## Fix

**Runtime (parse / save / apply):**
- Normalize missing/null `key` and `keycode` to `""` on parse and in the constructor
- Always serialize `"key": ""` in `toJSONForm()` (never omit the field)
- Skip `isEmpty()` shortcuts in `_applyShortcuts()` so unbound commands are not inserted into `zenKeyset`

**Migrations:**
- **v17** (retroactive for future 16→17 migrators): clear `key_duplicateTab` when `zen-duplicate-tab` is added, since Firefox's entry has no default binding
- **v19** (unchanged): disable `key_duplicateTab` in favour of `cmd_zenDuplicateTab`
- **v20** (new): run `repairMalformedUnboundKey()` on all shortcuts (clears literal `"undefined"` keys) and re-apply `key_duplicateTab` unbind/disable

`replaceWithChild()` on `dev` already avoids setting an empty `key` attribute; this completes the fix for the parse/save/apply/migrate path.

## Test plan

- [ ] Profile at CKS v18 with `key_duplicateTab` missing `key` field → upgrade runs v19+v20 → no `<key key="undefined">` in `zenKeyset`
- [ ] Profile at CKS v19 with corrupted bindings → v20 repair clears literal `"undefined"` keys
- [ ] Bare `u` on a normal page does not duplicate the tab
- [ ] Bound shortcuts (e.g. `accel+u` view source) still work
- [ ] Keyboard shortcut settings can still bind/unbind Duplicate Tab via `zen-duplicate-tab`

### DIFF
--- a/src/zen/kbs/ZenKeyboardShortcuts.mjs
+++ b/src/zen/kbs/ZenKeyboardShortcuts.mjs
@@ -316,6 +316,14 @@ class KeyShortcut {
   #reserved = false;
   #internal = false;
 
+  static #normalizeKey(key) {
+    return (key ?? "").toLowerCase();
+  }
+
+  static #normalizeKeycode(keycode) {
+    return keycode ?? "";
+  }
+
   constructor(
     id,
     key,
@@ -329,8 +337,8 @@ class KeyShortcut {
     internal = false
   ) {
     this.#id = id;
-    this.#key = key?.toLowerCase();
-    this.#keycode = keycode;
+    this.#key = KeyShortcut.#normalizeKey(key);
+    this.#keycode = KeyShortcut.#normalizeKeycode(keycode);
 
     if (!window.VALID_SHORTCUT_GROUPS.includes(group)) {
       throw new Error("Illegal group value: " + group);
@@ -373,8 +381,8 @@ class KeyShortcut {
   static #parseFromJSON(json) {
     return new KeyShortcut(
       json.id,
-      json.key,
-      json.keycode,
+      json.key ?? "",
+      json.keycode ?? "",
       json.group,
       nsKeyShortcutModifiers.parseFromJSON(json.modifiers),
       json.action,
@@ -547,8 +555,8 @@ class KeyShortcut {
   toJSONForm() {
     return {
       id: this.#id,
-      key: this.#key,
-      keycode: this.#keycode,
+      key: this.#key ?? "",
+      keycode: this.#keycode ?? "",
       group: this.#group,
       l10nId: this.#l10nId,
       modifiers: this.#modifiers.toJSONString(),
@@ -628,6 +636,16 @@ class KeyShortcut {
     this.#key = "";
     this.#keycode = "";
     this.#modifiers = new nsKeyShortcutModifiers(false, false, false, false);
+  }
+
+  /**
+   * Repair shortcuts whose unbound state was persisted without a "key" field and
+   * later applied as the literal string "undefined" in the keyset (#13902).
+   */
+  repairMalformedUnboundKey() {
+    if (this.#key === "undefined") {
+      this.clearKeybind();
+    }
   }
 
   setNewBinding(shortcut) {
@@ -848,7 +866,7 @@ class nsZenKeyboardShortcutsLoader {
 }
 
 class nsZenKeyboardShortcutsVersioner {
-  static LATEST_KBS_VERSION = 19;
+  static LATEST_KBS_VERSION = 20;
 
   constructor() {}
 
@@ -1226,6 +1244,14 @@ class nsZenKeyboardShortcutsVersioner {
           "zen-duplicate-tab-shortcut"
         )
       );
+      // Firefox's key_duplicateTab has no default binding; keep it unbound now
+      // that cmd_zenDuplicateTab is the Zen-managed duplicate-tab action.
+      for (let shortcut of data) {
+        if (shortcut.getID() == "key_duplicateTab") {
+          shortcut.shouldBeEmpty = true;
+          break;
+        }
+      }
     }
 
     if (version < 18) {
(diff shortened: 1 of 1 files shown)


margathon:
Closing — upstream 1.20.1b already addresses this via v19 and replaceWithChild. Profile workaround applied locally.

PR #13950, opened 2026-05-29
Adds library toolbar icon & adjustments to existing keyframes

JustAdumbPrsn:
Extends #12655 
1. Adds animated library toolbar icon.
2. Adjusted keyframes for media icon so the anim doesnt feel choppy.
3. Adjusted keyframes for arrow of downloads icon.
4. Adjusted keyframes for history to make the lid rotate more.


JustAdumbPrsn:
I might mess with the keyframe timings further

JustAdumbPrsn:
@mr-cheffy I dont think further adjustments are needed to the keyframes of existing icons

JustAdumbPrsn:
There is one issue with this new toolbar icon for library, The blur gets slightly clipped at the top, because of restricting viewbox which is only fixed if I make the icon smaller :((
And I am not sure how the animation handles when someone unhovers mid animation.

mr-cheffy (approved):

PR #13953, opened 2026-05-30
gh-13941: Fixed boost button viable when non-effective

### LINKED ISSUE #13941: Boost button appears on extension pages but does nothing
### Preliminary Checks

### What happened?

The Boost button is visible on extension pages (`moz-extension://...`) but clicking it does nothing.
I tested this with multiple extensions:

- uBlock Origin
- Stylus
- Feedbro

On each extension page, the Boost button appears in the toolbar/menu as if the page supports Boosts, but selecting it has no effect and no feedback is shown to the user.

Since extension pages are special privileged pages, I understand that Boosts may not be technically supported on them. However, if that is the case, the UI should not expose the Boost action on pages where it cannot function.

### Expected behavior

Boost should either work on supported moz-extension:// pages or be hidden/disabled on pages where it is not supported to avoid confusing users.

### Actual behavior

The Boost button is displayed on extension pages, but clicking it does nothing. No Boost is created, no error is shown, and there is no indication that the feature is unsupported on the current page.

### Steps to reproduce

1. Open Zen Browser.
2. Install an extension that has its own page (for example: Feedbro, Stylus, or uBlock Origin).
3. Open the extension page (`moz-extension://...` URL).
4. Attempt to create or use a Boost on that page.
5. Observe that the Boost UI appears, but clicking it has no effect.

### Screenshots and videos

https://github.com/user-attachments/assets/e20fb576-2577-4ff7-be72-763e6b0dfb8e

### Version

1.20.1b

Linux (Flatpak)

Other

```shell

```

sporocyst:
This PR attempts to adds an attribute `disabled="true"` to boost button when uri is not legit for boost.
The button would then become dimmed and not interactable to cursor based on existing CSS rules.

Fix #13941

### DIFF
--- a/src/zen/urlbar/ZenSiteDataPanel.sys.mjs
+++ b/src/zen/urlbar/ZenSiteDataPanel.sys.mjs
@@ -251,9 +251,6 @@ export class nsZenSiteDataPanel {
     if (!canBoostSite) {
       boostButton.removeAttribute("boosting");
       boostButton.setAttribute("disabled", "true");
-    }
-
-    if (!canBoostSite) {
       return;
     }
 
--- a/src/zen/urlbar/ZenSiteDataPanel.sys.mjs
+++ b/src/zen/urlbar/ZenSiteDataPanel.sys.mjs
@@ -250,11 +250,14 @@ export class nsZenSiteDataPanel {
     const boostButton = this.document.getElementById("zen-site-data-boost");
     if (!canBoostSite) {
       boostButton.removeAttribute("boosting");
+      boostButton.setAttribute("disabled", "true");
     }
 
     if (!canBoostSite) {
       return;
     }
+
+    boostButton.removeAttribute("disabled");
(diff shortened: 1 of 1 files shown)


src/zen/urlbar/ZenSiteDataPanel.sys.mjs
```diff
@@ -250,12 +250,15 @@ export class nsZenSiteDataPanel {
     const boostButton = this.document.getElementById("zen-site-data-boost");
     if (!canBoostSite) {
       boostButton.removeAttribute("boosting");
+      boostButton.setAttribute("disabled", "true");
     }
```
mr-cheffy: Now that I see it, can we early return here? Not sure why there's a second if statement

mr-cheffy (changes requested):

mr-cheffy (approved):

PR #13956, opened 2026-05-30
gh-13839: Fix icons when using custom fonts

### LINKED ISSUE #13839: Material Symbols are broken when using custom font through boosts
### Preliminary Checks

### What happened?

Material Symbols are broken when using custom font

With Custom Font

Image

Without Custom Font

Image

### Expected behavior

Symbols are properly rendered

### Actual behavior

Symbols/Icons are not rendered correctly

### Steps to reproduce

1. Go to https://fonts.google.com/icons
2. Open boosts and change your font to something else for example Arial

### Screenshots and videos

_No response_

### Version

1.20b

macOS - aarch64

Other

```shell

```

FlorianButz:


src/zen/boosts/ZenBoostStyles.sys.mjs
```diff
@@ -82,7 +82,7 @@ export class nsZenBoostStyles {
 
     if (fontCase != "" || fontFamily != "") {
       style += `/* Text Format */\n`;
-      style += `body * {\n`;
+      style += `body *:not(span):not(gf-load-icon-font):not(mat-icon):not(i) {\n`;
```
mr-cheffy: i'd remove "span"

mr-cheffy (changes requested):

src/zen/boosts/ZenBoostStyles.sys.mjs
FlorianButz: Google maps as well as other websites (I think) contain their icons in spans, removing it will break it the same way as on the Material Symbols site

mr-cheffy (approved):

src/zen/boosts/ZenBoostStyles.sys.mjs
```diff
@@ -82,7 +82,7 @@ export class nsZenBoostStyles {
 
     if (fontCase != "" || fontFamily != "") {
       style += `/* Text Format */\n`;
-      style += `body * {\n`;
+      style += `body *:not(.google-symbols, gf-load-icon-font, mat-icon, .google-material-icons) {\n`;
```
mr-cheffy, line 85: We should rather have some array.join as a follow up, if we wan't to keep adding to the exclusion list

mr-cheffy (approved):

PR #13975, opened 2026-06-01
gh-13941: Fixed boost button viable when non-effective, CSS part

### LINKED ISSUE #13941: Boost button appears on extension pages but does nothing

sporocyst:
In PR #13953, I thought there is a CSS rule for disabled buttons to prevent them from pressed.
Turns out, there is, but only applies on this area 
image

So in this PR, same rules are added to this part
image

### DIFF
--- a/src/zen/common/styles/zen-single-components.css
+++ b/src/zen/common/styles/zen-single-components.css
@@ -542,8 +542,7 @@
     margin: 0;
 
     &[disabled] {
-      opacity: 0.5;
-      pointer-events: none;
+      visibility: hidden;
     }
   }
 }
--- a/src/zen/common/styles/zen-single-components.css
+++ b/src/zen/common/styles/zen-single-components.css
@@ -540,6 +540,11 @@
 
   & toolbarbutton {
     margin: 0;
+
+    &[disabled] {
+      opacity: 0.5;
+      pointer-events: none;
+    }
   }
 }
 


src/zen/common/styles/zen-single-components.css
```diff
@@ -540,6 +540,11 @@
 
   & toolbarbutton {
     margin: 0;
+
+    &[disabled] {
```
mr-cheffy, line 544: What if we just not show it at all? `visibility: hidden;`

mr-cheffy (changes requested):

mr-cheffy (approved):

PR #13983, opened 2026-06-01
gh-13982: fix boosts size button not working for "100%" webpage scale

### LINKED ISSUE #13982: size override doesnt reset zoom to default after disabling it
### Preliminary Checks

### What happened?

When disabling the size override function in boosts, it goes from 90% to disable. But the zoom stays at 90% and doesnt change to the default page zoom until a reload.

### Expected behavior

size should be set to default after disabling

### Actual behavior

size doesn't change to default

### Steps to reproduce

1. Change size in boosts
2. Observe webpage size when going from 90% to disable/default

### Screenshots and videos

https://github.com/user-attachments/assets/3b141ee0-a9f8-4525-9754-569d0a847984

### Version

v1.21t

Linux (Tarball)

Other

```shell

```

ashvwinn:
fixes #13982


mr-cheffy:
This will always set the boost to 100% though, even if the value is the default one.

ashvwinn:
> This will always set the boost to 100% though, even if the value is the default one.

I dont understand what you mean by that.
What do you mean by "default" here? The default page size set by the user?

ashvwinn:
Couldn't find a good fix for this, but still added the "100%" size override value in this PR #14014 which "kind of" fixes the issue but not really.

PR #13598, opened 2026-05-07
Add space and container sync

kristijanribaric:
- Created sync engine for syncing workspaces and containers
- patched `ContextualIdentityService` to allow creating containers with specific IDs

PR #13984, opened 2026-06-01
gh-14470: Add tab and folder sync support

### LINKED ISSUE #14470: [META]: Cross-device sync

kristijanribaric:
Continuation of https://github.com/zen-browser/desktop/pull/13598

Added support for tabs, folders and tab splits sync via Firefox Sync


al-indigo:
@mr-cheffy I know my comments may be annoying but review it please? As far as I see @kristijanribaric requested a review half a month ago and I'm using now Arc and Zen-browser at the same time; that's really inconvenient

src/zen/sessionstore/ZenWindowSync.sys.mjs
```diff
@@ -1328,6 +1331,27 @@ class nsZenWindowSync {
...
+    if (item.isZenFolder && !item.isLiveFolder) {
+      lazy.ZenSyncStore.markFolderChanged(item.id);
+      return;
+    }
+
+    if (lazy.gSyncOnlyPinnedTabs && !item.pinned) {
+      return;
+    }
+
+    this.#maybeFlushTabState(item).finally(() => {
```
bdbch: Just a quick question: How are split tabs handled in Zen? Do we need to extra-guard here or add logic that handles split tab groups?

src/zen/sessionstore/ZenWindowSync.sys.mjs
```diff
@@ -1382,12 +1413,17 @@ class nsZenWindowSync {
...
     }
+    this.#notifySyncItemChanged(aEvent.target);
+
     return this.#delegateGenericSyncEvent(aEvent, SYNC_FLAG_LABEL);
   }
 
   on_TabHide(aEvent) {
     const tab = aEvent.target;
     const window = tab.ownerGlobal;
+    const id = aEvent.target.id;
```
bdbch: I think this is unused, right?

src/zen/spaces/ZenSpaceManager.mjs
```diff
@@ -204,6 +211,812 @@ class nsZenWorkspaces {
...
+    tab.id = syncId;
+    if (tab.id !== syncId) {
+      tab.setAttribute("id", syncId);
+    }
+
+    return tab.id === syncId;
+  }
+
+  #getSyncedTabState(tab) {
+    return JSON.parse(SessionStore.getTabState(tab));
```
bdbch: Do we need try/catch blocks here so we don't run into thrown errors? (Or where the helper is called)

src/zen/spaces/ZenSpaceManager.mjs
```diff
@@ -204,6 +211,812 @@ class nsZenWorkspaces {
...
+    }
+  }
+
+  #applyIncomingTabNavigation(tab, tabData) {
+    const incomingEntry = this.#getSyncedTabActiveEntry(tabData);
+    if (!incomingEntry?.url || tab.hasAttribute("zen-empty-tab")) {
+      return;
+    }
+
+    const currentState = this.#getSyncedTabState(tab);
```
bdbch: See above - this could throw when JSON.parse fails to parse content correctly

src/zen/sessionstore/ZenWindowSync.sys.mjs
```diff
@@ -1527,12 +1572,36 @@ class nsZenWindowSync {
     });
   }
 
+  /**
+   * Fired by tabbrowser for top-level location changes in any tab.
+   * We use this to mark the tab as changed so Firefox Sync can persist
+   * URL/history updates even when no tab label/icon event fires.
+   */
+  onLocationChange(aBrowser, aWebProgress, _aRequest, _aLocation, _aFlags) {
```
bdbch, line 1619: Should we debounce this / avoid firing this on every navigation change immediately? I'm not sure about Firefoxs sync server quotas but it would suck to run into server issues when too many sync requests are send.

src/zen/spaces/ZenSpaceManager.mjs
```diff
@@ -204,6 +211,812 @@ class nsZenWorkspaces {
...
+        !Array.isArray(splitData.tabs) ||
+        splitData.tabs.length < 2
+      ) {
+        continue;
+      }
+
+      const existingGroup = localSplitGroupsById.get(splitData.groupId);
+      if (
+        existingGroup &&
+        JSON.stringify(existingGroup) === JSON.stringify(splitData)
```
bdbch: I'm a bit worried here that while comparing both stringified JSON states keys won't match up which means this could return true all the time. Should we normalize this beforehand so we can be sure property keys don't cause problems later?

src/zen/split-view/ZenViewSplitter.mjs
```diff
@@ -1884,6 +1903,16 @@ class nsZenViewSplitter extends nsZenDOMOperatedFeature {
         document.removeEventListener("mousemove", dragFunc);
         window.setCursor("auto");
         this.tabBrowserPanel.removeAttribute("zen-split-resizing");
+        if (this.currentView >= 0) {
+          const group = this._data[this.currentView];
+          this.#dispatchItemEvent(
+            "ZenSplitViewGroupUpdated",
+            group.tabs[0].group,
```
bdbch, line 1925: I think we should add some guarding here in case `this._data[this.currentView]` returns null values.

`group`, `group.tabs`, `group.tabs[index]` and `group.tabs[index].group` could all potentially be nullish or undefined

src/zen/sessionstore/ZenWindowSync.sys.mjs
```diff
@@ -1664,8 +1741,15 @@ class nsZenWindowSync {
 
   on_ZenSplitViewTabsSplit(aEvent) {
     const tabGroup = aEvent.target;
+    if (!tabGroup?.id) {
```
bdbch, line 1789: Since we do an early return here, "runOnAllWindows" will not run anymore. Is this actually wanted behavior?

bdbch (changes requested):
I have some notes, questions or worries.

src/zen/sessionstore/ZenWindowSync.sys.mjs
```diff
@@ -256,6 +258,7 @@ class nsZenWindowSync {
     for (let eventName of EVENTS) {
       aWindow.addEventListener(eventName, this, true);
     }
+    aWindow.gBrowser?.addTabsProgressListener(this);
```
mr-cheffy, line 277: This needs to removed on unload, no?

mr-cheffy (changes requested):

src/zen/spaces/ZenSpaceManager.mjs
mr-cheffy, line 1: I feel like these should be moved to a separate file, to keep it clean

mr-cheffy (changes requested):
I think we should just be syncing pinned tabs? Without depending on `gSyncOnlyPinnedTabs`

arturict:
I built a safety-hardened integration on top of this PR for a real multi-device Windows/Linux setup: [arturict/zen-browser-desktop@`codex/cross-device-sidebar-sync`](https://github.com/arturict/zen-browser-desktop/tree/codex/cross-device-sidebar-sync).

I am not opening a duplicate tabs/folders PR. The reusable changes in that branch address the current review concerns and edge cases:

- sync only pinned tabs and Essentials, independent of the same-window preference
- serialize only the original pinned URL/title/icon, never live history, forms, scroll state, or ordinary tabs
- never navigate an existing local tab because of a remote update
- persist changed IDs so offline edits survive restart
- create a pre-apply recovery snapshot
- reconcile pre-existing Spaces/folders/pins/splits deterministically and tombstone obsolete aliases
- preserve local-only tabs when a remote folder or Space is deleted
- map containers through opaque IDs rather than numeric `userContextId`

The landed container-ID issue is split into focused draft PR #14551 / issue #14550. I am happy to extract any of the remaining hardening into smaller follow-ups after this PR is rebased.

mr-cheffy (changes requested):
I've pinged you on discord about a bug when creating new tabs, other than that, I think this is ready for merge?

src/zen/sessionstore/ZenWindowSync.sys.mjs
```diff
@@ -1351,6 +1374,29 @@ class nsZenWindowSync {
...
+    }
+
+    if (item.isZenFolder && !item.isLiveFolder) {
+      lazy.ZenSyncStore.markFolderChanged(item.id);
+      return;
+    }
+
+    if (lazy.gSyncOnlyPinnedTabs && !item.pinned) {
+      return;
+    }
```
bdbch: When `sync-only-pinned-tabs` is enabled (which is the default) this will cause no tombstone to be synced to other clients, but the tab is unpinned on this client.

This means other clients will keep the tab pinned.

src/zen/sync/ZenWorkspacesSync.sys.mjs
bdbch: Change suppression doesn't cover deferred events which can cause sync echo loops

`applyIncomingBatch` only sets `#ignoreChanges` / `tracker.ignoreAll` while the apply is awaited. But `ZenWindowSync.handleEvent` processes most events async through a promise chain. So events caused by the apply itself (`TabPinned`, `TabMove`, `TabGroupUpdate`, `TabClose` from `pinTab`, `addTabs`, `removeTab`, etc.) often run *after* the suppression flags are already reset.

Result: items we just downloaded get marked as changed again and re-uploaded on the next session save. And since `position` is the tab's global index in `gBrowser.tabs` (which includes local-only tabs), each device uploads slightly different data — devices can ping-pong updates back and forth instead of settling.

Suggestion: wait for the ZenWindowSync event queue to drain (e.g. expose `lastHandlerPromise`) before resetting `#ignoreChanges`, or tag items created during apply so their events are ignored.

src/zen/sync/ZenWorkspacesSync.sys.mjs
```diff
@@ -92,12 +101,31 @@ class ZenWorkspacesStore extends Store {
...
+      if (tab.zenSyncId && (!pinnedOnly || tab.pinned)) {
+        ids[createRecordId(RECORD_TYPES.TAB, tab.zenSyncId)] = true;
+      }
+    }
+
+    for (const folder of sidebar.folders || []) {
+      if (folder.id) {
+        ids[createRecordId(RECORD_TYPES.FOLDER, folder.id)] = true;
+      }
+    }
```
bdbch: Split views inside folders get synced twice, once as a folder, once as a split

`ZenFolders.storeDataForSessionStore()` puts `tab-group[split-view-group]` elements into `sidebar.folders` (marked with `splitViewGroup: true`). The loop here (and `createRecord` / `itemExists`) doesn't filter them out, so a split view inside a folder is uploaded as both a `folder` record and a `split` record.

On the receiving device, `#processIncomingFolder` finds no zen-folder with that id and calls `gZenFolders.createFolder([], {id})` — creating a phantom folder whose DOM id collides with the split group. The split-apply step later removes the phantom again (`removeTabGroup`), so it mostly self-heals, but you get folder create/remove churn on every sync, and the wrong `folder` record stays on the server forever.

Suggestion: skip entries with `splitViewGroup: true` when enumerating and creating folder records, and have the resolver ignore incoming folder data with that flag set.

src/zen/sync/ZenSpaceSyncResolver.sys.mjs
```diff
@@ -0,0 +1,985 @@
...
+
+    const incomingEntries = Array.isArray(tabData.entries)
+      ? tabData.entries.map(entry => ({ ...entry }))
+      : [{ ...incomingEntry }];
+    const incomingIndex = Math.min(
+      Math.max(typeof tabData.index === "number" ? tabData.index : 1, 1),
+      incomingEntries.length
+    );
+
+    const newState = {
```
bdbch: Incoming tab updates wipe the tab's back/forward history

Outgoing tab records only carry the current entry (`entries: [currentEntry], index: 1`). When that record is applied here, the tab's session state is replaced with that single entry whenever the URL differs — so navigating a synced tab on device A erases the entire back/forward history of that tab on device B.

This also hits at rollout: the engine version bump (2 → 3) forces a full re-download, so existing tabs — including ones the user is actively using — can lose their history the moment this ships.

Suggestion: append the incoming entry to the tab's existing history instead of replacing `entries`.

bdbch (commented):
Since I'm out of the development of this feature, I wanted to help with an automated review and since this is very critical I thought to send Fable onto this to do a review. It found **only 4 blockers**).

I attached it's comments on the right lines.

It also added some notes around code quality:

> - ZenSpaceSyncResolver.sys.mjs:568-573: tab.setUserContextId doesn't exist anywhere in the tree (containers are immutable per-browser), so container changes on existing tabs silently no-op behind the typeof guard. Either implement it (recreate the tab) or remove the dead path.
> - #removeSyncedItems calls removeGroup(groupIndex) without {suppressEvents: true}, while #applyIncomingSplitViewData passes it — inconsistent, and it feeds finding 2.
> - applyIncomingBatch silently drops incoming tabs/folders when no non-private window is open (firstSyncedWindow null) yet returns success, so the engine marks the records applied. Pre-existing pattern from #13598, but the blast radius is bigger now — worth a follow-up.
> - Nice touches: parseRecordId hardening (sep <= 0), [...group.tabs].reverse() fixing the in-place mutation in removeGroup, preserving pre-existing tab ids in on_TabOpen, and centralizing constants in ZenSyncConstants.sys.mjs.

Hopefully this can help with getting this over the finish line.

bdbch:
I'm looking into the few open points from my last review right now. I can create a PR pointing at this PR, right?

bdbch:
I opened a PR to the fork of @kristijanribaric. I think this is how this is supposed to work? :D Never had to contribute to an open PR of another person.

https://github.com/kristijanribaric/desktop/pull/1

This PR includes fixes for all open points I still saw unfixed from my last review.

mr-cheffy (approved):

Davis8483:
peak 🔥

bdbch:
Holy shit, it's happening.

JonathanXDR:
> Holy shit, it's happening.
the-office-panic

PR #13981, opened 2026-06-01
gh-14044: Implement Space Routing (gh-13981)

FlorianButz:
This PR aims to implement a version of Arc's Air Traffic Control feature natively into Zen.

> [!NOTE]
> This PR is based on #9042 
> Thanks for providing the necessary pointers in your PR.

---
### Structure

| Area | File | Purpose |
|---|---|---|
| Core | `src/zen/air-traffic/ZenAirTrafficManager.sys.mjs` | Stores an array of all routes and manages which URLs get routed where. |
| Browser Glue | `src/zen/air-traffic/ZenAirTrafficControlIntegration.mjs` | Hooks into `gBrowser.addTab`, asks the manager, then moves the tab to the correct Space. |
| Control Dialog | `src/zen/air-traffic/(ZenAirTrafficDialog.mjs,zen-air-traffic.(css,inc.xhtml))` | Provides the UI and talks to the manager. |

---
### Behavior

Users can
1. create routes of the type `Contains`, `Equal To` and `RegEx`
2. specify the route destination (Most recent Space, A specific Space, Lil Zen (once it is implemented))
3. specify the reference (the url, regex)

Once addTab fires internally the integration will call `ZenAirTrafficManager.routeUri()`, which returns a string that is either `most-recent-space`, `lil-zen` or the Space UUID.

If no rule matches a new tab will default to the current Space.
If a tab is opened from an external source (Discord, etc..), the "Default route for external links" route will be used.

Bildschirmfoto 2026-06-01 um 7 45 56 PM

---
### Missing Content
- ~~Localization is not fully complete.~~
- ~~The internal engine still uses the name "Air Traffic" while the new name is "Control Flow"~~

PR #13988, opened 2026-06-02
feat: add arrow-down-turn-right.svg

reizumii:
For #13981


mr-cheffy:
@FlorianButz could you please use this icon?

PR #14001, opened 2026-06-03
no-bug: bypass workspace container coercion for extension-opened tabs (gh-14001)

rishabshah0:
When a WebExtension (e.g. Grammarly) opens a tab via browser.tabs.create() without specifying a cookieStoreId, it expects the tab to open in the default (No Container) context. Zen's workspace container logic was overriding this, forcing the tab into the active workspace's container. This broke extension sign-in flows because the extension's background script (running in container 0) couldn't access cookies set during login in the workspace container.

This change detects when triggeringPrincipal is an addon principal (isAddonOrExpandedAddonPrincipal) and skips the workspace container override, preserving the extension's intended container context.

### DIFF
--- a/configs/windows/mozconfig
+++ b/configs/windows/mozconfig
@@ -9,6 +9,12 @@ if test "$ZEN_CROSS_COMPILING"; then
   export WINE="$(echo ~)/win-cross/wine/bin/wine"
   export WINEDEBUG=-all
 
+  # Force Wine to load the genuine Microsoft d3dcompiler_47.dll that ships next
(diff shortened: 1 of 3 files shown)


mr-cheffy (approved):

mr-cheffy (approved):

barakor:
This caused a regression in the use case of an extension like `Open external links in a container` which is used to open links in a specific container, `Switch to workspace where container is set as default when opening container tabs` now doesn't the same way. the new link is opened in the specified container (as per the extension's function) but in whatever workspace is currently active, not the one assigned to the container.

PR #14010, opened 2026-06-03
gh-13027: Fix window-sync jam that left synced tabs blank in new windows

### LINKED ISSUE #13027: Switching to a tab in a second Zen Window shows "New Tab" and doesn't load the tab (shows a blank window)
### Preliminary Checks

### What happened?

When I have two Zen windows open, and try to open a tab in a separate window, it sometimes renders as "New Tab" and doesn't load. This doesn't happen all the time - but seems to occur after using Zen for some time (e.g. 30 minutes). I have to close and re-open the Zen window to fix it.

### Expected behavior

Clicking on an already open tab in another Zen window should load the tab

### Actual behavior

Tab stays with the title "New Tab" and doesn't load.

Note: this happens on many different web pages - does not seem to be isolated to any individual site.

### Steps to reproduce

1. Open first Zen Window
2. Open various tabs
3. Open second Zen Window
4. <keep using Zen, as this issue seems to happen over time>
5. In the second Zen Window, select one of the tabs you had opened in the first Zen window
6. See that the tab doesn't load, and stays named as "New Tab"

I have to close and re-open the Zen window to fix it.

### Screenshots and videos

Image

### Version

1.19.5b

macOS - aarch64

Tabs

```shell

```

joegoldin:
## Summary

Fixes #13027. With synced tabs spanning two windows, closing one window, opening a new window, and selecting a synced tab left it stuck on a blank `about:blank` "New Tab" instead of loading the page it was actually showing. Restored 2-window sessions were fine; only a window-close-then-open jammed it.

## Root cause

In `ZenWindowSync.sys.mjs`, `on_SSWindowClosing` set `#docShellSwitchPromise` to serialize the close-time tab hand-off and resolved it, but never reset it to `null`. A resolved promise is still truthy, so after any window close every subsequent `on_TabSelect` / `on_focus` hit `if (this.#docShellSwitchPromise) return;` and bailed — the cross-window content swap (`#onTabSwitchOrWindowFocus`) never ran again, leaving synced tabs blank.

## Changes

**`ZenWindowSync.sys.mjs`**
- Clear `#docShellSwitchPromise` in a `finally` in `on_SSWindowClosing` (the actual bug), and wrap the swaps in `on_focus` / `on_TabSelect` in `try/finally` so a thrown swap can't re-jam the field.
- Add `#hasLiveContent(tab)` and decide the swap from each tab's live `currentURI` rather than the stale `_zenContentsVisible` flag: only pull contents in when the selected tab is blank **and** another window genuinely holds the live page; never swap a page out of a tab that already shows it; clear a stale owner flag pointing at a blank tab.

**`ZenSessionManager.sys.mjs`**
- Add `#stripBlankSwapArtifact(tabData)` to drop a trailing `about:blank` swap artifact (clamping `index`), applied when collecting tabs for save/new-window and when restoring window data, so saved or cloned session state heals onto the real page (also covers sessions saved before this fix).

## Test plan

- [x] Verified against the #13027 repro on a patched 1.20.1b release build (omni.ja injection): close one window, open a new window, select a synced tab — it now loads the real page instead of a blank "New Tab".
- [x] `./mach lint zen` passes — **0 errors**; eslint + `@zen-browser/prettier` report **0 problems / 0 fixed** on both changed files. (The only `mach lint zen` warning is a pre-existing `test-manifest-toml` note in `zen/tests/mochitests/sandbox/browser.toml`, untouched by this change; the skipped `cargo-audit` / `clang-format` linters are C++/Rust-only and need a full build toolchain.)

### DIFF
--- a/src/zen/sessionstore/ZenSessionManager.sys.mjs
+++ b/src/zen/sessionstore/ZenSessionManager.sys.mjs
@@ -737,6 +737,54 @@ export class nsZenSessionManager {
     return tabData && !(tabData.zenIsEmpty && !tabData.groupId);
   }
 
+  /**
+   * Removes trailing "about:blank" entries from a tab's saved session history.
+   *
+   * When a tab's contents are swapped to another window, ZenWindowSync loads
+   * "about:blank" into the now-empty browser so Firefox isn't left with an
+   * unloaded-but-selected tab. That "about:blank" can get persisted as the
+   * tab's active session entry; when such state is later restored (most visibly
+   * into a newly opened window, which is cloned from saved state), the tab would
+   * open on the blank artifact and show an empty "New Tab" instead of the page
+   * it was actually displaying. This strips the trailing artifact so the tab
+   * restores onto its real last entry, as long as a real entry exists to fall
+   * back to. See gh-13027.
+   *
+   * @param {object} aTabData - The tab state object to sanitize.
+   * @returns {object} The original object when there is nothing to strip, or a
+   *   shallow copy with the trailing "about:blank" entries removed and the
+   *   index clamped onto a real entry.
+   */
+  #stripBlankSwapArtifact(aTabData) {
+    const entries = aTabData?.entries;
+    if (!Array.isArray(entries) || entries.length < 2) {
+      return aTabData;
+    }
+    const isBlank = entry => !entry?.url || entry.url === "about:blank";
+    // Leave genuinely blank tabs alone: there's no real entry to fall back to.
+    if (entries.every(isBlank)) {
+      return aTabData;
+    }
+    let lastReal = entries.length - 1;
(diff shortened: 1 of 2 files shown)


mr-cheffy:
This looks heavilly ai generated, sorry. Seems to be a workaround rather than fixing the actual issue

joegoldin:
> This looks heavilly ai generated, sorry.

No worries.  I couldn't find a policy in the repo on AI generated submissions -- do you have one handy? To be clear, I've thoroughly tested this, and it is not completely AI generated. 

> Seems to be a workaround rather than fixing the actual issue

Did some more tinkering locally and got it slimmed down to these few changes (verified this is the minimal working set of changes on my machine): https://github.com/zen-browser/desktop/pull/14012

PR #14012, opened 2026-06-04
gh-13027: fix window sync lockup that left synced tabs blank in new windows (gh-14012)

### LINKED ISSUE #13027: Switching to a tab in a second Zen Window shows "New Tab" and doesn't load the tab (shows a blank window)

### LINKED ISSUE #14022: Multi-window tab synchronization breaks after window destruction or moving shared tabs between Workspaces
### Preliminary Checks

### What happened?

Several operations can break tab synchronization between browser windows.

The primary issue is window desynchronization. After desynchronization occurs, tabs stop behaving as shared tabs and instead behave as separate instances. Newly opened tabs only work in the window where they were created, while the same tabs in other windows may appear as empty "New Tab" pages or remain locked as if they were still active in another window.

Additionally, some operations involving Workspaces may cause different consistency issues. For example, moving a shared tab or folder between Workspaces can corrupt workspace state, break tab navigation, or leave windows showing conflicting information.

In all synchronization-related cases, the issue persists until the entire browser is restarted.

### Expected behavior

Tabs shared between multiple browser windows should remain synchronized at all times.

### Actual behavior

Certain actions (closing private windows, closing synchronized windows, or moving shared items between Workspaces) corrupt multi-window tab synchronization, resulting in desynchronized tabs, inconsistent Workspace state, and in some cases broken or inaccessible tabs.

### Case 1: Closing a Private Window

1. Open two normal Zen windows.
2. Verify that tab synchronization works correctly.
3. Open a Private Window (Ctrl+Shift+P).
4. Close the Private Window.

### Actual Behavior

The two normal windows become desynchronized.

Symptoms include:

* Newly created tabs only work in the window where they were opened.
* The same tabs may appear as empty "New Tab" pages in the second window.
* Some tabs appear locked as if another window still owns them.
* Switching focus between windows does not resolve the issue.
* Closing the second window does not resolve the issue.
* Only a full browser restart restores synchronization.

---

### Case 2: Closing a Synchronized Window

1. Open two normal Zen windows.
2. Verify that tab synchronization works correctly.
3. Close one of the windows.

### Actual Behavior

Same result as Case 1.

The remaining window enters a desynchronized state and reopening additional windows does not restore synchronization.

Only a full browser restart restores normal behavior.

---

### Case 3: Moving a Shared Tab to Another Workspace

1. Open two normal Zen windows.
2. Open a tab that is visible in both windows.
3. Move that tab to another Workspace.

### Actual Behavior

The same desynchronization described in Case 1 occurs.

Additionally:

* Tab navigation may stop working in one of the windows.
* Tabs may become inaccessible.
* After restarting the browser, some tabs may disappear.
* Some tabs may be converted into "New Tab" entries.

---

### Case 4: Moving a Folder to Another Workspace

1. Open two normal Zen windows.
2. Create a folder containing tabs.
3. Move the folder to another Workspace.

### Actual Behavior

Window consistency becomes corrupted.

For example:

* The first window shows the folder in the original Workspace.
* The second window shows the folder in the new Workspace.

After restarting the browser:

1. The folder appears in the Workspace where it was moved.
2. Clicking tabs inside the folder may redirect to the original Workspace.

Deleting the folder restores normal behavior.

### Screenshots and videos

_No response_

### Version

1.20.2b

Windows - x64

Tabs

```shell

```

### LINKED ISSUE #14080: [Window Sync] Opening synced tab in new window creates seperate page instance
### Preliminary Checks

### What happened?

I'm on Linux and have window sync enabled. When I have a few tabs open, and open a second window, the tabs stay synced across both windows, but opening the tab freshly loads the webpage again, instead of providing access to the same tab. The effect of opening a synced tab in a new window is pretty much the same as opening an unloaded tab. 

You can see the in the images

### Expected behavior

Window Sync should cause actual tab instances to be shared across all windows

### Actual behavior

The tabs are synced, but opening a tab freshly opens a new instance of the page

### Steps to reproduce

- On Linux, start zen
- Set "Sync only pinned tabs in workspace" to false
-  Open a few tabs, and visit a few pages
- Open another instance of zen.
- Try opening a synced tab - It will load a freshly installed version
- Try closing a synced tab - It will close across both windows, despite having different instances of the page.

### Screenshots and videos

Image
Obfuscated some of the tabs for privacy reasons. Notice the same "synced" tab having seperate page states. 

### Version

1.20.2b (Firefox 151.0.3) (64-bit)

Linux (AppImage), Linux (Tarball)

Tabs

```shell

```

joegoldin:
Fixes #13027

`on_SSWindowClosing` never reset `#docShellSwitchPromise` to `null`, so after a window close every `on_TabSelect`/`on_focus` saw the resolved-but-truthy promise and skipped the window swap.

Now it clears on window close and the swap only happens when the new tab is blank and another window holds the live page.

### DIFF
--- a/src/zen/sessionstore/ZenWindowSync.sys.mjs
+++ b/src/zen/sessionstore/ZenWindowSync.sys.mjs
@@ -200,19 +200,6 @@ class nsZenWindowSync {
   }
 
   /**
-   * Whether a tab currently displays live page content rather than the
-   * "about:blank" placeholder loaded into a browser whose contents were
-   * swapped out to another window.
-   *
-   * @param {MozTabbrowserTab} aTab - The tab to check.
-   * @returns {boolean} True unless the tab is blank or has no current URI.
-   */
-  #hasLiveContent(aTab) {
-    const spec = aTab?.linkedBrowser?.currentURI?.specIgnoringRef;
-    return !!(spec && spec !== "about:blank");
-  }
-
-  /**
    * Called when a browser window is about to be shown.
    * Adds event listeners for the specified events.
    *
@@ -534,10 +521,13 @@ class nsZenWindowSync {
     if (flags & SYNC_FLAG_ICON) {
       aTargetItem.zenStaticIcon = aOriginalItem.zenStaticIcon;
       if (gBrowser.isTab(aOriginalItem)) {
-        gBrowser.setIcon(
-          aTargetItem,
-          aOriginalItem.getAttribute("image") || gBrowser.getIcon(aOriginalItem)
-        );
+        try {
+          gBrowser.setIcon(
+            aTargetItem,
+            aOriginalItem.getAttribute("image") ||
+              gBrowser.getIcon(aOriginalItem)
+          );
+        } catch {}
       } else if (aOriginalItem.isZenFolder) {
         // Icons are a zen-only feature for tab groups.
         gZenFolders.setFolderUserIcon(aTargetItem, aOriginalItem.iconURL);
@@ -1167,16 +1157,12 @@ class nsZenWindowSync {
         aWindow,
         selectedTab.id
       );
-      const selHasContent = this.#hasLiveContent(selectedTab);
-      const otherHasContent = this.#hasLiveContent(otherSelectedTab);
       selectedTab._zenContentsVisible = true;
       if (otherSelectedTab) {
         delete otherSelectedTab._zenContentsVisible;
-        if (!selHasContent && otherHasContent) {
-          promises.push(
-            this.#swapBrowserDocShellsAsync(selectedTab, otherSelectedTab)
-          );
-        }
+        promises.push(
+          this.#swapBrowserDocShellsAsync(selectedTab, otherSelectedTab)
+        );
       }
     }
     await Promise.all(promises);
--- a/src/zen/sessionstore/ZenWindowSync.sys.mjs
+++ b/src/zen/sessionstore/ZenWindowSync.sys.mjs
@@ -200,6 +200,19 @@ class nsZenWindowSync {
   }
 
   /**
+   * Whether a tab currently displays live page content rather than the
+   * "about:blank" placeholder loaded into a browser whose contents were
+   * swapped out to another window.
+   *
+   * @param {MozTabbrowserTab} aTab - The tab to check.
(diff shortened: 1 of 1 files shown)


src/zen/sessionstore/ZenWindowSync.sys.mjs
```diff
@@ -1154,12 +1167,16 @@ class nsZenWindowSync {
...
       );
+      const selHasContent = this.#hasLiveContent(selectedTab);
+      const otherHasContent = this.#hasLiveContent(otherSelectedTab);
       selectedTab._zenContentsVisible = true;
       if (otherSelectedTab) {
         delete otherSelectedTab._zenContentsVisible;
-        promises.push(
-          this.#swapBrowserDocShellsAsync(selectedTab, otherSelectedTab)
-        );
+        if (!selHasContent && otherHasContent) {
```
mr-cheffy: This is not needed, we shouldn't care about the current url, imo the fix at `docShellSwitchPromise` should be enough

mr-cheffy (changes requested):

mr-cheffy (approved):

joegoldin:
Thanks!

mr-cheffy:
@joegoldin does this fix https://github.com/zen-browser/desktop/issues/14022 ?

PR #14014, opened 2026-06-04
no-bug: add a "100%" size override value in boosts

ashvwinn:
Adds a new "100%" size override value in boosts. I think a lot of people has their default zoom set to something else other than "100%", so they might need it sometimes.

This is how the colos look:

Light mode
image

Dark mode
image

### DIFF
--- a/src/zen/boosts/ZenBoostsEditor.mjs
+++ b/src/zen/boosts/ZenBoostsEditor.mjs
@@ -639,7 +639,7 @@ ${cssSelector} {
    * Handles the size toggle button press, cycling through size override options
    */
   onBoostSizePressed() {
-    if (this.currentBoostData.sizeOverride == 1) {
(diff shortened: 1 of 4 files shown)


ashvwinn:
Should I add the contrast change in this PR? Its a small change so idk

mr-cheffy:
Not too sure about this tbh, this would just add more complication, looks more like a workaround rather than fixing the actual issue. That, without saying that 100% is the default zoom anyways

ashvwinn:
I know it's not a fix which is why I closed the last PR and didn't link the issue here either, but I still think adding a 100% size option is good.
I myself use 110% as the default size so if I want to set the size as 100% with boosts, I just can't without the option.

mr-cheffy:
I dont know this seems like a workaround and honestly I dont even know what the actual solution could be. Sorry

PR #14023, opened 2026-06-04
no-bug: Add script to map Firefox version to Zen release tag

iskunk:
Zen's versioning scheme, being independent of Firefox's, makes answering the question "I have Firefox version X, which Zen tag Y do I need to build against it?" a bit tricky to answer. (The reverse mapping is easier: switch to tag Y, find the Firefox version X in `surfer.json`.)

I'd like to contribute this script, which makes use of some Git-fu to get the answer. E.g.
```
$ ./scripts/ff-version-to-zen-tag.sh 145.0
1.17.7b
```
If there are multiple tags that correspond to the specified Firefox version, then this will return the newest one.

### DIFF
--- a/scripts/ff-version-to-zen-tag.sh
+++ b/scripts/ff-version-to-zen-tag.sh
@@ -1,0 +1,62 @@
+#!/usr/bin/env bash
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+#
+# Find the most recent Zen release tag that targets a specified
+# Firefox version. (The tag must be present in the local Git repo
+# in order for this script to find it.)
+#
+# Example shell-script usage:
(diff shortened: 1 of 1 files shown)


mr-cheffy:
I don't think we need this script, it just bloats up the repo imo, you should always just use the latest version.

iskunk:
If _you_ needed this script, you'd have written it yourself :-]

I'm setting up a CI process to build Zen against a packaged version of Firefox (specifically Debian's). The package comes out a bit after Mozilla releases the new version. In the meanwhile, Mozilla could put out a newer version... and/or Zen could push a new tag, for either version. I can't just use the latest Debian package and the HEAD of your repo and call it a day; I need to match up the two, checking the mapping between the two versions, and this solves the problem when the Firefox version is the input to the function.

I can use this script in my own repo to address my need, but the reason I made this PR is because other integrators may similarly find the mapping to be a problem. Being able to get tags for older Firefox versions is a nice bonus, but it's not the real reason for this.

PR #14036, opened 2026-06-05
no-bug: Fix GitHub live folder crash when a PR row has no number

robbiebusinessacc:
### What
  The legacy HTML pull-request parser in `GithubLiveFolder.sys.mjs` called
  `.match(/#[0-9]+/).shift()` with no null check. When a pull-request row's
  "opened-by" text has no `#<number>` token, `match()` returns `null` and
  `.shift()` throws a `TypeError`. That error escapes `fetchItems` and is
  caught at its top-level handler, so the **entire** GitHub live folder
  renders `zen-live-folder-failed-fetch` instead of just skipping the row.

  ### Fix
  Guard the match exactly the way the sibling `parseIssues` already does
  (`numberMatch?.[0] ?? ""`), so a row with no number degrades to an empty
  suffix instead of taking down the whole folder.

  ### Tests
  Adds two cases to `browser_github_live_folder.js` covering the legacy
  HTML pull-request path (previously only the issues path was tested): one
  row with a number, and one without (the crash case), asserting the folder
  still returns items.

### DIFF
--- a/src/zen/live-folders/providers/GithubLiveFolder.sys.mjs
+++ b/src/zen/live-folders/providers/GithubLiveFolder.sys.mjs
@@ -167,9 +167,9 @@ export class nsGithubLiveFolderProvider extends nsZenLiveFolderProvider {
           activeRepos.add(repo);
         }
 
-        const idMatch = authors[i].parentElement.textContent
-          .match(/#[0-9]+/)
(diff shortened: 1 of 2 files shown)


mr-cheffy:
Why would a PR have no number?

PR #14039, opened 2026-06-05
Add experimental scroll tab switcher

pascalwiemers:
## Summary

Draft/demo PR for fast tab switching by scrolling.

This implementation was AI-generated/AI-assisted. The goal is to demonstrate the UX and start a maintainer/design conversation, not to claim this is production-ready or already shaped correctly for Zen internals.

This adds an experimental native browser-chrome tab switcher behind `zen.tabs.scroll-switcher.enabled`:

- `Alt+X` / `Option+X` opens a centered tab list
- scroll wheel or trackpad scroll live-previews tabs
- `Alt+X` / `Option+X` again or `Enter` commits
- `Escape` cancels and restores the original tab

The intent is to show the UX in the native layer. Extension prototypes could not get the same feel because protected pages, live preview, and full-window scroll capture all need browser chrome access.

## Notes

- Default pref is `false`.
- No search UI in this version.
- Current-window visible tabs only.
- This is intentionally a draft; I expect this may need to move or be reshaped to match Zen architecture/design patterns.

## Testing

- `node --check src/zen/tab-switcher/ZenScrollTabSwitcher.mjs`
- `git diff --check`

Manual test target:

- Set `zen.tabs.scroll-switcher.enabled` to `true`
- Restart/open Zen
- Press `Alt+X` / `Option+X`
- Scroll to preview tabs
- Commit with `Alt+X` / `Option+X` or `Enter`
- Cancel with `Escape`

PR #14060, opened 2026-06-07
Fast tab switching by scrolling

pascalwiemers:
## Summary

Sorry for the earlier rough PR attempt. I cleaned this up and changed the approach to lean on the native tabbox switching behavior instead of trying to recreate tab cycling manually.

This implementation is AI-generated/AI-assisted and intended as a draft UX prototype, not a final architecture claim.

Firefox/Zen already supports scroll-based tab switching when the pointer is over the tab list. This PR builds on that native behavior by adding a small keyboard-first Zen command mode around it: press `Alt+X` / `Option+X`, scroll anywhere in the window to live-preview tabs, then press `Alt+X` / `Option+X` or `Enter` to commit. `Escape` cancels and restores the original tab.

The reason I think this fits Zen is that it keeps tab browsing out of the way. You do not have to move your pointer to the tab list, open a heavy switcher, or break flow. It feels closer to a temporary command mode: enter it, skim tabs by scrolling, land where you want, and keep going.

## Behavior

- Disabled by default behind `zen.tabs.scroll-switcher.enabled`
- Uses `gBrowser.tabContainer.advanceSelectedTab(direction, true)` for tab switching
- Captures wheel/trackpad scrolling only while the switcher is open
- Supports scroll wheel and trackpad scrolling
- `Alt+X` / `Option+X` opens and commits
- `Enter` commits
- `Escape` restores the original tab
- No search UI in this draft

## Testing

- `node --check src/zen/tab-switcher/ZenScrollTabSwitcher.mjs`
- `git diff --check`

Manual testing:

- Enabled `zen.tabs.scroll-switcher.enabled`
- Opened the switcher with `Alt+X` / `Option+X`
- Scrolled through tabs with live preview
- Committed with `Alt+X` / `Option+X` and `Enter`
- Cancelled with `Escape`
- Verified normal scrolling resumes after closing

### DIFF
--- a/prefs/zen/zen.yaml
+++ b/prefs/zen/zen.yaml
@@ -52,3 +52,6 @@
 
 - name: zen.tabs.ctrl-tab.ignore-pending-tabs
   value: false
+
+- name: zen.tabs.scroll-switcher.enabled
+  value: false
--- a/src/browser/base/content/zen-assets.jar.inc.mn
+++ b/src/browser/base/content/zen-assets.jar.inc.mn
@@ -9,6 +9,7 @@
 #include ../../../zen/mods/jar.inc.mn
 #include ../../../zen/spaces/jar.inc.mn
 #include ../../../zen/tabs/jar.inc.mn
+#include ../../../zen/tab-switcher/jar.inc.mn
 #include ../../../zen/kbs/jar.inc.mn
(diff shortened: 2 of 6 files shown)


mr-cheffy:
duplicate PR https://github.com/zen-browser/desktop/pull/14039#issuecomment-4637884847

PR #14077, opened 2026-06-09
gh-13747: Fixed space icons overflow being inaccessible

### LINKED ISSUE #13747: Space icons overflow inaccessible unless sidebar is made wider than necessary
### Preliminary Checks

### What happened?

I use different spaces to separate tabs specific to different clients and usually have between 4-6 spaces but currently have 10 (heavy workload but not a lot of tabs)

This is the makeup of my spaces with the sidebar expanded to accommodate. Note, the first space (left) is called "Home" and the second space is called "BCF", which the next image references. 
Image

When the sidebar is made smaller, the icons change to dots (except for active workspace) and then the spaces area becomes scrollable horizontally. This is shown below however the leftmost workspace that I can access by hovering and scrolling horizontally to the left is not my leftmost space. In this case, I can only access the leftmost and rightmost spaces by keeping the sidebar extra wide, which I can't justify on even a 15 inch monitor. 

Image

Can these icons wrap instead of shrinking and then becoming an inaccessible scrollable area? 

While we're at it, can the user define the size of the workspace icons? I'd love to be able to make them larger like my essentials tabs, but I can write up a separate feature request. Thanks a ton

### Expected behavior

I should be able to see all of my spaces displayed on the sidebar unless the sidebar is hidden entirely. 

### Actual behavior

Sorry, wrote it out in first field. 

### Steps to reproduce

Create multiple workspaces and shrink the sidebar to a point where they become compressed. You may not be able to access all of them. 

### Screenshots and videos

Included in first field

### Version

1.19.12b

macOS - aarch64

Workspaces

```shell

```

PitterG4:
Fixes #13747 

## The root cause
It was the `justify-content: center`. 
I mean, It worked fine, but whenever the icons overflowed the flexbox was pushing the leftmost items into a negative position. Since the `scrollLeft` bottoms out at 0, those icons get trapped, you simply couldn't reach any icon with a position < 0.

## The solution
I only switched to `justify-content: safe center`, since it maintains everything centralized and uses the flex-start when there's negative positions.

## Other things done
The `safe center` worked, but the scrolling still felt a bit off. The manual offset (`this.scrollLeft = target.offsetLeft - 10;`) was creating an overly smooth walking to the right, making horizontal scrolling to the left feel very slow and rigid. So I did some additional tweaks:
- Replaced the manual offset calculation by scrollIntoView method;
- Added a 20px scroll-margin so the next icon is revealed smoothly;
- Added an exception for the Shift key in case anyone navigates the Spaces with Shift + Scroll.

## Tests
- Tested standard mouse hover and switching;
- Tested with an XP-Pen graphic tablet (it worked and it felt really good!).

overflow_test

### DIFF
--- a/src/zen/spaces/zen-workspaces.css
+++ b/src/zen/spaces/zen-workspaces.css
@@ -41,6 +41,7 @@
     fill-opacity: 0.6;
     -moz-context-properties: fill-opacity, fill;
     fill: currentColor;
+    scroll-margin: 0 20px;
 
     & .zen-workspace-icon {
       pointer-events: none;
@@ -105,7 +106,7 @@
 
   &[icons-overflow] {
     gap: 0 !important;
-    justify-content: center;
+    justify-content: safe center;
 
     & toolbarbutton {
       margin: 0;
--- a/src/zen/spaces/zen-workspaces.css
+++ b/src/zen/spaces/zen-workspaces.css
@@ -20,7 +20,6 @@
     display: none !important;
   }
 
-  border-radius: var(--button-border-radius) !important;
   background: transparent;
   appearance: unset !important;
   height: fit-content;
--- a/src/zen/spaces/ZenSpaceIcons.mjs
+++ b/src/zen/spaces/ZenSpaceIcons.mjs
@@ -15,12 +15,12 @@ class nsZenWorkspaceIcons extends MozXULElement {
 
     this.initDragAndDrop();
     this.addEventListener("mouseover", e => {
-      if (this.isReorderMode) {
+      if (e.shiftKey || this.isReorderMode) {
         return;
       }
       const target = e.target.closest("toolbarbutton[zen-workspace-id]");
       if (target) {
-        this.scrollLeft = target.offsetLeft - 10;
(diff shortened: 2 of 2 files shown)


src/zen/spaces/zen-workspaces.css
```diff
@@ -41,6 +41,7 @@
     fill-opacity: 0.6;
     -moz-context-properties: fill-opacity, fill;
     fill: currentColor;
+    scroll-margin: 0 20px;
```
mr-cheffy, line 43: What does this change exactly?

mr-cheffy (approved):

src/zen/spaces/zen-workspaces.css
PitterG4: Thanks for merging! =^)

This add a 20 pixel margin on both sides of each icon. It's not a physical margin that's interact visually, it just forces this extra space to fit on the screen!

So in this case, when your mouse is above an icon that one of the sides it's less than 20 pixels, the effect is like the icon is being pushed. If you leave this as zero you won't be able to iterate the whole list of them anymore as there's nothing pushing the next icon to be revealed.

PR #14101, opened 2026-06-10
Fix/collapsed toolbar rename

### LINKED ISSUE #14082: "Rename Folder" option unavailable in collapsed toolbar layout
### Preliminary Checks

### What happened?

When appearance is set to "Single toolbar" or "Multiple toolbars", right clicking a folder in the toolbar on the left gives a "Rename Folder" option. When in "Collapsed toolbar", that option is not missing.

I understand that this might have been intentional as the name of the folder is not easily visible in collapsed toolbar mode, however if you hover over a closed folder, you still get the popup to search tabs which displays something like "Search <folder name>". 

### Expected behavior

Rename folder option should be available in popup menu

### Actual behavior

It's not there

### Steps to reproduce

1. Change Browser Layout to Collapsed toolbar in Look and Feel settings
2. Right click a folder in the toolbar on the left

### Screenshots and videos

_No response_

### Version

1.20.2b

Linux (Tarball)

Tab Folders

```shell

```

sukritgoyal:
## Fix: "Rename Folder" now works in Collapsed Toolbar layout

Closes #14082

### Problem

When the sidebar is in **Collapsed Toolbar** mode (compact icon-only view), right-clicking a tab folder and selecting "Rename Folder" did nothing. The option was visible in the context menu but completely unresponsive.

The root cause was an early-return guard in `nsZenFolder.rename()`:

```js
rename() {
  if (!document.documentElement.hasAttribute("zen-sidebar-expanded")) {
    return; // silently bailed out in collapsed mode
  }
  ...
}
```

In collapsed mode the sidebar has no visible label input, so the inline rename flow (`gZenVerticalTabsManager.renameTabStart`) can't work — but instead of falling back gracefully, it just returned without any feedback to the user.

### Solution

When the sidebar is collapsed, fall back to a native `Services.prompt.prompt()` dialog pre-filled with the folder's current name. On confirmation, the result is passed to the existing `labelElement.onRenameFinished()` callback, which handles saving and firing the `ZenFolderRenamed` event — so all downstream persistence logic is unchanged.

When the sidebar is expanded, the existing inline rename flow is used as before.

### Changes

- **`src/zen/folders/ZenFolder.mjs`** — Modified `rename()` to show a prompt dialog instead of silently returning when `zen-sidebar-expanded` is absent.
- **`locales/en-US/browser/browser/zen-folders.ftl`** — Added `zen-folder-rename-prompt` Fluent string for the prompt message.

### Before / After

| Sidebar mode | Before | After |
|---|---|---|
| Expanded (Single/Multiple toolbar) | ✅ Inline rename works | ✅ Unchanged |
| Collapsed Toolbar | ❌ Silent no-op | ✅ Native prompt dialog |


src/zen/folders/ZenFolder.mjs
```diff
@@ -137,8 +137,25 @@ export class nsZenFolder extends MozTabbrowserTabGroup {
...
+        null,                      // checkMsg (no checkbox)
+        {}                         // checkState
+      );
+      if (confirmed && result.value.trim()) {
+        this.labelElement.onRenameFinished(result.value);
+      }
       return;
     }
+    // Expanded sidebar: use the existing inline rename flow
     gZenVerticalTabsManager.renameTabStart({
```
mr-cheffy, line 159: I think we should move this prompt functionality into `renameTabStart` so it also works for renaming spaces, tabs, etc.

src/zen/folders/ZenFolder.mjs
```diff
@@ -137,8 +137,25 @@ export class nsZenFolder extends MozTabbrowserTabGroup {
 
   rename() {
     if (!document.documentElement.hasAttribute("zen-sidebar-expanded")) {
+      // Collapsed toolbar mode: no inline label input is visible, so use a prompt dialog.
+      const currentName = this.label || "";
+      const result = { value: currentName };
+      const confirmed = Services.prompt.prompt(
+        window,
+        null,                      // title (null = use default window title)
+        gBrowser.ownerDocument.l10n.getValueSync("zen-folder-rename-prompt", {
```
mr-cheffy, line 146: Sync value fetching doesn't work in this context. Have you tested this?

Also, why use `gBrowser.ownerDocument` instead of the document object?

mr-cheffy (changes requested):

src/zen/folders/ZenFolder.mjs
sukritgoyal: Actually I am new to open source contributions. If you can guide me how do I test and anything which I need to know other than just coding and development. The codebase is very large.

sukritgoyal:
@mr-cheffy If you can guide me here, I will be really grateful to you!

ashvwinn:
@sukritgoyal You can read this for guidance https://docs.zen-browser.app/contribute/desktop

src/zen/folders/ZenFolder.mjs
mr-cheffy: Did you get it to build?

sukritgoyal: Nope this is the first time I working on open source contributions.

PR #14102, opened 2026-06-10
gh-14100: restore workspace routing for extension-opened tabs with explicit containers

### LINKED ISSUE #14100: Extension-opened container tabs no longer switch to the container's assigned space (regression in 1.20.2b, PR #14001)
Disclaimer: I've researched the issue and submitted this ticket with Claude's help. I've verified all of its claims and the issue draft following:

### Preliminary Checks

### What happened?

I have a container ("Work") set as the default container of a space, with "Switch to the space where the container is set as default when opening container tabs" enabled. Until 1.20.1b, any tab opened in the Work container was automatically moved to the Work space.

Since updating to 1.20.2b, tabs opened **by an extension** into the Work container (in my case, [Open external links in a container](https://addons.mozilla.org/en-US/firefox/addon/open-url-in-container/), which calls `browser.tabs.create()` with an explicit `cookieStoreId`) stay in the currently active space instead of being moved to the container's assigned space.

Reproduced on two machines (Ubuntu 25.10 and EndeavourOS, both x64), and verified with all Mods and Custom CSS removed — only the "Open external links in a container" add-on installed.

Tabs opened any other way still behave correctly:
- "Open in container → Work" from the context menu → opens in the Work space ✓
- Middle-clicking a link in a Work-container tab sitting in another space → opens in the Work space ✓
- Tab opened via the extension into the Work container → stays in the current space ✗

### Root cause

This bisects to PR #14001 ("bypass workspace container coercion for extension-opened tabs"), shipped in 1.20.2b. It added an early return in `getContextIdIfNeeded` (`src/zen/spaces/ZenSpaceManager.mjs`):

```js
if (
  triggeringPrincipal &&
  triggeringPrincipal.isAddonOrExpandedAddonPrincipal
) {
  return [userContextId, false, undefined];
}
```

This returns before the `shouldForceContainerTabsToWorkspace` branch, so `zenForcedWorkspaceId` is always `undefined` for extension-opened tabs — the container→space routing is skipped entirely.

The PR's goal was to stop Zen from forcing the active space's container onto extension tabs opened **without** a `cookieStoreId` (the Grammarly sign-in fix). That part is correct. But the check is too broad: when an extension opens a tab **with** an explicit `cookieStoreId`, the user's intent is unambiguous, and the container→space switch should still apply.

Suggested fix: only take the addon-principal early return when the extension did not specify a container, e.g.:

```js
if (
  triggeringPrincipal &&
  triggeringPrincipal.isAddonOrExpandedAddonPrincipal &&
  typeof userContextId === "undefined"
) {
  return [userContextId, false, undefined];
}
```

(or equivalently, only skip the *container coercion* for addon principals while still allowing the forced-workspace lookup when `userContextId` is already set).

### Expected behavior

A tab opened by an extension with an explicit `cookieStoreId` pointing to a container that is the default of a space is moved to that space, exactly like tabs opened via "Open in container" or middle-click.

### Actual behavior

Extension-opened container tabs stay in the currently active space. The container is applied, but the space switch never happens.

### Steps to reproduce

1. Create a container "Work" and set it as the default container of a space "Work", with the switch-to-space-on-container-tab setting enabled.
2. Install "Open external links in a container" (or any extension that calls `browser.tabs.create({ cookieStoreId })`).
3. While a different space is active, open an external link routed to the Work container by the extension.
4. Observe the tab opens in the Work container but stays in the current space. On 1.20.1b and earlier the tab was moved to the Work space.

### Version

1.20.2b (Firefox 151.0.3) (64-bit)

Linux - x64 (Ubuntu 25.10 and EndeavourOS)

Spaces (Workspaces)

rishabshah0:
The previous fix for the Grammarly sign-in issue (which bypassed container coercion for tabs opened by WebExtensions) broke extensions that explicitly target specific containers (such as 'Open external links in a container'). Those tabs opened in the active workspace instead of the workspace associated with the container.

This change refines the bypass condition to only apply when the extension does not specify a container (meaning default/No Container context is expected, i.e., userContextId is undefined). This preserves the Grammarly fix while restoring correct workspace switching for container-specific extension requests.

### DIFF
--- a/src/zen/spaces/ZenSpaceManager.mjs
+++ b/src/zen/spaces/ZenSpaceManager.mjs
@@ -3005,7 +3005,8 @@ class nsZenWorkspaces {
 
     if (
       triggeringPrincipal &&
-      triggeringPrincipal.isAddonOrExpandedAddonPrincipal
+      triggeringPrincipal.isAddonOrExpandedAddonPrincipal &&
+      typeof userContextId === "undefined"
     ) {
       return [userContextId, false, undefined];
     }


mr-cheffy (approved):
Thanks!

PR #14104, opened 2026-06-10
no-bug: add --space command line flag to open a link in a specific workspace

barakor:
Adds a `--space <space>` command line flag that switches to the workspace matching the given name (case-insensitive) or UUID:

    zen --space work
    zen --space "Space Name"
    zen --space c2d1e574-3f04-4bd5-9721-2e6dd7f1adf9 # uuid of workspace

Behavior:
- Browser running: delegates to `gZenWorkspaces.changeWorkspaceFromCommandLine` in the most recent non-private window, which switches to the matching workspace.
- Cold start: the workspace is stashed in the `zen.workspaces.cmdline-initial-workspace` pref, which is consumed (and cleared) by the first window that restores its workspaces, so the browser starts on the requested space.
- Unknown workspace: no-op — the active workspace stays unchanged.
- When used together with a url invocation, the url would open in the specified space. 

Implementation: extends the `BrowserContentHandler.sys.mjs` patch (same pattern as the existing `--blank-window` flag) and adds `resolveWorkspaceFromString` / `changeWorkspaceFromCommandLine` to `ZenSpaceManager`. Includes mochitests covering name/UUID resolution, workspace switching, and the unknown-workspace no-op.
When used together with a url invocation, the url would open in the specified space. 

Useful for OS shortcuts and scripting, e.g. a `.desktop` entry per workspace.

Disclaimer: this was written by Claude, I reviewed and tested the code myself.

### DIFF
--- a/src/browser/components/BrowserContentHandler-sys-mjs.patch
+++ b/src/browser/components/BrowserContentHandler-sys-mjs.patch
@@ -2,7 +2,7 @@ diff --git a/browser/components/BrowserContentHandler.sys.mjs b/browser/componen
 index 4bfb4199dafac004486dfc61db954173742a010a..542c156432b9d7c2190c0d14ce9cfd09814422bf 100644
 --- a/browser/components/BrowserContentHandler.sys.mjs
 +++ b/browser/components/BrowserContentHandler.sys.mjs
-@@ -603,6 +603,88 @@ nsBrowserContentHandler.prototype = {
+@@ -603,6 +603,67 @@ nsBrowserContentHandler.prototype = {
        }
      }
  
@@ -30,29 +30,6 @@ index 4bfb4199dafac004486dfc61db954173742a010a..542c156432b9d7c2190c0d14ce9cfd09
 +    try {
 +      let zenWorkspaceParam = cmdLine.handleFlagWithParam("zen-workspace", false);
 +      if (zenWorkspaceParam) {
-+        let zenWorkspaceURL = null;
-+        try {
-+          zenWorkspaceURL = cmdLine.handleFlagWithParam("url", false);
-+        } catch (e) {}
-+        if (zenWorkspaceURL === null) {
-+          for (let i = cmdLine.length - 1; i >= 0; i--) {
-+            const arg = cmdLine.getArgument(i);
-+            if (arg && arg[0] != "-") {
-+              zenWorkspaceURL = arg;
-+              cmdLine.removeArguments(i, i);
-+              break;
-+            }
-+          }
-+        }
-+        let uri = null;
-+        let principal = null;
-+        if (zenWorkspaceURL !== null) {
-+          const resolved = resolveURIInternal(cmdLine, zenWorkspaceURL);
-+          if (shouldLoadURI(resolved.uri)) {
-+            uri = resolved.uri;
-+            principal = resolved.principal;
-+          }
-+        }
 +        let targetWin = Services.wm.getMostRecentWindow("navigator:browser");
 +        if (
 +          !targetWin?.gZenWorkspaces?.workspaceEnabled ||
@@ -70,16 +47,18 @@ index 4bfb4199dafac004486dfc61db954173742a010a..542c156432b9d7c2190c0d14ce9cfd09
 +          }
 +        }
 +        if (targetWin) {
-+          targetWin.gZenWorkspaces.openLinkInWorkspaceFromCommandLine(
-+            zenWorkspaceParam,
-+            uri ? uri.spec : null
++          targetWin.gZenWorkspaces.changeWorkspaceFromCommandLine(
++            zenWorkspaceParam
 +          );
-+          targetWin.focus();
++          cmdLine.preventDefault = true;
 +        } else {
-+          let win = openBrowserWindow(cmdLine, principal, uri ? uri.spec : null);
-+          win._zenInitialWorkspace = zenWorkspaceParam;
++          // Cold start: stash the workspace so the first window that
++          // restores picks it up, and let normal startup open that window.
++          Services.prefs.setStringPref(
++            "zen.workspaces.cmdline-initial-workspace",
++            zenWorkspaceParam
++          );
 +        }
-+        cmdLine.preventDefault = true;
 +      }
 +    } catch (e) {
 +      if (e.result != Cr.NS_ERROR_INVALID_ARG) {
@@ -91,12 +70,12 @@ index 4bfb4199dafac004486dfc61db954173742a010a..542c156432b9d7c2190c0d14ce9cfd09
      var searchParam = cmdLine.handleFlagWithParam("search", false);
      if (searchParam) {
        doSearch(searchParam, cmdLine);
-@@ -672,6 +754,8 @@ nsBrowserContentHandler.prototype = {
+@@ -672,6 +733,8 @@ nsBrowserContentHandler.prototype = {
        "  --new-window <url> Open <url> in a new window.\n" +
        "  --new-tab <url>    Open <url> in a new tab.\n" +
        "  --private-window [<url>] Open <url> in a new private window.\n";
 +    info += "  --blank-window [<url>]  Open the new blank window.\n";
-+    info += "  --zen-workspace <space> [<url>]  Open <url> in the workspace with the given name or UUID.\n";
++    info += "  --zen-workspace <space>  Switch to the workspace with the given name or UUID.\n";
      if (AppConstants.platform == "win") {
        info += "  --preferences      Open Options dialog.\n";
      } else {
--- a/src/browser/components/BrowserContentHandler-sys-mjs.patch
+++ b/src/browser/components/BrowserContentHandler-sys-mjs.patch
@@ -28,7 +28,7 @@ index 4bfb4199dafac004486dfc61db954173742a010a..542c156432b9d7c2190c0d14ce9cfd09
 +    }
 +
 +    try {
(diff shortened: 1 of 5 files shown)


code-IM-perfect:
Thanks @barakor had been waiting for this feature for a lot time

dor-eon:
@mr-cheffy Hi, any plans on landing this? The feature would be very useful to me :pray:

barakor:
Hey @mr-cheffy , is there a problem with the idea or the implementation? 
Another variation to achieve a similar goal could be a setting to choose default workspace for new tabs, currently used, specific one... and an option for "ask every time" that will pop a modal dialog with options.

src/browser/components/BrowserContentHandler-sys-mjs.patch
```diff
@@ -27,15 +27,76 @@ index 430c6877c81824d8ab318cacc997d4eda4cb6a11..e8a195e1fe09952b90a0d73d13c6c15d
 +      }
 +    }
 +
++    try {
++      let zenWorkspaceParam = cmdLine.handleFlagWithParam("zen-workspace", false);
++      if (zenWorkspaceParam !== null) {
```
mr-cheffy:
```suggestion
+      if (zenWorkspaceParam) {
```

src/browser/components/BrowserContentHandler-sys-mjs.patch
```diff
@@ -27,15 +27,76 @@ index 430c6877c81824d8ab318cacc997d4eda4cb6a11..e8a195e1fe09952b90a0d73d13c6c15d
...
++        let zenWorkspaceURL = null;
++        try {
++          zenWorkspaceURL = cmdLine.handleFlagWithParam("url", false);
```
mr-cheffy: Why do you need to parse the url? Isn't this already handled by FF?

mr-cheffy (changes requested):

src/browser/components/BrowserContentHandler-sys-mjs.patch
barakor: This is in the case that a workspace was given as an argument - 
ff would open it in whatever window/workspace is active if it's not removed, because this patch is not terminal (not returning or breaking) we have to parse the url out to avoid double opening
(note: `cmdLine.handleFlagWithParam` consumes the flag and value from cmdLine)

mr-cheffy: Just consume the space's name and let firefox handle the url parsing though?

barakor: I honestly didn't think of this flow initially, it's much nicer, updated to this approach

src/browser/components/BrowserContentHandler-sys-mjs.patch
```diff
@@ -27,15 +27,55 @@ index 4bfb4199dafac004486dfc61db954173742a010a..542c156432b9d7c2190c0d14ce9cfd09
 +      }
 +    }
 +
++    try {
++      let zenWorkspaceParam = cmdLine.handleFlagWithParam("zen-workspace", false);
++      if (zenWorkspaceParam) {
++        let targetWin = Services.wm.getMostRecentWindow("navigator:browser");
```
mr-cheffy: Use `lazy.BrowserWindowTracker.getTopWindow({ private: false, allowPopups: false })` and no need for the for loop and then check for `privateWindowOrDisabled`

src/browser/components/BrowserContentHandler-sys-mjs.patch
```diff
@@ -27,15 +27,55 @@ index 4bfb4199dafac004486dfc61db954173742a010a..542c156432b9d7c2190c0d14ce9cfd09
...
++            ) {
++              targetWin = win;
++            }
++          }
++        }
++        if (targetWin) {
++          targetWin.gZenWorkspaces.changeWorkspaceFromCommandLine(
++            zenWorkspaceParam
++          );
++          cmdLine.preventDefault = true;
```
mr-cheffy: Wouldn't this prevent it from parsing the url?

src/zen/spaces/ZenSpaceManager.mjs
```diff
@@ -1583,6 +1616,20 @@ class nsZenWorkspaces {
...
+   * Handles the `--zen-workspace` command line flag for an already running
+   * browser: switches to the workspace matching the given name or UUID.
+   *
+   * @param {string} workspaceMatch - The workspace UUID or name to switch to
+   */
+  async changeWorkspaceFromCommandLine(workspaceMatch) {
+    await this.promiseInitialized;
+    const workspace = this.resolveWorkspaceFromString(workspaceMatch);
+    if (workspace && workspace.uuid !== this.activeWorkspace) {
+      await this.changeWorkspaceWithID(workspace.uuid);
```
mr-cheffy, line 1750: This wont move the tab to the target space, no?

mr-cheffy (changes requested):
I would also rename the argument to `--space`, no?

src/browser/components/BrowserContentHandler-sys-mjs.patch
barakor: cool!

src/browser/components/BrowserContentHandler-sys-mjs.patch
barakor: no, it still parses and handles it, I dropped it as it seems to not have any effect anymore.

src/zen/spaces/ZenSpaceManager.mjs
barakor: No, after your earlier comment I revisited the flow to only move to the workspace before the tab opens, then letting ff deal with everything else - it opens the tab in workspace we switched to.

barakor:
Hey @mr-cheffy , is there anything else you think is need of changing?

src/zen/spaces/ZenSpaceManager.mjs
barakor: Hey @mr-cheffy  is this an issue? can I resolve this?

MadaraUchiha:
Really looking forward to this 🙏

karmicdude:
+1

barakor:
@mr-cheffy hey, anything I can do to get this going forward?

src/browser/components/BrowserContentHandler-sys-mjs.patch
```diff
@@ -27,15 +27,45 @@ index 4bfb4199dafac004486dfc61db954173742a010a..542c156432b9d7c2190c0d14ce9cfd09
...
++      if (zenWorkspaceParam) {
++        let targetWin = lazy.BrowserWindowTracker.getTopWindow({
++          private: false,
++          allowPopups: false,
++        });
++        if (
++          targetWin?.gZenWorkspaces &&
++          !targetWin.gZenWorkspaces.privateWindowOrDisabled
++        ) {
++          targetWin.gZenWorkspaces.changeWorkspaceFromCommandLine(
```
mr-cheffy: Wouldn't it be better to just call `changeWorkspaceWithID`?

src/browser/components/BrowserContentHandler-sys-mjs.patch
```diff
@@ -27,15 +27,45 @@ index 4bfb4199dafac004486dfc61db954173742a010a..542c156432b9d7c2190c0d14ce9cfd09
...
++            zenWorkspaceParam
++          );
++        } else {
++          // Cold start: stash the space so the first window that
++          // restores picks it up, and let normal startup open that window.
++          Services.prefs.setStringPref(
++            "zen.workspaces.cmdline-initial-workspace",
```
mr-cheffy, line 52: Seems a bit hacky, but it would be alright

mr-cheffy (changes requested):

src/browser/components/BrowserContentHandler-sys-mjs.patch
barakor: We don't have the id here (user inputs a string with the name of the space), so we have to first resolve the space id, which is done inside of `changeWorkspaceFromCommandLine`

src/browser/components/BrowserContentHandler-sys-mjs.patch
barakor: I was trying to think of what would be the most "native" way to let zen handle it, this seemed like the most ergonomic way to do it

src/zen/spaces/ZenSpaceManager.mjs
```diff
@@ -795,6 +814,20 @@ class nsZenWorkspaces {
...
+    const cmdLineWorkspace = Services.prefs.getStringPref(
+      "zen.workspaces.cmdline-initial-workspace",
+      ""
+    );
+    if (cmdLineWorkspace) {
+      // Set by the `--zen-workspace` command line flag on a cold start,
+      // see BrowserContentHandler.sys.mjs. Consumed by the first window
+      // that restores its workspaces.
+      Services.prefs.clearUserPref("zen.workspaces.cmdline-initial-workspace");
+      const initialWorkspace = this.resolveWorkspaceFromString(cmdLineWorkspace);
```
mr-cheffy, line 830: No need for this check

mr-cheffy (changes requested):

mr-cheffy:
Could you please open up a new PR without claude authoring the commits? We don't allow that unfortunately. But after that, this should be pretty much ready for merge!

PR #14144, opened 2026-06-12
gh-8309: Fix moving essential tabs to new window

### LINKED ISSUE #8309: Moving Essential Tab to New Window removes it from Essentials

wieli99:
Fixes #8309.

updates the move-to-new-window behavior for Essential tabs. For normal user moves, keeps the original Essential tab in place and opens a regular copy in the new window.

I also added coverage for the related edge cases that came up while testing this path.

### DIFF
--- a/src/browser/components/tabbrowser/content/tabbrowser-js.patch
+++ b/src/browser/components/tabbrowser/content/tabbrowser-js.patch
@@ -832,17 +832,57 @@ index 08b5b56e069d038d72c87355920c4ce8a55ed805..87f41c8583c26f364530def5bd5d8333
          aTab.selected ||
          aTab.closing ||
          // Tabs that are sharing the screen, microphone or camera cannot be hidden.
-@@ -6834,7 +7037,8 @@
+@@ -6834,7 +7037,48 @@
       * @param {object} [aOptions={}]
       *   Key-value pairs that will be serialized into the features string.
       */
 -    replaceTabWithWindow(aTab, aOptions = {}) {
++    #normalizeDuplicatedEssentialTab(tab) {
++      tab.removeAttribute("zen-essential");
++      tab.removeAttribute("zenDefaultUserContextId");
++      if (tab.pinned) {
++        tab.ownerGlobal.gBrowser.unpinTab(tab);
++      }
++    }
++
++    async #duplicateEssentialTab(tab) {
++      const duplicatedTab = this.duplicateTab(tab, true, {
++        tabIndex: tab._tPos + 1,
++      });
++      this.#normalizeDuplicatedEssentialTab(duplicatedTab);
++
++      return new Promise(resolve => {
++        const onRestored = () => {
++          this.#normalizeDuplicatedEssentialTab(duplicatedTab);
++          resolve(duplicatedTab);
++        };
++
++        duplicatedTab.addEventListener("SSTabRestored", onRestored, {
++          once: true,
++        });
++      });
++    }
++
 +    replaceTabWithWindow(aTab, aOptions = {}, zenForceSync = false) {
 +      if (!this.isTab(aTab)) return; // TODO: Handle tab groups
++      if (
++        !zenForceSync &&
++        aTab.hasAttribute("zen-essential")
++      ) {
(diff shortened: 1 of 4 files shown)


mr-cheffy:
There seems to be git conflicts, could you please fix it?

wieli99:
Whoops, yeah should be fixed now :)

mr-cheffy:
Hi, could you please fix the git issues again? Sorry about that.

wieli99:
No worries. Should be ready to merge again.

wieli99:
I fixed the conflicts again. Is there anything else you need from me to merge this @mr-cheffy ?

mr-cheffy (commented):
Why's there 2 different normalizeDuplicatedEssentialTab going on? What does each branch do

wieli99:
They were doing the same thing in the single-tab and multi-tab paths. I moved that into one shared helper and also made both paths wait for SSTabRestored before moving the duplicate. Added tests fr it too.

src/browser/components/tabbrowser/content/tabbrowser-js.patch
```diff
@@ -829,38 +829,123 @@ index 08b5b56e069d038d72c87355920c4ce8a55ed805..555ffd4772d9d4903491fdff9f368285
...
++          this.#normalizeDuplicatedEssentialTab(duplicatedTab);
++          resolve(duplicatedTab);
++        };
++        const onClose = () => {
++          cleanup();
++          resolve(null);
++        };
++
++        duplicatedTab.addEventListener("SSTabRestored", onRestored);
++        duplicatedTab.addEventListener("TabClose", onClose);
```
mr-cheffy: No need to, im pretty sure SSTabRestored is guaranteed to be fired

src/browser/components/tabbrowser/content/tabbrowser-js.patch
```diff
@@ -829,38 +829,123 @@ index 08b5b56e069d038d72c87355920c4ce8a55ed805..555ffd4772d9d4903491fdff9f368285
...
 -    replaceTabWithWindow(aTab, aOptions = {}) {
++    #normalizeDuplicatedEssentialTab(tab) {
++      tab.removeAttribute("zen-essential");
++      tab.removeAttribute("zenDefaultUserContextId");
++      if (tab.pinned) {
++        tab.ownerGlobal.gBrowser.unpinTab(tab);
++      }
++    }
++
++    async #duplicateEssentialTab(tab) {
```
mr-cheffy: Move these to zen's space manager to make patches smaller, same goes for normalizeDuplicatedEssentialTab

mr-cheffy (changes requested):

src/browser/components/tabbrowser/content/tabbrowser-js.patch
wieli99: Yep, moved both helpers into `gZenWorkspaces` so the tabbrowser patch only has the call sites now.

src/browser/components/tabbrowser/content/tabbrowser-js.patch
wieli99: Removed that branch and now just wait for SSTabRestored with { once: true }.

wieli99:
Since dev keeps moving and this PR has picked up conflicts a few times now, I’m going to hold off on rebasing again until you’re ready to take another look @mr-cheffy .

PR #14151, opened 2026-06-12
fix: compact sidebar on the right was covering the page scrollbar

thisis-manan:
## What

When compact mode is enabled with the sidebar on the right side, the sidebar's shown position was `right: -4px` — extending past the window edge and completely covering the page scrollbar, making it unclickable and undraggable.

## How

The sidebar's shown position now parks against the left edge of the scrollbar instead of the window edge.

- **CSS**: `right` now uses `--zen-page-scrollbar-width` with `env(scrollbar-inline-size, 0px)` as fallback. On Windows/Linux, `env()` resolves to the real scrollbar width automatically. On macOS overlay scrollbars, it resolves to 0 and JS handles it.
- **JS**: `_initScrollbarTracking()` sets `--zen-page-scrollbar-width` on `:root`. For classic scrollbars (Windows/Linux) it measures once on init. For macOS overlay scrollbars, it listens to `wheel` events (which bubble from content to chrome) and fades the gap out ~1.5s after scrolling stops — matching the native overlay scrollbar fade timing. The gap width is configurable via `zen.view.compact.overlay-scrollbar-width` (default 15px).
- **Flicker fix**: `onLeave` now returns early if the cursor is within the scrollbar gap on the right edge, preventing the sidebar from rapidly hiding and re-showing when the user moves their cursor to the scrollbar.

All changes are gated on `sidebarIsOnRight` — no effect on left-side or non-compact layouts.

## Test

1. Enable compact mode, move sidebar to the right
2. Open a page with a scrollbar
3. Verify you can click and drag the scrollbar while the sidebar is shown

### DIFF
--- a/prefs/zen/compact-mode.yaml
+++ b/prefs/zen/compact-mode.yaml
@@ -20,6 +20,9 @@
 - name: zen.view.compact.sidebar-keep-hover.duration
   value: 150
 
+- name: zen.view.compact.overlay-scrollbar-width
+  value: 15
+
 - name: zen.view.compact.animate-sidebar
   value: true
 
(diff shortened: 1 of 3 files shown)


mr-cheffy:
This looks VERY AI generated, I dont think this would be optimal fix to the solution imo

PR #14155, opened 2026-06-12
Fix shift + tab not focusing search engine dropdown while on search bar

### LINKED ISSUE #14126: Search bar: Shift+Tab selects a search suggestion instead of switching search engine
### Preliminary Checks

### What happened?

Shift+Tab selects a search suggestion instead of switching search engine

### Expected behavior

Pressing Shift+Tab should move focus to / cycle through the available search engines (one-off engine selector), not through the search suggestions

### Actual behavior

Shift+Tab selects a search suggestion instead of a search engine

### Steps to reproduce

1. Click into the search box.
2. Type a query (e.g. test).
3. Press Shift+Tab.

### Screenshots and videos

_No response_

### Version

1.21b

Linux (Tarball)

Other

```shell

```

Zharkan:
fix #14126


mr-cheffy:
I feel like this would be better suited to be submitted to firefox? Would off-load work from us: https://firefox-source-docs.mozilla.org/contributing/how_to_submit_a_patch.html

Zharkan:
This is Zen specific though, upstream Firefox has scotchBonnet.enableOverride enabled by default, so Tab/Shift+Tab navigation to the search engine dropdown already works there. The issue only exists because Zen disables scotchBonnet, which gates that code path. The search bar widget still uses SearchModeSwitcher regardless of scotchBonnet, so we need this patch to wire up the Tab navigation for it on our side, right ?

mr-cheffy:
Hi, sorry for the delayed answer. Does the bug happen on firefox when the pref is turned off?

PR #14210, opened 2026-06-15
perf: core architecture optimizations (memory, IPC, I/O, UI)

Attmaca:
This PR tackles several deep architectural bottlenecks in core browser mechanisms (IPC, Disk I/O, and Memory) to fix CPU overhead, UI stuttering, and memory bloat.

### Fixes & Optimizations:

- **Fix IPC Flood in `ZenGlanceChild.sys.mjs`:** 
  The content process was broadcasting `ZenGlance:RecordLinkClickData` IPC messages on *every single mousedown event* anywhere on the page. Added a strict check to only send the IPC message if the target is an actual link (`href`).

- **Fix Unbounded JSON Bloat & Disk I/O Lock in `ZenUBResultsLearner.sys.mjs`:** 
  Unclicked URL bar suggestions were accumulating negative scores without ever being deleted, causing the database JSON to grow infinitely. This massive JSON was being synchronously written to `prefs.js` on every keystroke, locking the main thread. Deprioritized entries are now properly deleted from the pool.

- **Fix Orphaned DOM Tree Leak in `ZenHasPolyfill.mjs`:** 
  Disconnected `MutationObserver` instances were keeping hard references to DOM elements in the `this.observers` array, preventing the Garbage Collector from freeing destroyed UI components. Wrapped the DOM references in `WeakRef` to allow proper garbage collection.

- **Fix Global Resize Layout Thrashing in `ZenCompactMode.mjs`:** 
  The sidebar `ResizeObserver` was firing `window.dispatchEvent(new window.Event("resize"))` continuously during animations, causing forced synchronous layout recalculations for the entire browser window at 60fps. The global resize dispatch is now safely debounced using `requestAnimationFrame`.

- **Fix Jitter/Disk I/O during Trackpad Swipes in `ZenSpacesSwipe.mjs`:** 
  The animation state (`zen.swipe.is-fast-swipe`) was being synchronously written to `prefs.js` during trackpad workspace transitions, causing file-lock stuttering. Moved this state tracking to a standard DOM attribute (`zen-is-fast-swipe`) in memory.

*(Also includes minor cleanups: fixed invalid CSS string joins in Workspace animations, added teardown for zombie global event listeners, and ensured pending rAFs are canceled on UI unmount).*

### DIFF
--- a/src/zen/common/emojis/ZenEmojiPicker.mjs
+++ b/src/zen/common/emojis/ZenEmojiPicker.mjs
@@ -139,26 +139,34 @@ class nsZenEmojiPicker extends nsZenDOMOperatedFeature {
   #onSearchInput(event) {
     const input = event.target;
     const value = input.value.trim().toLowerCase();
-    // search for emojis.tags and order by emojis.order
-    const filteredEmojis = this.#emojis
-      .filter(emoji => {
-        return emoji.tags.some(tag => tag.toLowerCase().includes(value));
-      })
+    
+    // Create an O(1) lookup Map to prevent O(N^2) iteration over 3000 DOM nodes.
+    const filteredEmojiOrderMap = new Map();
+    const filtered = this.#emojis
+      .filter(emoji => emoji.tags.some(tag => tag.toLowerCase().includes(value)))
       .sort((a, b) => a.order - b.order);
+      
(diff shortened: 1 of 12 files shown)


mr-cheffy:
These all seem to be very AI generated and I can practically already see regressions just by looking at it.

PR #14213, opened 2026-06-15
Add new tab action for folders

edducon:
## 
- Add a new tab button to regular folder headers in the vertical tabs sidebar.
- Reuse Zen's native new-tab/urlbar flow so users can type a search or URL immediately.
- Make the New Tab command folder-aware when the selected tab is inside a regular folder.
## 
This keeps live folders unchanged and only targets regular Zen folders.


mr-cheffy:
Hi, thanks for the contribution! But unfortunately we aren't planning on adding this, sorry.

PR #14221, opened 2026-06-16
gh-14165: fix background tabs focusing on routed tabs

### LINKED ISSUE #14165: Can't open background tab via space routing
### Preliminary Checks

### What happened?

Usually, you can open a link into a background tab by ctrl+click.
By if the link is in the path of space routing and the destination is not the currently active space, ctrl+click will brings you to another space and shifts focus on the newly made tab

### Expected behavior

When you open a link using ctrl+click, and the tab is moved to other space via space routing, current portveiw focus should not be changed., as expected when you open a background tab.

### Actual behavior

When you open a link using ctrl+click, and the tab is moved to other space via space routing, current portveiw focus also shift to that tab

### Steps to reproduce

1. Establish a rule on space routing, such as move www.reddit.com to space 2
2. In space 1, open a link containing, www.reddit.com using ctrl + click

### Screenshots and videos

_No response_

### Version

1.21b

Windows - x64

Workspaces

```shell

```

ashvwinn:
fixes: #14165

Two things I want to ask:
- This does fix the issue but I wanted to ask if we should maybe change the background tab toast message to notify user that it was routed to another space.
- Another thing is that the background tab toast does not appear when the sidebar is in view which I know is intentional but maybe it should be shown even then when tabs are being routed.


ashvwinn:
image

locales/en-US/browser/browser/zen-space-routing.ftl
```diff
@@ -22,4 +22,6 @@ zen-space-routing-regex =
     .label = RegEx
 
 zen-space-routing-open-in = Open In
-zen-space-routing-url = URL
\ No newline at end of file
+zen-space-routing-url = URL
+
+zen-space-routing-tab-routed-toast = New tab opened in Space { $targetWorkspace }
```
mr-cheffy:
```suggestion
zen-space-routing-tab-routed-toast = New tab opened in { $targetWorkspace }
```

src/browser/components/tabbrowser/content/tabbrowser-js.patch
mr-cheffy, line 1: I think the patch file got so messed up because of auto formatting or something?

mr-cheffy (approved):

PR #14272, opened 2026-06-18
gh-11766: Fixed video fullscreen not working while in glance

### LINKED ISSUE #11766: Full screen mode not working in YouTube while using Glance mode
### Preliminary Checks

### What happened?

When using the "Glance" feature in Zen Browser to open a YouTube link, the full-screen button on the YouTube player does not function. It fails to expand the video to fill the screen or the glance window properly.

### Expected behavior

The YouTube video should enter full-screen mode (within the Glance window or the entire screen) when the full-screen icon is clicked.

### Actual behavior

Clicking the full-screen button does nothing, or the video remains windowed inside the Glance overlay.

### Steps to reproduce

1. Open Zen Browser.
2. Open any website and use the "Glance" feature to open a YouTube video link (or use the shortcut to open a link in Glance).
3. Try to click the "Full Screen" icon on the YouTube player.
4. Observe that the video does not expand.

### Screenshots and videos

Image

### Version

1.17.15b

Linux (Tarball)

Glance

```shell

```

connosssss:
fixes #11766 

- Small CSS tweaks to let videos full screen correctly while making sure it looks correctly full screened in glance mode
- Renamed Title and branch name

### DIFF
--- a/src/zen/glance/ZenGlanceManager.mjs
+++ b/src/zen/glance/ZenGlanceManager.mjs
@@ -1430,7 +1430,7 @@ class nsZenGlanceManager extends nsZenDOMOperatedFeature {
       return;
     }
 
-    if (event.target === browser || browser.contains(event.target)) {
+    if (event.target === browser) {
       this.fullyOpenGlance();
     }
   }
--- a/src/zen/glance/ZenGlanceManager.mjs
+++ b/src/zen/glance/ZenGlanceManager.mjs
@@ -53,6 +53,10 @@ class nsZenGlanceManager extends nsZenDOMOperatedFeature {
   #setupEventListeners() {
     window.addEventListener("TabClose", this.onTabClose.bind(this));
     window.addEventListener("TabSelect", this.onLocationChange.bind(this));
+    window.addEventListener(
+      "MozDOMFullscreen:Entered",
+      this.onFullscreenEntered.bind(this)
+    );
 
     document
       .getElementById("tabbrowser-tabpanels")
@@ -1411,6 +1415,23 @@ class nsZenGlanceManager extends nsZenDOMOperatedFeature {
   onTabClose(event) {
     if (event.target === this.#currentParentTab) {
       this.closeGlance({ onTabClose: true });
+    }
+  }
+
+  /**
+   * Handle DOM Fullscreen request while inside glance
+   *
+   * @param {Event} event - The MozDOMFullscreen:Entered event
+   */
+  onFullscreenEntered(event) {
+    const browser = this.#currentBrowser;
+
+    if (!browser) {
+      return;
+    }
+
+    if (event.target === browser || browser.contains(event.target)) {
+      this.fullyOpenGlance();
     }
   }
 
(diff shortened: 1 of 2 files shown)


src/zen/glance/zen-glance.css
mr-cheffy, line 120: wouldn't it be better to simply not style these if its fullscreen?

mr-cheffy (changes requested):

src/zen/glance/zen-glance.css
connosssss: You're right, doing that made it a lot simpler while still working. I'm updating it now

mr-cheffy:
I've been wondering, what if instead, we make the glance tab a "full tab" when entering full screen? Arc does it that way

mr-cheffy (changes requested):
I think we'll be going forward with https://github.com/zen-browser/desktop/pull/14272#issuecomment-4746440678

connosssss:
Alright, I could still implement that if you would like

mr-cheffy:
That would be great! We wouldn't need the CSS changes then, sorry for the inconvenience.

connosssss:
Sorry for the late response, I ended up fully rebuilding and was trying to better understand how rebasing works but now it opens up the full screen video in a new full tab like Arc does

src/zen/glance/ZenGlanceManager.mjs
```diff
@@ -1414,6 +1418,23 @@ class nsZenGlanceManager extends nsZenDOMOperatedFeature {
...
+   * @param {Event} event - The MozDOMFullscreen:Entered event
+   */
+  onFullscreenEntered(event) {
+    const browser = this.#currentBrowser;
+
+    if (!browser) {
+      return;
+    }
+
+    if (event.target === browser || browser.contains(event.target)) {
```
mr-cheffy: Why `contains`?

mr-cheffy (changes requested):

src/zen/glance/ZenGlanceManager.mjs
connosssss: I used contains because I was thinking that when the page requests to be full screen, event.target would be the element in the page that wanted to full screen but not the browser itself

mr-cheffy:
Also, what should happen when glancing but the page was already fullscreen before?

connosssss:
I was thinking that it would just work like normally, but also when you say the page was already fullscreen before I'm imagining like the browser being fullscreen with F11 so I think I also might have misunderstood it

mr-cheffy:
It should be fine

src/zen/glance/ZenGlanceManager.mjs
mr-cheffy: Browsers can't have child elements, so please remove it

mr-cheffy (changes requested):

src/zen/glance/ZenGlanceManager.mjs
mr-cheffy:
```suggestion
    if (event.target === browser) {
```

mr-cheffy (changes requested):

mr-cheffy (approved):

PR #14279, opened 2026-06-18
gh-14044: Implement 'Add Route for Domain' in tab context menu

FlorianButz:


src/zen/space-routing/ZenSpaceRoutingManager.sys.mjs
```diff
@@ -18,6 +18,22 @@ class nsZenSpaceRoutingManager {
...
+   *
+   * @param {nsIDOMWindow} window - The browser window that just started up
+   */
+  onDelayedBrowserStartup(window) {
+    const element = window.MozXULElement.parseXULToFragment(`
+        <menuseparator/>
+        <menuitem id="context_zen-add-domain-to-routing"
+                  data-lazy-l10n-id="tab-context-zen-add-domain-to-sr"
+                  data-l10n-args='{"tabCount": 1}'
+                  command="cmd_contextZenAddDomainToRouting"/>
```
mr-cheffy: I'd add an event listener directly instead of a new command and remove it when the window unloads

src/zen/space-routing/ZenSpaceRoutingManager.sys.mjs
```diff
@@ -403,6 +419,44 @@ class nsZenSpaceRoutingManager {
...
+      return;
+    }
+
+    const newRoute = this.createNewRoute();
+    let routeReference = "";
+
+    if (selectedTabs.length == 1) {
+      newRoute.matchType = "contains";
+      routeReference = selectedTabs[0].linkedBrowser.currentURI.host.replace(
+        "www.",
```
mr-cheffy: This seems to be a bit hacky, no? What if the user has `test.www.google.com` or `test-www.google.com`

mr-cheffy (changes requested):

src/zen/space-routing/ZenSpaceRoutingManager.sys.mjs
```diff
@@ -16,6 +16,39 @@ class nsZenSpaceRoutingManager {
 
   constructor() {
     this.#readFromDisk();
+    this._addSelectedToRouting = this.onAddSelectedToRouting.bind(this);
```
mr-cheffy: No need to create a binding here?

src/zen/space-routing/ZenSpaceRoutingManager.sys.mjs
```diff
@@ -16,6 +16,39 @@ class nsZenSpaceRoutingManager {
...
+    const element = window.MozXULElement.parseXULToFragment(`
+        <menuseparator/>
+        <menuitem id="context_zen-add-domain-to-routing"
+                  data-lazy-l10n-id="tab-context-zen-add-domain-to-sr"
+                  data-l10n-args='{"tabCount": 1}'/>
+      `);
+    window.document.getElementById("context_undoCloseTab")?.after(element);
+    window.document
+      .getElementById("context_zen-add-domain-to-routing")
+      .addEventListener("command", this._addSelectedToRouting);
```
mr-cheffy: Can just make this an anonymous function and call onAddSelectedToRouting from it. also, it needs to be unloaded

mr-cheffy (changes requested):

src/zen/space-routing/ZenSpaceRoutingManager.sys.mjs
FlorianButz: I don't think it has to be unloaded, since the element is removed in the dom when the window is closed anyway

src/zen/space-routing/ZenSpaceRoutingManager.sys.mjs
```diff
@@ -16,6 +16,39 @@ class nsZenSpaceRoutingManager {
...
+   * @param {nsIDOMWindow} window - The browser window that just started up
+   */
+  onDelayedBrowserStartup(window) {
+    const element = window.MozXULElement.parseXULToFragment(`
+        <menuseparator/>
+        <menuitem id="context_zen-add-domain-to-routing"
+                  data-lazy-l10n-id="tab-context-zen-add-domain-to-sr"
+                  data-l10n-args='{"tabCount": 1}'/>
+      `);
+    window.document.getElementById("context_undoCloseTab")?.after(element);
```
mr-cheffy: No need for the `?.` operator


```suggestion
    window.document.getElementById("context_undoCloseTab").after(element);
```

src/zen/space-routing/ZenSpaceRoutingManager.sys.mjs
mr-cheffy: No need to keep a new binding for this here, can just be binded when adding the event listener

src/zen/space-routing/ZenSpaceRoutingManager.sys.mjs
```diff
@@ -16,6 +16,39 @@ class nsZenSpaceRoutingManager {
...
+   * tabs to a new route is pressed
+   *
+   * @param {Event} event
+   */
+  onAddSelectedToRouting(event) {
+    const window = event.target.documentGlobal;
+    const tabs = window.TabContextMenu.contextTab.multiselected
+      ? window.gBrowser.selectedTabs
+      : [window.TabContextMenu.contextTab];
+    gZenSpaceRoutingManager.addRouteForSelected(tabs, window);
```
mr-cheffy: Shouldn't it be `this`?

mr-cheffy (changes requested):

src/zen/space-routing/ZenSpaceRoutingManager.sys.mjs
FlorianButz: Ah true, I just copied it over from the old location and forgot to change it

src/zen/space-routing/ZenSpaceRoutingManager.sys.mjs
```diff
@@ -403,6 +435,41 @@ class nsZenSpaceRoutingManager {
...
   }
 
+  /**
+   * Adds a new route for all given tabs
+   *
+   * @param {Array<object>} selectedTabs - The tabs that should be routed
+   * @param {Window} parentWindow - The window from which this is being executed
+   */
+  addRouteForSelected(selectedTabs, parentWindow) {
+    if (!selectedTabs) {
```
mr-cheffy: No need for this if statement

mr-cheffy (changes requested):

mr-cheffy (approved):

PR #14305, opened 2026-06-20
fix: Select renamed tab label from the start instead of caret-at-end

daberni:
Instead of starting of with the caret at the very end when editing a tab label, now the full text gets selected directly and the label isn't scrolled to the end automatically, effectively showing it the same way as in non-edit mode

### DIFF
--- a/src/zen/common/modules/ZenUIManager.mjs
+++ b/src/zen/common/modules/ZenUIManager.mjs
@@ -1708,7 +1708,11 @@ window.gZenVerticalTabsManager = {
       this._tabEdited.after(input);
     }
     input.focus();
-    input.select();
+    // Select all, but keep the viewport anchored at the start (show the
+    // beginning of long labels, like the non-editing display) instead of the
+    // caret-at-end that input.select() leaves.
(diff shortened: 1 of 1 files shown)


mr-cheffy (changes requested):
Please remove the useless comments and link the PR to an issue, like the rest of PRs in this repo. Same goes to all your other PRs. Also, please re-open these without claude being an author of the commits.

daberni:
> Please remove the useless comments and link the PR to an issue, like the rest of PRs in this repo. Same goes to all your other PRs. Also, please re-open these without claude being an author of the commits.

Will strip the comments a bit where overly verbose.

@mr-cheffy  I honestly wasn't sure about the desired issue-linking practice, as many other PRs do not reference any existing issue either?
Should I create an issue for each fix/improvement first? I happily follow your desired approach here. A CONTRIBUTE.md might could clear this up for the future.

mr-cheffy:
See https://github.com/zen-browser/desktop/pull/14293 as an example, just add the issue to the title

PR #14307, opened 2026-06-20
fix: Stop emoji picker flickering and reset to the Emojis page reliably

daberni:
The picker reset its page with a smooth scrollIntoView while the popup was
hiding, and positioned the page before the popup was laid out, so it could
animate/jump on open. Reset instantly on popupshown (after layout) instead,
and drop the on-hide scroll so there's no visible animation.


mr-cheffy:

PR #14308, opened 2026-06-20
fix: Target the right-clicked space for the icon picker and workspace actions

daberni:
The workspace actions menu resolved its target id from the actions button,
which doesn't carry zen-workspace-id, so the icon picker and actions hit the
active space instead of the right-clicked one. Resolve the id from the
nearest [zen-workspace-id] ancestor, and anchor the picker on whatever
triggered it (falling back to the target workspace's own indicator icon).


mr-cheffy:

PR #14309, opened 2026-06-20
fix: Show pinned-tab scroll-shadow dividers based on scroll position

daberni:
The top/bottom scroll-shadow dividers were derived from the arrowscrollbox's
scrolledtostart/scrolledtoend, which it computes from the first/last scrollable
element's edge. With collapsed pinned tabs those edges are unreliable, so the
divider either flashed at the top (when pinned tabs were counted) or only
appeared once scrolled past the pinned/normal separator (when excluded).

Drive the dividers from the real scrollTop instead, via dedicated
zen-overflow-top/zen-overflow-bottom attributes: visible as soon as there is
hidden content past that edge, hidden only when actually at the start/end
(including collapsed pinned tabs sitting at the top). Zen hides the scroll
buttons, so the built-in attributes are unused otherwise.

### DIFF
--- a/src/zen/spaces/ZenSpace.mjs
+++ b/src/zen/spaces/ZenSpace.mjs
@@ -139,6 +139,15 @@ export class nsZenWorkspace extends MozXULElement {
     this.scrollbox.addEventListener("underflow", this);
     this.scrollbox.addEventListener("overflow", this);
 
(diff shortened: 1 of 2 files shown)


mr-cheffy:

PR #14310, opened 2026-06-20
fix: Don't let background navigations clobber the focused urlbar

daberni:


mr-cheffy:

PR #14311, opened 2026-06-20
feat: allow editing pinned tab URLs directly and grouping pinned URL options

daberni:
Currently its not possible to edit a pinned tab's URL directly. Zen only offers to apply the current URL ("Replace <...> URL with current). This makes it impossible to configure it a URL which is usually e.g. redirecting - but maybe thats intended.

For this reason I implemented a dedicated "Edit URL" option, which allows the editing directly.
As there would be 3 options for those URLs now, I decided to group them together in a sub menu


mr-cheffy:

PR #14318, opened 2026-06-21
gh-14317: Show pinned-tab scroll-shadow dividers based on scroll position

### LINKED ISSUE #14317: Workspaces scroll-shadow divider is shown when pinned tabs are collapsed
### Preliminary Checks

### What happened?

The workspace expand/collapse scroll-shadow divider in the sidebar tab list misbehaves when pinned tabs are collapsed: a false divider shows at the top when not scrolled, and while scrolling within the collapsed pinned area the divider only appears after passing the pinned/normal separator.

### Expected behavior

Divider is shown when there is actual content hidden due to the scroll position

### Actual behavior

Divider is always shown when pinned tabs are collapsed no matter the other tabs' scroll position

### Steps to reproduce

1. Pin some tabs and have enough tabs open, so the section is scrollable even when pinned tabs are collapsed.
2. Collapse the pinned tabs.

### Screenshots and videos

Valid Divider shown (pinned tabs are scrolled):
Image

Divider shown unnecessarily (nothing is scrolled, unpinned tabs are scrolled to the top anyway):
Image

### Version

1.21.3b

macOS - aarch64

Workspaces

```shell

```

daberni:
The top/bottom scroll-shadow dividers were derived from the arrowscrollbox's scrolledtostart/scrolledtoend, which it computes from the first/last scrollable element's edge. With collapsed pinned tabs those edges are unreliable, so the divider either flashed at the top (when pinned tabs were counted) or only appeared once scrolled past the pinned/normal separator (when excluded).

Drive the dividers from the real scrollTop instead, via dedicated
zen-overflow-top/zen-overflow-bottom attributes: visible as soon as there is hidden content past that edge, hidden only when actually at the start/end (including collapsed pinned tabs sitting at the top), recomputed on scroll and on overflow/underflow. Zen hides the scroll buttons, so scrolledtostart/scrolledtoend are unused otherwise.

### DIFF
--- a/src/toolkit/content/widgets/arrowscrollbox-js.patch
+++ b/src/toolkit/content/widgets/arrowscrollbox-js.patch
@@ -1,5 +1,5 @@
 diff --git a/toolkit/content/widgets/arrowscrollbox.js b/toolkit/content/widgets/arrowscrollbox.js
-index b80d1049bb6ae305f2ac9c4c35fe975fd508031c..574149bffa49329e927c8db9db0c080eb6b87f5f 100644
+index b80d1049bb6ae305f2ac9c4c35fe975fd508031c..e64199362eb9ef94a6f78dbf26bce40c4113cecb 100644
 --- a/toolkit/content/widgets/arrowscrollbox.js
 +++ b/toolkit/content/widgets/arrowscrollbox.js
 @@ -98,6 +98,7 @@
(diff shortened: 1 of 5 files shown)


mr-cheffy (changes requested):
Now we are calculating on every scroll twice, I think we should rather fix the core issue instead of a workaround

daberni:
> Now we are calculating on every scroll twice, I think we should rather fix the core issue instead of a workaround

@mr-cheffy I do not fully understand your request. Where do we calculate on every scroll twice? The `[scrolledtostart]` and `[scrolledtoend]` appear to stem from the native firefox engine. But they do not consider the zen-only pinned tabs container.
When pinned tabs are collapsed, it shifts the whole pinned block (including the space-fake-collapsible-start start marker) out of the scroll flow — its measured top is −37 while the scrollbox top is 127, even at scrollTop: 0.

This is exactly what is getting addressed with this PR now

mr-cheffy:
It's calculated at https://searchfox.org/firefox-main/source/toolkit/content/widgets/arrowscrollbox.js#630-631, we should rather check why its ignoring the pinned tabs container

daberni:
> we should rather check why its ignoring the pinned tabs container

Its not ignoring the pinned tabs container at all. It just looks for the very first child of the workspace, which is the fake element of the collapsed tabs. And collapsing works by setting a negative margin on exactly that first child.

Therefore the first child, is outside of the viewport and indicates that the content is scrolled out.

@mr-cheffy I now followed a different approach in patching the scroll position determination (only) for workspace scroll containers now

mr-cheffy:
See comment in issue for explanation

PR #14321, opened 2026-06-21
gh-14320: fix workspace change icon picker state when opening

### LINKED ISSUE #14320: Workspace icon picker opening with weird animation
### Preliminary Checks

### What happened?

When I open the workspaces "Change Icon"-picker multiple times, while switching to the "built in icons" tab, when opening a second time the picker is getting reset to "emojis" with a weird animation.

### Expected behavior

Change Icon picker is getting opened in the correct state right away without animations from previous state

### Actual behavior

Emoji picker switches to default emoji tab with weird animation

### Steps to reproduce

1. Open Workspace Change Icon Picker
2. Switch to "Icons" Tab
3. Close Icon Picker
4. Open Workspace Change Icon Picker again

Previous "Icons" Tab content is shown for a brief moment before "Emojis" are shown

### Screenshots and videos

https://github.com/user-attachments/assets/d9c7986e-5fc5-4ccd-b14c-802aba8cd584

### Version

1.21.3b 

macOS - aarch64

Workspaces

```shell

```

daberni:
The picker reset its page with a smooth scrollIntoView while the popup was hiding, and positioned the page before the popup was laid out, so it could animate/jump on open. Reset instantly on popupshown (after layout) instead, and drop the on-hide scroll so there's no visible animation.

### DIFF
--- a/src/zen/common/emojis/ZenEmojiPicker.mjs
+++ b/src/zen/common/emojis/ZenEmojiPicker.mjs
@@ -6,28 +6,28 @@ import { nsZenDOMOperatedFeature } from "chrome://browser/content/zen-components
 
 // prettier-ignore
 const SVG_ICONS = [
-    "airplane.svg", "american-football.svg", "baseball.svg", "basket.svg", 
-    "bed.svg", "bell.svg", "bookmark.svg", "book.svg", 
-    "briefcase.svg", "brush.svg", "bug.svg", "build.svg", 
-    "cafe.svg", "call.svg", "card.svg", "chat.svg", 
-    "checkbox.svg", "circle.svg", "cloud.svg", "code.svg", 
-    "coins.svg", "construct.svg", "cutlery.svg", "egg.svg", 
-    "extension-puzzle.svg", "eye.svg", "fast-food.svg", "fish.svg", 
-    "flag.svg", "flame.svg", "flask.svg", "folder.svg", 
-    "game-controller.svg", "globe-1.svg", "globe.svg", "grid-2x2.svg", 
-    "grid-3x3.svg", "heart.svg", "ice-cream.svg", "image.svg", 
-    "inbox.svg", "key.svg", "layers.svg", "leaf.svg", 
-    "lightning.svg", "location.svg", "lock-closed.svg", "logo-rss.svg", 
-    "logo-usd.svg", "mail.svg", "map.svg", "megaphone.svg", 
-    "moon.svg", "music.svg", "navigate.svg", "nuclear.svg", 
-    "page.svg", "palette.svg", "paw.svg", "people.svg", 
-    "pizza.svg", "planet.svg", "present.svg", "rocket.svg", 
-    "school.svg", "shapes.svg", "shirt.svg", "skull.svg", 
-    "squares.svg", "square.svg", "star-1.svg", "star.svg", 
-    "stats-chart.svg", "sun.svg", "tada.svg", "terminal.svg", 
-    "ticket.svg", "time.svg", "trash.svg", "triangle.svg", 
-    "video.svg", "volume-high.svg", "wallet.svg", "warning.svg", 
-    "water.svg", "weight.svg", 
+    "airplane.svg", "american-football.svg", "baseball.svg", "basket.svg",
+    "bed.svg", "bell.svg", "bookmark.svg", "book.svg",
+    "briefcase.svg", "brush.svg", "bug.svg", "build.svg",
+    "cafe.svg", "call.svg", "card.svg", "chat.svg",
+    "checkbox.svg", "circle.svg", "cloud.svg", "code.svg",
+    "coins.svg", "construct.svg", "cutlery.svg", "egg.svg",
+    "extension-puzzle.svg", "eye.svg", "fast-food.svg", "fish.svg",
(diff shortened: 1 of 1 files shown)


src/zen/common/emojis/ZenEmojiPicker.mjs
```diff
@@ -6,28 +6,28 @@ import { nsZenDOMOperatedFeature } from "chrome://browser/content/zen-components
...
+    "logo-usd.svg", "mail.svg", "map.svg", "megaphone.svg",
+    "moon.svg", "music.svg", "navigate.svg", "nuclear.svg",
+    "page.svg", "palette.svg", "paw.svg", "people.svg",
+    "pizza.svg", "planet.svg", "present.svg", "rocket.svg",
+    "school.svg", "shapes.svg", "shirt.svg", "skull.svg",
+    "squares.svg", "square.svg", "star-1.svg", "star.svg",
+    "stats-chart.svg", "sun.svg", "tada.svg", "terminal.svg",
+    "ticket.svg", "time.svg", "trash.svg", "triangle.svg",
+    "video.svg", "volume-high.svg", "wallet.svg", "warning.svg",
+    "water.svg", "weight.svg",
```
daberni: trailing whitespace cleanup happened during formatting.

can revert this if it introduced too much noice

src/zen/common/emojis/ZenEmojiPicker.mjs
```diff
@@ -199,14 +206,19 @@ class nsZenEmojiPicker extends nsZenDOMOperatedFeature {
     }
   }
 
+  #onPopupShown(event) {
```
mr-cheffy, line 206: This shouldn't be needed anymore as we are already setting it when opening

mr-cheffy (changes requested):

src/zen/common/emojis/ZenEmojiPicker.mjs
daberni: Sadly its needed right now:

The resetting via `open()` actually does not have any effect yet - we could drop it. In the moment the popup gets restored, the browser is automatically setting the focus into the `searchInput`, causing it scrolling into view. We would have to tackle this independently if this is not intended? (I tried dropping the `setTimeout()`-`focus()` block, but it didn't have any effect, ~the browser is still autofocusing~)

So after(!) that auto-focusing is happening (`#onPopupShown`), we have to apply the scrolling without animation directly

daberni: @mr-cheffy I now followed a different approach and align the focusing of the `searchInput` right away. the `setTimeout` looked like a workaround anyway, so now the focus+scrolling is both now only done in `#onPopupShown`

another option would be to just omit the `this.#changePage(false, { animate: false });` call, then we would restore the previous selected tab (which kinda doesn't sound too bad either) - because then we do not have any issues with animations anyway. setting the focus here manually is still necessary, but doesn't have any impact on the tab-selection

src/zen/common/emojis/ZenEmojiPicker.mjs
```diff
@@ -199,14 +203,23 @@ class nsZenEmojiPicker extends nsZenDOMOperatedFeature {
...
+    if (event.target !== this.#panel) {
```
mr-cheffy, line 207: Why would popupshown be called for a different target?

mr-cheffy (changes requested):
Please remember to run `npm run lint:fix`

src/zen/common/emojis/ZenEmojiPicker.mjs
mr-cheffy: Ok, looks good!

src/zen/common/emojis/ZenEmojiPicker.mjs
mr-cheffy: Looks good, thanks

src/zen/common/emojis/ZenEmojiPicker.mjs
daberni: just followed the same safeguard of `#onPopupShowing()` here tbh

daberni:
> Please remember to run `npm run lint:fix`

Nothing popped up at my side, did I miss something?

mr-cheffy:
Could you show the output of the command?

It may already be linted correctly though. Lets wait for https://github.com/zen-browser/desktop/actions/runs/27904838544/job/82595141046?pr=14321

daberni:
> Could you show the output of the command?

```shell
~/Developer/other/zen-desktop  fix/emoji-picker-page-flicker $ nvm use
Found '/Users/bd/Developer/other/zen-desktop/.nvmrc' with version <22>
Now using node v22.22.2 (npm v10.9.7)
~/Developer/other/zen-desktop  fix/emoji-picker-page-flicker $ npm run lint:fix

> zen-desktop@1.0.0 lint:fix
> npm run lint -- --fix


> zen-desktop@1.0.0 lint
> cd engine && ./mach lint zen --fix

ERROR: Unexpected result type received: <class 'NoneType'>
ERROR: Unexpected result type received: <class 'NoneType'>
✖ 0 problems (0 errors, 0 warnings, 31 fixed)
```

The same 2 errors appear for me on the `dev` branch too.

mr-cheffy (approved):
Looks good, thanks!

PR #14324, opened 2026-06-21
gh-11245: Fix tab placement with Move Tab / Space Routing

### LINKED ISSUE #11245: "Change Tab to Workspace" does not respect "Move the new tab button to the top" setting
### Preliminary Checks

### What happened?

When using "Change Tab(s) to Workspace" from the right-click menu, the position of the tab in the workspace it was moved to will always be at the bottom, even when "Move the new tab button to the top" is enabled in settings.

When "Move the new tab button to the top" option is enabled, new tabs always open at the top of the tab list, even when opened by other means like Ctrl+T or clicking the Settings cog. The top of the tab list becomes the top of the "inbox" formed by a vertical tab bar. Moving tabs between workspaces doesn't respect this pattern. This makes for a very annoying UX issue because the only way to move tabs from the bottom to the top of a long list of tabs is to drag them all the way 

### Expected behavior

If the new tab button is on the top then the usual logical place for tabs to go when using "change tabs to workspace" first should be at the top.

### Actual behavior

Tabs go on the bottom no matter what setting you have

### Steps to reproduce

1. Create two workspaces and open some tabs in both of them.
2. Go to Settings ->Look and Feel -> Check "Move the new tab button to the top"
3. Right click on some tabs in one space and go to "Change Tab(s) to Workspace" and select the other space
4. Observe that the tabs always go on the bottom, regardless of the state of "Move the new tab button to the top"

### Screenshots and videos

_No response_

### Version

1.17.6b

Windows - x64

Workspaces

```shell

```

FlorianButz:
Fixes #11245

### DIFF
--- a/src/zen/spaces/ZenSpaceManager.mjs
+++ b/src/zen/spaces/ZenSpaceManager.mjs
@@ -1506,19 +1506,23 @@ class nsZenWorkspaces {
         continue;
       }
 
+      const newtabPlacement =
+        Services.prefs.getBoolPref("zen.view.show-newtab-button-top", false);
+      const insertElement = newtabPlacement ? container.firstChild : container.lastChild;
+
       if (container) {
         if (tab.group?.hasAttribute("split-view-group")) {
           gBrowser.zenHandleTabMove(tab.group, () => {
             for (const subTab of tab.group.tabs) {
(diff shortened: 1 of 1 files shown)


mr-cheffy (approved):
We should have tests for these but thanks! <3

PR #14327, opened 2026-06-21
gh-9186: Drop a tab onto a space switcher icon to move it there

daberni:
Moves the dragged tab(s) to the dropped space at the new-tab position. Placement (top/bottom) is handled by moveTabsToWorkspace now (gh-14324)

### DIFF
--- a/src/zen/drag-and-drop/ZenDragAndDrop.js
+++ b/src/zen/drag-and-drop/ZenDragAndDrop.js
@@ -125,6 +125,10 @@
           "dragover",
           this.handle_spaceIconDragOver.bind(this)
         );
+        gZenWorkspaces.workspaceIcons.addEventListener(
+          "drop",
+          this.#handle_spaceIconDrop.bind(this)
+        );
       }
       window.addEventListener(
         "dragleave",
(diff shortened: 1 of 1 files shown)


mr-cheffy:
This looks heavily vibecoded, sorry.

daberni:
> This looks heavily vibecoded, sorry.

@mr-cheffy Sure I use AI for development, but whats the actual issue with the feature? Its not like I did a oneshot PR, its reviewed and heavily sanitized already

mr-cheffy:
Its a bit more tricky than what you showcased here, I think it would require some more effort that I simply can't provide right now. In the future, maybe we can get back to this feature, but for now, lets just leave it unplanned

daberni:
Thanks for the response, maybe you can shine some light on your concerns quickly and I would be happy to address them. I really would like to contribute such small additions as long as its feasible.

From testing perspective this already looked really good, including the applying of the container when one is assigned to a workspace already, switching to the dropped workspace directly (hence the cancellation of the timer) etc.

I was only unsure about the `if (draggedTab.ownerGlobal === window)` part, the rest appears quite straight forward tbh

PR #14329, opened 2026-06-21
gh-12153: edit pinned-page url directly via submenu (gh-14329)

daberni:
Adressing idea/feature-request of https://github.com/zen-browser/desktop/discussions/12153

New menu structure:
image

Dialog:
Screenshot 2026-06-21 at 20 08 26

### DIFF
--- a/src/zen/sessionstore/ZenWindowSync.sys.mjs
+++ b/src/zen/sessionstore/ZenWindowSync.sys.mjs
@@ -1223,38 +1223,42 @@ class nsZenWindowSync {
     return this.#maybeFlushTabState(aTab).finally(() => {
       this.log(`Setting pinned initial state for tab ${aTab.id}`);
       let { entries, index } = this.#getTabEntriesFromCache(aTab);
+      let image =
+        aTab.getAttribute("image") ||
+        aTab.documentGlobal.gBrowser.getIcon(aTab);
       let activeIndex = typeof index === "number" ? index : entries.length;
       // Tab state cache gives us the index starting from 1 instead of 0.
       activeIndex--;
       activeIndex = Math.min(activeIndex, entries.length - 1);
       activeIndex = Math.max(activeIndex, 0);
       let entryToUse = (entries[activeIndex] || entries[0]) ?? null;
-      this.#setPinnedInitialState(aTab, {
-        url: entryToUse?.url,
-        title: entryToUse?.title,
-      });
+      this.#setPinnedInitialState(
+        aTab,
+        { url: entryToUse?.url, title: entryToUse?.title },
+        image
+      );
     });
   }
 
   /**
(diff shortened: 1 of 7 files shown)


src/zen/sessionstore/ZenWindowSync.sys.mjs
```diff
@@ -1223,31 +1223,46 @@ class nsZenWindowSync {
...
+    this.log(`Setting pinned url for tab ${aTab.id}`);
+    this.#setPinnedInitialState(aTab, {
+      url: aUrl,
+      title: aTab._zenPinnedInitialState?.entry?.title ?? aTab.label,
+    });
+  }
+
+  #setPinnedInitialState(aTab, aEntry) {
+    const image =
+      aTab.getAttribute("image") || aTab.documentGlobal.gBrowser.getIcon(aTab);
```
mr-cheffy: Shouldn't we be using a cached favicon if the url has been manually entered? I assume you can add any site and hence, there a chance that the favicon is wrong

mr-cheffy (changes requested):
Please run `npm run lint:fix` and I would like a test for this. But looks good!

src/zen/sessionstore/ZenWindowSync.sys.mjs
daberni: thats actually a good point, didn't think about it because my usual workflow just covers changing the path, but not the domain.

I will have a look on where to get the cached icon from, or do you already have a hint on that?

That potentially makes the extracted `#setPinnedInitialState` helper obsolete again, but I will look into it 👍

mr-cheffy: I believe you can use `PlacesUtils.favicons.getFaviconForPage` for that, it should only be used when the url is manually added, really. I dont see why it would make it obsolete though? You can just keep it

src/zen/tabs/ZenPinnedTabManager.mjs
```diff
@@ -246,6 +246,66 @@ class nsZenPinnedTabManager extends nsZenDOMOperatedFeature {
...
+      ).preferredURI;
+    } catch (_) {}
+    if (!uri) {
+      gZenUIManager.showToast("zen-pinned-tab-url-invalid");
+      return;
+    }
+    const url = uri.spec;
+
+    // Skip when the value wasn't actually changed from what was prefilled.
+    if (url === initialUrl) {
```
mr-cheffy:
```suggestion
    if (!url || url === initialUrl) {
```

mr-cheffy (approved):

mr-cheffy:
Could you please run `npm run lint:fix`?

mr-cheffy (approved):

mr-cheffy (approved):

PR #14343, opened 2026-06-22
no-bug: Fix glance tab positioning and prevent split-view (gh-14343)

octaviusz:
- Fixed: If you try to open a link in a new tab from Glance, it opens in the wrong place
- Fixed: Exclude glance tabs from split-view targets


mr-cheffy (approved):

mr-cheffy (approved):
<3

PR #14357, opened 2026-06-24
gh-9918: Fixes issue setting keybinds with shift + digit or symbol

### LINKED ISSUE #9918: Keyboard shortcuts like Ctrl+Shift+NumberRowKey depend on the active keyboard layout
### Preliminary Checks

### What happened?

I tried binding workspace switching to `Ctrl+Shift+1`, `Ctrl+Shift+2`, `Ctrl+Shift+3` (*it would be nice if these were the default workspace keybinds, by the way*), and so on. However, unlike other `Ctrl+Shift` shortcuts (such as `Ctrl+Shift+AnyLetter`), these combinations are shown as `Ctrl+Shift+!`, `Ctrl+Shift+@`, etc.

The main problem is that such shortcuts don’t work when a non-English keyboard layout is selected. For example, in the Russian layout, `Ctrl+Shift+3` produces `Ctrl+Shift+№` instead of `Ctrl+Shift+#`.

### Expected behavior

Shortcuts like `Ctrl+Shift+NumberRowKey` should work regardless of keyboard layout. Also such combinations should be shown as `Ctrl+Shift+1`, `Ctrl+Shift+2`, etc. instead of `Ctrl+Shift+!`, `Ctrl+Shift+@`, etc.

### Actual behavior

Such shortcuts work only with the keyboard layout that was active when they were created.

### Steps to reproduce

1. Open `about:settings`.
2. Go to **Keyboard Shortcuts**.
3. Bind any action to `Ctrl+Shift+NumberRowKey` (for example, `Ctrl+Shift+2`) while using the English layout.
4. Switch to the Russian layout (or any other layout where the number row produces different symbols).
5. Try pressing the shortcut – nothing happens.

### Screenshots and videos

_No response_

### Version

1.14.11b

Windows - x64

Keyboard Shortcuts

```shell

```

### LINKED ISSUE #13575: binding ctrl + shift + num to switching workspaces does not work
### Preliminary Checks

### What happened?

I have "Switch to Workspace 1" as ctrl + shift + 1

Image

Pressing this keybind does not work

I have tried binding it to shift + 0, this also does not work
I have tried binding it to q, this worked
I have tried binding it to ctrl + q, this worked

I know this working around janurary.

### Expected behavior

pressing ctrl + shift + 1 switches to workspace 1, ctrl + shift  + 2 switches to workspace 2, etc.

### Actual behavior

No visibile change

### Steps to reproduce

Keybind move to workspace 1 as ctrl + shift + 1
Move to new tab not in workspace 1 and press the keybind ctrl + shift + 1

### Screenshots and videos

Image

### Version

1.19.11b (Firefox 150.0.1) (64-bit)

Windows - x64

Workspaces

```shell

```

connosssss:
Fixes #13575 

Before there were issues with setting keybinds to anything where it was shift + a digit or symbol that gets changed when shift is pressed (issue was around ctrl+shift+1 where the 1 becomes an !)

- When writing the keybind to XUL, it now swaps the key to the shifted character and removes the Shift modifier so XUL can match and run it properly
- This way it also keeps the original keybind internally, so the settings menu still shows Ctrl+Shift+1 instead of Ctrl+!


mr-cheffy (changes requested):
Could you please run `npm run lint:fix`? Also, in what OS has this been tested on?

connosssss:
I just ran it and I tested it on Windows

mr-cheffy:
Maybe we should lock it only for windows, because I dont think this issue exists on other OSs

connosssss:
Now it is locked to just windows but the shifted keys map is still at the top of the class and not affecting anything while not windows. I thought it was a good place for it but I could also move it if you'd like

mr-cheffy (approved):

mr-cheffy (approved):

Issue #14330, opened 2026-06-21
Change icon on different workspace is sometimes changing the current workspace's icon


daberni:
### Preliminary Checks

- [x] I have read and understood the important section above.
- [x] I have searched existing issues and avoided creating duplicates.
- [x] I am not filing an enhancement request.
- [x] I have checked that this issue cannot be reproduced on Mozilla Firefox.
- [x] I have checked that this issue can be reproduced once I removed all my Mods and Custom CSS.

### What happened?

I do not have a full reproduction, but sometimes its happening that when wanting to change the Icon of a non-active workspace via the context menu, after selecting, the current workspace's icon is getting replaced, instead of the targets one.

It actually starts with opening the Icon Picker in the wrong location, but I also already had it, that its opening at the bottom, but for the wrong workspace.

### Expected behavior

The icon of the right-click targeted workspace should be changed

### Actual behavior

Sometimes the icon of the current workspace is getting changed

### Steps to reproduce

~not 100% reproducible:~

1. right click on workspace icon of non-active workspace
2. Workspace context menu opens
3. Hover over "Set Profile" item
4. Profiles/Container selection is getting expanded
5. Click on Change Icon
6. Select different Icon
7. Click outside to lose focus and save

### Screenshots and videos

https://github.com/user-attachments/assets/d2561de7-1162-4fae-95a9-b5efa2af9984

### Version

1.21.3b 

### What platform are you seeing the problem on?

macOS - aarch64

### What component is this issue related to?

Workspaces

daberni:
### Analysis

The moment the Profile/Container selection submenu is getting opened and hidden again, the `"popuphidden"` event is raised, which runs in an ungated reset of the `#contextMenuData`, although the related popup is still open:

https://github.com/zen-browser/desktop/blob/f4781daa49fca33c62940074fdb3fb0becf38208/src/zen/spaces/ZenSpaceManager.mjs#L1076-L1082

Changing the icon then resolves to `activeWorkspace` as a fallback:

https://github.com/zen-browser/desktop/blob/f4781daa49fca33c62940074fdb3fb0becf38208/src/zen/spaces/ZenSpaceManager.mjs#L946-L947

PR #14381, opened 2026-06-27
gh-14330: Fixes wrong workspace context in the right-click actions menu

daberni:
Populate #contextMenuData from the popup's triggerNode (walking up to the nearest [zen-workspace-id]), and pass the triggering event to the programmatic openPopup in onActionsCommand so that path sets triggerNode too.

Clear #contextMenuData with a single synchronous, popup-scoped popuphidden handler, replacing the deferred double-setTimeout and the duplicate {once:true} clearer.


mr-cheffy:
Please dont let claude be author of the commit

daberni:
Oh damn, I overlooked that, reopened with proper commit message now https://github.com/zen-browser/desktop/pull/14384

PR #14391, opened 2026-06-27
gh-14390: fix profile picker menu when creating new workspace

### LINKED ISSUE #14390: Profile picker for new Workspace is first unresponsive and then lists items twice
### Preliminary Checks

### What happened?

When creating a new Workspace and wanting to assign the workspace to a Profile/Container:

1. The first click(s) are not responding, taking some time until the profile dropdown opens
2. The dropdown has all configured profiles listed twice:

### Expected behavior

Responding on the first click and only listing items once

### Actual behavior

Takes several clicks and listing items twice

### Steps to reproduce

1. "Create Space"
2. Click on "Default" (Profile Selector) - nothing happens
3. Click on "Default" again once more - menu opens

### Screenshots and videos

Image

### Version

1.21.4b

macOS - aarch64

Workspaces

```shell

```

daberni:
Profile menu now opens directly and lists profiles only once

### DIFF
--- a/src/zen/spaces/ZenSpaceCreation.mjs
+++ b/src/zen/spaces/ZenSpaceCreation.mjs
@@ -182,8 +182,8 @@ class nsZenWorkspaceCreation extends MozXULElement {
         this.onProfileCommand.bind(this)
       );
       this.profilesPopup.addEventListener(
-        "popupshown",
-        this.onProfilePopupShown.bind(this)
+        "popupshowing",
+        this.onProfilePopupShowing.bind(this)
       );
       this.profilesPopup.addEventListener(
         "command",
@@ -192,7 +192,7 @@ class nsZenWorkspaceCreation extends MozXULElement {
 
       this.currentProfile = {
         id: 0,
-        name: "Default",
+        name: ContextualIdentityService.formatContextLabel("user-context-none"),
       };
     } else {
       this.inputProfile.parentNode.hidden = true;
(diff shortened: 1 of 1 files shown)


src/zen/spaces/ZenSpaceCreation.mjs
```diff
@@ -192,7 +192,7 @@ class nsZenWorkspaceCreation extends MozXULElement {
 
       this.currentProfile = {
         id: 0,
-        name: "Default",
+        name: ContextualIdentityService.formatContextLabel("user-context-none"),
```
mr-cheffy, line 195: `user-context-none` doesn't exist?

mr-cheffy (changes requested):

mr-cheffy:
We dont exclude profiles on purpose, ill close these as invalid

src/zen/spaces/ZenSpaceCreation.mjs
daberni: Sure it does? its part of the engine already and resolves to the usual `No Container`: 
[https://github.com/mozilla-l10n/firefox-l10n-source/blob/466c89c1f73b2707d2c21d5d91b2c46c08c9608a/toolkit/toolkit/global/contextual-identity.ftl#L24-L26](https://github.com/mozilla-l10n/firefox-l10n-source/blob/466c89c1f73b2707d2c21d5d91b2c46c08c9608a/toolkit/toolkit/global/contextual-identity.ftl#L24-L26)

daberni:
> We dont exclude profiles on purpose, ill close these as invalid

@mr-cheffy that comment only applies to one line of the PR — the rest is the actual bug fix and is unrelated to profile exclusion.

The main fix was changing `popupshown` → `popupshowing`.

I changed the `excludeUserContextId` line just for consistency, as the active profile/container is excluded everywhere else in Zen's container menus, so I matched that here. If that's intentional in this spot, no problem at all — I'll happily drop just that line.

A one-line "please remove the exclude" would have been enough to land the fix; instead closing the whole PR and the issue as invalid loses the actual repro fix too. Could we reopen? I'll push a version without the exclusion so it's purely the popupshowing fix.

I genuinely just want to fix bugs and contribute meaningful additions here. Fast rejections like this — over something that's one review comment away from mergeable — make that pretty discouraging.

PR #14410, opened 2026-06-29
fix (issue-14355): added cooldown for copy url shortcut

GodKode69:
Added a 1 second timer to "cmd_zenCopyCurrentURL" and "cmd_zenCopyCurrentURLMarkdown" stopping the spammed copy to clipboard command.


mr-cheffy:
We don't throttle these on purpose, its bad UX

PR #14412, opened 2026-06-29
gh-12948: Removed ai related stuff from settings search

### LINKED ISSUE #12948: AI controls not shown in settings by default but only when searching for them
### Preliminary Checks

### What happened?

The *AI controls* panel is not shown in the settings of Zen by default. To access them, users have to use the search bar. In Firefox, the table of contents (TOC), which is shown on the left of the settings page, includes the *AI controls* such that users can access them easily.

This is inconvenient and espacially impractical for users who don't know about these settings. Given the fact that these settings are hidden, I'm not even sure if they actually do anything. My opinion: 
- If the AI settings work and non-technical users should be able to change them, the settings should be shown by default. 
- If the AI settings are "experimental"/don't work as in Firefox, then maybe they shouldn't be available via the search bar but only via about:config? Or add a warning to the *AI controls* panel so users know about potential problems.

### Expected behavior

*AI Control* panel should be included in the settings' TOC on the left.

### Actual behavior

*AI Control* panel is hidden and can only be accessed via the settings' search bar.

### Steps to reproduce

open Zen settings, et voilà

### Screenshots and videos

Zen settings:
Image

Image

Firefox settings:
Image

### Version

1.19.4b

Windows - x64

Settings

```shell

```

GodKode69:
updated preferences-js.patch to set ai related stuff manually to false and added a new  preference in ai.yaml to set its visibility as false by default

image


src/browser/components/preferences/preferences-js.patch
```diff
@@ -10,7 +10,17 @@ index 57add34d876fb885275f1147209c6fbeee367a7c..be0ab43b299317c0022a5e719f47a070
        ],
        resizeCallback: async ({ title, frame }) => {
          // Search within main document and highlight matched keyword.
-@@ -437,6 +438,8 @@ const CONFIG_PANES = Object.freeze({
+@@ -255,8 +256,7 @@ const CONFIG_PANES = Object.freeze({
+     iconSrc: "chrome://global/skin/icons/highlights.svg",
+     groupIds: ["aiControlsDescription", "aiFeatures", "aiStatesDescription"],
+     module: "chrome://browser/content/preferences/config/aiFeatures.mjs",
+-    visible: () =>
+-      Services.prefs.getBoolPref("browser.preferences.aiControls", false),
```
mr-cheffy: We don't really need this? As the pref is already off.

mr-cheffy (commented):
You should also remove the `category-ai-features` selector from settings css

GodKode69:
@mr-cheffy the changes in the patch file are needed to force the yaml ruling for pre-existing profiles as that only affects the newly created profiels

mr-cheffy:
No? How are you deducing this? Please revert the patch file

GodKode69:
image

this is how the search results look without the patch changes

image

and this is how they look with the patch changes

if you wish i could revert the patch changes so the fix only applies to new profiles

mr-cheffy:
It doesn't make sense, maybe you didn't re-apply the change to the pref? You need to run `npm run import` again. Is it disabled in about:config?

GodKode69:
let me try npm run import once again, i only did npm run build

mr-cheffy (approved):
Looks better now!

Issue #14238, opened 2026-06-17
Space routing: external links open in last-active space instead of configured default after restart


TenBryBry2003:
### Preliminary Checks

- [x] I have read and understood the important section above.
- [x] I have searched existing issues and avoided creating duplicates.
- [x] I am not filing an enhancement request.
- [x] I have checked that this issue cannot be reproduced on Mozilla Firefox.
- [x] I have checked that this issue can be reproduced once I removed all my Mods and Custom CSS.

### What happened?

With space routing's default space for external links set to my personal space, external links only open in the correct space if the browser is currently running or was last closed while on that space. On a cold start where the previous session ended on a different space, external links open in that last active space instead of the configured Personal space. Routing seems to fall back to the restored/active space on startup rather than honoring the configured default. This may be tied to startup settings, I have `Open previous windows and tabs` enabled (_and `Continue where you left off` disabled_), so the browser restores to whatever space was last active.

### Expected behavior

External links should always open in the configured default external link space regardless of which space was active when the browser was last closed.

### Actual behavior

External links open in whichever space was active in the restored session, ignoring the configured default external link space.

### Steps to reproduce

1. In space routing settings, set the default space for external links to a specific space (e.g., Personal).
2. Switch to a different space (e.g., Work).
3. Close Zen.
4. Reopen Zen (with "Open previous windows and tabs" enabled, so it restores onto the other space).
5. Open an external link from another app.
6. Observe: the link opens in the restored space instead of the configured Personal space.

### Screenshots and videos

https://github.com/user-attachments/assets/ce1a7b76-3030-4dfd-bc6a-be27efcd8c47

The link is directed correctly in the first session, as the browser is already open. In the second one, however, the link is opened in the last opened space instead.

### Version

1.21.1b

### What platform are you seeing the problem on?

Windows - aarch64

### What component is this issue related to?

Other

### Relevant log output if applicable

```shell

```

PR #14414, opened 2026-06-29
gh-14238: external links not following space routing settings on initial browser load

ashvwinn:


src/browser/components/tabbrowser/content/tabbrowser-js.patch
```diff
@@ -100,16 +101,19 @@ index 08b5b56e069d038d72c87355920c4ce8a55ed805..87f41c8583c26f364530def5bd5d8333
        if (tabArgument && tabArgument.linkedBrowser) {
          remoteType = tabArgument.linkedBrowser.remoteType;
          initialBrowsingContextGroupId =
-@@ -760,6 +819,8 @@
+@@ -760,6 +819,11 @@
        this.tabpanels.appendChild(panel);
  
        let tab = this.tabs[0];
-+      gZenWorkspaces.handleInitialTab(tab, (!remoteType || remoteType === E10SUtils.PRIVILEGEDABOUT_REMOTE_TYPE) && !gZenUIManager.testingEnabled);
++      let uriToLoad = gBrowserInit.uriToLoadPromise;
```
mr-cheffy, line 107: Shouldn't we handle onbeforetab routing here? Because we might need to mutate the user context id

mr-cheffy (changes requested):

PR #14430, opened 2026-06-30
no-bug: show local and LAN dev URLs as a tab sublabel

Yash1hi:
- [x] Show the `host:port` of local/LAN dev servers as a sublabel under the tab title (Arc/Dia style)
- [x] Reuse the existing `.zen-tab-sublabel` element + `zen-show-sublabel` attribute — no new markup, CSS, or l10n strings
- [x] Gate behind a new `zen.tabs.show-localhost-url` preference (default **on**, live-toggleable from about:config)
- [x] Leave pinned tabs and live-folder items to their existing sublabel owners to avoid conflicts

### Motivation

As a web developer I usually have several `localhost` tabs open at once (`localhost:3000`, `localhost:8080`, a Storybook, an API, …). Their titles are often identical or generic ("React App", "Vite + React"), so the only way to tell them apart is to click in and read the address bar. Arc and Dia solve this by showing the local URL as a small subtitle under the tab title, and it's a genuinely nice quality-of-life touch for anyone doing local development.

This brings that behavior to Zen. When a tab points at a local or LAN address, its `host:port` is shown as a sublabel beneath the title; for everything else, nothing changes.

### How it works

The implementation is intentionally tiny because Zen already has all the pieces:

- Every tab already ships a `<label class="zen-tab-sublabel">`, and the `[zen-show-sublabel]` attribute already reveals it (`src/zen/tabs/zen-tabs/vertical-tabs.css`).
- The `zen-tab-sublabel` Fluent message already has an `*[other] { $tabSubtitle }` branch that renders any literal string, so no new localization entry is needed.
- `ZenLiveFoldersUI` already uses exactly this pattern to put custom text in a sublabel.

So the new `ZenLocalhostSublabel` module just listens for tab location changes (via `addTabsProgressListener`) and, for matching tabs, sets `zen-show-sublabel` + the sublabel arg — the same way live folders do.

**Matched hosts:** `localhost`, `*.localhost`, `*.local`, IPv4 loopback `127.0.0.0/8`, RFC 1918 ranges (`10/8`, `172.16/12`, `192.168/16`), link-local `169.254/16`, and IPv6 loopback/ULA/link-local (`::1`, `fc00::/7`, `fe80::/10`). Only `http`/`https` tabs are considered.

### Preference

`zen.tabs.show-localhost-url` (bool, default `true`). Flipping it in about:config updates all open tabs immediately.

### Notes / open questions

- **Default on vs off:** I've defaulted it on since it's unobtrusive (it only ever appears on local/LAN tabs) and that's the behavior I was after, but I'm very happy to default it off and/or surface a Settings-UI checkbox if maintainers prefer — just say the word.
- **Scope:** I included LAN/private ranges in addition to `localhost`, which is handy for hitting a dev server on another machine. Easy to narrow to `localhost`-only if that's preferred.
- **Pinned tabs / live folders:** these manage their own sublabel, so they're deliberately skipped.
- **Testing:** I haven't been able to do a full local Firefox build for this change; I'd appreciate a maintainer sanity-check on the build, and I'm glad to iterate. Manual behavior to verify: open `http://localhost:3000` → sublabel shows `localhost:3000`; open `https://example.com` → no sublabel; toggle the pref → sublabels appear/disappear live.

Happy to open a tracking issue or move this to a discussion first if that's the preferred process for new defaults.

### DIFF
--- a/prefs/zen/zen.yaml
+++ b/prefs/zen/zen.yaml
@@ -11,6 +11,9 @@
 - name: zen.tabs.rename-tabs
   value: true
 
+- name: zen.tabs.show-localhost-url
+  value: true # Show the host:port of local/LAN dev servers as a tab sublabel
+
 - name: zen.tabs.essentials.max
   value: 12
 
--- a/src/zen/common/ZenPreloadedScripts.js
+++ b/src/zen/common/ZenPreloadedScripts.js
@@ -23,6 +23,7 @@
     "chrome://browser/content/zen-components/ZenMediaController.mjs",
     "chrome://browser/content/zen-components/ZenGlanceManager.mjs",
     "chrome://browser/content/zen-components/ZenPinnedTabManager.mjs",
+    "chrome://browser/content/zen-components/ZenLocalhostSublabel.mjs",
     "chrome://browser/content/zen-components/ZenViewSplitter.mjs",
     "chrome://browser/content/zen-components/ZenFolders.mjs",
     "chrome://browser/content/zen-components/ZenEmojiPicker.mjs",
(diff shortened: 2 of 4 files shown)


mr-cheffy:
I personally dont think this is worth the overhead honestly and the subtitle should be reserved for more important usecases. Btw, did you use AI to do this pr?

Yash1hi:
My apologies, was trying to develop locally

PR #14473, opened 2026-07-04
gh-10749: Localize command palette actions, emoji picker, etc

### LINKED ISSUE #10749: Missing translations
### Preliminary Checks

### What happened?

Some string are just did not appear on Crowdin and cannot be translated

### Expected behavior

All strings should be translatable

### Actual behavior

Some string are just did not appear on Crowdin and cannot be translated

### Steps to reproduce

Be Crowdin translator and search for some strings:
"Search emojis" (from workspace icon changing menu)
All command bar actions like "Toggle Compact Mode" or "Manage Extensions"
"Unload Space" from space contect menu

### Screenshots and videos

Image

Image

Image

### Version

1.16.4b

Other

Other

```shell

```

InlitX:
Fixes #10749.

A few strings were still hardcoded in English, so they never showed up in
Crowdin and couldn't be translated. This moves them over to Fluent:

- The "Search emojis" placeholder in the emoji picker
  (`emojis-picker.inc`) now uses a `data-l10n-id`.
- The command bar actions ("Toggle Compact Mode", "Manage Extensions", etc.)
  in `ZenUBGlobalActions.sys.mjs` now use l10n ids, and the two dynamic ones
  ("Focus on", "Extension") in `ZenUBActionsProvider.sys.mjs` are resolved the
  same way.
- New strings go in `zen-general.ftl`, which is already in `crowdin.yml`, so
  no Crowdin changes and no other-language files touched.

While I was in there I also localized the "Default" container label in the
new-space dialog (`ZenSpaceCreation.mjs`) — it was hardcoded too. It now reuses
Firefox's own `user-context-none` string, so it matches what the container
dropdown already shows.

One thing worth noting: the action `commandId`s are built from the stable l10n
ids, not the labels, so they stay exactly the same and don't break the learned
ordering in `ZenUBResultsLearner`. The English text is unchanged, so nothing
looks different — the strings just become translatable. "Unload Space" from the
issue is already localized in code, so I left it alone.

Tested on a local build: opened the emoji picker and the command bar, checked
every label shows up translated, and made sure the actions still fire after the
id change (compact mode, new window, etc.). `mach lint zen` is clean.


src/zen/urlbar/ZenUBActionsProvider.sys.mjs
```diff
@@ -27,6 +27,10 @@ ChromeUtils.defineESModuleGetters(lazy, {
   UrlUtils: "resource://gre/modules/UrlUtils.sys.mjs",
 });
 
+ChromeUtils.defineLazyGetter(lazy, "l10n", () => {
+  return new Localization(["browser/zen-general.ftl"], true);
```
mr-cheffy: We should move all the actions into its own separate translation file

mr-cheffy (changes requested):

src/zen/spaces/ZenSpaceCreation.mjs
```diff
@@ -192,7 +192,10 @@ class nsZenWorkspaceCreation extends MozXULElement {
 
       this.currentProfile = {
         id: 0,
-        name: "Default",
+        // Match the localized label Firefox uses for the default (no-container)
```
mr-cheffy: No need for a comment here, classic AI

mr-cheffy (changes requested):

src/zen/urlbar/ZenUBActionsProvider.sys.mjs
InlitX: done! moved all the actions into their own zen-command-palette.ftl file. lmk if that's what you had in mind or if you'd prefer a different name

src/zen/spaces/ZenSpaceCreation.mjs
InlitX: haha the comment was actually mine, i just wrote it in spanish and used ai to translate it, which turned it into that huge block 😅 removed it anyway

mr-cheffy (approved):

PR #14475, opened 2026-07-04
no-bug: add ellipsis to create space and folder menu items

InlitX:
Create Space and Create Folder open the creation panel but were missing the
trailing … that the other menu items already use (New Subfolder…, Rename Folder…,
Change Name…). Just added it so they match — small follow-up to #14388.

Screenshot after the change:

tes


mr-cheffy (approved):

mr-cheffy:
These ellipsis hasn't been added on purpose unfortunately, but thanks for the contribution! We may still discuss it though

InlitX:
ahh got it, i thought it was just missing 😅 makes total sense, thanks for the heads up!

PR #14480, opened 2026-07-05
gh-14460: prevent nested glance tabs (gh-14480)

### LINKED ISSUE #14460: Pinned/essential tab crashes after Glance overlay gets stuck from nested navigation
### Preliminary Checks
- [x] I have searched existing issues and avoided creating duplicates. (Related but distinct: #7182, #7108, #9115 — see below)
- [x] I have checked that this issue cannot be reproduced on Mozilla Firefox. (Glance is a Zen-only feature; Firefox has no equivalent to test against)
- [x] I have checked that this issue can be reproduced once I removed all my Mods and Custom CSS. (Profile's zen-themes.css is empty/no active mods)

### What happened?
On a pinned tab, external links auto-trigger Glance (per `zen.glance.open-essential-external-links`, default `true`). If a link *inside* that Glance overlay navigates to a further page, and that inner page is closed, the Glance overlay becomes stuck/un-closable. Returning focus to the underlying pinned tab then crashes its content process — the tab shows Firefox's "Gah. Your tab just crashed." page. This is 100% reproducible.

This looks like the same nested-Glance-detachment defect reported in #7182 (stuck "half-glance" state) and the bulk-close crash in #9115, but escalated: instead of just a stuck UI state, the parent pinned tab's content process actually crashes.

### Expected behavior
Closing the inner page/overlay should cleanly return the pinned tab to its normal, non-Glance state without crashing.

### Actual behavior
The Glance overlay cannot be closed after its inner navigation is closed, and returning to the pinned tab underneath crashes that tab's content process (tabcrashed page).

### Steps to reproduce
1. Pin a tab to a site that lists external links (e.g. an RSS reader like Feedly).
2. Click an external link on that pinned tab — Glance auto-opens a floating preview (no modifier key needed, since this is a pinned/essential tab).
3. Inside the Glance overlay, click a link that navigates further to a full page.
4. Close that inner full page.
5. Attempt to close the Glance overlay — it is now locked/unresponsive to closing.
6. Return focus to the original pinned tab underneath — it crashes and shows the tab-crashed page.

### Screenshots and videos
None captured — happy to add if useful, but the repro above is deterministic.

### Version
1.21.4b

Linux (Tarball)

Glance

journalctl showed a generic "Crash or unclean shutdown detected" on next launch after hitting this; no glance-specific stack trace surfaced in the user-facing logs.

### Workaround
Setting `zen.glance.open-essential-external-links` to `false` in `about:config` avoids the crash entirely, at the cost of disabling Glance's auto-trigger on pinned/essential tabs (Glance still works normally via its trigger key on regular tabs).

ashvwinn:


mr-cheffy (approved):
Good looking PR, from a good looking dude. Today must be my lucky day

src/zen/glance/ZenGlanceManager.mjs
```diff
@@ -1551,6 +1551,7 @@ class nsZenGlanceManager extends nsZenDOMOperatedFeature {
     return (
       owner &&
       owner.pinned &&
+      !owner.glanceTab &&
```
ashvwinn:
```suggestion
      !owner.hasAttribute("glance-id") &&
```

wait, maybe this is better

mr-cheffy: Yes

Tomcatt:
Awesome.

PR #14498, opened 2026-07-06
gh-14495: Fix misaligned symbolic space icons in Space options menu on Linux

### LINKED ISSUE #14495: Misaligned Symbolic space icons under the Space options menu on Linux
### Preliminary Checks

### What happened?

When you open the Space options menu, if the space has a symbolic icon, it will appear on the ratio button Before the space name in the Spaces list instead of appearing the area between the raito button and the Space name like Emoji space icons.

### Expected behavior

Sympolic icons should be between the raito button and the space name in the Spaces list like Emoji space icons.

### Actual behavior

The symbolic space icon appear on the ratio button in the Spaces list.

### Steps to reproduce

1. Set a Symbolic icon for a space, and set an Emoji icon for  another space
2. Tap the Space menu button (The 3-dot button after the space name)
3. Look at the Spaces list and the spaces icons
4. Notice the difference between the alignment of both the Symbolic icon and the Emoji icon

### Screenshots and videos

Image

### Version

1.21.5b

Linux (Flatpak)

Customizable UI / Toolbars

```shell

```

nickleigh05:
Fixes #14495

Symbolic space icons were set via the native `image` attribute, which Firefox binds to the same `.menu-icon` element GTK uses to draw the `type="radio"` indicator on Linux, so the icon was overwriting the radio bubble instead of sitting next to it. Emoji icons never hit this since they're just plain text in the label.

Fix: stop setting `image` for SVG icons, and instead insert the icon as a separate `<span>` next to the label, styled via `mask-image` (same pattern already used in zen-omnibox.css). Tried a `::before` on the label first, but `.menu-text` is a native XUL element and that broke its text rendering, hence the separate span.

Tested on Arch Linux + Hyprland (Wayland): radio bubble + icon + name now render together for symbolic icons, matching emoji rows. Checked active/checked state,
no-icon spaces, long names, and repeated menu open/close. No regressions or duplicate icons.

(Side note: `icons.css` isn't covered by the `zen` lint scope, so there are some pre-existing unrelated stylelint errors in that file, not introduced by this change. I left those alone.)

I'm open to feedback! If there's a cleaner way to avoid touching `.menu-icon`, I'd love to hear it. This was the least invasive fix I could find that didn't break the native XUL label rendering.

**Before**
image

**After**
image

### DIFF
--- a/src/zen/spaces/ZenSpaceManager.mjs
+++ b/src/zen/spaces/ZenSpaceManager.mjs
@@ -1161,6 +1161,9 @@ class nsZenWorkspaces {
     const item = document.createXULElement("menuitem");
     item.className = "zen-workspace-context-menu-item";
     item.setAttribute("zen-workspace-id", workspace.uuid);
+    if (AppConstants.platform === "linux") {
+      disableCurrent = true;
+    }
     if (!disableCurrent) {
       item.setAttribute("type", "radio");
     }


mr-cheffy (changes requested):
This wouldn't work for native context menus (e.g. macos) though? Seems hacky and AI generated. I think the solution would be to see why it isn't getting indented unlike on other OSs

nickleigh05:
Thank you for your feedback, I didn't consider native context menus. such as macos which doesn't read the light dom at all, so it would've broken the icon there. The bug's actually gtk specific which means I over scoped the fix. Let me take another look and try again.

nickleigh05:
Ok pushed a fix to make sure the scope is Linux only, reverting back to not touching macos native context menu. Now on Linux it is just reordered with flexbox. 

I looked into why its not getting indented, but it wasn't indented on linux the radio bubble and icon both shared .menu-icon so there was nothing to indent rather i found gtk puts the radio bubble into the slot where the icon is and the icon then overwrote it. win and mac never shared that element for both so the gap was never lost.

mr-cheffy:
How's the dot on linux being shown? And why does it not happen on windows?

nickleigh05:
The dot on Linux is rendered natively via gtk, For windows it not showing because of windows native rendering. The icon is present when it's unselected, but the moment it's selected, the native rendering overlays the check mark on top of the icon.

#14495 is scoped to linux but the issues exists for other OSs also should i file separate issues ?

Screenshot 2026-07-08 073557
Screenshot 2026-07-08 073609

mr-cheffy:
That doesn't look like an issue to me. Maybe for linux, it doesnt have to be a radio type menu

nickleigh05:
Yeah, abandoning the radio type is something I never considered. I was so focused on not regressing. Thank you for guiding me in a direction I hadn't thought of.

For Linux, I reused the same background highlight pattern zen already uses for "selected" state, with a media query so it persists on the active space, plain css, linux only. Is something like this what you were looking for, or should we take away any type of radio confirmation entirely so it's just the space name?

greatest

nickleigh05:
I wanted to leave a video because i felt like the screenshot didn't really show that persistent highlight

https://github.com/user-attachments/assets/97a5664a-8992-4a14-8e22-79a352b632b4

mr-cheffy:
Why dont you just force `disableCurrent` for linux? This seems to be a bit overkill for no reason

nickleigh05:
I hadn't thought of that, let me look into it. I was under the impression that would disable the active space rather than just change it's indication

nickleigh05:
Yeah, totally right disableCurrent on linux was the way to go. way simpler no need for css or anything it does disable the active space but the move tab menu already does the same thing so it fits perfectly. thank you.

mr-cheffy (approved):

Issue #9666, opened 2025-07-28
"Edit Theme" color picker doesn't show a hue/rainbow wheel


chriskirknielsen:
### Preliminary Checks

- [x] I have read and understood the important section above.
- [x] I have searched existing issues and avoided creating duplicates.
- [x] I am not filing an enhancement request.
- [x] I have checked that this issue cannot be reproduced on Mozilla Firefox.
- [x] I have checked that this issue can be reproduced once I removed all my Mods and Custom CSS.

### What happened?

Hello,

When using the Edit Theme option, the color picker doesn't show the expected hue wheel/rainbow in the background. This used to work I believe, but for a few weeks now has not been behaving properly (it might coincide with the release that reworked this feature? I am not sure). I wanted to wait a bit to see if it'd be fixed but as it's still an issue and I cannot find open tickets for it, here's my report.

Please see the screenshot below to understand why it's very inconvenient to set colors "blindly". I emptied out my userChrome stylesheet beforehand to make sure I didn't have any weird style conflicts.

Thanks in advance@

### Expected behavior

There should be a hue wheel to make it easy for users to find the color they want to use.

### Actual behavior

The color picker is an empty dot matrix, and the only way to figure out which color you have select is to blindly click on the grid

### Steps to reproduce

1. Right click the tabs area
2. Select "Edit Theme" to open the color picker

### Screenshots and videos

<img width="376" height="527" alt="The Zen Theme color picker UI without any hue wheel in the background" src="https://github.com/user-attachments/assets/cae29188-1c9b-42a5-b02e-179eef171972" />

### Version

1.14.6b

### What platform are you seeing the problem on?

macOS - aarch64

### What component is this issue related to?

Customizable UI / Toolbars

### Relevant log output if applicable

```shell

```

Buckeyes23:
@chriskirknielsen  use this mod 
https://zen-browser.app/mods/03a8e7ef-cf00-4f41-bf24-a90deeafc9db/?q=color

chriskirknielsen:
Thanks for your reply @Buckeyes23! Sadly, even after restarting Zen, this mod does not fix the issue on my computer. Regardless, this feels like this should be fixed in the core app, not an mod, right?

mr-cheffy:
This is expected sorry. HSL is hard to represent on a 2D space. It's not just "a color wheel"

chriskirknielsen:
[...]

PR #14505, opened 2026-07-07
Add visual guidance to the theme color picker

CT4nk3r:
## A personal note

Hey! While setting up Zen, I encountered two issues. One was closed quite quickly because it is tracked internally and should be fixed in the future, cool! :D

The other had already been raised by someone else in an [earlier issue](https://github.com/zen-browser/desktop/issues/9666), but it was closed as a `feature request` rather than a bug. I therefore opened [a discussion](https://github.com/zen-browser/desktop/discussions/14485), and I have now implemented the missing color wheel.

There is a working build available on my fork if you would like to try it. I have also recorded myself testing the finished build, and it works great! You can, of course, close the PR if you do not feel it is the right direction, and I am open to criticism. ;)

Thanks for creating an awesome browser!
CT4nk3r

## Summary

The theme picker currently presents its main color-selection area as a neutral dotted surface, even though pointer position already controls hue, saturation, and lightness. This makes the available color range difficult to discover.

This change gives the picker a visual representation of the colors that its existing position calculations produce:

- the default mode shows hue around the circle and lightness from black at the center to white at the edge;
- fixed-lightness presets show hue around the circle and saturation fading toward the edge;
- grayscale presets show the available black-to-white range.

The visualization follows the primary color's mode and lightness without changing the existing color-selection behavior.

## Context

This follows the proposal and user-facing discussion in https://github.com/zen-browser/desktop/discussions/14485.

It is also related to the earlier report in https://github.com/zen-browser/desktop/issues/9666.

## Demo

https://github.com/user-attachments/assets/b31e05ed-eb1b-4105-8555-d23ae795bd46

## Validation

- Added a browser test covering default, fixed-lightness, and grayscale visualization state.
- Mozilla lint passes for the changed files.
- Built the ARM64 macOS application from this branch and manually verified the picker behavior in the installed build.

### DIFF
--- a/src/zen/spaces/ZenGradientGenerator.mjs
+++ b/src/zen/spaces/ZenGradientGenerator.mjs
@@ -580,6 +580,24 @@ export class nsZenThemePicker extends nsZenMultiWindowFeature {
     return JSON.stringify({ x: Math.round(x), y: Math.round(y) });
   }
 
+  updateThemePickerVisualization() {
+    const gradient = this.panel.querySelector(".zen-theme-picker-gradient");
+    const primaryDot = this.dots.find(dot => dot.ID === 0);
+    let mode = "lightness";
+
+    if (primaryDot?.type === EXPLICIT_LIGHTNESS_TYPE) {
+      mode = "saturation";
+    } else if (primaryDot?.type === EXPLICIT_BLACKWHITE_TYPE) {
+      mode = "grayscale";
+    }
+
+    gradient.setAttribute("data-color-mode", mode);
+    gradient.style.setProperty(
+      "--zen-theme-picker-lightness",
+      `${primaryDot?.lightness ?? this.#currentLightness}%`
+    );
+  }
+
   createDot(color, fromWorkspace = false) {
     const [r, g, b] = color.c;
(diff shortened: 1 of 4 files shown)


gordonn:
This doesn't really align with Zen's design since its based on Arc's color picker which is already quite intuitive. In addition I don't think this provides that much practical value anyways as the user can still obtain basically any color with a bit of experimentation. Having a color wheel is not worth sacrificing the current minimal and clean design as the intention of Zens color picker is not to have extremely direct control and pixel perfect colors, its just so that the user can get a consistent and clean looking themes easily even if it does require a bit of tweaking and eyeballing.

CT4nk3r:
@gordonn 
> Having a color wheel is not worth sacrificing the current minimal and clean design as the intention of Zens color picker is not to have extremely direct control and pixel perfect colors

https://github.com/user-attachments/assets/26c764a2-68d7-4eb0-bbab-9e2f754d4952


While that is true, the product ships with a color picker that doesn't have color right now, so it seems like just the frontend is the one missing.

The one video i linked is how it CURRENTLY is in the product.

PR #14506, opened 2026-07-07
gh-14371: Added workspace indicator to urlbar switch-to-tab suggestions

### LINKED ISSUE #14371: URL bar "switch to tab" suggestions don't indicate a tab's workspace/container, making same-site tabs across workspaces indistinguishable
### Preliminary Checks

### What happened?

When the same site is open in more than one workspace — each workspace bound to a different container — the URL bar shows multiple identical "switch to tab" suggestions with no indication of which workspace or container each tab belongs to. There is no way to tell, for example, a work instance from a personal one.

On top of that, selecting one of these suggestions for a tab that lives in another workspace pulls that tab into the current workspace instead of switching to the workspace where it already lives. This breaks the per-workspace container isolation that the workspace/container setup is supposed to provide.

Concrete example: Gemini is open in a "Work" space (bound to a Work container) and in a "Personal" space (bound to a Personal container). Typing "gemini" in the URL bar returns two identical "Google Gemini" / "Switch to tab" entries with the same name and icon, so I can't tell which account/container each one is. Picking the wrong one drops me into the other space.

### Expected behavior

URL bar tab suggestions should indicate the workspace and/or container each tab belongs to (e.g. a label, color, or container icon) so that identical sites open in different workspaces are distinguishable. Selecting a suggestion for a tab that lives in another workspace should switch to that workspace, not move the tab into the current one.

### Actual behavior

Suggestions for the same site across workspaces are identical and unlabeled, so they cannot be told apart. Selecting one can pull the tab into the current workspace, breaking the container/workspace isolation.

### Steps to reproduce

1. Create two workspaces, e.g. "Work" and "Personal".
2. Assign a different default container to each via Settings > Tab Management > Workspaces.
3. Open the same site (e.g. gemini.google.com) in both workspaces, signed into a different account in each.
4. From one workspace, open the URL bar and type the site name (e.g. "gemini").
5. Observe the "switch to tab" suggestions: two identical entries with no workspace/container indicator. Selecting the one from the other workspace pulls it into the current one.

### Screenshots and videos

Image

### Version

1.21b

macOS - aarch64

Workspaces

```shell

```

shaun6jrome:
Fixes #14371

**Description**
When multiple tabs are open on the same site across different workspaces (and containers), the 'Switch to Tab' suggestions in the URL bar previously looked identical. This PR leverages Firefox's native `.action-contextualidentity` element (which was hidden in Zen) and populates it with the target tab's workspace name and icon.

- Removed `display: none` for `.action-contextualidentity` in `zen-omnibox.css` and styled it to look like a workspace pill.
- Patched `UrlbarView.sys.mjs` to resolve the target tab via `gZenWorkspaces.allUsedBrowsers`, extract the workspace info, and render it in the suggestion row.

@mr-cheffy PTAL.

### DIFF
--- a/src/browser/components/urlbar/UrlbarView-sys-mjs.patch
+++ b/src/browser/components/urlbar/UrlbarView-sys-mjs.patch
@@ -40,3 +40,35 @@ index b665adb1a1ce8bbae8df4cbea6c3248c3e4fb431..1de7f9461b8ccbd4680b917e6dd5ba3c
      this.#rows.toggleAttribute("wrap", wrap);
      this.oneOffSearchButtons?.container.toggleAttribute("wrap", wrap);
    }
+@@ -2914,6 +2914,30 @@
+       return;
+     }
+     actionNode.innerHTML = "";
++
++    let tabWorkspace = null;
++    if (this.#window.gZenWorkspaces) {
++      let browsers = this.#window.gZenWorkspaces.allUsedBrowsers;
++      for (let browser of browsers) {
++        let tab = this.#window.gBrowser.getTabForBrowser(browser);
++        if (tab && tab.getAttribute("usercontextid") == result.payload.userContextId) {
++           if (result.payload.url === browser.currentURI?.spec || result.payload.url === browser.userTypedValue) {
++             let wsId = tab.getAttribute("zen-workspace-id");
++             if (wsId) {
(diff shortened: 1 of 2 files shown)


mr-cheffy:
This looks rather hacky and AI generated, sorry

Issue #14534, opened 2026-07-09
Pinned tab closes when all other tabs are closed


niclimcy:
### Preliminary Checks

- [x] I have read and understood the important section above.
- [x] I have searched existing issues and avoided creating duplicates.
- [x] I am not filing an enhancement request.
- [x] I have checked that this issue cannot be reproduced on Mozilla Firefox.
- [x] I have checked that this issue can be reproduced once I removed all my Mods and Custom CSS.

### What happened?

Pinned tab closes when all other tabs are closed even though you are focused on it.

### Expected behavior

Pinned tab remains open

### Actual behavior

Pinned tab closes

### Steps to reproduce

1. Pin a tab and keep it open
2. Have other tabs open
3. Close other tabs while on the pinned tab

### Screenshots and videos

https://github.com/user-attachments/assets/2611e1c7-496d-49ee-be73-3a51663ee907

### Version

1.21.6b

### What platform are you seeing the problem on?

macOS - aarch64

### What component is this issue related to?

Tabs

### Relevant log output if applicable

```shell

```

ashvwinn:
can reproduce with `zen.workspaces.open-new-tab-if-last-unpinned-tab-is-closed` turned on

PR #14535, opened 2026-07-09
gh-14534: Prevent pinned tab from closing

niclimcy:
handleTabBeforeClose when zen.workspaces.open-new-tab-if-last-unpinned-tab-is-closed is set to true tries to open new empty tab instead of staying of selected pinned tab.


niclimcy:
@mr-cheffy could you review this?

PR #14551, opened 2026-07-11
gh-14550: Use opaque container IDs in Spaces Sync

### LINKED ISSUE #14550: Spaces Sync can overwrite or delete unrelated local containers
### Problem

The current Spaces Sync implementation treats Firefox `userContextId` as a cross-device identity:

- container records are keyed as `c~<userContextId>`
- incoming records update or create the same numeric ID
- incoming tombstones call `ContextualIdentityService.remove(userContextId)`
- Space records also carry the numeric `containerTabId`

However, `userContextId` is allocated independently in every Firefox or Zen profile. The same number can represent different containers on two devices.

### Reproduction

1. On device A, create or use container ID 2 as "Work" and assign a Space to it.
2. On device B, let ID 2 represent a different container, such as "Banking".
3. Sign both profiles into the same Mozilla account with Spaces Sync enabled.
4. Sync both devices.

The incoming record from A updates the unrelated ID 2 on B and makes the synced Space point at the existing cookie jar on B. Deleting the container on A is worse: the remote tombstone reaches B and `ContextualIdentityService.remove(2)` invokes the Firefox clear-data service for the unrelated container on B.

### Expected behavior

- Never use profile-local numeric IDs as Sync record identity.
- Use an opaque Sync ID with a profile-local mapping.
- Translate Space and container references through that mapping.
- Never clear local cookies or site data in response to a remote tombstone without an explicit local user action.
- Bump the engine version so incompatible preview records are reset.

### Impact

This can attach a Space to the wrong local cookie or session boundary, rename an unrelated container, or clear local site data on another device.

I have a working implementation and regression coverage in [arturict/zen-browser-desktop@`codex/cross-device-sidebar-sync`](https://github.com/arturict/zen-browser-desktop/tree/codex/cross-device-sidebar-sync). I will follow up with a focused PR containing only the landed Spaces and container safety fix; the broader pins and folders work remains separate because #13984 already covers it.

arturict:
Fixes #14550

## Summary

- replace profile-local numeric container record IDs with random opaque Sync IDs
- persist a profile-local map from opaque IDs to `userContextId`
- serialize Space references as `containerSyncId` and resolve them locally on apply
- reconcile independently-created matching containers by semantic ordinal without collapsing intentional same-name duplicates
- preserve local containers for remote tombstones so Sync never clears cookies or site data on another device
- bump the Spaces engine from version 2 to 3 to reset incompatible preview records
- add browser coverage for opaque IDs, aliases, duplicate ordinals, and non-destructive tombstones

## Why

Firefox allocates `userContextId` independently per profile. Using `c~2` account-wide can rename an unrelated ID 2, attach a Space to the wrong local cookie jar, or call `ContextualIdentityService.remove(2)` and clear unrelated local site data.

## Verification

- `node --check` passes for both modified modules and the browser test
- Firefox ESLint rules pass for the production modules and test
- `git diff --check` passes
- the same code path passes full imported-tree ESLint in the integration fork

A local browser-chrome run could not launch because this Windows checkout has no compiled browser and Visual Studio Build Tools installation was blocked by UAC. The PR is draft so CI can validate the imported build and browser test before it is marked ready.

### DIFF
--- a/src/zen/sync/ZenSyncManager.sys.mjs
+++ b/src/zen/sync/ZenSyncManager.sys.mjs
@@ -3,6 +3,8 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 const lazy = {};
+const CONTAINER_SYNC_MAPPINGS_PREF = "zen.sync.container-id-mappings";
+const MAX_CONTAINER_SEMANTIC_ORDINAL = 128;
 
 ChromeUtils.defineESModuleGetters(lazy, {
   ZenSessionStore: "resource:///modules/zen/ZenSessionManager.sys.mjs",
@@ -14,6 +16,17 @@ ChromeUtils.defineESModuleGetters(lazy, {
 function normalizeUserContextId(value) {
   const normalized = typeof value === "string" ? Number(value) : value;
   if (!Number.isSafeInteger(normalized) || normalized <= 0) {
+    return null;
+  }
+  return normalized;
+}
+
+function normalizeContainerSyncId(value) {
+  if (typeof value !== "string") {
+    return null;
+  }
+  const normalized = value.trim();
+  if (!normalized || normalized.length > 64 || normalized.includes("~")) {
     return null;
   }
   return normalized;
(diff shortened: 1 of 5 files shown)


mr-cheffy:
Ill close this for cleanup, seems to be highly ai generated as well

PR #14560, opened 2026-07-12
gh-14317: Show pinned-tab scroll-shadow dividers based on scroll position

daberni:
The top/bottom scroll-shadow dividers were derived from the arrowscrollbox's scrolledtostart/scrolledtoend, which it computes from the first/last scrollable element's edge. With collapsed pinned tabs those edges are unreliable, so the divider either flashed at the top (when pinned tabs were counted) or only appeared once scrolled past the pinned/normal separator (when excluded).

### DIFF
--- a/src/toolkit/content/widgets/arrowscrollbox-js.patch
+++ b/src/toolkit/content/widgets/arrowscrollbox-js.patch
@@ -1,5 +1,5 @@
 diff --git a/toolkit/content/widgets/arrowscrollbox.js b/toolkit/content/widgets/arrowscrollbox.js
-index b80d1049bb6ae305f2ac9c4c35fe975fd508031c..574149bffa49329e927c8db9db0c080eb6b87f5f 100644
+index b80d1049bb6ae305f2ac9c4c35fe975fd508031c..bb06b68b9e023b14d1b0655289ad11dab3d35313 100644
 --- a/toolkit/content/widgets/arrowscrollbox.js
 +++ b/toolkit/content/widgets/arrowscrollbox.js
 @@ -98,6 +98,7 @@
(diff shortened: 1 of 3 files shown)


mr-cheffy:
I feel like you already opened a similar PR to this? As I explained last time, we don't plan on fixing this, the overhead doesn't seem to be worth it for this minor issue

daberni:
> I feel like you already opened a similar PR to this? As I explained last time, we don't plan on fixing this, the overhead doesn't seem to be worth it for this minor issue

@mr-cheffy  You are right, but this approach is now really clean an minimal. Also restoring the same behaviour as Arc is showing. The duplicate divider on collapsed Workspaces is really annoying me :D

src/toolkit/content/widgets/arrowscrollbox-js.patch
```diff
@@ -22,7 +22,32 @@ index b80d1049bb6ae305f2ac9c4c35fe975fd508031c..574149bffa49329e927c8db9db0c080e
      connectedCallback() {
        this.removeAttribute("overflowing");
  
-@@ -642,7 +648,7 @@
+@@ -594,17 +600,21 @@
+       let scrolledToStart = false;
+       let scrolledToEnd = false;
+ 
++      // This is enough slop (>2/3) so that no subpixel value should
++      // get us confused about whether we reached the end.
```
mr-cheffy: Revert this, change and declare the constant again to make the patches lower

mr-cheffy (changes requested):
This all seems to be *super* AI generated....

daberni:
> This all seems to be _super_ AI generated....

why do you think so? which part of the code does not meet your standards, or what is your issue exactly? 🧐 I am happy to address any concerns

src/toolkit/content/widgets/arrowscrollbox-js.patch
daberni: Okay, just thought that I wouldn't want to duplicate constants.

PR #14563, opened 2026-07-12
gh-9186: Support dropping tabs onto a space switcher icon

daberni:
Disclaimer: there was already a PR of an earlier version of this feature.

This now should really be a minimal and non-intrusive implementation for supporting drag and drop of tabs onto the space switcher icons (without having to wait for activation of that space and manually finding a drop-position for insertions).
It is using the default settings of the "add tab-position" for locating the target position of the dragged tabs (this was already implemented before, the same is used for the "move to space" functionality).

This PR is currently only supporting tabs (no folders) dragged inside of the same window and skipping essentials as usual. For unsupported tabs, the releasing of mouse button is a simple noop - not causing any harm - and it is falling back to the previous "hover => activate workspace"-behaviour.

When additional types should be supported, this can easily be implemented later into the existing architecture by extending `#supportsDirectSpaceIconDrop`.

### DIFF
--- a/src/zen/drag-and-drop/ZenDragAndDrop.js
+++ b/src/zen/drag-and-drop/ZenDragAndDrop.js
@@ -126,6 +126,10 @@
           "dragover",
           this.handle_spaceIconDragOver.bind(this)
         );
+        gZenWorkspaces.workspaceIcons.addEventListener(
+          "drop",
+          this.#handle_spaceIconDrop.bind(this)
+        );
       }
       window.addEventListener(
         "dragleave",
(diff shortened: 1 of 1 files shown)


mr-cheffy:
We already have someone working on this and their solution seems to be cleaner (not sure what  the logic behind supportsDirectSpaceIconDrop even is?). Sorry and thanks!

daberni:
> not sure what the logic behind supportsDirectSpaceIconDrop even is?

Code deduplication. why write the same guards twice?

PR #14564, opened 2026-07-12
no-bug: Ignore `zen-empty-tab` in `#isLastTabWindow`

octaviusz:
When dragging a tab from the blank window to the main window, the blank window doesn't close because '#isLastTabWindow' considers 'zen-empty-tab' as a normal tab.


mr-cheffy (approved):
<3

mr-cheffy (approved):

PR #14578, opened 2026-07-14
Fix: multi-tab "Move Tabs > (workspace)" reverses tab order

### LINKED ISSUE #14575: Tab order flips when moving tabs between workspaces
### Preliminary Checks

### What happened?

When selecting multiple tabs and using the `Move Tabs > (workspace)` context menu action, the tabs get reversed.

It seems like there are a couple issues pertaining to many-tab actions being done in reverse order, perhaps there is an underlying cause. See #13796, #13822, and #12159

### Expected behavior

The tabs should stay in the same order when moving them

### Actual behavior

The tabs get reversed

### Steps to reproduce

1. Select multiple tabs by shift-clicking them
2. Right click -> Move tabs -> pick any workspace

### Screenshots and videos

https://github.com/user-attachments/assets/8e5d2463-7ee4-43d2-8af1-fbdea842e8d9

### Version

1.21.6b

macOS - aarch64

Tabs

```shell

```

wasteofwifi:
# Fix #14575: Preserve tab order when moving multiple tabs between workspaces

## Summary

Moving multiple selected tabs could reverse their order because the insertion point was recalculated after each tab was moved.

This PR batches multi-tab moves into a single `moveTabsToWorkspace` call and computes the insertion anchor once per target workspace, preserving the original tab order regardless of the `zen.view.show-newtab-button-top` preference.

## Testing

* Verified tab order is preserved when moving multiple selected tabs.
* Tested with both values of `zen.view.show-newtab-button-top`.
* Tested moving tabs into both empty and populated workspaces.

### DIFF
--- a/src/zen/spaces/ZenSpaceManager.mjs
+++ b/src/zen/spaces/ZenSpaceManager.mjs
@@ -1589,8 +1589,28 @@ class nsZenWorkspaces {
   }
 
   moveTabsToWorkspace(tabs, workspaceID) {
+    const workspaceContainer = this.workspaceElement(workspaceID);
+    const newtabPlacement = Services.prefs.getBoolPref(
+      "zen.view.show-newtab-button-top",
+      false
+    );
+
+    // The insertion anchor must be captured once per container before any
+    // tab is moved. Recomputing firstChild/lastChild after each tab is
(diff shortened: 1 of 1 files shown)


mr-cheffy:
AI

PR #14668, opened 2026-07-21
Add native tree connectors for folders and regular tabs.

KeanDraeYaMajesty:
Port the Folder Tree Connectors SVG approach into Zen chrome so expanded sidebars draw L-branch trees for folder children and every visible unpinned tab, controlled by zen.tree-connectors.* prefs.


mr-cheffy:
Hi, sorry but we don't plan on adding this to Zen. Thanks for the contribution though!

Issue #114, opened 2024-07-31
Sidebery features or better integration


redrockvp:
Currently I am making use of the Sidebery extension in regular Firefox, along with the ArcWTF user chrome customisations, as a way to have vertical tabs and group projects (Panels) and different components of those projects (The tree structure and folders). It is currently the best way I've found to manage my tabs, as when combined with Auto Tab Discard and Container Tabs, it essentially means you can just leave everything open, but have it neatly organized. 

I really like the direction that Zen browser is going, but I'm missing some key features that are currently stopping me from trying it out full time. I'd love to see some features from Sidebery make their way into the browser. 

I believe the only thing missing from Zen, at least at a high level, is a tree style tab layout. I see that Tab Groups are being worked on, but it would be great to have tabs nestable under other tabs, along with nestable groups. 

The alternative for me would be better support for Sidebery, as currently, its a bit awkward to try and use it on Zen, as its a sidebar extension. Ideally, I'd want to be able to move the unique Zen buttons, e.g. the Zen Sidebar etc, to another location, such as the top bar or hide them on the right side of the browser, as having a double sidebar is a little awkward. 

Thanks for your work so far :)

Reference: https://github.com/kikaraage/arcwtf

Jak2k:
Maybe you could soft fork Sidebery for better integration (e.g. for split tabs) and apply some style tweaks…

Apart from stability, this is the only thing holding be back.

guigirl42:
Another vote here for incorporating the brilliant `Sidebery` functionality in Zen, please. I use it _extensively_ in `Firefox Nightly` & `Floorp`, & miss all its magic very badly in Zen. 

**Note**; IMO the "urgency" for this request is increasing, as i feel quite worried about the well-being of the `Sidebery` Project. Afaict at https://github.com/mbnuqw/sidebery, there's been no more project activity [at least, per _Commits_] for **four months**, which is very unusual compared to the earlier frequent activity. It would be a terrible shame if this amazing AddOn became unmaintained then broke with ongoing Firefox development, but building it into browsers like `Floorp` & `Zen` would ameliorate any such loss a bit.

GitMurf:
> I believe the only thing missing from Zen, at least at a high level, is a tree style tab layout. I see that Tab Groups are being worked on, but it would be great to have tabs nestable under other tabs, along with nestable groups.

[...]

PR #14746, opened 2026-07-28
no-bug: Native tree-style tabs for the vertical sidebar

joegoldin:
Native tree-style tabs for the vertical sidebar. Tabs nest into a real parent/child tree, drag as whole branches, survive restart, undo-close and window sync, and treat split views as first-class tree nodes. Adds a middle-mouse drag-select gesture for closing a range of tabs.

[Discussion #891](https://github.com/zen-browser/desktop/discussions/891) has been asking for this since 2024 (276 upvotes, 30 comments, carried over from issue #114): "tabs nestable under other tabs, along with nestable groups". This covers that nesting, groups included. It doesn't address the other half of that thread, which is about the Zen sidebar coexisting better with Sidebery's own.

## Screenshots

Four levels of nesting, and a branch folded away by its twisty:

| Nested tree | Branch collapsed |
|:--:|:--:|
| Tabs nested four levels deep | A collapsed branch showing its twisty |

A split view sitting in the tree as one row, and a multi-selection spanning a branch:

| Split view as a tree node | Multi-selection |
|:--:|:--:|
| A split view group nested as a single tree row | Three tabs in a branch multi-selected |

## The tree model

`src/zen/tab-tree/ZenTabTree.mjs` holds parent/child state on `_zenTreeParent`, `_zenTreeLevel` and `_zenTreeCollapsed`, mirrored onto `zen-tree-id`, `zen-tree-parent-id`, `zen-tree-collapsed` and `zen-tree-hidden` attributes. The strip is kept in depth-first order, so the DOM order and the visual tree never disagree.

Tabs auto-nest under the tab that opened them. Closing a parent either promotes its children or closes the subtree, controlled by a pref. There's a depth cap (`max-depth`), collapse and expand through a left-aligned twisty, and nesting across workspaces is rejected.

Auto-nesting had to survive one Firefox behaviour: `tab.owner` is cleared on each successive tab opened from the same opener, so only the first would keep it. The opener is captured in `tabbrowser` as `tab._zenOpenerTab`, where it's still known, which is what makes "open five links from one page" nest all five.

## Drag and drop

`src/zen/drag-and-drop/ZenDragAndDrop.js` handles three drop zones. Vertical edges reorder, with a live indent preview clamped to the level you can actually reach, and the tab lands exactly where the indicator showed it would. Dragging left outdents to root. The central band nests. Horizontal edges arm a split-left or split-right, but only on a deliberate hold, so a normal reorder never turns into a split by accident.

Dragging a parent moves its whole branch as one block with the hierarchy intact, and the branch animates from its dragstart layout to its dropped position as a single unit (a block FLIP) rather than tabs sliding through each other. Dragging a multi-selection nests and reorders every selected tab, not just the one under the cursor.

## Split views as tree nodes

A `tab-group[split-view-group]` is one node: it nests, indents, moves and collapses as a single row, and its inner tabs aren't independent nodes. When a nested tab becomes a split it hands its tree role, both parent and children, to the group. Dissolving a split returns the tabs and their children to the split's level.

## Middle-mouse drag-select

`src/zen/tab-tree/ZenTabMultiSelectDrag.mjs` adds a range-select gesture: hold the middle button and drag to select a range, release to close the selection. Right-click aborts to the context menu, and a clean middle-click still closes a single tab. It closes exactly the selection and never the active tab outside that range.

## Persistence, undo-close and window sync

Tree structure persists across restarts, rebuilds from the attributes, and mirrors into synced windows.

Ctrl+Shift+T restores the hierarchy rather than a flat list. The tree attributes ride along as restorable tab attributes, parents are matched by a stable id that survives window sync reassigning live ids, and the rebuild re-asserts depth-first order so restored tabs don't interleave with unrelated ones.

## Engine touchpoints

Two patches outside `src/zen`:

* `src/browser/components/sessionstore`: `TabState` and `TabAttributes` persist
and restore the `zen-tree-*` attributes on every path, including undo-close, which doesn't run `restoreInitialTabData`.
* `src/browser/components/tabbrowser`: captures the opener durably as
`tab._zenOpenerTab`, as described above.

## Prefs

`prefs/zen/tab-tree.yaml` adds indent width, `max-depth`, drag zone sizes, the nest-to-split hold delay, and a toggle for the middle-drag gesture.

## Tests

`src/zen/tests/tab-tree/` has 16 files covering the model, nesting, depth capping, collapse, drag, opener tracking, close behaviour, pinned reordering, reordering, persistence, split-as-node, multi-select detach, eligibility, workspace boundaries and middle-drag. `src/zen/tests/window_sync/browser_sync_tree.js` checks the tree survives window sync.

Built against Firefox 153 and run locally with `npm test -- <dir> --headless`.

`tab-tree`: **142 passed, 0 failed** across the 16 files listed above.

`window_sync`: **28 passed, 0 failed**, over three consecutive runs of the whole directory. Getting there turned up three sync bugs, fixed here: a tab close and a tab open were both being dropped when the other window happened to be mid-sync, and the icon sync read the source tab's window through a property that is undefined on a tab, so every propagation threw inside a bare `catch {}` and the mirror silently kept its old favicon.

### DIFF
--- a/prefs/zen/tab-tree.yaml
+++ b/prefs/zen/tab-tree.yaml
@@ -19,7 +19,10 @@
   value: 4 # maximum level; root = 0; 0 disables the cap
 
 - name: zen.tab-tree.drag-nest-to-split-delayMC
-  value: 300 # ms hold before nest escalates to split
+  value: 500 # ms hold in a tab's right edge zone before it escalates to split
+
+- name: zen.tab-tree.drag-split-zone
+  value: 25 # rightmost % of a tab that arms the split gesture
 
 - name: zen.tabs.middle-drag-select.enabled
   value: true
--- a/prefs/zen/tab-tree.yaml
+++ b/prefs/zen/tab-tree.yaml
@@ -13,7 +13,7 @@
   value: promote
 
 - name: zen.tab-tree.indent
-  value: 20 # px per nesting level
+  value: 14 # px per nesting level
 
 - name: zen.tab-tree.max-depth
   value: 4 # maximum level; root = 0; 0 disables the cap
--- a/prefs/zen/tab-tree.yaml
+++ b/prefs/zen/tab-tree.yaml
@@ -13,7 +13,7 @@
   value: promote
 
 - name: zen.tab-tree.indent
-  value: 14 # px per nesting level
+  value: 20 # px per nesting level
 
 - name: zen.tab-tree.max-depth
   value: 4 # maximum level; root = 0; 0 disables the cap
--- a/prefs/zen/tab-tree.yaml
+++ b/prefs/zen/tab-tree.yaml
@@ -1,0 +1,25 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+- name: zen.tab-tree.enabled
+  value: true
+
+- name: zen.tab-tree.auto-nest-by-opener
+  value: true
+
+# "promote" (children move up) or "close-subtree" (close descendants too)
+- name: zen.tab-tree.close-parent-behavior
+  value: promote
+
+- name: zen.tab-tree.indent
+  value: 14 # px per nesting level
+
+- name: zen.tab-tree.max-depth
+  value: 4 # maximum level; root = 0; 0 disables the cap
+
+- name: zen.tab-tree.drag-nest-to-split-delayMC
+  value: 300 # ms hold before nest escalates to split
+
+- name: zen.tabs.middle-drag-select.enabled
+  value: true
--- a/prefs/zen/tab-tree.yaml
+++ b/prefs/zen/tab-tree.yaml
@@ -24,5 +24,8 @@
 - name: zen.tab-tree.drag-split-zone
   value: 25 # rightmost % of a tab that arms the split gesture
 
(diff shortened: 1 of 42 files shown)


joegoldin:
Been running this (in some form or another) on my machine for the last couple months, I think I've ironed out all the weird bugs and edge-case behaviors. :crossed_fingers:

mr-cheffy:
Sorry, but we don't plan on adding this. It adds extra complexity and maintainability that we aren't looking forward right now unfortunately.

D3SOX:
@joegoldin Can you please maintain a fork then? Sidebery is literally the only thing keeping me from switching. This natively integrated would be even better.
I could help setting up build infra and maintain an AUR package

joegoldin:
> @joegoldin Can you please maintain a fork then? 

Yep, I'll get something spun up today in [my fork](https://github.com/joegoldin/zen-browser-desktop), I have a build pipeline already so I can pretty easily add AUR etc.

Thanks for taking a look @mr-cheffy, bummed but understand!

D3SOX:
I just wanted to look into creating the AUR package but I saw you're already working on it. Will test once it's out and report if I find any issues with it :)

12th-devs:
@joegoldin you could call it bonsai, like the tree

kaelendev:
Can't it be added as a zen mod or a firefox extension for the moment ? Adding it as a Zen mod would avoid maintainability et complexity issues. I'll try the fork ! Thanks @joegoldin

Issue #14600, opened 2026-07-15
Space routing: external links inherit source-space container instead of destination space's default (regression from #14482)


sarcasticbird:
### Preliminary Checks

- [x] I have read and understood the important section above.
- [x] I have searched existing issues and avoided creating duplicates.
- [x] I am not filing an enhancement request.
- [x] I have checked that this issue cannot be reproduced on Mozilla Firefox.
- [x] I have checked that this issue can be reproduced once I removed all my Mods and Custom CSS.

### What happened?

This is the flip side of #14242. #14482's userContextId approach fixes the explicit-container case, but it stamps the active (source-space) container onto the tab before routing moves it, so external links no longer land in the destination space's default container.

External links carry no explicit container intent — they should resolve to the routed space's default container, not inherit whatever was active at click time. Splitting explicit-container tabs from no-container tabs would satisfy both cases.

### Expected behavior

External link routed to the default space opens in that space's default container.

### Actual behavior

Lands in the correct space, but keeps the previously-active space's container.

### Steps to reproduce

- Default space D (container X), space Y (container Z); routing sends external links to D.
- Switch to Y.
- Click a link from another app.
- Tab lands in D but in container Z, not X. (Pre-1.21b: correctly X.)

### Version

1.21.6b

### What platform are you seeing the problem on?

macOS - aarch64

### What component is this issue related to?

Tabs

### Relevant log output if applicable

```shell

```

steffenagger:
When the described sequence happens, closing the tab and switching space (swipe on the sidebar) also bricks the sidebar and/or main window for me. Either no page is shown when selecting a tab, or the sidebar swipes halfway.

denzakharov:
Same for me: macOS, 1.21.8b

utopiaeh:
@mr-cheffy When I open a link from Slack, it redirects me to the correct container, but it also tries to open the same link in the active container after redirecting to the right one, and it remains open in both containers. I tried opening the same link from Telegram, and the behavior is fine.

PR #14749, opened 2026-07-28
no-bug: Seed Space Routing rules and Spaces from managed preferences

joegoldin:
## Why

Space Routing (gh-13981) persists its rules to an lz4-compressed profile file (`zen-space-routing.jsonlz4`) and references each target Space by a random per-profile UUID. Neither the rules nor the Spaces they point at can be expressed declaratively. The [Containerise](https://github.com/kintesh/containerise) extension solves the same problem differently: its rules live in a plain `storage.local` JSON keyed by container *name*, which a configuration-management setup can seed.

So there's no way to configure Space Routing, or the Spaces themselves, deterministically from config. This adds both, modelled on how Containerise rules are seeded, via two prefs:

* `zen.space-routing.managed-spaces` defines the Spaces (name, icon, container,
order).
* `zen.space-routing.managed-routes` defines the rules that route URLs into
them, by Space name.

Paired, they let a config own which Spaces exist and what opens where, referencing everything by stable names instead of per-profile UUIDs.

## Managed routes

`zen.space-routing.managed-routes` holds a JSON array of routes. Managed routes are evaluated before user-created routes and take precedence over them, so the configuration is the deterministic source of truth (the same intent as Containerise's `force = true`). They're never written back to the on-disk store.

A route targets a Space by name through `openInSpace`, resolved to that Space's id at routing time, which is what lets the config survive the random per-profile UUIDs. An `openIn` literal (`"most-recent-space"` or a known Space id) still works.

The Space Routing dialog renders managed routes read-only, so a rule the user never added by hand still has a visible explanation.

Malformed input never throws. A JSON parse error or an unexpected shape yields no managed routes and logs, so a bad pref can't break tab opening.

### Route schema

```jsonc
[
  // route github.com into the Space named "Work"
  { "reference": "github.com", "matchType": "contains", "openInSpace": "Work" },
  // matchType defaults to "contains"; openIn literal also works
  { "reference": "youtube.com", "openInSpace": "Personal" },
  { "reference": "^https://.*\\.internal/", "matchType": "regex", "openIn": "most-recent-space" }
]
```

`matchType` is one of `contains` | `equal-to` | `regex`, defaulting to `contains`. An object form `{ "routes": [ … ] }` is also accepted.

## Managed Spaces

`zen.space-routing.managed-spaces` holds a JSON array of Space definitions. A managed Space is created if missing and synced if present: on every window init a reconcile pass ensures a Space with each configured name exists and updates its icon and container to match the config.

It is never deleted. A Space absent from the config is left untouched, so the config is purely additive and can't destroy a user's own Spaces.

It is read-only in the UI. Managed Spaces can't be renamed, re-themed or deleted by hand: the context-menu rename input and theme picker are hidden, and attempting to delete shows a "managed by your configuration" toast. This mirrors how managed routes render.

Spaces are matched by name, so the config is stable across profiles. Malformed JSON, or an entry with an empty name, yields no managed Spaces and logs, so a bad pref can't break startup.

### Space schema

```jsonc
[
  // "briefcase" = a built-in selectable icon; container matched by its name
  { "name": "Work", "icon": "briefcase", "container": "Work", "position": 0 },
  // icon can be an emoji…
  { "name": "Personal", "icon": "🏠" },
  // …a built-in icon name / *.svg (expanded to a chrome:// selectable-icon URL),
  // or a full URL. position falls back to the array order.
  { "name": "Media", "icon": "play.svg" }
]
```

`name` is required and is the key used to match and sync an existing Space. `icon` takes an emoji (kept as-is), a built-in icon name or `*.svg` (expanded to `chrome://browser/skin/zen-icons/selectable/<name>.svg`), or a full URL. `container` takes a container's name, resolved to its `userContextId`; an unknown name means no container, and containers are never created here. `position` is the intended order and falls back to the entry's array index.

An object form `{ "spaces": [ … ] }` is also accepted.

## Implementation notes

`ZenSpaceRoutingManager` gains `getManagedRoutes()`, which parses, normalizes and memoizes the pref, re-parsing only when the raw value changes via a lazy pref getter. `routeUri()` takes an optional `win` argument and consults managed routes first, resolving `openInSpace` to an id through `win.gZenWorkspaces`. `getAllRoutes()`, which backs the editable user routes, is unchanged.

`ZenSpaceRoutingDialog` renders a read-only "Managed by your configuration" section above the editable routes.

`ZenManagedSpaces` is a new module. `getManagedSpaces()` parses, normalizes and memoizes the pref; `resolveContainerId()` maps a container name to a `userContextId`; `reconcile(win)` creates and syncs Spaces through `gZenWorkspaces.saveWorkspace`, which materializes both the cache entry and its sidebar section, and never deletes.

`ZenSpaceManager` runs the reconcile pass in `restoreWorkspacesFromSessionStore` once workspaces are live, before any routing resolves a Space by name, and enforces read-only managed Spaces through a `removeWorkspace` guard plus toast and hidden context-menu entries.

New prefs are declared in `prefs/zen/space-routing.yaml`, and the read-only toast string in `zen-workspaces.ftl`.

## Two routing fixes that came with it

Exercising the managed config against a real profile surfaced two pre-existing edge cases in the Space Routing to container/Space integration (the `tabbrowser.js` patch, gh-14044). Both are fixed here.

**No-Container routes were ignored and the tab was hijacked into the active container.** A route to a Space with no container resolves to `userContextId = 0`, but the integration gated on `if (beforeRouteResult.userContextId)`, a truthiness check, so `0` read as "no route". `getContextIdIfNeeded` then forced the tab into the active or inherited container's Space, switching Spaces and tearing a freshly-opened child tab out from under its opener, which broke tree nesting. The symptom was middle-clicking a `news.ycombinator.com` link and landing in a different container and Space than the rule specified, intermittently. The fix is to gate on `isRouteFound` and treat `0` as a real "No Container" decision, so a No-Container route stays authoritative. With the tab no longer relocated, `moveTabToWorkspace` no-ops and the nesting survives.

**An explicit "Open in New Container Tab" was overridden by a matching route.** Making routes authoritative over-reached: a deliberate container choice, which the context menu passes straight through to `addTab` as `userContextId`, lost to the rule. `onBeforeAddTab` now skips routing when the caller passed an explicit `userContextId`, so the chosen container and its own Space win through the default-on container-to-Space mapping in `getContextIdIfNeeded`. Routing only acts when no container was explicitly chosen. External links are exempt from that skip, so gh-14600's fix for external links still applies.

Net behaviour: no container specified plus a matching rule routes to the rule's Space (or No Container); an explicit "Open in New Container Tab → X" gives X's container and X's Space, and the rule is ignored.

## Tests

`browser_space_routing_managed.js` covers routes: name resolution, precedence over user routes, literal / `most-recent-space` / default targets, unknown-name fallback, `matchType` defaulting, the object-with-`routes` form, malformed-JSON tolerance, pref clearing, and the read-only dialog rendering.

`browser_managed_spaces.js` covers Spaces: parsing and normalization (icon expansion, emoji passthrough, position fallback, empty-name drop), container name resolution, reconcile create / sync-existing / never-duplicate / never-delete, the startup reconcile in a fresh window, and read-only enforcement, checking that a managed Space can't be deleted while a non-managed one still can. `head.js` gains a `withManagedSpaces` helper.

`browser_space_routing_no_container.js` covers the routing fixes with a real `addTab` test: a tab routed to a No-Container Space opens with no container rather than being hijacked into the active one, and an explicit container choice survives a No-Container route, keeping both its container and its own Space. `browser_space_routing_on_add_tab.js` gains a unit test that an explicit `userContextId` skips routing.

Built against Firefox 153 and run locally with `npm test -- space_routing --headless`: **23283 passed, 0 failed**, twice in a row over the whole directory. `mach lint zen` (eslint + stylelint) is clean.

This branch also repairs the existing suite. `onBeforeAddTab` grew a `targetWorkspaceName` key without the assertions being updated, so `browser_space_routing_on_add_tab.js` fails **7 subtests on dev** as it stands; those are corrected here.

The managed-Spaces tests were also leaking into the two files that follow them: Spaces outliving the task that made them (`removeWorkspace` settles on a microtask and three calls didn't await it), no removal of the Spaces the file introduced, and a teardown that took Zen's empty-tab placeholder with it. Zen rebuilds that placeholder lazily, and the null `linkedBrowser` at the front of the strip is what made the next file's first real `addTab()` return null.

### DIFF
--- a/locales/en-US/browser/browser/zen-space-routing.ftl
+++ b/locales/en-US/browser/browser/zen-space-routing.ftl
@@ -31,3 +31,7 @@ tab-context-zen-add-domain-to-sr =
             [one] Add Route for Domain
             *[other] Add Route for Domains
         }
+
+# Heading for the read-only list of routes that come from a managed policy or a
+# declarative configuration (e.g. a dotfiles setup) rather than the dialog.
+zen-space-routing-managed-section = Managed by your configuration
--- a/locales/en-US/browser/browser/zen-workspaces.ftl
+++ b/locales/en-US/browser/browser/zen-workspaces.ftl
@@ -87,6 +87,8 @@ zen-workspace-default-profile = Default
 zen-workspaces-delete-workspace-title = Delete Space?
 zen-workspaces-delete-workspace-body = Are you sure you want to delete { $name }? This action cannot be undone.
 
+zen-workspaces-managed-readonly-toast = This Space is managed by your configuration.
+
 # Note that the html tag MUST not be changed or removed, as it is used to better
 # display the shortcut in the toast notification.
 zen-workspaces-close-all-unpinned-tabs-toast = Tabs Closed! Use <span>{ $shortcut }</span> to undo.
--- a/prefs/zen/space-routing.yaml
+++ b/prefs/zen/space-routing.yaml
@@ -18,3 +18,14 @@
 # `openIn` instead to target a literal "most-recent-space" or a known Space id.
 - name: zen.space-routing.managed-routes
   value: ""
+
+# A JSON array of "managed" Spaces, seeded the same way as managed-routes so a
+# declarative config can define the Spaces themselves (not just route to them).
+# Each entry: { "name": "Work", "icon": "briefcase", "container": "Work",
+# "position": 0 }. `icon` is an emoji, a built-in icon name / *.svg (expanded to
+# a chrome:// selectable-icon URL), or a full URL. `container` is resolved by
+# label to a userContextId (unknown -> none; containers are never created).
+# Managed Spaces are created if missing, synced if present, never deleted, and
+# read-only in the UI.
+- name: zen.space-routing.managed-spaces
+  value: ""
--- a/prefs/zen/space-routing.yaml
+++ b/prefs/zen/space-routing.yaml
@@ -1,0 +1,20 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+# A JSON array of "managed" Space Routing rules, intended to be set by an
+# enterprise policy or a declarative configuration manager (e.g. a Nix /
+# home-manager dotfiles setup), the same way Containerise rules can be seeded.
+#
+# Managed routes are evaluated before — and take precedence over — the routes a
+# user creates in the Space Routing dialog, and they are never written back to
+# the on-disk store. Each entry looks like:
+#
+#   { "reference": "github.com", "matchType": "contains", "openInSpace": "Work" }
+#
+# `openInSpace` targets a Space by its name, which is resolved to the Space's id
+# at routing time. That indirection is what makes a declarative config
+# deterministic: Space ids are random per profile, but names are stable. Use
+# `openIn` instead to target a literal "most-recent-space" or a known Space id.
+- name: zen.space-routing.managed-routes
+  value: ""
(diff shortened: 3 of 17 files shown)


joegoldin:
I wanted this since I configure everything via nix: https://github.com/joegoldin/dotfiles/blob/main/modules/home/zen/default.nix#L373-L377

mr-cheffy:
Hi, thank you for your contribution! Unfortunately, we don't plan on adding this, and the UUIDs aren't random, those are the UUIDs of the spaces which can be duplicated as well. Sorry. This also seems to be highly AI generated

PR #14750, opened 2026-07-28
no-bug: Reopen tabs in the destination Space's container when moving between Spaces

joegoldin:
## What this fixes

Moving a tab into a Space that has a default container left the tab in its old container. The tab now sits in a Space whose container it doesn't belong to, and nothing in the UI says so, because the container marker was decided once at tab creation and never recomputed.

Two changes: move the tab into the destination Space's container, and make the marker reflect the Space you're actually looking at.

## Reopening in the destination container

Firefox can't change a tab's `userContextId` in place, so the tab has to be reopened. `reopenTabInWorkspaceContainerIfNeeded()` copies the tab's SessionStore state (URL, full session history, pinned state), sets the new container and Space on the copy, restores it, and closes the original. Selection follows the tab, and a tab that wasn't selected is recreated lazily so moving a batch of background tabs doesn't force a load.

It returns the original tab untouched when there's nothing to do: the destination Space has no default container, the tab is already in that container, or the tab is one whose DOM relationships would break if it were recreated (essentials, the empty-tab placeholder, glance tabs, and members of a split-view group). If the reopen throws partway through, selection is put back on the original tab and the half-built replacement is discarded, so a failure costs you nothing.

Both paths that move tabs between Spaces use it: the tab context menu's "move to Space" action, and dropping a tab onto a Space in the sidebar.

One incidental change in `changeTabWorkspace`: the tab to select afterwards is now read before `moveTabsToWorkspace` rather than after. That function reverses its argument array in place when `zen.view.show-newtab-button-top` is set, which defaults to true, so reading afterwards picked the first of a multi-tab selection instead of the last. Single-tab moves, the common case, are unaffected.

## Making the container marker mean something

The marker was driven by a `zenDefaultUserContextId` attribute written when the tab was created, so it only ever described the Space the tab was born in. Once a Space had a container configured, tabs whose container differed from it showed nothing.

`updateTabContainerIndicators()` now recomputes the marker for the active Space's tabs against that Space's default container, and runs whenever the reference container can change: on Space switch, when a Space's container is reconfigured, and when tabs are added or moved. A tab shows the marker when its container differs from the Space default and hides it when they match.

Container-less tabs needed a separate case. Firefox only colours the marker when a tab has a `userContextId`, so a tab with no container sitting in a Space that does have one would render nothing at all, which is exactly the case the user needs to see. Those get a `zen-neutral-container-marker` attribute and a gray marker, since there's no container colour to borrow.

## Testing

Built against Firefox 153. This branch adds no automated tests; the behaviour is exercised through the Space Routing suite, which covers the container resolution it builds on.

### DIFF
--- a/src/zen/drag-and-drop/ZenDragAndDrop.js
+++ b/src/zen/drag-and-drop/ZenDragAndDrop.js
@@ -1016,6 +1016,7 @@
         }
       }
       gZenWorkspaces.updateTabsContainers();
+      gZenWorkspaces.updateTabContainerIndicators();
     }
 
     #handle_dropCreateSplit(event) {
--- a/src/zen/drag-and-drop/ZenDragAndDrop.js
+++ b/src/zen/drag-and-drop/ZenDragAndDrop.js
@@ -988,8 +988,23 @@
         ) {
           if (isTab(draggedTab)) {
             const movingTabs = draggedTab._dragData?.movingTabs || [draggedTab];
-            for (let tab of movingTabs) {
-              tab.setAttribute("zen-workspace-id", activeWorkspace);
+            for (let i = 0; i < movingTabs.length; i++) {
+              const tab = movingTabs[i];
+              // Reopen the tab in the destination space's container (if it has
+              // one) instead of leaving it in its previous container.
+              const reopenedTab =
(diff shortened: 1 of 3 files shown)


joegoldin:
I've also been running this on my machine for a while without issues

PR #14751, opened 2026-07-28
no-bug: Add a Nix flake that builds Zen from source

joegoldin:
A Nix flake that builds Zen from source through nixpkgs' `buildMozillaMach`, so `nix build` produces a Zen binary without going near `surfer` or the network mid-build. It is modelled on the nixpkgs packaging effort in [NixOS/nixpkgs#496647](https://github.com/NixOS/nixpkgs/pull/496647).

## What it adds

`flake.nix` plus `nix/{package,zen-browser,assets}.nix`, exposing `packages.<system>.zen-browser-unwrapped` for x86_64 and aarch64 Linux and Darwin. Nothing outside those files is touched, and nothing in the existing build path changes: `surfer` and the npm scripts keep working exactly as they do today.

The build runs entirely inside the Nix sandbox, which means every input has to be fixed up front:

* The Firefox base is a fixed-output `fetchurl`, and its version is read from `surfer.json` so it tracks the tree rather than drifting in a second place.
* `ffprefs` is built as a Rust package and run to compile the pref YAMLs, standing in for `npm run ffprefs`.
* Service dumps are merged offline from `configs/dumps` instead of being fetched.
* The full `src/` patchset is applied with `git apply`.

## Using it

```
nix build github:zen-browser/desktop#zen-browser-unwrapped
nix run github:zen-browser/desktop
```

## Notes for review

`nix/zen-browser.nix` hand-maintains two values. `firefox-src.hash` is the sha512 of the Firefox tarball and has to move when the base version does; the build fails loudly with the expected hash when it doesn't match, so it can't silently build the wrong source. `zen-version` names the store path and the reported application version, and is hand-maintained because Zen's version lives in the release tag rather than a tracked file.

`rust-cbindgen` comes from a pinned nixpkgs rather than the main input: Firefox 153 requires 0.29.4 and nixos-26.05 still carries 0.29.2, so configure rejects it. That pin can go away once the main nixpkgs catches up.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,44 @@
+{
+  "nodes": {
+    "nixpkgs": {
+      "locked": {
+        "lastModified": 1780453794,
+        "narHash": "sha256-bXMRa9VTsHSPXL4Cw8R6JJLQeY3Y/IP4+YJCYVmQ7FY=",
+        "owner": "nixos",
+        "repo": "nixpkgs",
+        "rev": "6b316287bae2ee04c9b93c8c858d930fd07d7338",
+        "type": "github"
+      },
+      "original": {
+        "owner": "nixos",
+        "ref": "nixos-26.05",
+        "repo": "nixpkgs",
+        "type": "github"
+      }
+    },
+    "nixpkgs-cbindgen": {
+      "locked": {
+        "lastModified": 1784783405,
+        "narHash": "sha256-4IHyyLgLBdKefkljdKod4IMn023pQiDXAWJA187cmdY=",
+        "owner": "NixOS",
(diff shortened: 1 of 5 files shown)


joegoldin:
I consume this in my nix config here: https://github.com/joegoldin/dotfiles/blob/main/flake.nix#L33-L40

mr-cheffy:
This looks highly vibe coded, There are already high quality ones like https://github.com/0xc000022070/zen-browser-flake

joegoldin:
This builds from source, that flake patches a prebuilt package. All good, thanks for the review!

PR #14753, opened 2026-07-28
gh-14057: Fix unclosable tab when dropping files into the browser area

### LINKED ISSUE #14057: Dragging a file from Finder with no open tabs creates an uncloseable tab
### Preliminary Checks

### What happened?

When Zen has no tabs open and a file (image, folder, etc.) is dragged from Finder into Zens window, the file's content is displayed but no tab entry is created in the tab bar. The content cannot be hiden, ⌘W has no effect or just closes Zen and there's no close button (×). Opening and then closing a new tab does not resolve the state. 

### Expected behavior

The dragged file should open as a closeable tab, consistent with opening the same file via Finder → Open With → Zen, which correctly creates a tab that can be closed with ⌘W

### Actual behavior

The file opens in a tab-less display state with no entry in the tab bar.

### Steps to reproduce

1. Open Zen and ensure no tabs are currently open.
2. In Finder, locate any file — an image, a folder, or similar.
3. Drag the file from Finder and drop it into the Zen window.
4. Open a new tab (`⌘T`), then close it (`⌘W`).
5. Observe that the dragged file's content is still shown, with no corresponding tab entry in the tab bar and no way to close it.

### Screenshots and videos

https://github.com/user-attachments/assets/73cb8403-34a2-4abd-aaa7-7cfbce0895fc

### Version

1.20.2b 

macOS - aarch64

Tabs

```shell

```

bnquon:
when no visible tabs are open, file drops were replacing Zen’s hidden empty tab instead of creating a normal tab because the target tab’s empty state wasn’t checked. this keeps the empty tab available while preserving normal tab replacement behavior.

tested Finder drops, Open With in Nightly, Cmd+W, and URL drops onto normal tabs.

demo showing issue case and replacement still working

https://github.com/user-attachments/assets/15878f2e-2f6a-476b-a246-7a3951305149

also noticed this bug while creating other demo's but finally found the issue
closes #14057

### DIFF
--- a/src/browser/components/tabbrowser/content/tabbrowser-js.patch
+++ b/src/browser/components/tabbrowser/content/tabbrowser-js.patch
@@ -236,7 +236,7 @@ index c42b1a1a8df6b38886c17f71ea88e5aaa7eebc80..c741404ed973ee3ac246fe246c0f645e
        }
  
 -      if (replace) {
-+      if (replace && !(!targetTab && this.selectedTab?.hasAttribute('zen-empty-tab'))) {
++      if (replace && !((targetTab || this.selectedTab)?.hasAttribute('zen-empty-tab'))) {
          if (this.isTabGroupLabel(targetTab)) {
            throw new Error(
              "Replacing a tab group label with a tab is not supported"


mr-cheffy (approved):
Looks great, thank you!

PR #14778, opened 2026-07-29
gh-14207: Fix add shortcut glitch on homepage

### LINKED ISSUE #14207: "Add Shortcut" Glitch on Homepage
### Preliminary Checks

### What happened?

On MacOS, after adding one shortcut on the homepage (4 rows selected in homepage settings) the add shortcut button disappears and there's no option to add another.

### Expected behavior

The "Add Shortcut" button should keep moving along until all rows are full.

### Actual behavior

As above, the "Add Shortcut" option isn't available after adding one shortcut.

### Steps to reproduce

Add a shortcut on the homepage.

### Screenshots and videos

_No response_

### Version

1.21.1b

macOS - aarch64

Customizable UI / Toolbars

```shell

```

bnquon:
the add shortcut tile was defaulting to the last available index when there were no shortcuts. since Zen doesn't fill those slots with sponsored or default sites by default, the first shortcut was pinned at that last index and the add button had nowhere left to move.

this starts the fallback index at 0, so the add button moves forward as shortcuts are added and only disappears when the grid is actually full.

demo showing manually test with an empty homepage.

https://github.com/user-attachments/assets/9557e27b-6d81-4a29-b39a-57bad555c2e1

closes #14207


mr-cheffy:
I feel like this is more an issue with Firefox, please submit this patch to them if you'd like, thanks!

PR #14781, opened 2026-07-30
fix(split-view): prevent tab selection switch on closing split tab

### LINKED ISSUE #13761: Closing a Split Tab in one Zen window causes other windows' tabs to change to the remaining tab
### Preliminary Checks

### What happened?

When multiple Zen windows are open, closing a split tab in one window will cause the other windows displaying the same space to flash and open (or attempt to open) the remaining tab from the split. Ideally, closing the tab would only affect the window you are on (and the sidebars of the other windows), without changing the focus of the other windows to the remaining tab.

### Expected behavior

When closing a split tab in one window, I would expect it to close the tab in that window, filling that window with the remaining tab, while not affect what any other window is displaying.

### Actual behavior

Closing a tab from the split is causing all other open windows within the same space to load the remaining tab simultaneously as their active tabs.

### Steps to reproduce

1. Open two separate Zen windows using the same space
2. In one window, load any website
3. In the other window, load any two websites, and add them to a split
4. Close one website from the split; both windows will update to view the remaining tab

### Screenshots and videos

https://github.com/user-attachments/assets/a65aa575-92d6-41ad-bb8e-d70ed1da850e

### Version

1.19.13b (Firefox 150.0.3) (64-bit)

Windows - x64

Split View

```shell

```

SAKTHIVEL280:
• Title:
  fix(split-view): prevent tab selection switch on closing split tab
  • Description:
### Fixes #13761

    Summary of Changes:
    - Scoped `gBrowser` to `tab.ownerGlobal` in `ZenViewSplitter.mjs` to prevent tab selection state from leaking across windows.
    - Ensured active tab selection only switches when the tab being closed was actually the actively selected tab.

### DIFF
--- a/src/zen/split-view/ZenViewSplitter.mjs
+++ b/src/zen/split-view/ZenViewSplitter.mjs
@@ -230,7 +230,17 @@ class nsZenViewSplitter extends nsZenDOMOperatedFeature {
       }
       this.removeGroup(groupIndex);
       if (changeTab) {
-        gBrowser.selectedTab = remainingTabs[0];
+        const win = tab.ownerGlobal || window;
+        const winGBrowser = win.gBrowser || gBrowser;
+        const remainingUnsplitTabs = remainingTabs.filter(t => t !== tab);
+        if (
+          winGBrowser.selectedTab === tab ||
+          !remainingUnsplitTabs.includes(winGBrowser.selectedTab)
+        ) {
+          if (remainingUnsplitTabs.length > 0) {
+            winGBrowser.selectedTab = remainingUnsplitTabs[0];
+          }
+        }
         document
           .getElementById("cmd_zenNewEmptySplit")
           .removeAttribute("disabled");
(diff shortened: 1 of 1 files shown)


src/zen/split-view/ZenViewSplitter.mjs
```diff
@@ -230,7 +230,12 @@ class nsZenViewSplitter extends nsZenDOMOperatedFeature {
       }
       this.removeGroup(groupIndex);
       if (changeTab) {
```
mr-cheffy, line 232: I don't think changeTab is true here?

mr-cheffy (changes requested):

src/zen/split-view/ZenViewSplitter.mjs
SAKTHIVEL280: Hi @mr-cheffy!

Thanks for checking! You're right that during cross-window background sync (`ZenWindowSync.sys.mjs` line 1691), `{ changeTab: false }` is explicitly passed:

```javascript
win.gZenViewSplitter.removeTabFromGroup(targetTab, undefined, { changeTab: false });
```

The root cause of issue #13761 was that the old code was referencing global `gBrowser.selectedTab = remainingTabs[0]`, which mutated the active window's tab selection even when `changeTab: false` was passed from background window sync events.

By wrapping the selection logic inside `if (changeTab)` and scoping to `winGBrowser` (`(tab.ownerGlobal || window).gBrowser`), `changeTab: false` during cross-window sync now properly preserves tab selection across all open windows!

mr-cheffy: Sorry, but this looks very vibecoded and this AI generated answer doesn't really help.

SAKTHIVEL280: ok sorry for my mistake, i learnt a lesson :(

SAKTHIVEL280: This was my first contribution ever. I was just inspired by how dedicated you are to this project. so, i thought of helping but it went wrong.

PR #14782, opened 2026-07-30
gh-14770: Fix Bookmark menu styling

### LINKED ISSUE #14770: Broken input field in "Library" (Bookmarks)
### Preliminary Checks

### What happened?

When I click on a bookmark, the fill forms in the "Library" window (for name, url, tags and keyword) seems broken.

### Expected behavior

For it to not be broken.

### Actual behavior

Tags section being broken.

### Steps to reproduce

1. Open `Bookmarks -> Manage Bookmarks`.
2. Click on any bookmark.

### Screenshots and videos

How it looks in Firefox:

Expected look in Firefox

How it look on latest (as of 2026-07-29) Zen browser:

Actual look in Zen

### Version

1.21.9b

Linux (Tarball)

Bookmarks

```shell

```

bnquon:
restores the normal Firefox layout in Bookmarks > Manage Bookmarks.

Zen's field wrappers were treated as single grid cells, misaligning the labels and inputs. it also hides the unused Space selector controls in the Library. `editBookmark.js `disables the selector in this context but its related elements remained visible.

https://github.com/user-attachments/assets/f241f8f5-ff91-450a-9f53-9bfd74e117f4

(the before styling in the original issue is also on macOS which I reproduced on my own)
closes #14770

### DIFF
--- a/src/browser/themes/shared/places/organizer-shared-css.patch
+++ b/src/browser/themes/shared/places/organizer-shared-css.patch
@@ -1,0 +1,21 @@
+diff --git a/browser/themes/shared/places/organizer-shared.css b/browser/themes/shared/places/organizer-shared.css
+index fe6b79cfc2453927c324a044bd62040f866def95..a93813c76ba3633c365a739289b39187c0d1e881 100644
+--- a/browser/themes/shared/places/organizer-shared.css
++++ b/browser/themes/shared/places/organizer-shared.css
+@@ -52,6 +52,16 @@
+   }
+ }
+ 
++#places .zenEditBMPanel_fieldContainer {
++  display: contents;
++}
++
++#places label[control="editBMPanel_workspacesSelectorExpander"],
++#places #editBMPanel_workspaceDropdown,
++#places #editBMPanel_workspaceList {
++  display: none;
++}
++
+ #editBMPanel_itemsCountText {
+   grid-column: auto / span 2;
+   justify-self: center;


mr-cheffy (changes requested):
This would be better to be moved into zen-popups.css and guard them behind `:root[windowtype="..."]`

bnquon:
> This would be better to be moved into zen-popups.css and guard them behind `:root[windowtype="..."]`

hi @mr-cheffy, i tried moving the rules into `zen-popup.css`, but `places.xhtml` doesn’t currently load it. loading it there would also bring the existing unguarded popup styles into the Library, which could cause unrelated side effects.

would you prefer keeping the targeted `organizer-shared.css` patch, creating a dedicated stylesheet and loading it from `places.xhtml`, or another approach? either way after, i’ll update the rules to use the `:root[windowtype="Places:Organizer"]` guard.

mr-cheffy:
Its fine, don't worry. But no need to ping me each time

mr-cheffy (approved):

ashvwinn:
Rather than hiding the "Space Selector Controls", shouldn't we rather fix it?
Cuz there will be no way to edit an existing bookmark's spaces if we just hide it, right?
Well, can be done as a followup ig

bnquon:
yeah, i agree. although it’s currently disabled for `Places:Organizer` in `editBookmark-js.patch`, so i thought making it functional would be better handled as a follow-up rather than in this styling fix PR.

PR #14787, opened 2026-07-30
no-bug: Add keyboard shortcut for adding a tab to Essentials

django23:
Adds a keyboard shortcut for **Add to Essentials**. Discussion: #14784

The action already exists in the tab context menu and the command palette, so this
registers a shortcut for it rather than adding new essentials logic.

### Default binding: `accel+alt+shift+E`

| Combo | Status |
|---|---|
| `accel+E` | Taken, `key_findSelection` |
| `accel+alt+E` | Taken, `key_netmonitor` |
| `accel+shift+E` | Unsafe. `DevToolsStartup` registers devtools with `accel,alt` on Darwin and `accel,shift` elsewhere, so this is Network Monitor on Windows and Linux. The same flip hits `accel+shift+K`, `+I`, `+Z` and `+M`. |
| `accel+alt+shift+E` | **Free on every platform.** Only `C`, `I`, `S`, `]` and `}` are bound under `accel+alt+shift`. |

Registered in `ZEN_OTHER_SHORTCUTS_GROUP`, so it appears in Settings under Other and
stays user-rebindable.

### Notes for review

- New command rather than reusing `cmd_contextZenAddToEssentials`: with no argument,
  `addToEssentials()` falls back to `TabContextMenu.contextTab`, which is null until the
  context menu has been opened once and stale afterwards.
- Guarded on `canEssentialBeAdded`, so the shortcut refuses exactly when the context menu
  greys out and the palette action hides.
- `LATEST_KBS_VERSION` 19 to 20 with a `version < 20` migration, per the "add new default
  shortcuts inside the migration function" comment in `zenGetDefaultShortcuts()`, so
  existing profiles pick up the binding without a shortcut reset.

### Tests

`src/zen/tests/pinned/browser_essential_shortcut.js`, two tasks: the happy path, and the
`canEssentialBeAdded` no-op staged by setting `zen.tabs.essentials.max` to 0. The first
task also asserts no collision via `checkForConflicts`, which walks the merged registry
including Firefox's own keyset.

I could not run the suite locally - a build needs 40-60GB and I do not have the disk for
it - so these are unexecuted and rely on CI. Flagging that rather than implying otherwise.

### DIFF
--- a/locales/en-US/browser/browser/preferences/zen-preferences.ftl
+++ b/locales/en-US/browser/browser/preferences/zen-preferences.ftl
@@ -360,3 +360,4 @@ zen-close-all-unpinned-tabs-shortcut = Close All Unpinned Tabs
 zen-new-unsynced-window-shortcut = New Blank Window
 zen-duplicate-tab-shortcut = Duplicate Tab
 zen-key-find-selection = Find Selection
+zen-add-to-essentials-shortcut = Add Current Tab to Essentials
--- a/src/browser/base/content/zen-commands.inc.xhtml
+++ b/src/browser/base/content/zen-commands.inc.xhtml
@@ -38,6 +38,7 @@
   <command id="cmd_zenEditPinnedUrl" />
   <command id="cmd_contextZenAddToEssentials" />
   <command id="cmd_contextZenRemoveFromEssentials" />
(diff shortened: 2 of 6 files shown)


mr-cheffy:
AI slop

PR #14788, opened 2026-07-30
no-bug: Add keyboard shortcut for moving a tab into a folder

django23:
Adds a keyboard shortcut to move the selected tab into a sidebar folder.
Discussion: #14785

There are currently no folder commands in the shortcut registry, so a folder can only be
filled by dragging tabs in. The **Move to Folder** context submenu cannot be driven from a
command either, since its `popupshowing` reads `TabContextMenu.contextTab`.

> **Depends on #14787.** That PR takes `LATEST_KBS_VERSION` 19 to 20; this one takes it to
> 21. If this merges first, existing profiles jump 19 to 21 and #14787's `version < 20`
> block never runs for them, so its shortcut would silently never appear. Fresh profiles
> run every block, so manual testing would not catch it. Happy to rebase to whatever
> order suits you.

### Default binding: `accel+alt+shift+F`

| Combo | Status |
|---|---|
| `accel+alt+F` | Taken, `key_search2` / `Tools:Search` |
| `accel+shift+F` | Taken, `key_enterFullScreen_old` |
| `accel+alt+shift+F` | **Free on every platform.** Only `C`, `I`, `S`, `]` and `}` are bound under `accel+alt+shift`. |

Three modifiers also sidesteps the devtools platform flip: the bindings that swap
`accel,alt` on Darwin for `accel,shift` elsewhere are all two-modifier, and the one genuine
three-modifier devtools key (`key_browserToolbox`, `accel+alt+shift+I`) uses the same combo
everywhere.

### Targeting

Most recently used folder in the active space, falling back to the first folder in sidebar
order, no-op when the space has none. "Nearest folder" was rejected because a loose tab
sits below every folder, so it always resolves to the bottom-most one. The target folder
and its ancestors are expanded before the move, otherwise a keyboard-only action gives no
feedback.

### Notes for review

- `getTargetFoldersForTabs` is extracted from the context submenu's existing filter rather
  than duplicated. That extraction is the only change to existing code; the context-menu
  call site still calls `group.addTabs(tabs)` and behaves exactly as before.
- `moveTabsToFolder` pins first, mirroring `createFolder`, since folders live in the pinned
  area and `addTabs` leaves pinned state untouched.
- Separately, I think **Move to Folder** has a pre-existing bug: it is offered for unpinned
  tabs with no gating and leaves them unpinned inside the folder. Not touched here. Happy to
  file it and fix it in its own PR if useful.

### Tests

`src/zen/tests/folders/browser_folder_move_shortcut.js`, five tasks: registration plus a
`checkForConflicts` assertion, no-folders no-op, multi-folder fallback with an empty MRU
map, happy path, and last-used targeting in both directions.

Unexecuted - a build needs 40-60GB and I do not have the disk, so these rely on CI.

### DIFF
--- a/locales/en-US/browser/browser/preferences/zen-preferences.ftl
+++ b/locales/en-US/browser/browser/preferences/zen-preferences.ftl
@@ -360,3 +360,4 @@ zen-close-all-unpinned-tabs-shortcut = Close All Unpinned Tabs
 zen-new-unsynced-window-shortcut = New Blank Window
 zen-duplicate-tab-shortcut = Duplicate Tab
 zen-key-find-selection = Find Selection
+zen-move-tab-to-folder-shortcut = Move Tab to Folder
--- a/src/browser/base/content/zen-commands.inc.xhtml
+++ b/src/browser/base/content/zen-commands.inc.xhtml
@@ -61,6 +61,7 @@
   <command id="cmd_zenGlanceSplit" />
 
   <command id="cmd_zenOpenFolderCreation" />
+  <command id="cmd_zenMoveTabToFolder" />
 
   <command id="cmd_zenTogglePinTab" />
   <command id="cmd_zenCloseUnpinnedTabs" />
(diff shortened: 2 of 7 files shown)


mr-cheffy:
There's already a PR for it

PR #14789, opened 2026-07-30
no-bug: Add --workspace flag to open a URL in a named space

django23:
Adds `--workspace <name>` so a URL can be opened into a named space from the command line.
Discussion: #14786

```
zen --workspace "Work" --new-tab https://example.com
```

`--new-tab` currently always targets whichever space happens to be active, so there is no
way to place a tab in a specific space from a script.

### Behaviour

- Opens in the **background** of the target space, without switching to it. `#openInWorkspace`
  never calls `changeWorkspace`; it only sets `lastSelectedWorkspaceTabs[uuid]`, which decides
  the focused tab if and when the user switches there themselves. Switching would drag the
  user through one awaited animation per tab and serialise a batch.
- Unknown name reports an error, lists the known spaces, and opens nothing. It never falls
  back to the active space.
- Duplicate names: first match wins, stated in `helpInfo`.
- No running window with spaces enabled reports an error and creates no tab. One expression
  covers no window, no `gZenWorkspaces`, and spaces disabled (private windows, popups).

### Notes for review

- Implemented as an `nsICommandLineHandler` via `components.conf`, not a Firefox source patch.
- It registers as `b-zen-workspace` so it is asked before Firefox's `m-browser` and can claim
  `--new-tab`. This is the one design decision I would most like a second opinion on. When
  `--workspace` is absent it consumes nothing and returns before touching `preventDefault`, so
  `--new-window`, `-url`, `-private-window`, bare URLs and no-args are untouched - there is a
  test pinning exactly that. If handler ordering ever changed, the failure mode is the
  `--workspace`-without-`--new-tab` branch: a loud error and no tab, never a tab misfiled into
  the wrong space.
- The alternative was a self-contained flag carrying the URL, e.g. `--zen-open-in-space
  "Name=URL"`, which needs no interception. Rejected because `=` is legal in both space names
  and URLs, and because it would have to reimplement `--new-tab`'s URI resolution and then stay
  in sync with it. Swapping to that approach is about ten lines if you would rather not depend
  on registration order.
- `userContextId` is passed explicitly. Without it, `getContextIdIfNeeded` hands the tab the
  *active* space's container, so it would land in the right space with the wrong container.

### Tests

`src/zen/tests/spaces/browser_workspace_command_line.js`, seven tasks, including a tab opening
into a named non-active space (asserting `activeWorkspace` is unchanged), the unknown-name and
no-window error paths, duplicate names, and one that passes `--new-tab`, `--new-window`,
`--private-window` and `-url` together and re-reads all four to prove none were consumed.

Unexecuted - a build needs 40-60GB and I do not have the disk, so these rely on CI. Two things
I could not settle by reading: whether handler ordering resolves as the `b-` prefix implies, and
whether `moveTabToWorkspace` always finds a DOM section for the target space in a long-running
window.

### DIFF
--- a/src/zen/spaces/ZenWorkspaceCommandLine.sys.mjs
+++ b/src/zen/spaces/ZenWorkspaceCommandLine.sys.mjs
@@ -1,0 +1,188 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const lazy = {};
+
+ChromeUtils.defineESModuleGetters(lazy, {
+  BrowserWindowTracker: "resource:///modules/BrowserWindowTracker.sys.mjs",
+});
+
+const WORKSPACE_FLAG = "workspace";
+const NEW_TAB_FLAG = "new-tab";
+
+/**
+ * Command line handler for `--workspace <name>`.
+ *
+ * Firefox's own handler sends `--new-tab` to whichever space happens to be
+ * active. This handler is registered under the "command-line-handler" category
+ * as "b-zen-workspace" so that it is asked first (handlers run in alphabetical
+ * order of their category entry name, and Firefox registers itself as
+ * "m-browser"). When `--workspace` is present it consumes `--new-tab` as well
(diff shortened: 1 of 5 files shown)


mr-cheffy:
There's already a PR for it

PR #14794, opened 2026-07-31
perf: Improve cold startup and page-load speed on Windows

kotasiddharth:
## Summary
This change reduces cold/new-tab first-paint and page-load overhead, especially on Windows with Compact mode and light tab loads.

- **Throttled the sidebar `ResizeObserver`** and raised `zen.view.sidebar-height-throttle` to `50` ms.
- **Deferred non-critical `ZenUIManager` initialisation** (`gZenMediaController`, `gZenLiveFoldersUI`, `_debloatContextMenus`, `_initOmnibox`) out of the startup critical path.
- **Batched `ZenHasPolyfill` observer work** with `requestAnimationFrame` and disconnected the toolbar observer when Compact mode is disabled.
- **Lazy-loaded `ZenGlance` and `ZenBoosts` content actors** so they do less work on every page load.
- **Made workspace restore faster** by racing per-window `SSWindowRestored` instead of waiting for every window, and moved the bookmark refresh off the critical path via `idleDispatchToMainThread`.
- **Debounced `ZenWindowSync` tab events** with a 50 ms batch and deduplicated stale events for closing tabs.
- **Tuned session-store backup defaults** to reduce I/O.
- **Fixed the floating URL bar in animation** so it no longer double-applies the horizontal `translate` and added `in`/`out` keyframes plus a `zen-urlbar-fading-out` rule.

## Test plan
- [ ] `npm run lint` passes on CI (could not run locally because the full engine/Surfer toolchain needs `7z`).
- [ ] Cold start with a new profile feels snappier and first paint completes before all windows/tabs are restored.
- [ ] Compact mode hover show/hide still works.
- [ ] Workspaces with pinned/essential tabs restore correctly after restart.
- [ ] Glance still opens with the configured modifier.
- [ ] Boosts still apply per-site CSS.
- [x] Tabs can be moved/pinned/closed without dropping state.

## Known follow-up
The fade-out keyframe and CSS rule are in place, but `UrlbarInput`/`ZenUIManager` still needs to set `zen-urlbar-fading-out` and defer `zen-floating-urlbar`/`breakout-extend` removal for 300 ms so the out animation actually plays. This can be done in a small follow-up PR.

### DIFF
--- a/prefs/zen/session-store.yaml
+++ b/prefs/zen/session-store.yaml
@@ -13,3 +13,9 @@
 
 - name: browser.sessionstore.max_windows_undo
   value: 2
+
+- name: zen.session-store.max-backups
+  value: 5
+
+- name: zen.session-store.backup-hour-span
+  value: 6
--- a/prefs/zen/view.yaml
+++ b/prefs/zen/view.yaml
@@ -3,7 +3,7 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 - name: zen.view.sidebar-height-throttle
-  value: 0
+  value: 50
 
 - name: zen.view.sidebar-expanded.max-width
   value: 500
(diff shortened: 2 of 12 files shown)


mr-cheffy:
You haven't tested this, have you?

kotasiddharth:
I had only built it on my computer. Let me fix that mistake and send a working pr next time.

kotasiddharth:
Screenshot 2026-07-31 091604

PR #14795, opened 2026-07-31
gh-14168: Fix collapsed workspace toggle hitbox

### LINKED ISSUE #14168: Small hitbox for hiding space
### Preliminary Checks

### What happened?

The hitbox for hidding/extanding is pretty hidden and small

### Expected behavior

There should be a extra hitbox just for toggling

### Actual behavior

The hitbox for toggeling is behind the open settings hitbox

### Steps to reproduce

- Small sidebar
- Have at least one Space entry
- Try to toggle the space visability (Top left worked best for me)

### Screenshots and videos

https://github.com/user-attachments/assets/b1dc7f24-2adf-4e06-89a4-31f375ecfa6a

Expected behavior (big sidebar)
Image

### Version

1.21b

Linux (Flatpak)

Workspaces

bnquon:
follows the suggested approach from the original issue by preventing the workspace actions button from appearing on hover in the collapsed sidebar. this keeps the space icon visible and clickable for collapsing pinned tabs, while the actions menu remains right-clickable.

demo showing sidebar toggling and the style fixes for the hover

https://github.com/user-attachments/assets/62991364-5abe-4dea-a536-22b4a3dd701d

closes #14168


bnquon:
while testing these changes, i noticed that the workspace action in a collapsed unsynced/blank window has no visible icon or label in the current release. the tooltip appears and the action still works when clicked, but the button itself looks blank, as shown while hovering in the attached image.

this is unrelated to the changes in this pr, so i preserved the existing behavior. should this be addressed separately, or would you prefer that i include some suggested change in this PR?

from the current release, red arrow added to show where my cursor is hovering during SS, on top of the blank window workspace collapsed icon
image

mr-cheffy:
Sure, you can address it here

bnquon:
> Sure, you can address it here

added a dedicated icon class for the move to space action in blank windows. contemplated between this icon and the open icon, but went with the corner-down-right arrow since it’s less commonly used elsewhere and felt clearer for moving something. the tooltip and expanded sidebar behavior still work too

https://github.com/user-attachments/assets/dce94bdf-e97e-4c39-9906-8bd4056371cd

mr-cheffy (approved):

mr-cheffy:
Only concern I have so far is that maybe people are used to simply click on the three dots for the workspace actions, but thats not possible anymore...

bnquon:
yeah fair, although i do think this is cleaner than adding another button or finding a separate place for the collapsed action, and it better matches the click behavior of the expanded sidebar

mr-cheffy:
What about making it float to the top left? Like the close buttons on tabs

bnquon:
> What about making it float to the top left? Like the close buttons on tabs

i was able to get it looking like this by adding the same Firefox css from the collapsed tab close button. i checked the left alignment using Shottr’s (screenshot app) ruler, and it looks to have the same gap as the buttons on tabs. the only difference is that i used 1px less padding to make the three dots look a little bigger and clearer. this only applies to normal workspaces, the blank windows keep the move to space arrow from the earlier change.

if this looks fine and you want to go with this lmk and i can commit my local changes 

(hovered workspace icon in sidebar view)
image

mr-cheffy:
That looks great!

bnquon:
okay sounds good! committed those changes

src/zen/tabs/zen-tabs/vertical-tabs.css
```diff
@@ -715,15 +715,65 @@
...
+        --close-button-extra-background: transparent;
+        width: 15px;
+        height: 15px;
+        box-sizing: border-box;
+        padding: 2.5px;
+        margin: 0 !important;
+        -moz-context-properties: fill, fill-opacity;
+        fill: var(--tab-selected-textcolor);
+        background-color: var(--lwt-accent-color);
+        background-image: image(var(--close-button-extra-background)), image(var(--tab-background-color-selected)), image(var(--toolbox-background-color));
```
mr-cheffy, line 752: Whats up with this? Why so many fallbacks

src/zen/tabs/zen-tabs/vertical-tabs.css
```diff
@@ -715,15 +715,65 @@
...
+        color-scheme: var(--tab-selected-color-scheme);
+        outline-color: var(--tab-selected-outline-color);
+        box-shadow: var(--tab-selected-shadow);
+        border-radius: var(--border-radius-circle) !important;
+        transform: rotate(90deg);
+
+        &:-moz-window-inactive {
```
mr-cheffy: No need for this imo

src/zen/tabs/zen-tabs/vertical-tabs.css
```diff
@@ -715,15 +715,65 @@
...
+    & .zen-workspaces-actions:not(.zen-move-tabs-to-workspace-button) {
+      inset-inline-start: calc(var(--zen-toolbox-padding) - 2px);
+      top: -2px;
+      width: 15px;
+      height: 15px;
+      margin: 0 !important;
+      padding: 0 !important;
+      z-index: 1;
+
+      & .toolbarbutton-icon {
```
mr-cheffy, line 742: should show only on hover

mr-cheffy (changes requested):

src/zen/tabs/zen-tabs/vertical-tabs.css
bnquon: those aren’t fallbacks, it's background layers copied from Firefox’s collapsed tab close button styles in `engine/browser/themes/shared/tabbrowser/tabs.css`. they handle the button background and hover states across different themes since i was trying to match the close tab button styles exactly

src/zen/tabs/zen-tabs/vertical-tabs.css
bnquon: the visibility is handled in `zen-workspaces.css`

src/zen/tabs/zen-tabs/vertical-tabs.css
bnquon: removed, thx

src/zen/tabs/zen-tabs/vertical-tabs.css
```diff
@@ -715,15 +715,60 @@
...
+        --close-button-extra-background: transparent;
+        width: 15px;
+        height: 15px;
+        box-sizing: border-box;
```
mr-cheffy, line 746: What's up with these weird attributes being added? Seems really vibecoded. color-scheme as well, etc...

mr-cheffy (changes requested):

src/zen/tabs/zen-tabs/vertical-tabs.css
bnquon: these are taken from Firefox’s [base tab close button styles](https://searchfox.org/firefox-main/source/browser/themes/shared/tabbrowser/tabs.css#866-873) and [collapsed vertical tab close button styles](https://searchfox.org/firefox-main/source/browser/themes/shared/tabbrowser/tabs.css#2666-2705) for consistency

PR #14819, opened 2026-08-02
gh-14483: Fix boost font override breaking certain icons

### LINKED ISSUE #14483: Boosts: custom font override breaks all icon-based UI elements on Claude.ai (regression/gap in #13839 fix)
### Preliminary Checks

### What happened?

When applying a custom font via Boosts, all icon-based UI elements across Claude.ai break. This includes the message action buttons (copy, read aloud, retry), sidebar navigation icons (New chat, Chats, Projects, Artifacts, Code, Customize, Design), the Share button, and the chat input icons. Instead of rendering as icons, they show raw font codepoints (e.g. E05E, E0C1, E0F8, E10D).
This appears related to #13839 (Material Symbols are broken when using custom font through boosts), which was fixed in 1.20.2b via PR #13956. However, this issue still happens on 1.21.5b, so the fix doesn't seem to cover this icon set or this site.
Confirmed reproducible with all Mods and Custom CSS disabled, so this is isolated to Boosts' font override behavior itself, not a mod conflict.

### Expected behavior

Icon buttons should render normally no matter what font is selected in Boosts, since these are icon font glyphs, not regular text.

### Actual behavior

Icon buttons across the entire site show raw codepoint placeholders instead of icons once a custom font is applied through Boosts.

### Steps to reproduce

1. Go to claude.ai and start any conversation
2. Open Boosts on this site and apply a custom font (e.g. Arial)
3. Look at the sidebar icons, message action icons, and chat input icons
4. All of them show codepoint text instead of rendering as icons

### Screenshots and videos

Image

Image

### Version

1.21.5b (64-bit)

Windows - x64

Other

```shell

```

lokendra-singh-dhirawat:
Fix: #14483
Just adds a `span[data-cds='Icon']` element in the ignored elements in the styles applying the font override.
I also extracted the elements into an array as the list was getting a bit long.


src/zen/boosts/ZenBoostStyles.sys.mjs
```diff
@@ -81,8 +81,15 @@ export class nsZenBoostStyles {
     }
 
     if (fontCase != "" || fontFamily != "") {
+      const ignoredElements = [
+        ".google-symbols",
+        "gf-load-icon-font",
+        "mat-icon",
+        ".google-material-icons",
+        "span[data-cds='Icon']",
```
mr-cheffy: Why not ignore all `span` elements while we are at it?

mr-cheffy (changes requested):

src/zen/boosts/ZenBoostStyles.sys.mjs
lokendra-singh-dhirawat: I have ignored all `span` elements in the styles as instructed, but I did keep the previous elements that were ignored as removing them might break some sites.

mr-cheffy (approved):

mr-cheffy:
Reverted for https://github.com/zen-browser/desktop/issues/14899

PR #14822, opened 2026-08-03
gh-11291: Fix toolbar expansion in compact mode

### LINKED ISSUE #11291: Collapsed Toolbar: Ctrl+T Temporarily Expands Top Toolbar in Compact Mode
### Preliminary Checks

### What happened?

In collapsed toolbar, compact mode when I use CTRL + T, it shows top toolbar and it looks very unpleasing. 

Image

### Expected behavior

Should not have displayed the top toolbar in this configuration.

### Actual behavior

Displayed top toolbar when using CTRL + T

### Steps to reproduce

1. Change the configuration to collapsed toolbar, compact mode
2. Hit CTRL + T 

### Screenshots and videos

_No response_

### Version

1.17.6b

Windows - x64

Compact Mode

```shell

```

mcsharland:
Fixes #11291


src/zen/compact-mode/toolbar.inc.css
```diff
@@ -41,6 +41,11 @@
       transition: opacity var(--zen-hidden-toolbar-transition);
     }
 
+    /* Prevent fading after breakout-extend is removed */
+    & #urlbar[zen-newtab]:not([breakout-extend]) {
```
mr-cheffy, line 45: Shouldn't it be using `zen-floating-urlbar`?

mr-cheffy (changes requested):

src/zen/compact-mode/toolbar.inc.css
mcsharland: Making the toolbar hide in compact mode introduced a new issue where the top urlbar would briefly flash after the floating urlbar closed. `zen-floating-urlbar` is removed before `breakout-extend`, so the selector never matched. `zen-newtab` remains throughout the handoff. Just tested to confirm.

~Alternatively, I could try reversing their removal order and use `zen-floating-urlbar` if that's preferred~

E: That did not prevent the flash. The original implementation is what I'd stick with

PR #14828, opened 2026-08-03
gh-11210: Fix non-contiguous tab indices from browser.tabs.query

### LINKED ISSUE #11210: Broken [browser|chrome].tabs.query({}) making Tab Session Manager unusable
### Preliminary Checks

### What happened?

Tab Session Manager uses `await browser.tabs.query({});` to save the current session.
Zen Browser hasn’t worked as expected since versions **v1.10b** → **1.17.4b**.
Previous versions didn’t have this problem. (1.9.1b-1 on Arch Linux works)
Lookings for workarounds , `chrome.windows.getAll({ populate: true })` also don't work.

Issue downstream.
https://github.com/sienori/Tab-Session-Manager/issues/1542

A fresh installation without any extensions (except  Tab Session Manager) also experiences this problem although not consistently.
I suspect it might be related to the following settings (on a fresh installation):

```bash
zen.urlbar.replace-newtab
browser.tabs.closeWindowWithLastTab
```

The extension uses `browser` from `webextension-polyfill` but it also occurs on `chrome.tabs.query`

### Expected behavior

Session saved or all tabs returned.

### Actual behavior

Errors, session not saved, `Uncaught (in promise) Error: An unexpected error occurred undefined
`
### Steps to reproduce

- Install Tab Session Manager
- Load a session with  (or open new) several windows
- Try to save the session

or 
- open the debug console of a extension background
- execute `await browser.tabs.query({});`

### Screenshots and videos

tabs.query failing:
Image
windows.getAll failing:
Image
Brave (Chromium, working without problems):
Image
Extension failing:
Image

### Version

since v1.10b

macOS - aarch64, Linux (Tarball)

Tabs

Opening the dev tools on any extension background process (with 'tabs' and 'windows' permissions)
You can see `browser.tabs.query` don't work without filters.

```shell
await browser.tabs.query({currentWindow:true});
Array(4) [ {…}, {…}, {…}, {…} ]

await chrome.tabs.query({currentWindow:true});
Array(4) [ {…}, {…}, {…}, {…} ]

await browser.tabs.query({currentWindow:false, active: true});
Array(17) [ {…}, {…}, {…}, {…}, {…}, {…}, {…}, {…}, {…}, {…}, … ]

await browser.tabs.query({});
Uncaught (in promise) Error: An unexpected error occurred undefined
await chrome.tabs.query({});
Uncaught (in promise) Error: An unexpected error occurred undefined

await chrome.windows.getAll({ populate: true })
Uncaught (in promise) Error: An unexpected error occurred undefined
await browser.windows.getAll({ populate: true })
Uncaught (in promise) Error: An unexpected error occurred undefined
Background event page was not terminated on idle because a DevTools toolbox is attached to the extension.
```

80avin:
Fixes https://github.com/zen-browser/desktop/issues/11210

First observed after https://github.com/zen-browser/desktop/commit/43c98b7ddd78ff4c1beedf4348021da1ca63c071

Issue: The folders & "New Tab" were created as tabs but filtered from the `browser.tabs.query` API without adjusting the `tab.index` leading to non contiguous indices.

After this fix:
image

### DIFF
--- a/src/browser/components/extensions/parent/ext-browser-js.patch
+++ b/src/browser/components/extensions/parent/ext-browser-js.patch
@@ -1,35 +1,31 @@
 diff --git a/browser/components/extensions/parent/ext-browser.js b/browser/components/extensions/parent/ext-browser.js
-index 91adc6bc3649af47745818e6f9d00cbbdffc35a5..cd83459693067211f030b6354b839a555df2e822 100644
+index 91adc6bc3649af47745818e6f9d00cbbdffc35a5...4ff10dc750e037dbc9c342f616b4132ee6ef2865 100644
 --- a/browser/components/extensions/parent/ext-browser.js
 +++ b/browser/components/extensions/parent/ext-browser.js
-@@ -27,6 +27,59 @@ function isPrivateTab(nativeTab) {
+@@ -27,6 +27,40 @@ function isPrivateTab(nativeTab) {
    return PrivateBrowsingUtils.isBrowserPrivate(nativeTab.linkedBrowser);
  }
  
-+// Zen hides placeholder "empty tabs" (the workspace New Tab button and one
-+// per folder) from extensions, so extension-facing indices are "logical"
-+// indices that exclude them, translated to physical gBrowser.tabs positions
-+// at the boundaries below.
++// Zen hides placeholder "empty tabs" (workspace New Tab button, one per
++// folder) from extensions; _zenLogicalIndex holds the extension-facing index.
 +function zenIsEmptyTab(nativeTab) {
-+  return (
-+    nativeTab && nativeTab.hasAttribute && nativeTab.hasAttribute("zen-empty-tab")
-+  );
++  return nativeTab?.hasAttribute("zen-empty-tab");
 +}
 +
-+// Extension-visible index of nativeTab: count of non-empty tabs before it.
-+function zenLogicalTabIndex(nativeTab) {
-+  let gBrowser = nativeTab.documentGlobal.gBrowser;
++// Recompute _zenLogicalIndex for every tab; skip the closing tab, which is
++// still listed when TabClose fires.
++function zenSyncTabIndices(gBrowser, excludedTab) {
++  // gBrowser.tabs is cached and can be stale at event time.
++  gBrowser.tabContainer._invalidateCachedTabs();
 +  let logical = 0;
-+  let pos = nativeTab._tPos;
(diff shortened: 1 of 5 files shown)


src/browser/components/extensions/parent/ext-browser-js.patch
```diff
@@ -1,29 +1,112 @@
...
+   return PrivateBrowsingUtils.isBrowserPrivate(nativeTab.linkedBrowser);
+ }
+ 
++// Zen hides placeholder "empty tabs" (the workspace New Tab button and one
++// per folder) from extensions, so extension-facing indices are "logical"
++// indices that exclude them, translated to physical gBrowser.tabs positions
++// at the boundaries below.
++function zenIsEmptyTab(nativeTab) {
++  return (
++    nativeTab && nativeTab.hasAttribute && nativeTab.hasAttribute("zen-empty-tab")
```
mr-cheffy, line 15: Is this AI generated?

mr-cheffy (changes requested):

src/browser/components/extensions/parent/ext-browser-js.patch
80avin: Kind of.
I wrote such a pattern while debugging & trying fixes. Then used AI to clean/generalize it.

This pattern was referenced after looking at https://github.com/zen-browser/desktop/blob/952fb30547aae9ca1eb1e8ae690327e19287ead8/src/toolkit/content/widgets/findbar-js.patch#L9

mr-cheffy: Seems to be a bit expensive to run a loop every time the index gets accessed...

80avin: Fair call.

Then, I guess, the better approach could be to add logical index directly as an attribute, recomputed on tab creation/deletion.
It can also possibly replace the `zen-empty-tab` attribute as logical index of undefined/null/-1 can be made to mean same as zen-empty-tab.

Does it sound better ?

---

EDIT:
* Replacing `zen-empty-tab` is too much complexity as zen-empty-tab is used in ~80 places including CSS. 
* a cached _zenLogicalIndex property will save query time but reverse lookup will still have similar O(n) complexity.
* One suggestion is to accept the O(n) time complexity for a simple implementation. Else, let me know what kind of complexity is accepted here.

80avin: @mr-cheffy hope it suffices now.

PR #14832, opened 2026-08-03
gh-14171: Fix cross-space routing tab lifecycle races

### LINKED ISSUE #14171: Opening a link under a Space routing rule then closing the tab it opened makes the browser unusable
### Preliminary Checks

### What happened?

I open a link from another app (in this example reddit link) which I specified with space routing rules to open in a different work space than the one I'm currently operating in
If I close this tab I notice that the url still says "reddit.com" and the whole browser becomes so sluggish and don't open any tabs or do anything unless I close it and open it again
I notice that this doesn't happen if I changed to another tab before I close the one I just opened (it's like the browser doesn't know which tab should it go back to when this one closes)

### Expected behavior

It should let me close the tab and goes back to another one or to the original workspace

### Actual behavior

It becomes unusable

### Steps to reproduce

I open a tab in workspace A
click on a link from another program that is supposed to open in workspace B
Don't switch from this tab and close it
The browser becomes unusable

### Screenshots and videos

https://github.com/user-attachments/assets/7582831e-a2c5-4671-8641-3069ee0b4ee1

### Version

1.21.1b

Linux (Flatpak)

Workspaces

```shell

```

### LINKED ISSUE #15038: Space Routing: closing a routed tab with tabs.remove() leaves a dead AsyncTabSwitcher and breaks tab switching for the window
### Preliminary Checks

- [X] I have read and understood the important section above.
- [X] I have checked that this issue can be reproduced once I removed all my Mods and Custom CSS.

### What happened?

Closing a Space-Routed tab through the WebExtension `tabs.remove()` API leaves the window with a permanently broken `AsyncTabSwitcher`. Afterwards clicking any tab does nothing: the strip highlights the new tab but the content area never changes, and only restarting Zen clears it. The process stays alive and `sample` shows the main thread parked in the normal event loop, so it is broken state rather than a hang.

It is a race between the synchronous teardown `tabs.remove()` performs and the asynchronous space change Space Routing starts when a routed tab is closed:

1. `gBrowser.removeTab()` only defers `_endRemoveTab` behind `gZenVerticalTabsManager.animateItemClose` when the caller passes `animate`. Closing by mouse or keyboard passes it; the extension API does not. So `_endRemoveTab` runs in the same tick and calls `browser.destroy()`, nulling the tab's `linkedBrowser`.
2. The space change still in flight then builds a fresh `AsyncTabSwitcher` whose `requestedTab` is that already-destroyed tab.
3. `AsyncTabSwitcher.sys.mjs:165-167` dereferences it immediately:

```js
let initialTab = this.requestedTab;
let initialBrowser = initialTab.linkedBrowser;
let tabIsLoaded = !initialBrowser.isRemoteBrowser || …   // initialBrowser is null
```

The constructor throws, but `gBrowser._switcher` has already been assigned, so the window keeps a half-initialised switcher forever. `gZenWorkspaces.changeWorkspace` catches and logs the throw, so nothing surfaces to the user and `#updateWorkspaceState` never runs. Every later tab switch then fails inside that dead object: `preActions()` trips both of its `assert(this.tabbrowser._switcher …)` checks and `postActions()` throws on the null `linkedBrowser`. In the log below that repeats about 1670 times between the failure and the force-quit.

`_switcher` is per-window state with no recovery path, so one caught exception disables tab switching for the life of the window.

On a small profile the same `tabs.remove()` fails differently: when the routed tab is the only tab in its space, the window is left with no selected tab at all (`tabs.query()` reports none active, `captureVisibleTab()` throws) and paints nothing until another tab is opened. That variant reproduces deterministically with the steps below, on a brand-new profile with no mods and no custom CSS. I could not reproduce the `AsyncTabSwitcher` variant synthetically, since it needs a session large enough for a switcher to be in play, so that trace comes from a real session.

Not reproducible on Firefox: Space Routing and `gZenWorkspaces` are Zen-only, and the trigger is the space change that Space Routing starts on close, so there is no equivalent code path in Firefox to hit.

### Expected behavior

Closing a Space-Routed tab via `tabs.remove()` leaves tab switching working, exactly as closing it with the mouse does.

### Actual behavior

`AsyncTabSwitcher`'s constructor throws on the destroyed tab, `gBrowser._switcher` is left half-built, and every subsequent tab click silently fails until Zen is restarted.

### Steps to reproduce

1. Start a new profile, with no mods and no custom CSS. Create two spaces, `SpaceA` and `SpaceB`.
2. In Space Routing settings add one rule: `routetest`, match type `contains`, open in `SpaceB`.
3. Load any extension with the `tabs` permission that can run script from a background page.
4. With `SpaceA` active, run this from that background page:

```js
// routed to SpaceB, which also switches the window to SpaceB
const tab = await browser.tabs.create({ url: "https://example.com/?routetest", active: true });
await new Promise(r => setTimeout(r, 5000));
await browser.tabs.remove(tab.id);
```

5. The window is now on an empty `SpaceB` with no selected tab and nothing painted. `browser.tabs.query({currentWindow: true})` reports no active tab and `browser.tabs.captureVisibleTab()` throws.
6. For the `AsyncTabSwitcher` variant, do the same on a session large enough to have an active tab switcher (mine: 106 tabs across 5 spaces), with several other tabs already in `SpaceB`, and close the routed tab from the extension. Tab switching then breaks for the whole window.

Closing the same tab with the mouse or keyboard never triggers either variant, which matches `animate` being the only difference on the removal path.

### Screenshots and videos

_No response_

### Version

1.21.15b (126.8.18)

macOS - aarch64

Workspaces

```shell
Captured with devtools.console.stdout.chrome=true. Trimmed to the failure,
repeated frames collapsed, paths and hosts redacted.

console.debug: "ZenSessionManager:" "Restored sidebar data into window" ({tabs:106, groups:0, folders:0, spaces:5})

console.error: "gZenWorkspaces: Error changing workspace" (new TypeError("can't access property \"isRemoteBrowser\", initialBrowser is null", "moz-src:///browser/components/tabbrowser/AsyncTabSwitcher.sys.mjs", 169))
JavaScript error: chrome://global/content/elements/browser-custom-element.mjs, line 804: TypeError: can't access property "secureBrowserUI", this.browsingContext is null
Assertion failure
assert@moz-src:///browser/components/tabbrowser/AsyncTabSwitcher.sys.mjs:501:36
preActions@moz-src:///browser/components/tabbrowser/AsyncTabSwitcher.sys.mjs:568:10
handleEvent@moz-src:///browser/components/tabbrowser/AsyncTabSwitcher.sys.mjs:1113:12
Assertion failure
assert@moz-src:///browser/components/tabbrowser/AsyncTabSwitcher.sys.mjs:501:36
preActions@moz-src:///browser/components/tabbrowser/AsyncTabSwitcher.sys.mjs:569:10
handleEvent@moz-src:///browser/components/tabbrowser/AsyncTabSwitcher.sys.mjs:1113:12
JavaScript error: moz-src:///browser/components/tabbrowser/AsyncTabSwitcher.sys.mjs, line 626: TypeError: can't access property "isRemoteBrowser", this.requestedTab.linkedBrowser is null

... the frames above then repeat until the browser is force-quit:
      AsyncTabSwitcher assert failures (preActions)  : 1672
      TypeError requestedTab.linkedBrowser is null    : 836
      TypeError initialBrowser is null (constructor)  : 11

sample on the parent process in this state shows the main thread parked in
nextEventMatchingMask, confirming broken state rather than a hang.
```

patlux:
## Summary

Fixes #14171.

Cross-Space routing could leave Firefox's `AsyncTabSwitcher` referencing a routed tab after its `linkedBrowser` had been destroyed. Closing that tab could then make every tab's content area unusable until Zen was restarted.

## Root cause

Two related tab lifecycle races could occur:

1. `addTab()` selected a foreground tab before Zen moved it into its routed Space and container.
2. During animated tab removal, the routed tab could still be selected when its browser was destroyed.

This left `selectedTab` or `AsyncTabSwitcher.requestedTab` pointing to a tab without a live `linkedBrowser`.

## Changes

- Defer foreground selection when a tab is routed into a different active Space.
- Let the workspace transition select the routed tab after it belongs to the destination Space and container.
- Before destroying a still-selected tab's browser, synchronously select another connected, non-closing tab.
- Prefer a visible replacement tab, then fall back to any valid tab.
- Add browser-chrome regression coverage for routing selection order and animated close followed by an immediate Space change.

## Test plan

- Built Zen/Firefox successfully on macOS arm64.
- Ran the focused browser-chrome test:

  ```sh
  ./mach test zen/tests/space_routing/browser_space_routing_on_add_tab.js --headless
  ```

- Verified the new tests pass their functional assertions:
  - the source tab remains selected when the cross-Space transition starts;
  - the routed tab uses the destination Space's container;
  - closing the routed tab leaves a selected tab with a live `linkedBrowser`;
  - the closed routed tab is not selected again after returning to the source Space;
  - the final selected tab remains connected and usable.
- Verified no `requestedTab.linkedBrowser is null`, `oldBrowser is null`, or other `AsyncTabSwitcher` failure occurs with the fix.
- Verified the source patch reverses cleanly from the generated engine and the PR diff has no whitespace errors.

The local focused suite still exits non-zero because of unrelated existing `targetWorkspaceName` expectations and a missing installed Fluent resource (`menu-bookmark-tab`). These failures are outside the new lifecycle assertions.


mr-cheffy:
I doubt these changes actually fix the issue? Could you please showcase a video of before and after?

patlux:
Not saying its the best solution for the issue but it works. I'm already using it because zen is not usable without it for me. See attached recordings below.

I also committed a script to reproduce it automatically:

```bash
# Without the fix
ZEN_REPRO_HEADLESS=0 \
     ./scripts/reproduce-space-routing-close-race.sh baseline

# With the fix
ZEN_REPRO_HEADLESS=0 \
     ./scripts/reproduce-space-routing-close-race.sh fixed
```

| # | Video  |
|--------|--------|
| Before |  Screenshot_2026-08-04_10-56-41 |
| After |Screenshot_2026-08-04_10-54-39 |

OfryL:
THANK YOU!

scripts/reproduce-space-routing-close-race.sh
mr-cheffy, line 1: No need for this

src/zen/tests/space_routing/README-click-close-repro.md
mr-cheffy, line 1: No need for this

src/browser/components/tabbrowser/content/tabbrowser-js.patch
```diff
@@ -733,6 +738,27 @@ index c42b1a1a8df6b38886c17f71ea88e5aaa7eebc80..c741404ed973ee3ac246fe246c0f645e
        } else {
          TabBarVisibility.update();
        }
+@@ -6019,6 +6215,20 @@
+
+       var browser = this.getBrowserForTab(aTab);
+
++      if (this.selectedTab === aTab) {
```
mr-cheffy: Shouldn't this be handled on `blurTab`?

mr-cheffy (changes requested):

src/browser/components/tabbrowser/content/tabbrowser-js.patch
patlux: Let me try

patlux: Works 👌

unisons:
I am hugely affected by this bug. Is there any binary to test this patch? I tried but failed to build.

src/browser/components/tabbrowser/content/tabbrowser-js.patch
```diff
@@ -383,8 +383,13 @@ index c42b1a1a8df6b38886c17f71ea88e5aaa7eebc80..c741404ed973ee3ac246fe246c0f645e
...
        if (pinned) {
          this.#notifyPinnedStatus(t);
-@@ -3553,6 +3686,15 @@
-       if (!inBackground) {
+@@ -3553,6 +3686,19 @@
+-      if (!inBackground) {
++      const shouldDeferRoutedTabSelection =
++        beforeRouteResult.isRouteFound &&
++        beforeRouteResult.targetRoute !== gZenWorkspaces.activeWorkspace &&
++        window === Services.wm.getMostRecentWindow("navigator:browser");
```
mr-cheffy: Why check for the window exactly?

src/zen/tests/space_routing/browser_space_routing_on_add_tab.js
```diff
@@ -290,6 +290,216 @@ add_task(async function test_onAfterAddTab_ignores_missing_before_result() {
...
+    if (!targetIdentity) {
+      ok(
+        true,
+        "No second container is available for the routing regression test"
+      );
+      return;
+    }
+
+    const originalTestingEnabled = gZenUIManager.testingEnabled;
+    gZenUIManager.testingEnabled = false;
```
mr-cheffy: Why?

mr-cheffy (changes requested):

mr-cheffy:
@patlux Also, if you get the git conflicts fixed, we can get this merged right away :)

src/browser/components/tabbrowser/content/tabbrowser-js.patch
patlux: Selection is deferred only for the window that will perform the follow-up workspace change. `onAfterAddTab()` intentionally activates the target workspace only when `win` is the most-recent browser window; other windows only move the tab. If selection were deferred in those windows too, no later operation would select the foreground tab.

I moved this predicate into `shouldDeferTabSelection()` so the selection and routing paths now use the same condition.

src/zen/tests/space_routing/browser_space_routing_on_add_tab.js
patlux: Browser-chrome tests enable Zen testing mode, where `gZenWorkspaces.selectEmptyTab()` intentionally returns early. This regression needs the destination workspace's real empty fallback tab because it participates in the close/selection lifecycle.

I added an explanation and moved the temporary override into a shared `try/finally` helper so the global test state is always restored. The focused test now passes 28/28 assertions.

patlux:
@mr-cheffy All requested changes and review comments are addressed now. The branch is updated with `dev`, both review threads are answered/resolved, and the focused browser-chrome test passes 28/28 assertions with no unexpected results. Ready for another look when you have time.

src/browser/components/tabbrowser/content/tabbrowser-js.patch
```diff
@@ -805,7 +809,11 @@ index 39b5babee5942427a82a7a64ad69e1b16efebcc1..32beac13dd4215a16de3a783db6c32da
      }
  
      _blurTab(aTab) {
-@@ -6437,7 +6639,7 @@
+-      this.selectedTab = this._findTabToBlurTo(aTab);
++      gZenWorkspaces.selectTabToBlur(aTab, this._findTabToBlurTo(aTab));
```
mr-cheffy: Shouldn't we investigate why `_findTabToBlurTo` is returning a closing tab?

src/browser/components/tabbrowser/content/tabbrowser-js.patch
```diff
@@ -383,8 +383,12 @@ index 39b5babee5942427a82a7a64ad69e1b16efebcc1..32beac13dd4215a16de3a783db6c32da
...
        if (pinned) {
          this.#notifyPinnedStatus(t);
-@@ -3611,6 +3744,15 @@
-       if (!inBackground) {
+@@ -3608,9 +3741,18 @@
+ 
+       gSharedTabWarning.tabAdded(t);
+ 
+-      if (!inBackground) {
++      if (!inBackground && !gZenSpaceRoutingManager.shouldDeferTabSelection(beforeRouteResult, window)) {
```
mr-cheffy: Maybe we should be setting inBackground to true if `gZenSpaceRoutingManager.shouldDeferTabSelection` is true to avoid future issues?

mr-cheffy (changes requested):

1000Reasons:
Any updates on this? It's really annoying to use the space routing right now.

unisons:
I really need this patch!

patlux:
sorry I forgot about this bc I'm already using the fix 😜. thanks for the reminder!

src/browser/components/tabbrowser/content/tabbrowser-js.patch
patlux: Good point. Runtime tracing showed that `_findTabToBlurTo()` returned the live owner from the source Space, not a closing tab. `getTabsToExclude()` only scanned the active Space via `gBrowser.tabs`, so selecting that owner was a no-op. I now scan all stored tabs, removed the defensive fallback, restored the normal Firefox `_blurTab()` path, and added direct regression assertions. The focused test passes 31/31.

src/browser/components/tabbrowser/content/tabbrowser-js.patch
patlux: Agreed. Deferred routed tabs now use `initiallyInBackground` for owner assignment and initial selection. The original `inBackground` value remains unchanged for routing and Compact Mode, so foreground tabs still activate their destination Space without triggering background-tab behavior. The focused test passes 31/31.

patlux:
@mr-cheffy done. please review :)

888markus:
When is this finally coming i am suffering because of this bug so much ...

mr-cheffy (changes requested):
Ugh, sorry about that. Could you please resolve conflicts one more time? I can merge right after. But please use `npm run export <file>` so they are exported correctly.

unisons:
@patlux Just a gentle reminder to check on this when you have a moment. No rush!

patlux:
@mr-cheffy done :)

src/browser/components/tabbrowser/content/tabbrowser-js.patch
```diff
@@ -306,6 +306,7 @@ index 43180487a2fac15b7a65fac94f233eeed9094990..ef6d30eb681caeae48d15280687b3bc3
 +        return null;
 +      }
 +
++      const initiallyInBackground = inBackground || gZenSpaceRoutingManager.shouldDeferTabSelection(beforeRouteResult, window);
```
mr-cheffy: Why don't we set `inBackground ||= gZenSpaceRoutingManager.shouldDeferTabSelection(beforeRouteResult, window);`?

mr-cheffy (changes requested):

src/browser/components/tabbrowser/content/tabbrowser-js.patch
patlux: I tested that assignment: both lifecycle tests fail because `onAfterAddTab()` receives `inBackground: true` and skips activating the destination Space. `initiallyInBackground` only controls initial selection and default owner assignment; the original value is preserved for routing and Compact Mode.

mr-cheffy: I'd rather we take track of something like `originallyInBackground` (ugly name, just an example) and override inBackground for 2 reasons: 1. Less patch size, 2. in case we miss some extra background-specific functionality between updates.

patlux: Agreed, updated as suggested: I save `originallyInBackground`, override `inBackground` for Firefox’s tab-creation logic, and pass the original value to routing and Compact Mode.

mr-cheffy (approved):
Looks good, thanks!

Sorry for the delays.

PR #14855, opened 2026-08-05
gh-13761: Fix closing split tab in one window changing focus of other windows

### LINKED ISSUE #13761: Closing a Split Tab in one Zen window causes other windows' tabs to change to the remaining tab

ashvwinn:
Fixes: #13761


ashvwinn:
honestly, i cannot figure out any other way to solve this

mr-cheffy (approved):

PR #14858, opened 2026-08-05
gh-14300: Fix URL in addressbar not restored

### LINKED ISSUE #14300: URL in the addressbar not restored
### Preliminary Checks

### What happened?

When I'm using ⌘+L (on mac) and change the URL, when I then cancel the operation (press ESC) the URL in the main addressbar doesn't change back to the original, but contains the new value. See animation.

### Expected behavior

The URL in address bar should be reverted

### Actual behavior

The URL in the address bar shows the changed URL that was never "submitted"

### Steps to reproduce

- open a website (e.g. google.com)
- ⌘ + L 
- change the URL but press ESC after the changes
- observe the value in browser's addressbar

### Screenshots and videos

Image

### Version

1.21.3b (Firefox 152.0.1) (aarch64)

macOS - aarch64

URL Bar

```shell

```

mcsharland:
Presently, when the URL bar is opened with ⌘L / Ctrl + L, closing it is handled by Firefox. Pressing Escape closes the URL bar view, but does not revert the address bar to the current URI or its blank state.

This behavior makes sense in Firefox, because closing the view only closes the search suggestions and the address bar remains focused. This change in turn makes behavior visually consistent with opening the URL bar through a new tab and with Arc.

Fixes: #14300

### DIFF
--- a/src/browser/components/urlbar/UrlbarController-sys-mjs.patch
+++ b/src/browser/components/urlbar/UrlbarController-sys-mjs.patch
@@ -19,3 +19,12 @@ index a05894d593e4149097c473822a38f87b2220625f..1660106c2a6548920fdffd2656e5a1f1
          }
          break;
        }
+
+@@ -362,6 +362,7 @@
+    switch (event.keyCode) {
+      case KeyEvent.DOM_VK_ESCAPE:
+        if (executeAction) {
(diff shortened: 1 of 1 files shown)


mr-cheffy:
See comment in issue

Issue #14862, opened 2026-08-06
Mod+scroll space switching also switches tabs when switchByScrolling is enabled


jupjohn:
### Preliminary Checks

- [x] I have read and understood the important section above.
- [x] I have searched existing issues and avoided creating duplicates.
- [x] I am not filing an enhancement request.
- [x] I have checked that this issue cannot be reproduced on Mozilla Firefox.
- [x] I have checked that this issue can be reproduced once I removed all my Mods and Custom CSS.

### What happened?

When `toolkit.tabbox.switchByScrolling` is enabled, changing workspaces with <kbd><kbd>`zen.workspaces.scroll-modifier-key`</kbd>+<kbd>Scroll Up/Down</kbd></kbd> also causes tabs in the originating workspace to be switched, as if normally scrolling. As a weird side-effect, previously open tabs become visually selected - it's purely visual as the context menu doesn't have any actions that selected tabs normally have.

### Expected behavior

The previous/next workspace is switched to, retaining the workspace's selected tab

### Actual behavior

The workspace switch occurs, but the tabs in the previous workspace also switch & visually select

### Steps to reproduce

1. Set `toolkit.tabbox.switchByScrolling` to `true`
2. Set `zen.workspaces.scroll-modifier-key` to `ctrl`
3. Create 2 workspaces
4. Create multiple tabs in each workspace
5. Select the middle-ish tab in each workspace
6. Use <kbd><kbd>Ctrl</kbd>+<kbd>Scroll Up/Down</kbd></kbd> to scroll between the workspaces

### Screenshots and videos

Timeline:
- **[00:07]** - Scrolling between spaces with <kbd><kbd>Ctrl</kbd>+<kbd>Scroll Up/Down</kbd></kbd>
- **[00:17]** - Showing context menu to confirm tabs are not actually selected, just visually
- **[00:20]** - Clicking on visually-selected tabs which "deselects" them
- **[00:26]** - Explicitly selecting multiple tabs, and showing context menu to compare real selection against

https://github.com/user-attachments/assets/1c7564ee-d333-4215-9c06-c0d9d54358f9

### Version

1.21.10b (Firefox 153.0.1) (64-bit)

### What platform are you seeing the problem on?

Windows - x64

### What component is this issue related to?

Tabs

### Relevant log output if applicable

```shell
N/A
```

PR #14881, opened 2026-08-07
gh-14862: Fix tab switching when switching spaces with mouse wheel

ashvwinn:


mr-cheffy:
Ill close this for cleanup

PR #14916, opened 2026-08-11
fix(tabs): keep collapsed sidebar centered when Windows always shows scrollbars

### LINKED ISSUE #12402: Windows: Scroll bar in tab list causes layout shift when always show scroll bars is enabled
### Preliminary Checks

### What happened?

Windows has the accessibility option to always show all scrollbars. 
(Windows Settings -> Accessibility -> Visual Effects -> Always show scrollbars)

The tab list on the left shows a scroll bar when there are more tabs than fit in the vertical space.
This works great when the sidebar is expanded:

Image

But when the sidebar is collapsed, the Tab list is offset to the left by the width of the scrollbar, losing its left margin, and the new tab button is cut off on the left side by the edge of the window:

Image

Without this accessibility feature turned on, the sidebar works as expected, all buttons are aligned:

Image

### Expected behavior

Keep buttons aligned, expand the sidebar to make space for the scroll bar or shrink the tab buttons instead of offsetting

### Actual behavior

Buttons get misaligned by scrollbar

### Steps to reproduce

1. Go to Windows Settings -> Accessibility -> Visual Effects -> Turn of "Always show scrollbars"
2. Collapse the sidebar
3. Hover over the new tab button. It is now misaligned, pushed too far to the left.

### Screenshots and videos

_No response_

### Version

1.18.7b

Windows - x64

Tabs

```shell

```

feichuans:
Closes #12402

## Problem
With Windows *Always show scrollbars* enabled, the collapsed sidebar tab list shifts left and the new-tab button is clipped when the active workspace overflows. The tab strip is only about 60px wide, so the classic scrollbar consumes a large part of its inline space.

## Fix
Use the platform-standard `scrollbar-gutter: stable both-edges` only when the collapsed tab strip has the existing `[overflow]` state:

- The native scrollbar remains visible and draggable.
- The opposite gutter keeps the icon-only content centered in the full sidebar.
- Collapsed tabs, folders, split-view groups, and the new-tab button can shrink to the available content width instead of overflowing and being clipped.
- When the tab strip does not overflow, the existing layout is unchanged.
- Overlay scrollbars do not consume a gutter, so macOS/Linux overlay behavior is not artificially widened.

The existing `#tabbrowser-tabs[overflow]` state is maintained by `ZenSpaceManager.updateOverflowingTabs()` from the active workspace scrollbox, so this does not add a new observer or scrollbar-width calculation.

## Verification
- Static CSS brace balance: passed.
- `git diff --check`: passed.
- Browser geometry smoke test with a 60px container, classic 17px scrollbar, and `stable both-edges`: content and controls remain centered while the scrollbar remains present.
- ⚠️ Still needs Windows Zen verification with *Always show scrollbars* enabled, including regular tabs, folders, split-view tabs, tab scrolling/dragging, and the no-overflow state.

### DIFF
--- a/src/zen/tabs/zen-tabs/vertical-tabs.css
+++ b/src/zen/tabs/zen-tabs/vertical-tabs.css
@@ -479,11 +479,16 @@
   }
 
   /* Windows "Always show scrollbars" reserves layout space for the
-     scrollbar, shifting the collapsed tab list left and clipping the
-     new-tab button (issue #12402). Hide it in the collapsed sidebar
-     where tabs are icon-only; wheel scrolling still works. */
-  :root:not([zen-sidebar-expanded="true"]) & {
-    scrollbar-width: none;
+     scrollbar, shifting the collapsed icon-only tab list left and
+     clipping the new-tab button (issue #12402). The collapsed sidebar
+     is only ~60px wide; hide the classic scrollbar there so tabs stay
+     centered. Wheel scrolling still works, and the expanded sidebar
+     keeps its scrollbar. Limited to Windows: macOS/Linux use overlay
+     scrollbars that never reserve space. */
+  @media (-moz-platform: windows) {
+    :root:not([zen-sidebar-expanded="true"]) & {
+      scrollbar-width: none;
+    }
   }
 }
 
--- a/src/zen/tabs/zen-tabs/vertical-tabs.css
+++ b/src/zen/tabs/zen-tabs/vertical-tabs.css
@@ -477,6 +477,14 @@
     margin-left: calc(-1 * var(--zen-toolbox-padding));
     width: calc(100% + var(--zen-toolbox-padding) * 2);
   }
+
+  /* Windows "Always show scrollbars" reserves layout space for the
+     scrollbar, shifting the collapsed tab list left and clipping the
+     new-tab button (issue #12402). Hide it in the collapsed sidebar
+     where tabs are icon-only; wheel scrolling still works. */
+  :root:not([zen-sidebar-expanded="true"]) & {
+    scrollbar-width: none;
+  }
 }
 
 #pinned-tabs-container {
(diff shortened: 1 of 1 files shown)


src/zen/tabs/zen-tabs/vertical-tabs.css
```diff
@@ -477,6 +477,19 @@
...
+  /* Windows "Always show scrollbars" reserves layout space for the
+     scrollbar, shifting the collapsed icon-only tab list left and
+     clipping the new-tab button (issue #12402). The collapsed sidebar
+     is only ~60px wide; hide the classic scrollbar there so tabs stay
+     centered. Wheel scrolling still works, and the expanded sidebar
+     keeps its scrollbar. Limited to Windows: macOS/Linux use overlay
+     scrollbars that never reserve space. */
+  @media (-moz-platform: windows) {
+    :root:not([zen-sidebar-expanded="true"]) & {
+      scrollbar-width: none;
```
mr-cheffy: We still want a scrollbar

mr-cheffy (changes requested):

feichuans:
I replaced the hidden-scrollbar approach in b15d84e. The scrollbar remains visible and draggable now. The fix uses the existing #tabbrowser-tabs[overflow] state plus scrollbar-gutter: stable both-edges, and shrinks only collapsed icon-only controls when the gutter is active. This avoids a new JS observer and preserves the no-overflow layout. Windows verification is still needed.

mr-cheffy:
Your new solution also seems super hacky and vibecoded. Sorry but im not going to play the guessing game with claude.

feichuans:
You’re totally correct. I made speculative tweaks without reproducing and testing the issue on Windows, and I leaned on Claude’s suggestions too heavily, resulting in blind guesswork.
I’ll stop guessing now and won’t reopen the PR without solid reproduction evidence. Thanks a lot for pointing this out.

Issue #13860, opened 2026-05-25
Zen does not appear in Local Network permissions and cannot access local router IP


Nickleics:
### Preliminary Checks

- [x] I have read and understood the important section above.
- [x] I have searched existing issues and avoided creating duplicates.
- [x] I am not filing an enhancement request.
- [x] I have checked that this issue cannot be reproduced on Mozilla Firefox.
- [x] I have checked that this issue can be reproduced once I removed all my Mods and Custom CSS.

### Environment
- Zen Browser: 1.20b
- macOS: 26.5
- Device: MacBook Pro
- URL affected: `http://10.0.0.1/` / `https://10.0.0.1/`

### Problem
Zen cannot access my router admin page at `10.0.0.1`. Safari can access the same address successfully.

Zen shows its “Unable to connect” page and specifically suggests checking macOS Local Network permissions. However, Zen does not appear in System Settings > Privacy & Security > Local Network, even after macOS showed the Local Network prompt and I clicked Allow.

### Investigation
The Zen app bundle appears to be missing `NSLocalNetworkUsageDescription`:

```bash
/usr/libexec/PlistBuddy -c "Print :NSLocalNetworkUsageDescription" "/Applications/Zen.app/Contents/Info.plist"
# Print: Entry, ":NSLocalNetworkUsageDescription", Does Not Exist
```

Nested helper apps also appear to be missing it, including:

```text
/Applications/Zen.app/Contents/MacOS/plugin-container.app
Bundle ID: app.zen-browser.plugincontainer
```

Direct access fails in Zen, but a workaround succeeds if I avoid exposing the local/private IP to Zen:

```bash
sudo ifconfig lo0 alias 203.0.113.10 255.255.255.255
socat TCP-LISTEN:18443,bind=203.0.113.10,reuseaddr,fork TCP:10.0.0.1:443
```

Then this works in Zen:

```text
https://203.0.113.10:18443/
```

This suggests the router is reachable, but Zen/macOS Local Network permission registration is broken for the app or its networking helper.

### Expected behavior

Zen should appear in macOS Local Network settings and be able to access `10.0.0.1` after permission is granted.

### Actual behavior

Zen does not appear in Local Network settings and cannot access `10.0.0.1`, while Safari can.

### Steps to reproduce

Zen cannot access my router admin page at `10.0.0.1`. Safari can access the same address successfully.

[...]

Issue #11201, opened 2025-11-08
Unable to connect to local network as of 1.17.5b


genericbar:
### Preliminary Checks

- [x] I have read and understood the important section above.
- [x] I have searched existing issues and avoided creating duplicates.
- [x] I am not filing an enhancement request.
- [x] I have checked that this issue cannot be reproduced on Mozilla Firefox.
- [x] I have checked that this issue can be reproduced once I removed all my Mods and Custom CSS.

### What happened?

As of 1.17.5b, I'm no longer able to connect to any devices on the local network. I've also tried with other local link with no success.

### Expected behavior

Able to connect to the device on the local network.

### Actual behavior

Unable to connect error.

### Steps to reproduce

Steps:
1. Update to latest Zen version.
2. Attempt to connect to any device on the local network via localhost or their IP. 

Additional info:
- Running macOS Tahoe 26.1
- Experienced as of Zen 1.17.5b
- I was able to connect via other browsers.
- Local Network permission is granted via the macOS Privacy & Security settings.

### Screenshots and videos

<img width="850" height="369" alt="Image" src="https://github.com/user-attachments/assets/01ef9399-0d63-455b-b4bb-79a58ed84f11" />

### Version

1.17.5b

### What platform are you seeing the problem on?

macOS - aarch64

### What component is this issue related to?

Security

### Relevant log output if applicable

```shell

```

uplg:
Works for me, same OS (26.0.1), arch but version 1.17.6b

Edit : Upgraded to macOS 26.1, works perfect also.

genericbar:
I haven't done anything different though. I was working on a local network link, stopped to update the app, and after updating I could no longer connect..

swelborn:
also seeing this

update: this was a different problem on my end. my `vscode` was running a phantom process on that port.

uplg:
In System settings > Privacy & Security > Local network

Is Zen checked ?

genericbar:
> In System settings > Privacy & Security > Local network
> 
> Is Zen checked ?

It is!

steynovich:
Same issue here. However as a workaround I can use Zen to open local network resources when launching Zen manually from the console. I'm using:
```/Applications/Zen.app/Contents/MacOS/zen```

But I can't open local network resources if I use MacOS Launcher.

ElieM90:
Same issue for me.
In System Settings → Privacy & Security → Local Network, Zen doesn't even appear in the list.

[...]

PR #14929, opened 2026-08-11
fix: declare local network usage on macOS

TJKANZLER:
## Summary

- add `NSLocalNetworkUsageDescription` to the macOS application plist through Zen's Firefox patch set
- register the patch in the external-patches manifest

## Problem

Zen users on macOS report that LAN addresses fail while the same addresses work in other browsers, and in some cases Zen does not appear correctly under System Settings > Privacy & Security > Local Network. This is tracked in #13860 and #11201.

The signed Zen 1.21.13b bundle currently has a valid application identifier and unique executable UUIDs, but its main `Info.plist` does not contain `NSLocalNetworkUsageDescription`. Apple recommends that any application which accesses the local network directly or indirectly include this key so macOS can explain and track the permission correctly:

https://developer.apple.com/documentation/BundleResources/Information-Property-List/NSLocalNetworkUsageDescription

Firefox is also tracking the missing metadata upstream in https://bugzilla.mozilla.org/show_bug.cgi?id=1936468. Keeping this as an external Firefox patch lets Zen address the gap until it lands upstream.

## Impact

Signed macOS builds will declare why Zen needs local-network access, allowing macOS to present a meaningful permission prompt for websites and services hosted on the user's LAN. This does not reset already-corrupted or stale macOS permission records.

## Validation

- applied the patch after Zen's existing `no_liquid_glass_icon.patch` against the frozen Firefox 153.0.3 source
- ran the Zen Node 22 patch-import sequence and confirmed the external patch applies
- validated the resulting `Info.plist.in` with `plutil -lint`
- ran `git diff --check`
- applied the final submission patch to a clean Zen checkout

### DIFF
--- a/src/external-patches/firefox/macos_local_network_usage_description.patch
+++ b/src/external-patches/firefox/macos_local_network_usage_description.patch
@@ -1,0 +1,13 @@
+diff --git a/browser/app/macbuild/Contents/Info.plist.in b/browser/app/macbuild/Contents/Info.plist.in
+index 2795b177f4943d58da987153d3f3b66b81bd16ab..eea4e19e95e9e392a021493a8f57444d62e6fd7b 100644
+--- a/browser/app/macbuild/Contents/Info.plist.in
++++ b/browser/app/macbuild/Contents/Info.plist.in
+@@ -319,4 +319,8 @@
+   <key>NSNetworkVolumesUsageDescription</key>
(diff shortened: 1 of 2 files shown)


mr-cheffy:
Hi, please submit (or we can as well) this to firefox so they can take a look and we'll just pull it from upstream. Thanks!

PR #14932, opened 2026-08-12
gh-14922: Collapse toolbar after modal prompts

### LINKED ISSUE #14922: Top bar gets stuck opened when warn on close is turned on
### Preliminary Checks

### What happened?

So whenever that dialog box appears, it can be a site asks "are you sure you want to leave?" or "Close X tabs?", top bar gets stuck.
I couldn't make it happen till 0:30 in attached video, but it happens frequently when I am not trying to show it :D

### Expected behavior

Top bar auto-hide

### Actual behavior

Top bar stucks open (while auto-hide is on)

### Steps to reproduce

1. Make any tab do "processing..". Can be uploading / Turn on warn on close and open multiple tabs
2. Click on "close", let that dialog box appear
3. Get disappointed with top bar auto-hide

### Screenshots and videos

https://github.com/user-attachments/assets/e1e71d23-e224-48dc-9642-38838aa2481e

(It appeared after a few attempts, exact time is 0:30)

### Version

1.21.10b

Linux (Tarball)

Customizable UI / Toolbars

```shell

```

fzlzjerry:
## Summary

- clear stale compact-mode hover state when modal prompts close
- preserve genuinely hovered toolbars through the existing hover check
- add browser regression coverage for the modal-close event

Fixes #14922

## Testing

- `cd engine && ./mach build faster`
- `python3 scripts/run_tests.py compact_mode/browser_compact_mode_dialog.js --headless --force` (2/2 subtests passed)
- `cd engine && ./mach lint --warnings=soft zen/compact-mode/ZenCompactMode.mjs zen/tests/compact_mode/browser_compact_mode_dialog.js`

### DIFF
--- a/src/zen/compact-mode/ZenCompactMode.mjs
+++ b/src/zen/compact-mode/ZenCompactMode.mjs
@@ -129,6 +129,12 @@ window.gZenCompactModeManager = {
       this._clearAllHoverStates()
     );
 
+    // Modal prompts can interrupt mouseleave delivery and leave a stale hover
+    // state behind after the prompt closes.
+    window.addEventListener("DOMModalDialogClosed", () =>
+      this._clearAllHoverStates()
+    );
+
     // Hide any element kept open by the outside mouse tracking as soon as the
     // window loses focus
     window.addEventListener("deactivate", () => this._collapseTrackedElement());
--- a/src/zen/compact-mode/ZenCompactMode.mjs
+++ b/src/zen/compact-mode/ZenCompactMode.mjs
@@ -129,12 +129,6 @@ window.gZenCompactModeManager = {
       this._clearAllHoverStates()
     );
 
-    // Modal prompts can interrupt mouseleave delivery and leave a stale hover
-    // state behind after the prompt closes.
-    window.addEventListener("DOMModalDialogClosed", () =>
-      this._clearAllHoverStates()
-    );
-
     // Hide any element kept open by the outside mouse tracking as soon as the
     // window loses focus
     window.addEventListener("deactivate", () => this._collapseTrackedElement());
--- a/src/zen/tests/compact_mode/browser_compact_mode_dialog.js
+++ b/src/zen/tests/compact_mode/browser_compact_mode_dialog.js
@@ -5,9 +5,12 @@
 
 add_task(async function test_toolbar_collapses_after_modal_dialog_closes() {
   const toolbar = document.getElementById("zen-appcontent-navbar-wrapper");
-  const browser = gBrowser.selectedBrowser;
 
-  EventUtils.synthesizeMouseAtCenter(browser, { type: "mousemove" }, window);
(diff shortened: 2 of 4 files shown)


src/zen/tests/compact_mode/browser_compact_mode_dialog.js
```diff
@@ -0,0 +1,29 @@
...
+  );
+
+  gZenCompactModeManager._setElementExpandAttribute(toolbar, true);
+  Assert.ok(
+    toolbar.hasAttribute("zen-has-hover"),
+    "The toolbar starts with a stale hover state"
+  );
+
+  // Firefox dispatches this event after beforeunload and other modal prompts.
+  window.dispatchEvent(new CustomEvent("DOMModalDialogClosed"));
```
mr-cheffy: We should open an actual dialog, not fire the event directly

mr-cheffy (changes requested):
"Good first issue"s are for new people to learn and get used to zen's source code, not so you can just slap claude into it... but ig it works

src/zen/tests/compact_mode/browser_compact_mode_dialog.js
fzlzjerry: Updated the regression test to open and accept an actual window-modal prompt through `Services.prompt`. It now waits for both `DOMModalDialogClosed` and the embedded dialog container cleanup before asserting the toolbar state, instead of dispatching the event directly.

The focused browser test passes 2/2 with no unexpected results or leaks.

src/zen/compact-mode/ZenCompactMode.mjs
```diff
@@ -129,6 +129,12 @@ window.gZenCompactModeManager = {
       this._clearAllHoverStates()
     );
 
+    // Modal prompts can interrupt mouseleave delivery and leave a stale hover
+    // state behind after the prompt closes.
+    window.addEventListener("DOMModalDialogClosed", () =>
+      this._clearAllHoverStates()
```
Ivaon: The issue is not coming from hover state. But the `has-popup-menu` attribute

https://github.com/zen-browser/desktop/issues/14922#issuecomment-5279090568

Ivaon (changes requested):

src/zen/compact-mode/ZenCompactMode.mjs
fzlzjerry: Good catch — the stale state is has-popup-menu, not zen-has-hover. In onPopupHidden(), removal was deferred to the next mousemove while the close warning owned hover, so the attribute could remain indefinitely if the pointer stayed still.

3cb30c4 moves the modal-close cleanup into ZenUIManager, cancels the pending mousemove listener, and updates the regression test to exercise the multi-tab close warning and popup-tracking path. Focused browser test: 4/4 passed, 0 unexpected, no leaks.

PR #14945, opened 2026-08-13
gh-13383: Keep GitHub Live Folder tabs and repositories when a fetch returns no results

### LINKED ISSUE #13383: Erratic Live Folder behaviour, repo-specific or behaviour-based folders are flaky
### Preliminary Checks

### What happened?

The last week or so I have been working with 1 Live Folder for each of the usual repos I visit and one that gathers the rest. I moved away from it the other day cuz it frequently bugged out in several different ways:

- Tabs did not load on startup
  - reloading the folder via rightclick does nothing
  - Re-selecting the repo or any of the assigned or created by me _usually_ resolved it
- The repo tab is grayed out
  - I don't fully understand why this is happening
  - Sometimes the other repos, except the selected one, are shown
  - Sometime no repos are shown at all

### Expected behavior

Folder should keep fetching PRs from specific repos

### Actual behavior

Folder don't fetch PRs and the reload "options" are flaky at best

### Steps to reproduce

Mostly this has been happened when left overnight. I dont know if this is due to our Company Policy where we have to re-auth every morning.

### Screenshots and videos

No good picture of the issue, will update this issue whenever the error occurs again. 

### Version

1.19.8b

macOS - aarch64

Tab Folders

```shell

```

srgnft:
Fixes: https://github.com/zen-browser/desktop/issues/13383

GitHub answers `200` with an empty result set when the session no longer has access to the repositories (an expired SSO session is the reported trigger). The provider takes that at face value: every tab is removed as outdated, `state.repos` is overwritten with an empty `Set` — which leaves the *Repositories* submenu permanently disabled — and no error is surfaced, so *Refresh* looks like a no-op. Evidence and full trace: https://github.com/zen-browser/desktop/issues/13383#issuecomment-5253318476.

Two guards in `fetchItems`:

- an empty result is trusted only once a second consecutive fetch confirms it; the first one raises the existing `zen-live-folder-github-no-auth` state, whose action button already opens `github.com/login`
- the known repository list is kept when a fetch reports no repositories

This mirrors the guard `onLiveFolderFetch` already applies to `dismissedItems`.

Trade-off: someone with genuinely no open pull request gets one "Not signed in to GitHub" cycle before the empty result is accepted. Happy to switch to an explicit counter or a dedicated string instead.

Not built locally — `mach lint zen` needs the full engine checkout. The file is prettier-clean with the tree's options (`--arrow-parens avoid --print-width 80 --trailing-comma es5`).


src/zen/live-folders/providers/GithubLiveFolder.sys.mjs
```diff
@@ -92,7 +92,22 @@ export class nsGithubLiveFolderProvider extends nsZenLiveFolderProvider {
...
 
-      this.state.repos = combinedActiveRepos;
+      // A 200 with an empty result set is indistinguishable from a session that
+      // lost access to the repositories, so only trust a second empty fetch.
+      if (
+        !combinedItems.size &&
+        this.state.repos.size &&
+        this.state.lastErrorId !== "zen-live-folder-github-no-auth"
+      ) {
+        return "zen-live-folder-github-no-auth";
```
mr-cheffy, line 102: This will make it show an error that they aren't logged in though?

mr-cheffy (changes requested):

PR #14947, opened 2026-08-13
gh-14872: Fix mobile bookmarks item alignment in the bookmarks menu

### LINKED ISSUE #14872: Mobile bookmarks item slightly misaligned in Bookmarks menu
### Preliminary Checks

### What happened?

In the Bookmarks menu, the "Mobile bookmarks" item is slightly misaligned (both left and right edges) compared to the other items in the list.

_Zen 1.21.11b_
Image

_Firefox 153.0.3_
Image

### Expected behavior

"Mobile bookmarks" should be aligned with the other items, as it is in Firefox.

### Actual behavior

"Mobile bookmarks" is slightly offset on both sides relative to the other menu items.

### Steps to reproduce

1. Open the Bookmarks menu
2. Observe the "Mobile bookmarks" item compared to other items in the list

### Screenshots and videos

_No response_

### Version

1.21.11b

Windows - x64

Bookmarks

```shell

```

guirguispierre:
fixes #14872

upstream markup gives the mobile bookmarks menu an extra subviewbutton class that its siblings in the bookmarks popup do not have, so it picks up the arrowpanel padding of 8px instead of the 6px that every other row in that popup gets. the result is a row that sits slightly inset on both edges compared to the rest of the menu.

this scopes a rule to that item so it uses the same menuitem padding variable as the rest of the popup. the id selector wins on specificity, so no important flag is needed.

### DIFF
--- a/src/zen/common/styles/zen-panels/bookmarks.css
+++ b/src/zen/common/styles/zen-panels/bookmarks.css
@@ -68,3 +68,8 @@ hbox.editBMPanel_folderRow {
 #PersonalToolbar {
   background: transparent;
 }
+
+/* Also a .subviewbutton unlike its siblings, so it gets arrowpanel padding */
+menupopup > #BMB_mobileBookmarks {
+  padding: var(--menuitem-padding);
+}


src/zen/common/styles/zen-panels/bookmarks.css
```diff
@@ -68,3 +68,8 @@ hbox.editBMPanel_folderRow {
 #PersonalToolbar {
   background: transparent;
 }
+
+/* Also a .subviewbutton unlike its siblings, so it gets arrowpanel padding */
+menupopup > #BMB_mobileBookmarks {
```
mr-cheffy, line 73: no need for `menupopup`

mr-cheffy (changes requested):

PR #14950, opened 2026-08-13
Fix space routing containers for new tabs

jochimvd:
## Summary

- apply a routed space's container when a new tab inherited its opener's container
- recreate initially blank tabs through the routed `addTab` path so their immutable origin attributes use the destination container
- preserve explicitly selected containers and guard against duplicate replacement tabs

## Testing

- `./mach lint zen/common/sys/ui/ZenSpaceRoutingNavigation.sys.mjs zen/tests/space_routing/browser_space_routing_initial_tab_container.js`
- `MOZ_HEADLESS=1 ./mach test --force zen/tests/space_routing/browser_space_routing_initial_tab_container.js` (7/7 assertions passed)
- manually verified routed links retain the authenticated destination-container session


mr-cheffy:
AI

Issue #13703, opened 2026-05-15
[1.19.13b] Workspace swiping will ocassionally be stuck.


gordonn:
### Preliminary Checks

- [x] I have read and understood the important section above.
- [x] I have searched existing issues and avoided creating duplicates.
- [x] I am not filing an enhancement request.
- [x] I have checked that this issue cannot be reproduced on Mozilla Firefox.
- [x] I have checked that this issue can be reproduced once I removed all my Mods and Custom CSS.

### What happened?

When swiping to switch workspace it will occasionally get stuck in between switches.

### Expected behavior

Should scroll smoothly.

### Actual behavior

Sometimes gets stuck and requires multiple swipes or clicking on the workspace icon to switch which makes tabs un-clickable.

### Steps to reproduce

Try swiping to change workspace with a trackpad.

### Screenshots and videos

https://github.com/user-attachments/assets/325398be-850a-44fa-8b56-7b669960255f

### Version

1.19.13b

### What platform are you seeing the problem on?

Windows - x64

### What component is this issue related to?

Workspaces

### Relevant log output if applicable

```shell

```

gitjuice:
This is happening to me also on zen linux since last update 1.19.13b. I'm using the tarball installation.

Specifically, it only happens when trying to switch:
- from the last workspace to the first 
- from the first workspace to the last

gordonn:
> * from the last workspace to the first
> * from the first workspace to the last

Yeah I have the exact same issue.

gitjuice:
> > * from the last workspace to the first
> > * from the first workspace to the last
> 
> Yeah I have the exact same issue.

I must add that since restarting the app the issue seems to be gone, dunno if that is also the case for the windows version.

edit: nope, it's still there but only if I swipe using touchpad gestures, not with the mouse horizontal scroll wheel. Hope it helps.

Edit 2. Also, tabs become unresponsive after the glitch, which requires a browser restart. Video below

https://github.com/user-attachments/assets/07217825-8984-4188-8cc2-9292a172f123

mr-cheffy:
Does it only happen when swiping from the *last* space to the right?

gordonn:
> Does it only happen when swiping from the _last_ space to the right?

It seems to happen more often when I swipe the last to right but as shown in the video I attached swiping the first to left also occasionally gets stuck as well.

PR #14985, opened 2026-08-15
gh-14984: Fix tab strip staying offset when a space swipe is interrupted

### LINKED ISSUE #14984: Interrupted space swipe leaves the tab strip offset and stops the sidebar from scrolling
### Preliminary Checks

### What happened?

Swiping horizontally on the sidebar (horizontal scroll wheel or trackpad) starts a space
swipe. If that gesture is interrupted before it finishes, switching to another window or
app, or a popup opening, the browser never finishes the swipe and never cleans it up.

The tab strip stays parked on the partial swipe offset, and the tab list stops reacting to
scrolling entirely. Nothing recovers it on its own.

This happens regardless of how many spaces exist. I only have the default one.

### Expected behavior

An interrupted space swipe snaps the tab strip back onto the current space and leaves the sidebar scrollable.

### Actual behavior

The tab strip stays parked on the partial swipe offset and the tab list no longer responds to scrolling.

### Steps to reproduce

1. Open Zen with the sidebar expanded and enough tabs that the tab list scrolls vertically.
2. Start a horizontal swipe over the tab list — a horizontal scroll wheel or a two-finger
   trackpad swipe — but do not complete it.
3. While the swipe is still in progress, switch to another window or application
   (Cmd-Tab, or click a window of another app).
4. Switch back to Zen.
5. The tab strip sits a few pixels off to the side, and scrolling the tab list vertically
   does nothing.

Deterministic variant, in the Browser Toolbox console (chrome context) — this dispatches the
start of a swipe and then simply never delivers the end events, which is exactly what an
interrupted gesture looks like:

```js
const strip = document.getElementById("tabbrowser-tabs");
const r = strip.getBoundingClientRect();
const x = Math.round(r.left + r.width / 2), y = Math.round(r.top + r.height / 2);
const fire = (type, delta) => {
  const e = document.createEvent("SimpleGestureEvent");
  e.initSimpleGestureEvent(type, true, true, window, 0, x, y, x, y,
    false, false, false, false, 0, null, 0, SimpleGestureEvent.DIRECTION_RIGHT, delta, 0);
  strip.dispatchEvent(e);
};
fire("MozSwipeGestureMayStart", 0);
fire("MozSwipeGestureStart", 0);
fire("MozSwipeGestureUpdate", 0.28);
// no MozSwipeGesture / MozSwipeGestureEnd -- this is what the interruption does
```

Then check the state, which stays like this indefinitely:

`document.querySelector("zen-workspace").style.transform;              // "translateX(13.6309%)"`
`document.documentElement.getAttribute("swipe-gesture");               // "true"`
`getComputedStyle(document.getElementById("tabbrowser-arrowscrollbox")).pointerEvents;  // "none"`

### Screenshots and videos

_No response_

### Version

1.21.14b

macOS - aarch64

Workspaces

```shell

```

obi12341:
Closes #14984

`ZenSpacesSwipe` only cleans up in `onSwipeGestureAnimationEnd()`, which is wired to
`MozSwipeGestureEnd`. When a gesture is abandoned — the window is deactivated, or a popup
takes over — that event never arrives and nothing recovers the state:

- the `translateX(...)` applied by `_handleSwipeUpdate` stays on the strip, and
- `swipe-gesture="true"` stays on `documentElement`, keeping
  `:root[swipe-gesture] #tabbrowser-arrowscrollbox { pointer-events: none }` (gh-13703)
  active, which is what stops the tab list from scrolling.

This adds `#abortGesture()`, which puts the strip back on the current space, clears the
essentials transforms and runs the existing cleanup. It is triggered on `deactivate` and on
`visibilitychange` while hidden. `_popupOpenHandler` now goes through it as well — that path
had the same defect, since it reset the swipe state but never the transform.


mr-cheffy:
Have you tested this fix?

src/zen/spaces/ZenSpacesSwipe.mjs
```diff
@@ -22,6 +22,35 @@ export class ZenSpacesSwipe {
   constructor() {
     this.#attachWorkspaceSwipeGestures(gNavToolbox);
     this._popupOpenHandler = this._popupOpenHandler.bind(this);
+    this.#attachGestureAbortTriggers();
+  }
+
+  // Cleanup only runs on MozSwipeGestureEnd, which never arrives when the window
+  // stops receiving the rest of the gesture. The strip then keeps its offset and
+  // the tab strip its disabled pointer events, see gh-13703 for those.
+  #attachGestureAbortTriggers() {
```
mr-cheffy, line 28: These are not needed imo

src/zen/spaces/ZenSpacesSwipe.mjs
```diff
@@ -22,6 +22,35 @@ export class ZenSpacesSwipe {
...
+    if (!this._swipeState?.isGestureActive) {
+      return;
+    }
+    const ws = gZenWorkspaces;
+    // Put the strip back on the space we never left
+    ws._organizeWorkspaceStripLocations(ws.getActiveWorkspaceFromCache(), true);
+    for (const container of document.querySelectorAll(
+      "#zen-essentials .zen-workspace-tabs-section"
+    )) {
+      container.style.removeProperty("transform");
```
mr-cheffy: This doesn't seem like a good idea? Why did you change this? We should animate back to the original space. Why wasn't `onSwipeGestureAnimationEnd` working out?

mr-cheffy (changes requested):

src/zen/spaces/ZenSpacesSwipe.mjs
obi12341: You're right about animating back. Switched to `_cancelSwipeAnimation()` and dropped both
the manual `_organizeWorkspaceStripLocations` call and the essentials loop.

On `onSwipeGestureAnimationEnd`: in this scenario it never runs at all, because it is only
wired to `MozSwipeGestureEnd`, and the widget never sends that once the window stops
receiving the gesture.

obi12341:
> Have you tested this fix?

Yes. Measured on a build of this branch with a clean profile, driving the gesture with
synthetic `MozSwipeGesture*` events plus a minimize/restore cycle.

src/zen/spaces/ZenSpacesSwipe.mjs
mr-cheffy: Not sure, you marked this as resolved but the function is still here?

mr-cheffy (changes requested):

src/zen/spaces/ZenSpacesSwipe.mjs
obi12341: Sorry, I read your comment as being about the comment block above the function and
removed that.

Do you mean the helper itself? I can fold those two addEventListener calls into
#attachWorkspaceSwipeGestures so there is no extra function, same behaviour.

mr-cheffy (changes requested):
Same as before

mr-cheffy:
It all looks very ai generated and half of the stuff doesnt make much sense

obi12341:
@mr-cheffy The PR fixes a real bug and includes a test that demonstrates it. If you don't understand the bug itself you could have asked.

I asked how you wanted the code structured so it would be acceptable to you. I got no answer for a week, then a close.
That is not a review, and it is not a reasonable way to treat people contributing work to this project.
We've been supporting Zen on Patreon. We'll be stopping.

mr-cheffy:
@obi12341 Im truly sorry, I missed that part. Seems like the comments got marked as resolved and hence stayed collapsed / hidden. I'll reopen and answer.

src/zen/spaces/ZenSpacesSwipe.mjs
mr-cheffy: Yes, I meant the helper itself. Since the chrome window wont ever call `visibilitychange` and deactivate is already handled by the swipe manager.

PR #9441, opened 2025-07-13
feat: improve tab navigation and cycling

nikvnt:
This PR includes commits based on the feature request from #8681 

It sets up custom fields for tab navigation within the 'Keyboard Shortcuts' section and also introduces a new section with two radio options under 'Tab Management'. These options allow users to decide whether they should cycle through unloaded tabs while navigating with shortcuts. Additionally, this same configuration can be bound to a shortcut for quick toggling.

I am aware someone tried to add custom shortcuts for tab navigation many months ago, but it was discontinued and unfinished, so I've made this.

I attempted to set the default shortcuts for tab cycling to 'Ctrl + Tab' and 'Ctrl + Shift + Tab', so that people who were already used to those would not have to go through the struggle of setting these manually on the custom fields. However, the 'Ctrl + Shift + Tab' combination did not function as expected. Because of that, I've decided to leave these default keybinds undefined for the time being and submit this PR for Cheff to check it out. 

The l10n PR related to this is right [here](https://github.com/zen-browser/l10n-packs/pull/173).

PR #3599, opened 2024-12-11
Shortcuts for Next/Prev Tab Navigation

normaltaro:
Added 2 shortcuts for tab navigation.
1. Switching to Next Tab (Unbounded by default)
2. Switching to Previous Tab (Unbounded by default)

Using extensions such as `Tridactyl` or `Vimium` to do this stops working on protected pages such as New Tab, Settings, etc and breaks the flow. But this keyboard shortcut will work on any opened tab. It is a better user experience to those who really like the keyboard.
Relevant Issue: https://github.com/zen-browser/desktop/issues/1678

I know English. I have submitted the relevant PR to the l10n repo: https://github.com/zen-browser/l10n-packs/pull/110

PR #15009, opened 2026-08-18
Add configurable next/previous tab shortcuts

jolanglinais:
Adds "Next Tab" and "Previous Tab" as configurable keyboard shortcuts in the Keyboard Shortcuts settings.

These were previously hardcoded in Firefox's `ShortcutUtils` system (`⌘`+`⌥`+`↓`/`↑` on macOS, `Ctrl`+`Alt`+`↓`/`↑` on other platforms). This change moves them into Zen's configurable shortcut system so users can rebind them.

### Changes
- Added `cmd_zenNextTab` and `cmd_zenPreviousTab` commands
- Added migration v21 with default bindings (Accel+Alt+Down/Up)
- Registered shortcuts under "Window & Tab Management" group
- Extended `ShortcutUtils` patch to remove hardcoded handler, preventing double-firing
- Added l10n strings for the settings UI

### DIFF
--- a/locales/en-US/browser/browser/preferences/zen-preferences.ftl
+++ b/locales/en-US/browser/browser/preferences/zen-preferences.ftl
@@ -365,4 +365,6 @@ zen-devtools-toggle-accessibility-shortcut = Toggle Accessibility
 zen-close-all-unpinned-tabs-shortcut = Close All Unpinned Tabs
 zen-new-unsynced-window-shortcut = New Blank Window
 zen-duplicate-tab-shortcut = Duplicate Tab
+zen-next-tab-shortcut = Next Tab
+zen-previous-tab-shortcut = Previous Tab
 zen-key-find-selection = Find Selection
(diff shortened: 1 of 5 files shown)


mr-cheffy:
Hi, sorry there are already existing PRs for this... I'll close this one for cleanup. Thanks for trying to contribute though!

jolanglinais:
I don't see any PRs that cover this specific thing @mr-cheffy - could you link one so I can keep updated with it being implemented?

jolanglinais:
Ah I think I see. There are older PRs, but they're all closed because Zen is waiting on Firefox to implement this.

- #11258 closed because waiting for Firefox
- #9441 closed because waiting for Firefox
- #3599 redirects to #9441
- #12804 closed and redirects to #11258

Source: https://bugzilla.mozilla.org/show_bug.cgi?id=2012210

jolanglinais:
It would definitely be nice to have this soon and not wait for it. Tab navigation is different than any other browser I've used, with the navigation between tabs taking me between workspaces instead, and I can't change it until Firefox does.

PR #15016, opened 2026-08-18
feat: add window blur support on wayland

abachmann:
Adds  blur support using the wayland [ext-background-effect-v1](https://gitlab.freedesktop.org/wayland/wayland-protocols/-/raw/main/staging/ext-background-effect/ext-background-effect-v1.xml) protocol.

When `zen.widget.linux.transparency` is enabled and the compositor supports the blur effect, the toplevel surface region is sent to the compositor so it can apply the blur. The region is updated on map, resize, move and scale.

I tested the changes with `GDK_BACKEND=x11 MOZ_ENABLE_WAYLAND=0` for X11 (no blur) and `MOZ_ENABLE_WAYLAND=1` for Wayland. Both cases worked fine.

Files in `src/widget/gtk/wayland` are generated with:
```bash
wayland-scanner client-header ext-background-effect-v1.xml ext-background-effect-v1-client-protocol.h
wayland-scanner private-code ext-background-effect-v1.xml ext-background-effect-v1-protocol.c
```


mr-cheffy:
Hi, we'd appreciate if you sent this patch to the firefox team so they can also take a look and we can just pull it from upstream. Thanks!

https://firefox-source-docs.mozilla.org/setup/contributing_code.html

Issue #8822, opened 2025-06-03
Tab auto unload not working with recent changes


Arctomachine:
### Preliminary Checks

- [x] I have read and understood the important section above.
- [x] I have searched existing issues and avoided creating duplicates.
- [x] I am not filing an enhancement request.
- [x] I have checked that this issue cannot be reproduced on Mozilla Firefox.
- [x] I have checked that this issue can be reproduced once I removed all my Mods and Custom CSS.

### What happened?

Looks like memory based unloader prevents tabs from auto unloading entirely. No matter how much time passes, tabs in background are always active, switching to them is instant. Even if whole system runs out of memory and starts using pagefile/swap

### Expected behavior

Auto unload tabs more agressively (or at all). Also return dim on auto unloaded tabs

### Actual behavior

Tabs are not auto unloaded at all

### Steps to reproduce

Open few tabs, use them, then leave them in background.
After some time switch to them to check if they were unloaded - they were not.

### Version

1.12.10b

### What platform are you seeing the problem on?

Windows - x64

### What component is this issue related to?

Tab unloading

### Relevant log output if applicable

```shell

```

Jai-JAP:
I can confirm the same issue on Linux.

Zen v1.12.10b on Arch Linux.

The background tabs don't refresh at all after around an hour of inactivity.

bl4ckswordsman:
Auto tab unloading stopped working for me too recently.

Linux (Ubuntu 25.04), Zen v1.12.10b

loaykhalifa:
me too
macOS 14.5

GitHYang:
me too,
Linux (debian12.11), v1.12.10b

Propheticus:
Zen has changed to the Firefox native tab unloader in v1.12.9b 
https://zen-browser.app/release-notes/#1.12.9b

![Image](https://github.com/user-attachments/assets/4b4ff359-71f5-47c1-866d-4680b6d656fa)

More info about its working https://firefox-source-docs.mozilla.org/browser/tabunloader/
Summary: It will only unload tabs when available memory is becoming low but there's (active) demand for it. It'll then unload stale/inactive tabs starting with the least recently used ones. It's no longer simply time based. By default it only considers tabs inactive after at least 10 minutes.

You can monitor which tabs are the first candidates to be unloaded (if free memory is becoming too low)  in `about:unloads`

By default no fading is applied unless tabs have been explicitly unloaded by the user.
[...]

PR #15053, opened 2026-08-21
feat: add automatic inactive tab unloading

benfirad:
## Summary

- add an opt-in automatic inactive-tab unloading mode with 5/15/30/60/120 minute delays
- reuse Firefox's `TabUnloader.getSortedTabs()` eligibility and weighting instead of maintaining a second discard heuristic
- unload only base-weight 0 tabs through non-forced `discardBrowser`, preserving Firefox's unload protections
- exclude selected, pinned, Essential, media/PiP/WebRTC/sharing, Glance, private, and internal tabs
- process up to five tabs per minute to avoid UI stalls
- add browser tests for disabled, eligible, recent, and protected tabs

## Why this differs from #12465

I understand timer-based unloading was previously declined in favor of memory-pressure-based unloading. This version is disabled by default and layers an explicit user choice on top of Firefox's existing candidate weighting and safety checks; the native low-memory path remains unchanged.

## Validation

- Surfer imported all 69 Zen patches into Firefox 154
- targeted Mozilla ESLint: 0 problems
- `test-manifest-toml`: 0 problems
- Fluent parser: 0 junk nodes
- XHTML/YAML parsing and JavaScript syntax checks passed
- browser-chrome test added; full execution was not run locally because Rust/Cargo and LLD are absent

Related: #8822, https://github.com/zen-browser/desktop/discussions/10494, #12465

### DIFF
--- a/locales/en-US/browser/browser/preferences/zen-preferences.ftl
+++ b/locales/en-US/browser/browser/preferences/zen-preferences.ftl
@@ -73,6 +73,23 @@ zen-tabs-cycle-ignore-pending-tabs =
   .label = Ignore Pending tabs when cycling with Ctrl+Tab
 zen-tabs-cycle-by-attribute-warning = Ctrl+Tab will cycle by recently used order, as it is enabled
 
+pane-zen-auto-unload-tabs-title = Tab Unloading
+zen-auto-unload-tabs-header = Automatically unload inactive tabs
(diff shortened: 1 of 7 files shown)


mr-cheffy:
same reason as https://github.com/zen-browser/desktop/pull/12465#issuecomment-3941787824 + this looks heavily ai generated

PR #15064, opened 2026-08-22
gh-15062: Fix spaces getting stuck with a theme that can't be changed

### LINKED ISSUE #15062: Theme stuck on Green
### Preliminary Checks

### What happened?

Zen 1.21.15b on MacOS Tahoe 26.5.2
I tried making a few tweaks to my main workspace's theme, and now I'm stuck with a dark green/black gradient and I can't get back to a solid light color (my 2 other workspaces are fine).  I've tried deleting all the theme circles, but I'm still left with a predominantly green theme.  
I'm probably doing something simple wrong, but I can't figure out how to get back to where I was.  Can anyone help?

(I have no mods and no CSS)

### Expected behavior

Able to adjust themes

### Actual behavior

Can no longer adjust them on this workspace

### Steps to reproduce

I just tried to Edit Theme...

### Screenshots and videos

Sidebar: Image

Status Bar:Image

Theme Editor:Image

### Version

1.21.5b

macOS - aarch64

Mods / Themes

```shell

```

tuna4ll:
Closes #15062

## Problem

Removing the last color from a space left its theme impossible to change. The
add button showed a `not-allowed` cursor and nothing brought the space back.

## Cause

Two layers hit the same dead end:

- `onWorkspaceChange` disabled every button in the color actions when
  `gradientColors` was empty, the add one included.
- Even enabled, adding did nothing: `calculateCompliments` returns an empty
  array when there are no dots to derive the

## Changes

- Add is only disabled once the wheel is full, and it drops the first dot
  itself when there is none.
- Add and the harmony toggle count the colors that actually have a dot. Custom
  colors don't get one, so a space holding only custom colors read as full
  while its wheel was empty.
- Remove drops the last custom color when the wheel is empty, since their list
  is hidden behind `zen.theme.gradient.show-custom-colors`.
- A theme we can't read no longer throws on eft
  the space painted with whatever was there before and impossible to theme.

### DIFF
--- a/src/zen/spaces/ZenGradientGenerator.mjs
+++ b/src/zen/spaces/ZenGradientGenerator.mjs
@@ -580,16 +580,28 @@ export class nsZenThemePicker extends nsZenMultiWindowFeature {
     return JSON.stringify({ x: Math.round(x), y: Math.round(y) });
   }
 
+  /**
+   * Where to drop a dot the user didn't place themselves. It's the same spot
+   * the single color presets use.
+   */
+  get #defaultDotPosition() {
+    const dotPad = this.panel.querySelector(".zen-theme-picker-gradient");
+    const rect = window.windowUtils.getBoundsWithoutFlushing(dotPad);
+    return { x: rect.width * 0.63, y: rect.height * 0.63 };
+  }
+
   createDot(color, fromWorkspace = false) {
-    const [r, g, b] = color.c;
+    // A color we can't read would throw and leave the space without any dot
+    // to edit, so just skip it instead.
+    if (!color?.c || (!color.isCustom && !Array.isArray(color.c))) {
+      return;
+    }
+    const [r, g, b] = color.isCustom ? [] : color.c;
     const dot = document.createElement("div");
     if (color.isPrimary) {
       dot.classList.add("primary");
     }
     if (color.isCustom) {
-      if (!color.c) {
-        return;
-      }
       dot.classList.add("custom");
       dot.style.opacity = 0;
       dot.style.setProperty("--zen-theme-picker-dot-color", color.c);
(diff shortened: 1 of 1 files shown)


mr-cheffy:
This looks very spitted out by AI, and I dont see how it fixes the issue sorry

tuna4ll:
Thanks for taking a look, @mr-cheffy. The important part of the PR is the empty-wheel state.

Let me explain that.

The main fix is that when the last color is removed, the workspace enters a zero-dot state where `+` is disabled and `calculateCompliments()` can't create the first dot. The PR makes that state recoverable by letting `+` create the first dot directly.

The other changes are just safeguards for related theme edge cases.

btw i use ai for translations (both all my texts and comment lines) that might be cause of the ai look :D

mr-cheffy:
Still, the issue isnt related to "empty states"

tuna4ll:
@mr-cheffy  - I think there's a slight misunderstanding about what this PR targets, so let me clarify.

You're right that the green tint on macOS, That is a separate issue, and I completely agree it needs its own fix.

However, the user in #15062 reported two distinct problems, not one:

The green tint.

The UI becomes completely unusable and they cannot add any new color, the "+" button shows a not-allowed cursor, and clicking does absolutely nothing.

The second problem is directly caused by the "empty state" deadlock.

So the empty state is the root cause of the UI lock, and the macOS accent fallback is what makes it visually green. The PR fixes the empty-state deadlock so the user can add a new color and overwrite the green

mr-cheffy:
Again, its unrelated as you said, and the UI lock is done on purpose.

tuna4ll:
Anyway, I won't dwell on it any further, but I didn't understand how this was done on purpose.

> The user deleted all colors and now they cannot add a single new colors the workspace is permanently stuck

Thanks for everything

PR #15074, opened 2026-08-24
feat: add --space command line flag to open a link in a specific space

barakor:
reopening of #14104

Adds a --space <space> command line flag that switches to the workspace matching the given name (case-insensitive) or UUID:

zen --space work
zen --space "Space Name"
zen --space c2d1e574-3f04-4bd5-9721-2e6dd7f1adf9 # uuid of workspace

Behavior:

    Browser running: delegates to gZenWorkspaces.changeWorkspaceFromCommandLine in the most recent non-private window, which switches to the matching workspace.
    Cold start: the workspace is stashed in the zen.workspaces.cmdline-initial-workspace pref, which is consumed (and cleared) by the first window that restores its workspaces, so the browser starts on the requested space.
    Unknown workspace: no-op — the active workspace stays unchanged.
    When used together with a url invocation, the url would open in the specified space.

Implementation: extends the BrowserContentHandler.sys.mjs patch (same pattern as the existing --blank-window flag) and adds resolveWorkspaceFromString / changeWorkspaceFromCommandLine to ZenSpaceManager. Includes mochitests covering name/UUID resolution, workspace switching, and the unknown-workspace no-op. When used together with a url invocation, the url would open in the specified space.

Useful for OS shortcuts and scripting, e.g. a .desktop entry per workspace.

### DIFF
--- a/src/browser/components/BrowserContentHandler-sys-mjs.patch
+++ b/src/browser/components/BrowserContentHandler-sys-mjs.patch
@@ -2,7 +2,7 @@ diff --git a/browser/components/BrowserContentHandler.sys.mjs b/browser/componen
 index 4bfb4199dafac004486dfc61db954173742a010a..542c156432b9d7c2190c0d14ce9cfd09814422bf 100644
 --- a/browser/components/BrowserContentHandler.sys.mjs
 +++ b/browser/components/BrowserContentHandler.sys.mjs
-@@ -603,6 +603,28 @@ nsBrowserContentHandler.prototype = {
+@@ -603,6 +603,61 @@ nsBrowserContentHandler.prototype = {
        }
      }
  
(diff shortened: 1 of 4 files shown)


src/browser/components/BrowserContentHandler-sys-mjs.patch
```diff
@@ -27,15 +27,49 @@ index 4bfb4199dafac004486dfc61db954173742a010a..542c156432b9d7c2190c0d14ce9cfd09
...
 +
++    try {
++      let zenWorkspaceParam = cmdLine.handleFlagWithParam("space", false);
++      if (zenWorkspaceParam) {
++        let targetWin = lazy.BrowserWindowTracker.getTopWindow({
++          private: false,
++          allowPopups: false,
++        });
++        if (
++          targetWin?.gZenWorkspaces &&
```
mr-cheffy, line 38: Why would gZenWorkspaces not exist?

src/browser/components/BrowserContentHandler-sys-mjs.patch
```diff
@@ -27,15 +27,49 @@ index 4bfb4199dafac004486dfc61db954173742a010a..542c156432b9d7c2190c0d14ce9cfd09
...
++          !targetWin.gZenWorkspaces.privateWindowOrDisabled
++        ) {
++          targetWin.gZenWorkspaces
++            .changeWorkspaceFromCommandLine(zenWorkspaceParam)
++            .catch(console.error);
++          targetWin.focus();
```
mr-cheffy, line 44: Shouldnt we focus before switching?

src/zen/spaces/ZenSpaceManager.mjs
```diff
@@ -675,6 +675,25 @@ class nsZenWorkspaces {
     }
   }
 
+  resolveWorkspaceFromString(value) {
```
mr-cheffy:
```suggestion
  resolveWorkspaceFromCLIString(value) {
```

src/zen/spaces/ZenSpaceManager.mjs
```diff
@@ -675,6 +675,25 @@ class nsZenWorkspaces {
...
+    if (!value) {
+      return null;
+    }
+    try {
+      const workspaces = this.getWorkspaces();
+      return (
+        workspaces.find(workspace => workspace.uuid === value) ||
```
mr-cheffy: Can't we also check the name here? Seems a bit excessive to do 2 separate iterations.

mr-cheffy (changes requested):

src/browser/components/BrowserContentHandler-sys-mjs.patch
barakor: In private and popup windows for example

barakor: We do filter them out for coverage, but there might be other kinds I am not aware of ^

src/browser/components/BrowserContentHandler-sys-mjs.patch
barakor: I don't think it matters?

src/zen/spaces/ZenSpaceManager.mjs
barakor: we can, will merge them (the rational was that a uuid match is stronger, but honestly who names a space in uuid format....)

MadaraUchiha:
@mr-cheffy can this be landed? I've been following this ticket for a while, it'd really help me workflow. Thanks :pray:

src/browser/components/BrowserContentHandler-sys-mjs.patch
```diff
@@ -27,15 +27,49 @@ index 4bfb4199dafac004486dfc61db954173742a010a..542c156432b9d7c2190c0d14ce9cfd09
 +      }
 +    }
 +
++    try {
++      let zenWorkspaceParam = cmdLine.handleFlagWithParam("space", false);
++      if (zenWorkspaceParam) {
++        let targetWin = lazy.BrowserWindowTracker.getTopWindow({
```
mr-cheffy, line 33: Does this respect the `--new-window` flag?

mr-cheffy (changes requested):

PR #15080, opened 2026-08-24
Prevent partial Zen Mod installations

kadrxy:
## Summary

- Download Zen Mod assets into a unique staging directory.
- Move the staging directory into the final mod location only after every asset has downloaded successfully.
- Clean up staged files and propagate the error when installation fails.
- Add a browser-chrome regression test covering a failure after the stylesheet has already been downloaded.

## Problem

`installMod()` created the final mod directory before downloading its assets, while the download helper swallowed errors after exhausting its retries.

If a later asset failed to download, Zen could leave a partially populated mod directory behind. Since mod synchronization uses the directory's existence to determine whether installation is needed, that incomplete installation could be treated as complete and its partial CSS could be applied.

## Testing

- Added `browser_mod_install.js`, which successfully downloads a stylesheet and then forces the readme download to fail.
- Verified that the installation rejects.
- Verified that neither the final mod directory nor the staging directory remains.
- `node --check src/zen/mods/ZenMods.mjs`
- `node --check src/zen/tests/mods/browser_mod_install.js`
- `git diff --check`
- Validated the new `browser.toml` with Python's TOML parser.

### DIFF
--- a/src/zen/mods/ZenMods.mjs
+++ b/src/zen/mods/ZenMods.mjs
@@ -358,6 +358,7 @@ class nsZenMods extends nsZenPreloadedFeature {
             url,
             e
           );
+          throw e;
         } else {
           console.warn(
             `[ZenMods]: Download failed (attempt ${attempt} of ${maxRetries}), retrying in ${retryDelayMs}ms...`,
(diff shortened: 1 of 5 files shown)


mr-cheffy:
Is there an open issue about this?

mr-cheffy:
https://github.com/zen-browser/desktop/pull/15083#issuecomment-5408022432

PR #15081, opened 2026-08-24
glance: Clean up resources when browser windows close

kadrxy:
## Summary

- Unregister each Glance manager's global quit observer when its browser window unloads.
- Release object URLs owned by tracked Glance entries during cleanup.
- Keep quit-request handling separate from window teardown so cancelled quit requests remain supported.
- Add browser-chrome coverage for observer registration and removal.

## Problem

Every browser window creates its own Glance manager and registers it as a strong observer for `quit-application-requested`.

The observer was never removed when an individual browser window closed. As a result, the observer service could retain the manager together with its closed window and DOM resources.

Glance entries were also cleared during application shutdown without releasing their generated object URLs.

## Testing

- Added `browser_glance_window_cleanup.js`.
- The test opens a second browser window and verifies that its Glance manager is registered.
- It then closes the window and verifies that the observer registration has been removed.
- `node --check src/zen/glance/ZenGlanceManager.mjs`
- `node --check src/zen/tests/glance/browser_glance_window_cleanup.js`
- `git diff --check`
- Validated the updated `browser.toml` with Python's TOML parser.

### DIFF
--- a/src/zen/glance/ZenGlanceManager.mjs
+++ b/src/zen/glance/ZenGlanceManager.mjs
@@ -21,6 +21,7 @@ class nsZenGlanceManager extends nsZenDOMOperatedFeature {
   #glances = new Map();
   #currentGlanceID = null;
   #confirmationTimeout = null;
+  #observingQuitRequests = false;
 
   // Animation flags
   animatingOpen = false;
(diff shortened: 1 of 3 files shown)


mr-cheffy:
These code changes doesnt really make much sense to me unfortunately and for some reason there's now 2 duplicate destructor functions? Sorry but the logic behind me also doesn't make much sense.

PR #15082, opened 2026-08-24
emoji-picker: Improve search and rendering performance

kadrxy:
## Summary

- Replace the emoji search filter/sort/find pipeline with a single pass over rendered buttons.
- Normalize and associate search tags with each button once during rendering.
- Assign emoji ordering once instead of recalculating it on every search.
- Batch emoji and SVG button insertion with document fragments.
- Add browser-chrome coverage for filtering and clearing the search.

## Problem

The emoji picker contains 1,915 emoji records.

For every search input, the previous implementation:

1. Filtered and sorted the complete emoji dataset.
2. Iterated over every rendered button.
3. Searched the filtered array again with `find()` for each button.
4. Repeated tag normalization while filtering and updating buttons.

This resulted in quadratic lookup behavior. Clearing the search performed approximately 1,834,570 `find()` comparisons. A broad search such as `face` performed approximately 302,445 comparisons.

The new implementation associates normalized tags with buttons during rendering and searches them in one pass. Clearing the search now requires 1,915 tag checks, while the `face` search requires roughly 9,000 tag checks.

The emoji data and rendered elements are still released when the popup closes, so this does not introduce a permanent per-window DOM cache.

## Testing

- Added `browser_emoji_search.js`.
- Verified that a matching rocket emoji remains visible.
- Verified that a non-matching emoji is hidden.
- Verified that clearing the search restores every emoji.
- Verified the picker cleanup and promise rejection path.
- `node --check src/zen/common/emojis/ZenEmojiPicker.mjs`
- `node --check src/zen/tests/emojis/browser_emoji_search.js`
- `git diff --check`
- Validated the new `browser.toml` with Python's TOML parser.

### DIFF
--- a/src/zen/common/emojis/ZenEmojiPicker.mjs
+++ b/src/zen/common/emojis/ZenEmojiPicker.mjs
@@ -32,6 +32,7 @@ const SVG_ICONS = [
 
 class nsZenEmojiPicker extends nsZenDOMOperatedFeature {
   #panel;
+  #emojiSearchTags = new WeakMap();
 
   #anchor;
   #emojiAsSVG = false;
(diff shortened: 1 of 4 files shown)


mr-cheffy:
https://github.com/zen-browser/desktop/pull/15083#issuecomment-5408022432

PR #15083, opened 2026-08-24
gh-14732: Preserve pinned extension order in single toolbar

### LINKED ISSUE #14732: Pinned extension icons order gets reversed after opening Customize Toolbar
### Preliminary Checks

### What happened?

When extension icons are pinned to the toolbar, opening the "Customize Toolbar" panel causes the order of the pinned extension icons to reverse (e.g. an order of 1234 becomes 4321). This happens even if no changes are made and no save action is taken — simply opening the Customize Toolbar panel and pressing Esc to exit is enough to trigger it. The only way to restore the correct order is to fully close and reopen the browser.

This bug only occurs when the Browser Layout is set to "Only Sidebar". I tested other layouts with a top toolbar and the pinned icon order remains correct there.

### Expected behavior

Opening and closing the Customize Toolbar panel without saving should not alter the order of pinned extension icons.

### Actual behavior

The order of pinned extension icons is reversed every time the Customize Toolbar panel is opened, when using the "Only Sidebar" layout.

### Steps to reproduce

1. Set Browser Layout to "Only Sidebar"
2. Pin several extension icons to the toolbar in a specific order
3. Open Customize Toolbar
4. Press Esc to close the panel without making any changes or saving
5. Observe that the pinned extension icons are now in reverse order
6. Restart the browser — the pinned icon order returns to normal.

### Screenshots and videos

https://github.com/user-attachments/assets/daff8b83-251c-4325-ba4e-9453dc55cb6f

### Version

1.21.9b

Windows - x64

Customizable UI / Toolbars

```shell

```

kadrxy:
## Summary

- Preserve the original order of pinned extension buttons when moving them into the single-toolbar layout.
- Continue reverse-processing regular toolbar items that are inserted after the shared separator.
- Add a browser-chrome regression test covering extension button ordering.

## Testing

- `node --check src/zen/common/modules/ZenUIManager.mjs`
- `node --check src/zen/tests/customizable_ui/browser_extension_order.js`
- `git diff --check`
- Validated the new browser test manifest with Python's TOML parser.

Fixes #14732

### DIFF
--- a/src/zen/common/modules/ZenUIManager.mjs
+++ b/src/zen/common/modules/ZenUIManager.mjs
@@ -1348,16 +1348,35 @@ window.gZenVerticalTabsManager = {
 
       if (isSingleToolbar) {
         this._navbarParent = navBar.parentElement;
-        let elements = document.querySelectorAll(
+        const elements = document.querySelectorAll(
           '#nav-bar-customization-target > :is([cui-areatype="toolbar"], .chromeclass-toolbar-additional):not(#urlbar-container):not(toolbarspring)'
         );
-        elements = Array.from(elements).reverse();
+        const overflowWebExtensions = Services.prefs.getBoolPref(
+          "zen.view.overflow-webext-toolbar",
+          true
+        );
+        const extensionButtons = [];
+        const otherButtons = [];
+        for (const button of elements) {
+          if (
+            button.hasAttribute("data-extensionid") &&
+            overflowWebExtensions
+          ) {
+            extensionButtons.push(button);
+          } else {
+            otherButtons.push(button);
+          }
+        }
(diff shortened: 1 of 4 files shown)


mr-cheffy:
Hi, have you manually verified this? Could I please see a recording?

kadrxy:
Hi! Not yet — I verified the ordering logic with the regression test and static checks, but I haven't manually tested it in a running build because my checkout did not include the generated engine tree.

I'll set up a full build, verify the issue before and after the change, and attach a recording here.

mr-cheffy:
Ok, please resubmit these PRs once you actually verified the fix manually because the logic here seems wrong. Also, not sure how you verified the regression test without the engine tree?

PR #15118, opened 2026-08-26
gh-15115: Tests for urlbar close and split view

### LINKED ISSUE #15115: Creating a new empty split view with an already open url bar is broken
### Preliminary Checks

### What happened?

When the URL bar is open and creating a split view via the "New Empty Split View" action is attempted, the url bar instantly closes and "about:blank" page comes in focus.

This bug is caused by a sequence of indirections in the logic for creating a split view

Everything begins at 
[gBrowser.selectedTab = emptyTab](https://github.com/zen-browser/desktop/blob/bf40c31149a72fa2f3781b4e20f6277717956ba7/src/zen/split-view/ZenViewSplitter.mjs#L2531)

Assuming an already open url bar, this causes a synchronous invocation of `ZenUiManager::handleUrlbarClose`, which schedules a timeout for clearing out the bar [here](https://github.com/zen-browser/desktop/blob/bf40c31149a72fa2f3781b4e20f6277717956ba7/src/zen/common/modules/ZenUIManager.mjs#L609)

Lets call this Task A. 
Task queue:  [Task A]
After that returns, the setTimeout right below our `gBrowser.selectedTab = emptyTab` call gets scheduled, call this Task B, 
task queue: [Task A, Task B]. 
Task B is responsible for adding an event listener for the `ZenURLBarClosed` tab and invoking `gZenUIManager.handleNewTab`

When Task A executes, it schedules [Task C](https://github.com/zen-browser/desktop/blob/bf40c31149a72fa2f3781b4e20f6277717956ba7/src/zen/common/modules/ZenUIManager.mjs#L657). That task is responsible for dispatching the `ZenURLBarClosed` event.
Task queue: [Task B, Task C]

Task B executes and adds the `ZenURLBarClosed` handler.
task queue: [Task C]

Task C executes, and sends `ZenURLBarClosed` which was supposed to be the event for closing the first url bar, but it ends up being interpreted as the current one being closed.

At that point the url bar just closes and about:blank is opened.

*Note that I have marked my zen browser version below, but this is still not patched in the dev branch.

I already have a fix for this and can prepare a PR.

### Expected behavior

A split view after triggering the "New Empty Split View" action

### Actual behavior

A full screen about:blank page instantly after triggering the "New Empty Split View" action.

### Steps to reproduce

The bug is consistent and should be easily reproducible. Just open a floating url bar and try to split the current window.

### Screenshots and videos

Here is a video reproducing the issue and running everything through the debugger to show the exact path described above

https://github.com/user-attachments/assets/8dc31b79-3fc0-4c04-98a7-e7f49008f13c

### Version

1.21.15b

Linux (Tarball)

Split View

```shell

```

iexavl:
Follow up for https://github.com/zen-browser/desktop/pull/15116

### DIFF
--- a/src/zen/tests/split_view/browser_split_view_empty.js
+++ b/src/zen/tests/split_view/browser_split_view_empty.js
@@ -7,50 +7,149 @@ const { UrlbarTestUtils } = ChromeUtils.importESModule(
   "resource://testing-common/UrlbarTestUtils.sys.mjs"
 );
 
+/**
+ * @param {object} [staleToken] - Token of a session the pane will supersede.
+ */
+async function stageEmptySplit(staleToken = null) {
+  let staleClosed = null;
+
+  if (staleToken) {
+    staleClosed = BrowserTestUtils.waitForEvent(
+      window,
+      "ZenURLBarClosed",
+      false,
+      event => gZenUIManager.matchesCloseToken(staleToken, event)
+    );
+  }
+  const staleClose = gURLBar._zenHandleUrlbarClose;
+  document.getElementById("cmd_zenNewEmptySplit").doCommand();
+  await TestUtils.waitForCondition(
+    () =>
+      gURLBar._zenHandleUrlbarClose &&
+      gURLBar._zenHandleUrlbarClose !== staleClose,
+    "Waiting for the staged pane to open its own urlbar session"
+  );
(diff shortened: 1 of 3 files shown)


mr-cheffy (approved):
Thanks!

PR #15126, opened 2026-08-27
fix(urlbar): include inactive Essential tab titles

Jish2:
## Summary
- search Zen's stored workspace tabs when resolving live switch-to-tab titles
- preserve distinct container-specific Essential labels while another workspace is active
- add regression coverage for active and inactive workspace Essentials sharing a URL

## Test plan
- [x] validate the Surfer patch against the clean Firefox source index
- [x] run URLbar and regression-test lint
- [x] build the browser UI
- [x] reproduce two container-specific Essentials in separate workspaces and verify Cmd+T returns both live labels

Made with [Cursor](https://cursor.com)


Jish2:
Opened against upstream by mistake; this follow-up is being validated in the fork first.

PR #15130, opened 2026-08-27
gh-15128: Fix dragging splitter while creating new split view

### LINKED ISSUE #15128: Trying to drag split view splitter while a new empty split is staged permanently messes up tab layout
### Preliminary Checks

### What happened?

Attempting to drag a pane's splitter while the url bar is open causes the url bar and staged empty tab to close, but the empty pane geometry stays in the layout.

The cause of this that when the url bar closes [createEmptySplit](https://github.com/zen-browser/desktop/blob/3d53b37e8b40580d441826073a589c63c737b2eb/src/zen/split-view/ZenViewSplitter.mjs#L2567) closes the staged empty pane, but no one tells the [drag listener](https://github.com/zen-browser/desktop/blob/3d53b37e8b40580d441826073a589c63c737b2eb/src/zen/split-view/ZenViewSplitter.mjs#L1900) to stop what it's doing, so it keeps adjusting the layout for a non-existing pane. It only stops after mouse up, by which point the layout is already  messed up.

*Note that I have marked browser version 1.21.15b, but it's still not patched in the dev branch.

### Expected behavior

Clicking on the splitter of a staged split pane should close the url bar and the staged pane.

### Actual behavior

Url bar and pane tab close, but the pane geometry stays.

### Steps to reproduce

Try to make a new empty split with the "New Empty Split View" action. Then, with the url bar open, start dragging the splitter that appeared between the two panes.

### Screenshots and videos

Here is a video, displaying the issue.
https://github.com/user-attachments/assets/dc32234a-c587-483b-a739-fe8d593d197f

### Version

1.21.15b

Linux (Tarball)

Split View

```shell

```

iexavl:
Closes https://github.com/zen-browser/desktop/issues/15128

The cause of the bug is described in the issue.

I decided to generalize for gestures in this fix.

This fix in particular can probably be done with just one field to hold the abort controller for the drag, but I think this is better since other types of gestures that need their state preserved for whatever reason (in this case being able to abort the event prematurely) can be later added (and existing ones can be put in there).

Here is a video showing  the fix doing its job:

https://github.com/user-attachments/assets/fde12bef-66d1-4716-8823-2f4d280c5b45

### DIFF
--- a/src/zen/split-view/ZenViewSplitter.mjs
+++ b/src/zen/split-view/ZenViewSplitter.mjs
@@ -70,6 +70,134 @@ class nsSplitNode extends nsSplitLeafNode {
   }
 }
 
+/**
+ * @typedef {object} nsSplitViewGesture
+ * @property {Function} begin - gesture start
+ * @property {Function} cancel - gesture stop
+ */
+
+/**
+ * @implements {nsSplitViewGesture}
+ */
+class nsSplitterDrag {
+  /**
+   * The slot this gesture occupies in nsZenViewSplitter._gestures.
+   *
+   * @type {string}
+   */
+  static GESTURE = "splitterDrag";
+
+  _controller = new AbortController();
+
+  /**
+   * The node whose children the splitter sits between.
+   *
+   * @type {nsSplitNode}
+   */
+  node;
+  /**
+   * Sizes of node.children at mousedown.
+   *
+   * @type {number[]}
+   */
+  originalSizes;
+  /**
+   * Which pair of panes the dragged splitter separates: it sits between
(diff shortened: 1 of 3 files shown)


mr-cheffy:
Not too sure about this one... seems like a pretty big change for not a very significant issue... I'll think about this one

iexavl:
> Not too sure about this one... seems like a pretty big change for not a very significant issue... I'll think about this one

generally speaking I scratched my head a bit too, but I definitely feel like this is a less band-aid solution. Since it needs a member field anyway might as well make it clear what that member field is for. And I think it's nice this can be reused for other gestures where needed instead of having to spam abort controllers or other random state in member fields. Your call though.

PR #15135, opened 2026-08-27
gh-15134: Fixes tab removed from group getting focused

### LINKED ISSUE #15134: Removing tab from group does not check which window it focuses
### Preliminary Checks

### What happened?

I don't think this is a bug right now, simply because I don't think it's reachable, but it's a bug waiting to happen.
[removeTabFromGroup](https://github.com/zen-browser/desktop/blob/3d53b37e8b40580d441826073a589c63c737b2eb/src/zen/split-view/ZenViewSplitter.mjs#L233) in the branch where it removes the group just focuses the first "remaining tab". The problem is that the removed tab is still in the list of remaining tabs. So if the removed tab just so happens to be the first tab in the list, it will get the focus. This doesn't make sense to begin with, but it becomes even more obvious this isn't desired as in the other branch this case is [explicitly checked](https://github.com/zen-browser/desktop/blob/3d53b37e8b40580d441826073a589c63c737b2eb/src/zen/split-view/ZenViewSplitter.mjs#L251).

*Note that I have specified version 1.21.15b but this is still not patched on the dev branch.

### Expected behavior

Removing a tab from a group focuses a tab that's not the removed one

### Actual behavior

Removing a tab from a group (potentially) focuses a tab that is  the removed one

### Steps to reproduce

Open the  browser toolbox and call `gZenViewSplitter.createEmptySplit("left")`. This will stage an empty split. Then close the url bar and the window that's left in focus is the empty one that supposedly should be removed.

### Screenshots and videos

https://github.com/user-attachments/assets/24146fc5-6d2a-4965-896f-6607b63ccc34

### Version

1.21.15b

Linux (Tarball)

Split View

```shell

```

iexavl:
Closes https://github.com/zen-browser/desktop/issues/15134

Fix video:
https://github.com/user-attachments/assets/1c06ea7f-d482-4d4a-8bbe-df41daa27616

### DIFF
--- a/src/zen/split-view/ZenViewSplitter.mjs
+++ b/src/zen/split-view/ZenViewSplitter.mjs
@@ -230,7 +230,10 @@ class nsZenViewSplitter extends nsZenDOMOperatedFeature {
       }
       this.removeGroup(groupIndex);
       if (changeTab) {
-        gBrowser.selectedTab = remainingTabs[0];
+        const tabToSelect = remainingTabs.find(remaining => remaining !== tab);
+        if (tabToSelect) {
+          gBrowser.selectedTab = tabToSelect;
+        }
         document
           .getElementById("cmd_zenNewEmptySplit")
           .removeAttribute("disabled");


mr-cheffy (approved):
Looks reasonable, thanks!

PR #15147, opened 2026-08-28
gh-15146: Fix bad current view index after removing a group

### LINKED ISSUE #15146: Bad group index after removing group
### Preliminary Checks

### What happened?

Right now when a group is removed we only check if the index of the removed group is the index of the currently viewed one [see](https://github.com/zen-browser/desktop/blob/f28e0688fbaccc6799550f56939706053fb0648b/src/zen/split-view/ZenViewSplitter.mjs#L1151).

But if `groupIndex < this.currentView` the array of groups shifts one down, so the currentView index becomes one above what it should be. Safe to say, this basically breaks the entire view splitter.

*Note that I have marked 1.21.15b version below, but this still isn't patched on the dev branch.

### Expected behavior

Picks correct group index after removal

### Actual behavior

Picks wrong group index after removal

### Steps to reproduce

Open the browser and create two split tabs. Close the first one you created. Watch the carnage.

### Screenshots and videos

https://github.com/user-attachments/assets/c62866c6-0603-4c35-8037-bf9363e91930

### Version

1.21.15b

Linux (Tarball)

Split View

```shell

```

iexavl:
Closes https://github.com/zen-browser/desktop/issues/15146

I want to write tests for this, but upon writing the tests I discovered another thing which would change the structure of the tests a little bit and I'm 99% sure this is not intended behavior, but I might as well ask.

In [handleTabClose](https://github.com/zen-browser/desktop/blob/f28e0688fbaccc6799550f56939706053fb0648b/src/zen/split-view/ZenViewSplitter.mjs#L163) we pass `{ forUnsplit: true }` which makes `removeTabFromGroup` default to `changeTab = true`, but this means we will change tabs even if the tab we are closing isn't the current view.
I think the call should look like this:
```javascript
this.removeTabFromGroup(tab, groupIndex, {
  forUnsplit: true,
  changeTab: groupIndex === this.currentView,
});
```

Here is a video of the fix:

https://github.com/user-attachments/assets/b5adc592-0153-46c0-8079-e48de2c87dfa

### DIFF
--- a/src/zen/split-view/ZenViewSplitter.mjs
+++ b/src/zen/split-view/ZenViewSplitter.mjs
@@ -358,7 +358,9 @@ class nsZenViewSplitter extends nsZenDOMOperatedFeature {
     ) {
       return;
     }
-    const currentView = this._data[this._lastOpenedTab.splitViewValue];
+    const currentView = this._data.find(group =>
+      group.tabs.includes(this._lastOpenedTab)
+    );
     if (currentView?.tabs.length >= this.MAX_TABS) {
       return;
     }
@@ -543,7 +545,9 @@ class nsZenViewSplitter extends nsZenDOMOperatedFeature {
     const panelsWidth = panelsRect.width;
     const panelsHeight = panelsRect.height;
     let numOfTabsToDivide = 2;
-    const currentView = this._data[this._lastOpenedTab.splitViewValue];
+    const currentView = this._data.find(group =>
+      group.tabs.includes(this._lastOpenedTab)
+    );
     if (currentView) {
       numOfTabsToDivide = currentView.tabs.length + 1;
     }
@@ -1106,7 +1110,6 @@ class nsZenViewSplitter extends nsZenDOMOperatedFeature {
    */
   resetTabState(tab, forUnsplit) {
     tab.splitView = false;
-    delete tab.splitViewValue;
     tab.removeAttribute("split-view");
     tab.linkedBrowser.zenModeActive = false;
     const container = tab.linkedBrowser.closest(".browserSidebarContainer");
@@ -1150,6 +1153,8 @@ class nsZenViewSplitter extends nsZenDOMOperatedFeature {
     }
     if (this.currentView === groupIndex) {
       this.deactivateCurrentSplitView();
+    } else if (this.currentView > groupIndex) {
+      this.currentView--;
     }
     for (const tab of this._data[groupIndex].tabs) {
       this.resetTabState(tab, true);
@@ -1641,7 +1646,6 @@ class nsZenViewSplitter extends nsZenDOMOperatedFeature {
   applyGridToTabs(tabs) {
     tabs.forEach(tab => {
       tab.splitView = true;
-      tab.splitViewValue = this.currentView;
       tab.setAttribute("split-view", "true");
       const container = tab.linkedBrowser?.closest(".browserSidebarContainer");
       container.setAttribute("is-zen-split", "true");


src/zen/split-view/ZenViewSplitter.mjs
```diff
@@ -1150,6 +1153,8 @@ class nsZenViewSplitter extends nsZenDOMOperatedFeature {
     }
     if (this.currentView === groupIndex) {
       this.deactivateCurrentSplitView();
+    } else if (this.currentView > groupIndex) {
```
mr-cheffy, line 1156: I don't really understand why this change?

iexavl: Because of the [array splice](https://github.com/zen-browser/desktop/blob/f28e0688fbaccc6799550f56939706053fb0648b/src/zen/split-view/ZenViewSplitter.mjs#L1157) right below it. `this.currentView` indexes into the _data array. Say you have two groups
[A, B] and `this.currentView` is index 1.
Then, if we call `removeGroup(0)` at the end of it the _data array has only 
[B], but `this.currentView` is 1, so we get an index out of bounds when trying to access. That's what you see on the video in the linked issue. To fix it we need to decrement `this.currentView` if our group index is smaller than it, because that's what causes the indices to shift left by one. Thus, this change.

mr-cheffy: Shouldn't we do like a Math.min() thing then?

iexavl: Uh.. I'm not sure what you are referring to. If you mean setting `this.currentView = Math.min(this._data.length - 1)` that won't work. It does cover the out of bounds access, so it will work on the specific example I gave above, but it breaks if there's a next element. It won't be an out of bounds access, but we will index into the wrong group. E.g,
Groups: [A, B, C]
this.currentView = 1 (in other words, group B)
remove group A (index 0)
Groups: [B, C]
this.currentView = 1 (in other words, group C, when it's supposed to be group B)
And Math.min() doesn't fix it, because in this case the Math.min call would look like Math.min(1,1).
We need to decrement this.currentView.

If this wasn't your idea and I misunderstood, I'm sorry, but could you clarify?

mr-cheffy: I meant as `this.currentView = Math.min(this.currentView, groupIndex)`

iexavl: > I meant as `this.currentView = Math.min(this.currentView, groupIndex)`

Mm, no. I don't think so. Imagine we have 
[A, B, C, D]
groupIndex = 0
this.currentView = 2 (group C) 
After removing A (index 0)
[B, C, D]
Math.min(0, 2) = 0
We would index into group B, when we are supposed to be on group C

mr-cheffy: Right, ok... I trust in you on this one

mr-cheffy (approved):

src/zen/split-view/ZenViewSplitter.mjs
iexavl: > Right, ok... I trust in you on this one

If you still aren't sure about something here, ask away. If there's a problem with this I don't see it should be fixed.

mr-cheffy: It should be fine, we can always revert if something goes wrong. Thanks!

PR #15161, opened 2026-08-29
Add per-container download folders

Visual-Studio-Coder:
## Summary

- Add a settings section for download folders by container.
- Let users choose or clear a local folder for each container.
- Apply the selected folder to the top browsing context for new and restored tabs.
- Add a browser test for applying a configured folder to a container tab.

## UI

The folder controls are added directly to each row in the existing Container settings page. The native folder picker opens when the folder button is selected. A live screenshot is not included because this machine cannot compile Firefox.

## Testing

- `node --check` passes for the changed JavaScript files.
- `git diff --check` passes.
- The full Firefox lint and browser test could not run to completion. Firefox configuration requires a linker that is not installed on this machine, and no Firefox build metadata is available for the focused test.

## Notes

The native folder picker stores the selected absolute local path in the `zen.downloads.container-folders` preference. Downloads from a tab use the path configured for that tab's container. If no path is configured, Zen uses the normal download folder.


mr-cheffy:
Sorry, we don't plan on adding this as we think it would add too much complexity. Regardless, you could try submitting it to firefox and it would be pulled down from then into zen. https://firefox-source-docs.mozilla.org/contributing/how_to_submit_a_patch.html

Visual-Studio-Coder:
@mr-cheffy why?? I need my "junk" space to have its own separate folder. the download situation in zen is just so terrible right now. so hard to organize.

mr-cheffy:
I understand, but again, it would add too much complexity for normal users

PR #15171, opened 2026-08-30
gh-15170: Fix ZenWindowSync mutual exclusion during docshell swap

### LINKED ISSUE #15170: Bad mutual exclusion when swapping docshells in ZenWindowSync
### Preliminary Checks

### What happened?

Right now there is a [promise](https://github.com/zen-browser/desktop/blob/e89bd7796e2dcecaf0c483a795225ed9ec549bbd/src/zen/sessionstore/ZenWindowSync.sys.mjs#L104). in `ZenWindowSync` which plays the role of a mutex when swapping docshells.

Unfortunately, in the way it's implemented right now there are multiple places where it completely breaks and leaves the docshell switching coroutine open to races. One example of this is [here](https://github.com/zen-browser/desktop/blob/e89bd7796e2dcecaf0c483a795225ed9ec549bbd/src/zen/sessionstore/ZenWindowSync.sys.mjs#L1545)

Right above this line we add a `TabSelect` listener pointing to a local handler. If this handler ends up getting invoked, we presumably have the `#docShellSwitchPromise` set to reference the promise that was assigned in this on_focus function, which is why we pass `ignorePromise` to `on_TabSelect`.
Lets call that promise P1 
`on_TabSelect` assigns a new promise to `this.#docShellSwitchPromise` (lets call that P2)
and ends up first executing the promise that was previously stored in `this.#docShellSwitchPromise` (in other words P1)  [here](https://github.com/zen-browser/desktop/blob/e89bd7796e2dcecaf0c483a795225ed9ec549bbd/src/zen/sessionstore/ZenWindowSync.sys.mjs#L1566), which jumps back  [to P1](https://github.com/zen-browser/desktop/blob/e89bd7796e2dcecaf0c483a795225ed9ec549bbd/src/zen/sessionstore/ZenWindowSync.sys.mjs#L1547). In P1, we end up nulling out `this.#docShellSwitchPromise`, but remember that by this point `this.#docShellSwitchPromise` is set to P2, not P1. So the docshell swap in `on_TabSelect` ends up running completely unguarded.

Not to mention that in [this case](https://github.com/zen-browser/desktop/blob/e89bd7796e2dcecaf0c483a795225ed9ec549bbd/src/zen/sessionstore/ZenWindowSync.sys.mjs#L1750) the "mutex" isn't even used at all.

### Expected behavior

Correct locking when swapping docshells

### Actual behavior

Incorrect locking when swapping docshells

### Steps to reproduce

It would be hard to reproduce an issue caused by this, because it's a race condition. But I think it can probably cause blank windows/windows with placeholders where there are supposed to be active ones.

### Screenshots and videos

_No response_

### Version

1.21.15b

Linux (Tarball)

Sync

```shell

```

iexavl:
Closes https://github.com/zen-browser/desktop/issues/15170

Slightly refactors how swaps are scheduled. Use a promise queue to make sure two swap requests can never run at the same time. As long as the `#enqueueSwap` function is used, there should be no races. Note that there is still one case that is still not handled properly and that is on window close, since the swap has to happen synchronously and immediately. I have multiple fixes related to window syncing, including a fix for this and can open follow up PRs.


mr-cheffy:
Does it also fix any issues inside https://github.com/zen-browser/desktop/issues/7079? or is this hypothetical?

iexavl:
> Does it also fix any issues inside #7079? or is this hypothetical?

It's honestly pretty difficult to tell. This bug is practically undefined behavior and has quite a bit of different paths it can take at different points. From what I can tell it does have a path for replicating this https://github.com/zen-browser/desktop/issues/7079?issue=zen-browser|desktop|14083

so its possible the patchset I have will fix that issue along with some others.
Again, note that this PR isn't the full fix. I just decided to split it into 2 or 3 PRs for easier review.

src/zen/sessionstore/ZenWindowSync.sys.mjs
```diff
@@ -1528,46 +1541,21 @@ class nsZenWindowSync {
     ) {
       return;
     }
-    if (this.#docShellSwitchPromise) {
-      return;
-    }
-    const onTabSelect = event => {
```
mr-cheffy, line 1534: So this is not needed anymore?

src/zen/sessionstore/ZenWindowSync.sys.mjs
```diff
@@ -1577,15 +1565,12 @@ class nsZenWindowSync {
       window.removeEventListener(eventName, this);
     }
     delete window.gZenWindowSync;
-    const { promise, resolve } = Promise.withResolvers();
-    this.#docShellSwitchPromise = promise;
+    // We cannot queue this. This swap needs to happen synchronously.
```
mr-cheffy, line 1568: Pretty sure this was done on purpose? Like closing a window while selecting a different tab on the other window

mr-cheffy (changes requested):

src/zen/sessionstore/ZenWindowSync.sys.mjs
iexavl: Not really. And it never actually did anything useful in the first place, I described why in the original issue.

src/zen/sessionstore/ZenWindowSync.sys.mjs
iexavl: this block doesn't really do anything, because `this.#moveAllActiveTabsToOtherWindowsForClose(window);` runs synchronously, so there is no way for another request to come in the meantime anyway. And even if`this.#moveAllActiveTabsToOtherWindowsForClose(window);` wasn't synchronous, this would still be broken, because it would prevent new requests from coming in, but if there was already a docshell swap  in flight, then a window close comes in we still get two docshell swaps happening at the same time, because we cannot afford to queue this request.
This is still an issue and I have a patch for it, but it's not in this PR.

mr-cheffy:
I'll merge this after the new release cycle

src/zen/sessionstore/ZenWindowSync.sys.mjs
mr-cheffy: Ok, in that case, it should be fine. But please remove the comment, it doesnt serve much purpose.

src/zen/sessionstore/ZenWindowSync.sys.mjs
mr-cheffy: This was introduced at https://github.com/zen-browser/desktop/commit/5b790a431ad776ac140412b8800c3a724982bdda. As long as the bug is not reproducible anymore, im fine with it.

mr-cheffy (changes requested):

src/zen/sessionstore/ZenWindowSync.sys.mjs
iexavl: > This was introduced at [5b790a4](https://github.com/zen-browser/desktop/commit/5b790a431ad776ac140412b8800c3a724982bdda). As long as the bug is not reproducible anymore, im fine with it.

Okay, I'm not actually sure about this one. It's true that the code as is right now doesn't do much for mutual exclusion, but it's possible that one of the checks here like `if (event.detail?.previousTab === event.target)` could be preventing this and we could regress on that account. I could of course add the check that prevents this, but I'm not sure what originally caused this issue so I don't know which one to add. Do you remember what was the cause of this bug?

mr-cheffy: I think the issue was caused when you focused a window *while* switching a tab.

So you have Tab A selected in window 1, and you have Tab B selected in window 2 (focused). Then you would have to click on Tab B from window 1 so window 1 gets focused AND switches tabs at the same time.

iexavl: Yeah, this was gonna regress. c23946ceb398e5a8017b4f5b0a0285f73b89099c should fix it.

src/zen/sessionstore/ZenWindowSync.sys.mjs
iexavl: fixed

mr-cheffy:
Is this ready for another review?

iexavl:
> Is this ready for another review?

Yeah, should be

PR #15183, opened 2026-08-31
gh-15182: Fix tab changing when handling tab close when group collapses

### LINKED ISSUE #15182: Removing unfocused group causes us to switch to the remaining tab if any
### Preliminary Checks

### What happened?

When the closing of a tab causes a group to collapse, even if that group isn't in focus if there is a remaining tab it gets focus.

### Expected behavior

Stay on the current tab

### Actual behavior

Remaining tab gets focused

### Steps to reproduce

Open a split view in one tab
focus another tab
close a tab in the opened split view (while still focused on the other tab)
the remaining tab gets focus

### Screenshots and videos

https://github.com/user-attachments/assets/0c259683-6438-4e04-92ae-47ed2cf9b266

### Version

nightly

Linux (Tarball)

Split View

```shell

```

iexavl:
Closes https://github.com/zen-browser/desktop/issues/15182

https://github.com/user-attachments/assets/515834f2-8e48-40fb-a982-bfa62f2db54a

### DIFF
--- a/src/zen/split-view/ZenViewSplitter.mjs
+++ b/src/zen/split-view/ZenViewSplitter.mjs
@@ -160,7 +160,10 @@ class nsZenViewSplitter extends nsZenDOMOperatedFeature {
     if (groupIndex < 0) {
       return;
     }
-    this.removeTabFromGroup(tab, groupIndex, { forUnsplit: true });
+    this.removeTabFromGroup(tab, groupIndex, {
+      forUnsplit: true,
+      changeTab: groupIndex === this.currentView,
+    });
   }
 
   /**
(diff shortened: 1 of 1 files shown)


src/zen/split-view/ZenViewSplitter.mjs
```diff
@@ -160,7 +160,10 @@ class nsZenViewSplitter extends nsZenDOMOperatedFeature {
     if (groupIndex < 0) {
       return;
     }
-    this.removeTabFromGroup(tab, groupIndex, { forUnsplit: true });
+    this.removeTabFromGroup(tab, groupIndex, {
+      forUnsplit: true,
+      changeTab: groupIndex === this.currentView,
```
mr-cheffy: Shouldn't this be the default for `changeTab` if it didnt get specified?

mr-cheffy (changes requested):

src/zen/split-view/ZenViewSplitter.mjs
iexavl: done

mr-cheffy (approved):
Thanks!

PR #15207, opened 2026-09-02

### LINKED ISSUE #15204: Creating new split causes the tabs to flash before settling in the view
### Preliminary Checks

### What happened?

This flashing is caused by two things.
The first one is that the opened tab goes full-bleed initially before it's fit into it's position in the split
The second one is that the split is rebuilt instead of only swapping the tab in the split node of the staged empty tab.

### Expected behavior

No flashing on screen when creating a split

### Actual behavior

Flashing on screen when creating a split

### Steps to reproduce

Create a split via the "New Empty Split View" action, type something in the url bar and press enter.

### Screenshots and videos

https://github.com/user-attachments/assets/84e936bc-25e5-41a5-8c02-edf4a653aa5a

### Version

nightly

Linux (Tarball)

Split View

```shell

```

iexavl:
Closes https://github.com/zen-browser/desktop/issues/15204

Note that this solution is kind of hacky in terms of parking the newly picked tab into the staged pane before commiting. To get it working a new  `ZenURLBarAboutToClose` was introduced that's fired synchronously at `handleUrlbarClose` before any timeouts. This is done, because `updateSplitView` is invoked synchronously after `handleUrlbarClose`
I checked this under `browser/components/urlbar/content/UrlbarInput.mjs` (`UrlbarInput::pickResult`)
That means when we get a `ZenURLBarAboutToClose` with `onElementPicked = true` it's safe to assume the next invocation of `updateSplitView` is going to be for the tab the url bar picked. In that case instead of letting the tab take up the whole screen we immediately position it into the staged pane.

I think the most proper solution here would be to stamp every tab with the url bar's session id so we know exactly which tab was picked by the url bar, but I couldn't really figure out a way to do this without a pretty big patch surface on top of the firefox source, which isn't worth it in my opinion.

Here is a video of this fixed:

https://github.com/user-attachments/assets/0a623d8f-5d79-4651-80c2-ac0374ec3a64

### DIFF
--- a/src/zen/common/modules/ZenUIManager.mjs
+++ b/src/zen/common/modules/ZenUIManager.mjs
@@ -619,7 +619,7 @@ window.gZenUIManager = {
    * corresponds to a close event.
    *
    * @param {object} closeToken - Token previously passed to handleNewTab.
-   * @param {CustomEvent} event - A ZenURLBarClosed event.
+   * @param {CustomEvent} event - A ZenURLBarAboutToClose or ZenURLBarClosed event.
    * @returns {boolean}
    */
   matchesCloseToken(closeToken, event) {
@@ -647,6 +647,13 @@ window.gZenUIManager = {
     }
 
     const isFocusedBefore = gURLBar.focused;
+
+    window.dispatchEvent(
+      new CustomEvent("ZenURLBarAboutToClose", {
+        detail: { onSwitch, onElementPicked, closeSeq },
+      })
+    );
+
     setTimeout(() => {
       /* If someone else opened the url bar in the meantime don't be
        * stingy and leave it open for them.*/
--- a/src/zen/split-view/ZenViewSplitter.mjs
+++ b/src/zen/split-view/ZenViewSplitter.mjs
@@ -84,6 +84,8 @@ class nsZenViewSplitter extends nsZenDOMOperatedFeature {
 
   _lastOpenedTab = null;
 
+  #stagedPickPending = false;
+
   MAX_TABS = 4;
 
   init() {
@@ -1523,11 +1525,23 @@ class nsZenViewSplitter extends nsZenDOMOperatedFeature {
   updateSplitView(tab) {
     const oldView = this.currentView;
     const newView = this._data.findIndex(group => group.tabs.includes(tab));
+    const stagingSplit = this.#stagedPickPending;
+    this.#stagedPickPending = false;
 
     if (newView === oldView && oldView < 0) {
       return;
     }
     if (newView < 0 && oldView >= 0) {
+      /* The tab for the staged split has arrived. Don't tear the
+       * group down or we will have to rebuild it which will cause a
+       * visual flash of the tabs expanding to full size and getting
+       * back into position.
+       * Park the tab where it's supposed to be in the split instead.
+       * */
+      if (stagingSplit) {
+        this.#parkInStagedPane(tab);
+        return;
+      }
       this.deactivateCurrentSplitView();
       return;
     }
@@ -1717,7 +1731,11 @@ class nsZenViewSplitter extends nsZenDOMOperatedFeature {
       const browserContainer = splitNode.tab.linkedBrowser.closest(
         ".browserSidebarContainer"
       );
-      browserContainer.style.inset = `${nodeRootPosition.top}% ${nodeRootPosition.right}% ${nodeRootPosition.bottom}% ${nodeRootPosition.left}%`;
+      const inset = `${nodeRootPosition.top}% ${nodeRootPosition.right}% ${nodeRootPosition.bottom}% ${nodeRootPosition.left}%`;
+      browserContainer.style.inset = inset;
+      if (browserContainer.hasAttribute("zen-split-staged")) {
+        this.#setStagedPaneSlot(inset);
+      }
       this._tabToSplitNode.set(splitNode.tab, splitNode);
       return;
     }
@@ -1760,6 +1778,39 @@ class nsZenViewSplitter extends nsZenDOMOperatedFeature {
       }
     });
     this.maybeDisableOpeningTabOnSplitView();
+  }
+
+  /**
+   * Set the rectangle of the staged split
+   *
+   * @param {?string} inset - The staged pane's inset, as written to its container.
+   */
+  #setStagedPaneSlot(inset) {
+    if (!inset) {
+      this.#stagedPickPending = false;
+      this.tabBrowserPanel.removeAttribute("zen-split-staging");
+      this.tabBrowserPanel.style.removeProperty("--zen-staged-pane-inset");
+      this.#parkInStagedPane(null);
+      return;
+    }
+    this.tabBrowserPanel.setAttribute("zen-split-staging", "true");
+    this.tabBrowserPanel.style.setProperty("--zen-staged-pane-inset", inset);
+  }
+
+  /**
+   * Parks a tab's container in the staged pane's rectangle so we don't
+   * get a full-bleed over the split before the pane commits.
+   *
+   * @param {?Tab} tab - The tab to park, or null to empty the slot.
+   */
+  #parkInStagedPane(tab) {
+    const container = tab?.linkedBrowser?.closest(".browserSidebarContainer");
+    const parked = this.tabBrowserPanel.querySelector("[zen-split-parked]");
+    if (parked === container) {
+      return;
+    }
+    parked?.removeAttribute("zen-split-parked");
+    container?.setAttribute("zen-split-parked", "true");
   }
 
   /**
@@ -1967,6 +2018,10 @@ class nsZenViewSplitter extends nsZenDOMOperatedFeature {
    */
   resetContainerStyle(container, removeDeckSelected = false) {
     container.removeAttribute("zen-split");
+    if (container.hasAttribute("zen-split-staged")) {
+      container.removeAttribute("zen-split-staged");
+      this.#setStagedPaneSlot(null);
+    }
     if (removeDeckSelected) {
       container.classList.remove("deck-selected");
     }
@@ -2541,6 +2596,9 @@ class nsZenViewSplitter extends nsZenDOMOperatedFeature {
     };
     this.#withoutSplitViewTransition(() => {
       this._data.push(data);
+      emptyTab.linkedBrowser
+        .closest(".browserSidebarContainer")
+        .setAttribute("zen-split-staged", "true");
       this.activateSplitView(data);
       gBrowser.selectedTab = emptyTab;
       setTimeout(() => {
@@ -2559,6 +2617,16 @@ class nsZenViewSplitter extends nsZenDOMOperatedFeature {
           const command = document.getElementById("cmd_zenNewEmptySplit");
           command.removeAttribute("disabled");
         };
+        window.addEventListener(
+          "ZenURLBarAboutToClose",
+          event => {
+            if (!gZenUIManager.matchesCloseToken(closeToken, event)) {
+              return;
+            }
+            this.#stagedPickPending = event.detail.onElementPicked;
+          },
+          { signal: controller.signal }
+        );
         window.addEventListener(
           "ZenURLBarClosed",
           event => {
--- a/src/zen/split-view/ZenViewSplitter.mjs
+++ b/src/zen/split-view/ZenViewSplitter.mjs
@@ -2583,6 +2583,66 @@ class nsZenViewSplitter extends nsZenDOMOperatedFeature {
     }
   }
 
+  /**
+   * Adopt tabs into a given group without disolving the group.
+   * This assumes the count of the passed tabs and the count of the tabs in
+   * the group is the same, because it works by swapping the tabs of each split node
+   * in the group.
+   *
+   * @param {object} group - The group
+   * @param {Array<Tab>} tabs - The tabs to end up with, in the group's order.
+   * @returns {boolean} False if the split has to be rebuilt instead.
+   * That can be done with {@link nsZenViewSplitter#splitTabs}
+   */
+  #adoptStagedPane(group, tabs) {
+    if (group?.tabs.length !== tabs.length) {
+      return false;
+    }
+    if (tabs.some(tab => tab.splitView && !group.tabs.includes(tab))) {
+      // The pick already holds a pane elsewhere, which is a merge, not an
+      // adoption. splitTabs is the one that knows how to do that.
+      return false;
+    }
+    const pinned = tabs.filter(tab => tab.pinned).length;
+    if (
+      (pinned && pinned !== tabs.length) ||
+      tabs.some(tab => tab.hasAttribute("zen-live-folder-item-id"))
+    ) {
+      // if we have to duplicate it defeats the purpose
+      return false;
+    }
+    const nodes = group.tabs.map(tab => this.getSplitNodeFromTab(tab));
+    if (nodes.some(node => !node)) {
+      return false;
+    }
+    const splitGroup = this._getSplitViewGroup(tabs);
+    if (!splitGroup) {
+      return false;
+    }
+    this.#withoutSplitViewTransition(() => {
+      group.tabs.forEach((oldTab, i) => {
+        const newTab = tabs[i];
+        if (oldTab === newTab) {
+          return;
+        }
+        nodes[i].tab = newTab;
+        this._tabToSplitNode.delete(oldTab);
+        this._tabToSplitNode.set(newTab, nodes[i]);
+        group.tabs[i] = newTab;
+        this.resetTabState(oldTab, false);
+      });
+      for (const tab of tabs) {
+        if (tab.group !== splitGroup) {
+          gBrowser.moveTabToExistingGroup(tab, splitGroup);
+        }
+      }
+      group.groupId = splitGroup.id;
+      this.activateSplitView(group, true);
+    });
+    this.#dispatchItemEvent("ZenSplitViewTabsSplit", splitGroup);
+    return true;
+  }
+
   createEmptySplit(side = "right") {
     const selectedTab = gBrowser.selectedTab;
     const emptyTab = gZenWorkspaces._emptyTab;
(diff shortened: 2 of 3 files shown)


mr-cheffy:
I think there could be a much cleaner solution than the one you proposed. Why is there a transition to begin with? Maybe we should wrap something with `withoutSplitViewTransition`

iexavl:
> I think there could be a much cleaner solution than the one you proposed. 

I seriously hope there is

> Why is there a transition to begin with? Maybe we should wrap something with `withoutSplitViewTransition`

yeah, I initially thought about that too, but it won't work. The transition is caused by the tab selected by the url bar coming into focus full-bleed as all tabs usually do. Then, after the two timeouts in [handleUrlbarClose](https://github.com/zen-browser/desktop/blob/3655afbe13dff96a00af4c789a1d3180ceecc79c/src/zen/common/modules/ZenUIManager.mjs#L704), the tab is moved into it's position in the split. 

So we can't use `withoutSplitViewTransition`, because the selected tab isn't even initially part of the split view and we can't wait until the `ZenURLBarClosed` to add it there, because by that time it would have already been shown full-bleed, causing this unsightly transition when we do add it there.

The only way to deal with this is (at least that I found) is to identify which window was picked and modify its insets so it sits in the correct position before it's ever shown on screen. That way in the first frame it is ever shown it's in the correct position.

Here's a video showing what I mean:

https://github.com/user-attachments/assets/31b2fbbe-462f-44de-8695-52e57fc84677

As you can see, by the time we reach the `ZenURLBarClosed` event, the picked window is already shown full-bleed. And from what I gather pretty much the only place where we can prevent that from happening is the `onTabSelect` fired synchronously when a tab is picked *before* it's ever shown on screen (thus, `updateSplitView`, invoked a little bit further down the line) which is why I have put that check there.

And here is a video with the fix:

https://github.com/user-attachments/assets/23bcc7b0-3521-463d-9116-e34982435d66

Pretty much the only thing that happened is the window being rendered where it should be to begin with

mr-cheffy:
Could we explicitly set `zen-split-view-no-transition`?

iexavl:
> Could we explicitly set `zen-split-view-no-transition`?

We can't. Or at least we would do it in the same way I am setting the tab insets right now. Because it's not an issue of "can we do it" it's an issue of "what do we do it on". We simply don't know which tab the url bar picked. That's the entire reason for this ugly workaround. Besides, even if we did set `zen-split-view-no-transition` that wouldn't do anything, because again, It's the tab going from full-bleed to its place in the split. So even if our transition length is 0 the tab would still flash on the screen briefly. The only thing that fixes this is to not have it appear full-bleed in the first place.

mr-cheffy:
Hi! I think the issue lies here: https://github.com/zen-browser/desktop/blob/6af79fd621c96b0c8884fd8b96eaba5a7ee583cd/src/zen/common/modules/ZenUIManager.mjs#L702-L704

This is my take on this issue, which I personally its a cleaner way to tackle it:

```patch
diff --git a/src/zen/common/modules/ZenUIManager.mjs b/src/zen/common/modules/ZenUIManager.mjs
index 770ffbb22..f3ccc2e81 100644
--- a/src/zen/common/modules/ZenUIManager.mjs
+++ b/src/zen/common/modules/ZenUIManager.mjs
@@ -647,68 +647,67 @@ window.gZenUIManager = {
     }
 
     const isFocusedBefore = gURLBar.focused;
+
+    let isCloseSeqOwner = this._isOwner(closeSeq);
+    if (!isCloseSeqOwner) {
+      return;
+    }
+    window.dispatchEvent(
+      new CustomEvent("ZenURLBarClosed", {
+        detail: { onSwitch, onElementPicked, closeSeq },
+      })
+    );
     setTimeout(() => {
       /* If someone else opened the url bar in the meantime don't be
        * stingy and leave it open for them.*/
-      if (this._isOwner(closeSeq)) {
-        // We use this attribute on Tabbrowser::addTab
-        gURLBar.removeAttribute("zen-newtab");
-
-        // Safely restore tab visual state with proper validation
-        if (
-          this._lastTab &&
-          !this._lastTab.closing &&
-          this._lastTab.documentGlobal &&
-          !this._lastTab.documentGlobal.closed &&
-          gBrowser.selectedTab === this._lastTab
-        ) {
-          this._lastTab._visuallySelected = true;
-          this._lastTab = null;
-        }
+      // We use this attribute on Tabbrowser::addTab
+      gURLBar.removeAttribute("zen-newtab");
 
-        // Reset newtab buttons
-        for (const button of this.newtabButtons) {
-          button.removeAttribute("in-urlbar");
-        }
+      // Safely restore tab visual state with proper validation
+      if (
+        this._lastTab &&
+        !this._lastTab.closing &&
+        this._lastTab.documentGlobal &&
+        !this._lastTab.documentGlobal.closed &&
+        gBrowser.selectedTab === this._lastTab
+      ) {
+        this._lastTab._visuallySelected = true;
+        this._lastTab = null;
+      }
 
-        // Handle search data
-        if (onSwitch) {
-          this.clearUrlbarData();
-        } else {
-          this._lastSearch = gURLBar._untrimmedValue || "";
+      // Reset newtab buttons
+      for (const button of this.newtabButtons) {
+        button.removeAttribute("in-urlbar");
+      }
 
-          if (this._clearTimeout) {
-            clearTimeout(this._clearTimeout);
-          }
+      // Handle search data
+      if (onSwitch) {
+        this.clearUrlbarData();
+      } else {
+        this._lastSearch = gURLBar._untrimmedValue || "";
 
-          this._clearTimeout = setTimeout(() => {
-            this.clearUrlbarData();
-          }, this.urlbarWaitToClear);
+        if (this._clearTimeout) {
+          clearTimeout(this._clearTimeout);
         }
 
-        // Safely restore URL bar state with proper validation
-        if (this._prevUrlbarLabel) {
-          gURLBar.setURI({
-            uri: this._prevUrlbarLabel,
-            dueToTabSwitch: onSwitch,
-            isSameDocument: !onSwitch,
-          });
-        }
+        this._clearTimeout = setTimeout(() => {
+          this.clearUrlbarData();
+        }, this.urlbarWaitToClear);
+      }
 
-        gURLBar.handleRevert();
+      // Safely restore URL bar state with proper validation
+      if (this._prevUrlbarLabel) {
+        gURLBar.setURI({
+          uri: this._prevUrlbarLabel,
+          dueToTabSwitch: onSwitch,
+          isSameDocument: !onSwitch,
+        });
       }
 
+      gURLBar.handleRevert();
+
       if (isFocusedBefore) {
         setTimeout(() => {
-          window.dispatchEvent(
-            new CustomEvent("ZenURLBarClosed", {
-              detail: { onSwitch, onElementPicked, closeSeq },
-            })
-          );
-
-          if (!this._isOwner(closeSeq)) {
-            return;
-          }
           this._urlbarOwner = null;
 
           gURLBar.view.close({ elementPicked: onElementPicked });
diff --git a/src/zen/split-view/ZenViewSplitter.mjs b/src/zen/split-view/ZenViewSplitter.mjs
index e464e16cd..291788cd9 100644
--- a/src/zen/split-view/ZenViewSplitter.mjs
+++ b/src/zen/split-view/ZenViewSplitter.mjs
@@ -2524,7 +2524,7 @@ class nsZenViewSplitter extends nsZenDOMOperatedFeature {
     } finally {
       requestAnimationFrame(() => {
         this.tabBrowserPanel.classList.remove("zen-split-view-no-transition");
-      }, 0);
+      });
     }
   }
 
@@ -2570,29 +2570,32 @@ class nsZenViewSplitter extends nsZenDOMOperatedFeature {
             const groupIndex = this._data.findIndex(group =>
               group.tabs.includes(emptyTab)
             );
-            const newSelectedTab = gBrowser.selectedTab;
-            if (onElementPicked) {
-              if (
-                newSelectedTab === emptyTab ||
-                newSelectedTab === selectedTab ||
-                selectedTab.getAttribute("zen-workspace-id") !==
-                  newSelectedTab.getAttribute("zen-workspace-id")
-              ) {
-                cleanup(onSwitch, groupIndex);
-                return;
-              }
-              this.removeTabFromGroup(emptyTab, groupIndex, {
-                forUnsplit: true,
-              });
-              gBrowser.selectedTab = selectedTab;
-              this.resetTabState(emptyTab, false);
-              this.splitTabs(
-                topOrLeft
-                  ? [newSelectedTab, selectedTab]
-                  : [selectedTab, newSelectedTab],
-                gridType,
-                topOrLeft ? 0 : 1
-              );
+            if (onElementPicked && onSwitch) {
+              const onTabSelect = () => {
+                const newSelectedTab = gBrowser.selectedTab;
+                if (
+                  newSelectedTab === emptyTab ||
+                  newSelectedTab === selectedTab ||
+                  selectedTab.getAttribute("zen-workspace-id") !==
+                    newSelectedTab.getAttribute("zen-workspace-id")
+                ) {
+                  cleanup(onSwitch, groupIndex);
+                  return;
+                }
+                this.removeTabFromGroup(emptyTab, groupIndex, {
+                  forUnsplit: true,
+                });
+                gBrowser.selectedTab = selectedTab;
+                this.resetTabState(emptyTab, false);
+                this.splitTabs(
+                  topOrLeft
+                    ? [newSelectedTab, selectedTab]
+                    : [selectedTab, newSelectedTab],
+                  gridType,
+                  topOrLeft ? 0 : 1
+                );
+              };
+              window.addEventListener("TabSelect", onTabSelect, { once: true });
             } else {
               cleanup(onSwitch, groupIndex);
             }
```

We'd had to change here (https://github.com/zen-browser/desktop/blob/6af79fd621c96b0c8884fd8b96eaba5a7ee583cd/src/browser/components/urlbar/content/UrlbarInput-mjs.patch#L70) so that the second argument (onSwitch) actually informs whether the tab will be changed (search, tab switch, etc). Because for example, typing a math formula like `1+2` and clicking on the "copy to keyboard" result, it will remain with the split open.

iexavl:
> Hi! I think the issue lies here:
> 
> https://github.com/zen-browser/desktop/blob/6af79fd621c96b0c8884fd8b96eaba5a7ee583cd/src/zen/common/modules/ZenUIManager.mjs#L702-L704
> 
> This is my take on this issue, which I personally its a cleaner way to tackle it:
> 
> ```diff
> diff --git a/src/zen/common/modules/ZenUIManager.mjs b/src/zen/common/modules/ZenUIManager.mjs
> index 770ffbb22..f3ccc2e81 100644
> --- a/src/zen/common/modules/ZenUIManager.mjs
> +++ b/src/zen/common/modules/ZenUIManager.mjs
> @@ -647,68 +647,67 @@ window.gZenUIManager = {
>      }
>  
>      const isFocusedBefore = gURLBar.focused;
> +
> +    let isCloseSeqOwner = this._isOwner(closeSeq);
> +    if (!isCloseSeqOwner) {
> +      return;
> +    }
> +    window.dispatchEvent(
> +      new CustomEvent("ZenURLBarClosed", {
> +        detail: { onSwitch, onElementPicked, closeSeq },
> +      })
> +    );
>      setTimeout(() => {
>        /* If someone else opened the url bar in the meantime don't be
>         * stingy and leave it open for them.*/
> -      if (this._isOwner(closeSeq)) {
> -        // We use this attribute on Tabbrowser::addTab
> -        gURLBar.removeAttribute("zen-newtab");
> -
> -        // Safely restore tab visual state with proper validation
> -        if (
> -          this._lastTab &&
> -          !this._lastTab.closing &&
> -          this._lastTab.documentGlobal &&
> -          !this._lastTab.documentGlobal.closed &&
> -          gBrowser.selectedTab === this._lastTab
> -        ) {
> -          this._lastTab._visuallySelected = true;
> -          this._lastTab = null;
> -        }
> +      // We use this attribute on Tabbrowser::addTab
> +      gURLBar.removeAttribute("zen-newtab");
>  
> -        // Reset newtab buttons
> -        for (const button of this.newtabButtons) {
> -          button.removeAttribute("in-urlbar");
> -        }
> +      // Safely restore tab visual state with proper validation
> +      if (
> +        this._lastTab &&
> +        !this._lastTab.closing &&
> +        this._lastTab.documentGlobal &&
> +        !this._lastTab.documentGlobal.closed &&
> +        gBrowser.selectedTab === this._lastTab
> +      ) {
> +        this._lastTab._visuallySelected = true;
> +        this._lastTab = null;
> +      }
>  
> -        // Handle search data
> -        if (onSwitch) {
> -          this.clearUrlbarData();
> -        } else {
> -          this._lastSearch = gURLBar._untrimmedValue || "";
> +      // Reset newtab buttons
> +      for (const button of this.newtabButtons) {
> +        button.removeAttribute("in-urlbar");
> +      }
>  
> -          if (this._clearTimeout) {
> -            clearTimeout(this._clearTimeout);
> -          }
> +      // Handle search data
> +      if (onSwitch) {
> +        this.clearUrlbarData();
> +      } else {
> +        this._lastSearch = gURLBar._untrimmedValue || "";
>  
> -          this._clearTimeout = setTimeout(() => {
> -            this.clearUrlbarData();
> -          }, this.urlbarWaitToClear);
> +        if (this._clearTimeout) {
> +          clearTimeout(this._clearTimeout);
>          }
>  
> -        // Safely restore URL bar state with proper validation
> -        if (this._prevUrlbarLabel) {
> -          gURLBar.setURI({
> -            uri: this._prevUrlbarLabel,
> -            dueToTabSwitch: onSwitch,
> -            isSameDocument: !onSwitch,
> -          });
> -        }
> +        this._clearTimeout = setTimeout(() => {
> +          this.clearUrlbarData();
> +        }, this.urlbarWaitToClear);
> +      }
>  
> -        gURLBar.handleRevert();
> +      // Safely restore URL bar state with proper validation
> +      if (this._prevUrlbarLabel) {
> +        gURLBar.setURI({
> +          uri: this._prevUrlbarLabel,
> +          dueToTabSwitch: onSwitch,
> +          isSameDocument: !onSwitch,
> +        });
>        }
>  
> +      gURLBar.handleRevert();
> +
>        if (isFocusedBefore) {
>          setTimeout(() => {
> -          window.dispatchEvent(
> -            new CustomEvent("ZenURLBarClosed", {
> -              detail: { onSwitch, onElementPicked, closeSeq },
> -            })
> -          );
> -
> -          if (!this._isOwner(closeSeq)) {
> -            return;
> -          }
>            this._urlbarOwner = null;
>  
>            gURLBar.view.close({ elementPicked: onElementPicked });
> diff --git a/src/zen/split-view/ZenViewSplitter.mjs b/src/zen/split-view/ZenViewSplitter.mjs
> index e464e16cd..291788cd9 100644
> --- a/src/zen/split-view/ZenViewSplitter.mjs
> +++ b/src/zen/split-view/ZenViewSplitter.mjs
> @@ -2524,7 +2524,7 @@ class nsZenViewSplitter extends nsZenDOMOperatedFeature {
>      } finally {
>        requestAnimationFrame(() => {
>          this.tabBrowserPanel.classList.remove("zen-split-view-no-transition");
> -      }, 0);
> +      });
>      }
>    }
>  
> @@ -2570,29 +2570,32 @@ class nsZenViewSplitter extends nsZenDOMOperatedFeature {
>              const groupIndex = this._data.findIndex(group =>
>                group.tabs.includes(emptyTab)
>              );
> -            const newSelectedTab = gBrowser.selectedTab;
> -            if (onElementPicked) {
> -              if (
> -                newSelectedTab === emptyTab ||
> -                newSelectedTab === selectedTab ||
> -                selectedTab.getAttribute("zen-workspace-id") !==
> -                  newSelectedTab.getAttribute("zen-workspace-id")
> -              ) {
> -                cleanup(onSwitch, groupIndex);
> -                return;
> -              }
> -              this.removeTabFromGroup(emptyTab, groupIndex, {
> -                forUnsplit: true,
> -              });
> -              gBrowser.selectedTab = selectedTab;
> -              this.resetTabState(emptyTab, false);
> -              this.splitTabs(
> -                topOrLeft
> -                  ? [newSelectedTab, selectedTab]
> -                  : [selectedTab, newSelectedTab],
> -                gridType,
> -                topOrLeft ? 0 : 1
> -              );
> +            if (onElementPicked && onSwitch) {
> +              const onTabSelect = () => {
> +                const newSelectedTab = gBrowser.selectedTab;
> +                if (
> +                  newSelectedTab === emptyTab ||
> +                  newSelectedTab === selectedTab ||
> +                  selectedTab.getAttribute("zen-workspace-id") !==
> +                    newSelectedTab.getAttribute("zen-workspace-id")
> +                ) {
> +                  cleanup(onSwitch, groupIndex);
> +                  return;
> +                }
> +                this.removeTabFromGroup(emptyTab, groupIndex, {
> +                  forUnsplit: true,
> +                });
> +                gBrowser.selectedTab = selectedTab;
> +                this.resetTabState(emptyTab, false);
> +                this.splitTabs(
> +                  topOrLeft
> +                    ? [newSelectedTab, selectedTab]
> +                    : [selectedTab, newSelectedTab],
> +                  gridType,
> +                  topOrLeft ? 0 : 1
> +                );
> +              };
> +              window.addEventListener("TabSelect", onTabSelect, { once: true });
>              } else {
>                cleanup(onSwitch, groupIndex);
>              }
> ```
> 
> We'd had to change here (
> 
> https://github.com/zen-browser/desktop/blob/6af79fd621c96b0c8884fd8b96eaba5a7ee583cd/src/browser/components/urlbar/content/UrlbarInput-mjs.patch#L70
> ) so that the second argument (onSwitch) actually informs whether the tab will be changed (search, tab switch, etc). Because for example, typing a math formula like `1+2` and clicking on the "copy to keyboard" result, it will remain with the split open.

Mm, this won't work. I actually noticed something that makes my original solution not work either. Not to mention the thing where we enter math formulas instead of actually switching tabs.
There are two main problems here. First, we don't know whether the onTabSelect is for the tab selected by the url bar or some other tab that came in the meantime.
Second, by the time onTabSelect gets invoked, the tab is already shown. Which means we still get that initial full-bleed.
This is actually an issue with my solution too, which I hadn't noticed.
However, this gave me another idea. I thought there was some reason the `ZenURLBarClosed` was deferred inside the timeouts, but if that's not the case I think there's a way to solve all of this in one fell swoop.

Another thing to note by the way, in your patch the event dispatch is gated by whether the invoked handler holds the seq number of the current owner. We shouldn't do that. The event should always be invoked and the responsibility to ignore it if the handler is not the owner should be up to the callee. Because if we don't fire the event the handler will hang forever, which is a memory leak. Not to mention that if the handler itself has to do some cleanup like how it's done in `createSplitView` that would never run either.

iexavl:
https://github.com/zen-browser/desktop/pull/15262#issue-5367815994

PR #15222, opened 2026-09-03
feat: focus existing tab instead of opening a new one for external links

effati:
When clicking a link from an external app, if a tab with that URL is already open, focus it and switch to its workspace instead of opening a new tab.

The setting is opt-in (default off) and lives under Settings > Tab Management, backed by the `zen.tabs.focus-existing-tab` pref. only applies to external links (`fromExternal`), not in-browser navigation. Includes trailing slash normalization so `github.com` and `github.com/` match the same tab.

I'm also thinking of doing a similar setting for the "New Tab" dialog ("Switch to tab" instead of opening a new tab).

(for transparency, this code was AI assisted. let me know if it needs fixing!)

### DIFF
--- a/locales/en-US/browser/browser/preferences/zen-preferences.ftl
+++ b/locales/en-US/browser/browser/preferences/zen-preferences.ftl
@@ -64,6 +64,9 @@ zen-tabs-select-recently-used-on-close =
 zen-tabs-close-on-back-with-no-history =
     .label = Close tab and switch to its owner tab (or most recently used tab) when going back with no history
 
+zen-tabs-focus-existing-tab =
+    .label = Focus an existing tab instead of opening a new one when clicking links from external apps
+
 zen-settings-workspaces-sync =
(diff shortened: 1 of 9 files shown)


mr-cheffy:
Hi, thanks for contributing! Unfortunately, we think this new feature would simply add more complexity and overload to users that we are not sure we want. I'll close this for cleanup, but FR can be submitted at https://github.com/zen-browser/desktop/discussions so it can get voted and if it gets enough popularity, we can take another look at this. Sorry.

effati:
> Hi, thanks for contributing! Unfortunately, we think this new feature would simply add more complexity and overload to users that we are not sure we want. I'll close this for cleanup, but FR can be submitted at https://github.com/zen-browser/desktop/discussions so it can get voted and if it gets enough popularity, we can take another look at this. Sorry.

No worries. Can I ask what the complexity and overload would be?

PR #15224, opened 2026-09-03
Add pinned tabs to URL bar results, loaded or unloaded

### LINKED ISSUE #15223: Unloaded pinned tabs are not searchable in the URL bar (`%` mode only finds loaded tabs)
### Preliminary Checks

### What happened?

Pinned tabs that are unloaded (pause icon in the sidebar, `pending` attribute) cannot be found from the URL bar. This includes pins with a custom name and pins inside collapsed folders. The 1.12.4b release notes describe renamed pinned tabs as searchable, but that only holds while the pin is loaded.

### Expected behavior

A pinned tab is part of the workspace whether or not it is loaded. Typing its custom name or hostname should offer it as a switch-to-tab result, loaded or not, ideally above history and search suggestions (the Arc behavior discussed in #2829).

### Actual behavior

No result for an unloaded pin, in normal mode and in `%` (tabs) restrict mode. After clicking the pin so it loads, the same query shows it as "Switch to Tab".

### Steps to reproduce

1. Pin a tab, rename it to `GH` (right click, Rename), put it in a folder, collapse the folder.
2. Unload it (`Cmd+W` with the default "switch" behavior, or the sidebar unload action). Confirm the pause icon is shown.
3. `Cmd+T`, type `GH`. No pin-related result.
4. `Cmd+T`, type `% GH`. Still nothing.
5. Load the pin by clicking it. Repeat steps 3 and 4: the pin now shows as "Switch to Tab".

Cause, from the 1.21.16b source: native "Switch to Tab" rows come from `UrlbarProviderPlaces` joining the `moz_openpages_temp` table, which only contains browsers registered through `UrlbarProviderOpenTabs.registerOpenTab`. Unloaded pins are not in that table, so no URL bar provider can see them. The pinned URL and label are available on the tab (`tab._zenPinnedInitialState.entry.url`, `tab.zenStaticLabel`), so a small `UrlbarProvider` in `src/zen/urlbar/` that walks `gZenWorkspaces.allStoredTabs` for `tab.pinned` entries can cover them without that table. Selecting such a result should go through `gBrowser.selectedTab = tab` rather than `switchToTabHavingURI`, which opens a new tab when the browser's current URI does not match (the unloaded case).

Working proof of concept as an fx-autoconfig userscript, with notes on the relevant internals: https://github.com/spirosbax/zen-pinsearch. Happy to send a PR that ports it in-tree next to `ZenUBActionsProvider`.

### Screenshots and videos

_No response_

### Version

1.21.16b

macOS - aarch64

URL Bar

```shell

```

spirosbax:
Fixes #15223

## What

Pinned tabs and Essentials of the current space show up as switch-to-tab rows in the URL bar when the typed text matches their custom label, hostname or URL, whether or not the tab is loaded. Pins inside collapsed folders are included, with the folder path shown as a breadcrumb.

Today the native "Switch to Tab" rows come from the open pages table, which unloaded pins are not in, so a renamed pin stops being findable the moment it is unloaded (see the issue for repro steps on 1.21.16b).

## How

- New `ZenUrlbarProviderPinnedTabs` in `src/zen/urlbar/`, registered next to the global actions provider and written in the same style (heuristic-type provider, DYNAMIC rows, `getViewTemplate` / `getViewUpdate`, `onEngagement`).
- Matching walks `gZenWorkspaces.allStoredTabs` and scores label, host and URL (exact, prefix, word prefix, substring, fuzzy). Ties keep sidebar order, pins before Essentials.
- The best match is the heuristic row, so `Enter` switches to it. When the input looks like a URL (scheme, or a dot and no spaces) the pins sit below Firefox's own heuristic instead. In `%` tabs mode they are listed first but not pre-selected, since heuristics are dropped in search modes.
- Picking a row runs `gBrowser.selectedTab = tab`, the same path as a sidebar click: the space switches if needed and session store restores an unloaded pin. The rows are deliberately DYNAMIC rather than `TAB_SWITCH`, because `pickResult` for `TAB_SWITCH` calls `switchToTabHavingURI(uri, true)` and that opens a new tab while a restoring pin is still at `about:blank`.
- Patches: `ZEN_PINNED_TAB` result group (first root group) and `ZEN_PINNED_TABS` source in `UrlbarShared`, group mapping in `UrlbarUtils.getResultGroup`, the source accepted in `updateSourcesIfEmpty` (also under `%`), heuristic span exemption in the muxer like `ZEN_ACTIONS`, and a muxer rule that drops native switch-to-tab and history rows pointing at the same page as a pinned tab row so loaded pins are not listed twice.
- Prefs: `zen.urlbar.suggestions.pinned-tabs` (default `true`) and `zen.urlbar.suggestions.pinned-tabs.all-spaces` (default `false`, searches every space and prefixes the space name).
- Reuses Firefox's `urlbar-result-action-switch-tab` string, so no l10n change.

## Testing

- `src/zen/tests/ub-actions/browser_pinned_tabs_search.js` covers: custom label match with `Enter` switching and no new tab, an unloaded pin found by host with URL-like input keeping Firefox's heuristic first and the pin restoring on pick, and the pref turning the provider off.
- I have not built Zen locally, so the mochitest has only been checked for syntax and formatting, not run. The provider logic itself has been running for a day as an fx-autoconfig userscript on 1.21.16b (same code paths, source at https://github.com/spirosbax/zen-pinsearch); the in-tree version differs only in using proper result group and source constants instead of borrowing the actions group, and a muxer rule instead of wrapping `sort`. Happy to adjust anything CI or review turns up.
- The five patch files were regenerated with `git diff --full-index` against pristine Firefox 155.0 sources with the current dev patches applied, and re-verified to apply cleanly.

## Notes for review

- Placement of the `ZEN_PINNED_TAB` group ahead of `ZEN_ACTION` is a judgment call (a pin named exactly what you typed beats a learned command). It is a one-line move if you prefer actions first.
- While a pin is the heuristic, Firefox's autofill for that query does not apply (autofill only fires when the autofill result is first). URL-like input restores it.

https://claude.ai/code/session_01Sg99ACeu6sQL5t2mMSWnGC

### DIFF
--- a/src/browser/components/urlbar/UrlbarMuxerStandard-sys-mjs.patch
+++ b/src/browser/components/urlbar/UrlbarMuxerStandard-sys-mjs.patch
@@ -1,8 +1,16 @@
 diff --git a/browser/components/urlbar/UrlbarMuxerStandard.sys.mjs b/browser/components/urlbar/UrlbarMuxerStandard.sys.mjs
-index 65bd41b2ea14c26d490632d4a5348d6cf33ee9dd..80fbd685156949b6c7e93ed9d0ef56350dd1a0fe 100644
+index 82c9d771ed1fb5c98a4a594db5f1badbafa284ba..b03be688dc4153b1b2eb05a0f7ada50f811e9e3c 100644
 --- a/browser/components/urlbar/UrlbarMuxerStandard.sys.mjs
 +++ b/browser/components/urlbar/UrlbarMuxerStandard.sys.mjs
-@@ -925,11 +925,16 @@ class MuxerUnifiedComplete extends UrlbarMuxer {
+@@ -119,6 +119,7 @@ class MuxerUnifiedComplete extends UrlbarMuxer {
+       urlToTabResultType: new Map(),
+       addedRemoteTabUrls: new Set(),
+       addedSwitchTabUrls: new Set(),
++      zenPinnedTabUrls: new Set(),
+       addedResultUrls: new Set(),
+       canShowPrivateSearch: unsortedResults.length > 1,
+       canShowTailSuggestions: true,
+@@ -919,11 +920,16 @@ class MuxerUnifiedComplete extends UrlbarMuxer {
        result.providerName == "UrlbarProviderHeuristicFallback" &&
        state.context.heuristicResult?.providerName !=
          "UrlbarProviderHeuristicFallback"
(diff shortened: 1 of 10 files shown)


mr-cheffy:
Sorry, we don't accept claude generated code

PR #15239, opened 2026-09-04
feat: add quick attachment tray for downloads and clipboard images

gustavohnsv:
relates to discussion #15160

Instead of immediately triggering the system's native file picker, clicking a file upload field now opens a tray at the bottom of the page. This tray allows users to quickly attach an image currently in their clipboard or recently downloaded files from the browser. 

It supports image thumbnail previews; for non-image downloaded files, it renders the native system icon for that extension. The feature is named **Snap** and can be enabled or disabled in the settings under the "Look and Feel" section, designed to feel as native as possible. I've also included error logging for unexpected edge cases, as well as automated tests.

Both the tray styling and the maximum file display limit remain open for feedback; once the core concept is validated, I'll continue refining them.

This is especially useful when taking frequent screenshots to attach to forms that don't support direct pasting (like Discord's web app), or when downloading multiple files saved across different system folders, allowing for much faster attachments. 

I've attached a video in the linked discussion showing how it works in practice (view here: https://github.com/user-attachments/assets/0bc45a34-4fa4-4915-a3ef-c8abbc2ffc6e).


gustavohnsv:
i have implemented the core functionality of the feature, but I would like to gather feedback before refining the UI, checking more edge cases, and delivering the final result.

mr-cheffy:
Hi, sorry but I dont think we'll want this feature into zen as it adds extra complexity. Have you discussed this with the team or at least what the community thinks?

gustavohnsv:
thanks for the feedback 

that's fair, I initially built this to solve a personal workflow need and left discussion #15160 open, but since there wasn't much community traction yet (probably due to the time), I understand the concern about adding maintenance overhead to zen's core

anyway, I appreciate the review. I'll close the PR for now. thanks

mr-cheffy:
Thanks for understanding!

PR #15240, opened 2026-09-04
gh-14238: Route external links during startup

### LINKED ISSUE #14238: Space routing: external links open in last-active space instead of configured default after restart

ReyadRamahi:
Summary

External links received during startup could skip Space Routing or load into the current tab without going through addTab(). This could leave the link in the last-active Space instead of its configured destination.

This PR:

Allows tabs marked fromExternal through the startup routing guard while continuing to skip restored tabs.

Opens a single external startup URI in a new tab when it resolves to an existing destination Space, so the normal routing path can apply the destination and container. Keeps the current-tab behavior for the most-recent-space default or a missing destination.

Adds startup regression coverage and updates existing test expectations to include the workspace name.

Fixes #14238.

Testing

Tested locally on macOS 26.6.2, Apple Silicon, with commit d1d5ff01ed9690a51d82b9527b44d760c325f98d.

npm run build:ui completed successfully.

npm test -- space_routing/browser_space_routing_on_add_tab.js --headless --force passed: 27 checks (26 subtests and 1 test), 0 unexpected results.

Coverage includes external routing before startup is ready, restored-tab skipping, a specific external default, an explicit matching route, the most-recent-space default, a missing destination, and non-external startup behavior.

Manually tested an external https://example.com link from Apple Notes after quitting the local Nightly build with Work last active and Personal configured as the default external destination. The link opened in Personal and the Work tab was restored.

These results cover the targeted browser-chrome test and the macOS manual scenario; they do not represent a full test-suite or cross-platform run.

Demo

https://github.com/user-attachments/assets/5ae544c1-cf28-4a6b-a73b-00c6a0641eb0


mr-cheffy:
Hi, have you tested this? Could I get a video of the fix?

ReyadRamahi:
https://github.com/user-attachments/assets/e4edfd7e-6ae7-40fe-a8c2-0bf87f7ea6e3

I believe this is the intended behavior?

src/zen/space-routing/ZenSpaceRoutingManager.sys.mjs
```diff
@@ -123,6 +123,32 @@ class nsZenSpaceRoutingManager {
...
+   * @param {boolean} fromExternal - Whether the URI came from another application
+   * @param {Window} win - The browser window receiving the startup URI
+   * @returns {boolean} True when the URI should open in a routed tab
+   */
+  shouldOpenExternalInNewTab(uriString, fromExternal, win) {
+    if (!fromExternal) {
+      return false;
+    }
+
+    const targetRoute = this.routeUri(uriString, { fromExternal: true });
```
mr-cheffy:
```suggestion
    const targetRoute = this.routeUri(uriString, { fromExternal });
```

src/zen/space-routing/ZenSpaceRoutingManager.sys.mjs
```diff
@@ -123,6 +123,32 @@ class nsZenSpaceRoutingManager {
...
+    return (
+      targetRoute !== "most-recent-space" &&
+      !!win?.gZenWorkspaces?.getWorkspaceFromId?.(targetRoute)
```
mr-cheffy: Why would `gZenWorkspaces` not exist? And if it exists, why would getWorkspaceFromId not exist? And why would `win` be undefined?

Are workspaces initialized at this stage?

mr-cheffy (changes requested):

src/zen/space-routing/ZenSpaceRoutingManager.sys.mjs
ReyadRamahi: You're right, those optional checks weren't needed. `win` is passed directly from the startup handler, and the workspace manager and its method already exist at that point.

The workspace data isn't necessarily loaded yet, though. I checked the startup order, and on a cold start this can run before SessionStore has populated it. I've updated the helper to wait for `gZenWorkspaces.promiseInitialized` before looking up the destination and removed the optional chaining. I also added a regression test for that timing case.

PR #15241, opened 2026-09-05
feat: Add Content Link Handling options

chirivelli:
Adds configurable modifier-click shortcuts for opening page links in Glance, Split View, a new tab, or a new window on macOS, Windows, and Linux. Glance and split clicks share the ContentLinkHandling actors, including links inside frames. Opening a split keeps the source page beside the focused new pane.

### Defaults

| Action | macOS | Windows / Linux |
| --- | --- | --- |
| Glance | Disabled | Disabled |
| Split View | Disabled | Disabled |
| New Tab | Cmd + Click | Ctrl + Click |
| New Window | Shift + Click | Shift + Click |

### Settings and conflicts

- Adds shortcut-recording fields under **Look and Feel → Content Link Handling**. Select a field, then press and release one or more modifier keys.
- Escape or leaving the field cancels recording. Clear, Backspace, or Delete disables a shortcut; Restore defaults resets all four assignments.
- Conflicting assignments show an inline warning naming the other action, with **Swap shortcuts** and **Cancel**. The previous assignment stays active until the conflict is resolved. Conflicts introduced outside Settings do not silently choose a winner.
- Removes the separate Enable Glance checkbox and restart prompts. Shortcut changes apply immediately to already-open pages; disabling a shortcut leaves context-menu actions available.

### Implementation

Moves shared link-click handling into `content-link-handling` and groups all four shortcut preferences under `zen.content-link-handling.*`. Modifier combinations match exactly. New Tab and New Window reuse Firefox's existing opening path, while split links reuse the context-menu split operation. Trusted-click, drag, and link-load security checks remain in place.

### Validation

- Local macOS UI build succeeded, including the final preference rename.
- Local browser integration run passed **52 assertions**, covering page and iframe links, live changes, remapped tab/window actions, recording, clearing, cancellation, and inline conflict swaps.
- JavaScript syntax and `git diff --check` passed.
- Native Windows and Linux runs have not been performed.


mr-cheffy (changes requested):
Why specifically for MacOS only? Maybe we can have this on the same file as glance's logic and rename it to ContentLinkHandling?

chirivelli:
Screenshot 2026-09-05 at 8 40 22 PM

Update the settings page

Shift is the default for "Open in Glance"
Option/Alt is the default for "Open in Split View"

chirivelli:
sorry about all the force pushes 😅

chirivelli:
@mr-cheffy 

I have mapped these similar to what Arc does.

I use all 3 modifers like following
- `Cmd` + Click for open in New Tab
- `Opt` + Click for open in Split View
- `Shift` + Click for open in Glance/Peek

mr-cheffy:
And for windows? What about opening links on new windows?

chirivelli:
> And for windows? What about opening links on new windows?

well honestly didn't think of that, do you want me to add that?

chirivelli:
Screenshot 2026-09-07 at 7 44 23 PM

@mr-cheffy now they are configurable by recording keyboard

the SS shows the defaults.

mr-cheffy:
Shouldn't they be drop downs?

chirivelli:
drop down was way too long, we can use combination of multiple keys.

so I thought keyboard recording would be a better UX

chirivelli:
@mr-cheffy do try it in your machine and let me know if possible!

chirivelli:
> Shouldn't they be drop downs?

Screenshot 2026-09-07 at 7 24 25 PM

drop down was way too long, we can use combination of multiple keys.

so I thought keyboard recording would be a better UX

PR #15246, opened 2026-09-05
gh-15155: Fix tab moving to bottom of tab list

### LINKED ISSUE #15155: Moving a tab to the new tab button moves it to the bottom of the list
### Preliminary Checks

### What happened?

When dragging a tab onto the new tab button, it gets moved to the bottom

### Expected behavior

it should have done nothing as there's nowhere for the tab to go

### Actual behavior

it moves the tab to the bottom of the list

### Steps to reproduce

1. have multiple tabs in workspace
2. drag the first tab onto the new button at a certain position (where the line for tab reposition doesn't show up), then drop it
3. notice the tab having move to the bottom

### Screenshots and videos

https://github.com/user-attachments/assets/7060b90c-a041-454d-9bd2-f6a1dfe0cfc0

### Version

1.21.15b (Firefox 154.0) (64-bit)

Windows - x64

Tabs

```shell

```

bnquon:
when there wasn’t a valid drop target, `ZenDragAndDrop.js` hid the indicator but still used the hidden empty tab as a fallback and dropping after that ended up moving the tab to the bottom of the list.

changed it to return `[undefined, null]` when there’s still no target after checking the previous drag target, so dropping there leaves the tab where it is.

demo showing it doesn't move tab to bottom anymore when no valid drop target:

https://github.com/user-attachments/assets/41f6bf14-0b99-42a3-87a8-235ae0c181af

closes #15155

### DIFF
--- a/src/zen/drag-and-drop/ZenDragAndDrop.js
+++ b/src/zen/drag-and-drop/ZenDragAndDrop.js
@@ -1334,10 +1334,14 @@
         dropElement = dragData.dropElement;
         dropBefore = dragData.dropBefore;
       }
-      // Essentials should be properly handled by ::animateVerticalPinnedGridDragOver
-      if (!dropElement || dropElement.hasAttribute("zen-essential")) {
+      if (!dropElement) {
         this.clearDragOverVisuals();
-        // If dropElement is null or essential, dropElement should be the empty tab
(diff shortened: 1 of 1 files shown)


mr-cheffy (approved):
Looks good, thanks!

PR #15251, opened 2026-09-06
gh-15188: Fixed invisible download progress bar in Library when unselected

### LINKED ISSUE #15188: Download manager theme don't let you see the progress bar when unselected
### Preliminary Checks

### What happened?

Hi,
First, sorry for my english,

It seem that the dark mode for the library (Download manager), have a little issue with the progress bar
It seem that the color is missing for the done part of the progress bar when not selected

Image
Image
Image
Image

```
Linux arch 7.1.9-arch1-2 #1 SMP PREEMPT_DYNAMIC Fri, 21 Aug 2026 22:18:59 +0000 x86_64 GNU/Linux
```
```
zen-browser --v
Mozilla Zen 1.21.15b
```

### Expected behavior

The color should be visible when the entry is not selected

### Actual behavior

The color is missing, and when unselected you can't really see the download progress

### Steps to reproduce

Start a download,
Open the download part and open the library
Switch to a dark theme (No tested on white theme)
I don't have CSS or mods or such, only the dark theme (over the whole OS) is used

### Screenshots and videos

_No response_

### Version

1.21.15b

Linux (AppImage)

Customizable UI / Toolbars

```shell

```

Iron0utlaw:
Fixes #15188

## Problem

The download progress bar's completed portion is invisible in the Library window when the entry is not selected (dark themes on Linux).

## Root cause

Firefox colors the fill via `progressmeter.css` with no fallback:

    --download-progress-fill-color: var(--toolbarbutton-icon-fill-attention);

Zen overrides that variable globally in `zen-theme.css`:

    --toolbarbutton-icon-fill-attention: var(--toolbar-color) !important;

`--toolbar-color` is not defined in `places.xhtml`, so the property becomes invalid and the fill computes to transparent. Verified in the Browser Toolbox: the variable computes to an empty string in Zen's Library window, while vanilla Firefox has a real value there.

It only manifests on Linux because Firefox's Windows and macOS themes hardcode `#3c9af8` on the fill in their `allDownloadsView.css`, masking the invalid variable. The Linux theme has no such override. Selected rows are unaffected because a separate contrast rule switches them to `currentColor`.

## Fix

Add a `currentColor` fallback to the override so windows without `--toolbar-color` degrade to a visible color instead of an invalid property. This keeps the intent of the override (attention icons follow the toolbar text color) and applies the same behavior the existing selected-row contrast rule already uses.

## Testing

The equivalent change was verified by the issue reporter on Arch Linux via userChrome.css: applying `--toolbarbutton-icon-fill-attention: currentColor` in `places.xhtml` makes the unselected progress bar visible. See [this comment](https://github.com/zen-browser/desktop/issues/15188#issuecomment-5554413318).

### DIFF
--- a/src/zen/common/styles/zen-theme.css
+++ b/src/zen/common/styles/zen-theme.css
@@ -101,7 +101,7 @@
   ) !important;
   --focus-outline-color: var(--color-accent-primary) !important;
 
-  --toolbarbutton-icon-fill-attention: var(--toolbar-color) !important;
+  --toolbarbutton-icon-fill-attention: var(--toolbar-color, currentColor) !important;
   --toolbarbutton-icon-fill: currentColor !important;
 
   --button-background-color-primary: var(--in-content-primary-button-background) !important;


mr-cheffy (changes requested):
Have you verified this fix? I don't think it works

Iron0utlaw:
Hey, thanks for taking a look!

I can't test a Linux build myself, but two verifications so far: 

1. the reporter confirmed the equivalent change (`--toolbarbutton-icon-fill-attention: currentColor` in places.xhtml via userChrome) makes the bar visible on his Arch setup ([comment](https://github.com/zen-browser/desktop/issues/15188#issuecomment-5554413318))
2. I checked in the Browser Toolbox that `--toolbar-color `is undefined in places.xhtml, so the var() fallback does fire there.

What did you test on? A few things that would make it look like it doesn't work: on macOS/Windows the Library bar is hardcoded `#3c9af8` by Firefox's platform theme so this change is invisible there by design, paused downloads use GrayText via a separate variable, and selected rows were never broken. The bug only shows on Linux, unselected, actively downloading.

Happy to simplify to `--toolbarbutton-icon-fill-attention: currentColor !important`; directly (matching the `--toolbarbutton-icon-fill` line below it) if you'd prefer, that's the exact form the reporter verified.

mr-cheffy (approved):
Sounds good, thanks!

Issue #15123, opened 2026-08-26
1.21.15b: runtime.connectNative() never spawns any native messaging host (regression from 1.21.14b)


patrickrconnor:
### Preliminary Checks

- [X] I have read and understood the important section above.
- [X] I have searched existing issues and avoided creating duplicates.
- [X] I am not filing an enhancement request.
- [X] I have checked that this issue cannot be reproduced on Mozilla Firefox.
- [X] I have checked that this issue can be reproduced once I removed all my Mods and Custom CSS.

### What happened?

In 1.21.15b, `browser.runtime.connectNative()` never spawns the native messaging host process. The returned port is created, but nothing else happens: no `onMessage`, no `onDisconnect`, no error, and no host process. Downgrading to 1.21.14b fully restores it, so this is a regression in 1.21.15b.

This affects **every** native messaging host, not one specific integration. I reproduced it with a minimal unsigned Python host that has no code signature, entitlements, or launch constraints (see Steps to reproduce), as well as with Apple's iCloud Passwords host. Reproduced on two separate Macs, both after auto-updating to 1.21.15b, and in a fresh profile with no Mods or Custom CSS.

Evidence that the host process is never started:

- The test host's first action is appending to a marker file. That file is never created. Running the identical script standalone immediately afterwards writes the marker normally, so the host itself works.
- Polled `ps -Ao pid,ppid,command` every 50 ms for 150 s, matching on the full command path, across multiple `connectNative` calls: zero matches. Also enumerated all of Zen's child processes: nothing.
- `onDisconnect` never fires, with no error object. Verified across three separate hosts/attempts.
- No `exec` reaches the kernel. Apple's `com.apple.passwordmanager` host carries AMFI launch constraints, and macOS logs a `Launch Constraint Violation` within milliseconds when an unauthorized parent tries to spawn it (confirmed by deliberately spawning it from `/bin/zsh`). When Zen calls `connectNative`, AMFI logs nothing at all.
- Firefox 154 on the same machine works with the same host manifests.

Ruled out: profile state (fresh profile fails), extension state (clean reinstall fails), manifest location (tested `~/Library/Application Support/Mozilla/NativeMessagingHosts/`, `~/Library/Application Support/zen/NativeMessagingHosts/`, and the system-level `/Library/Application Support/Mozilla/`), manifest validity, host executability, MDM policy (empty policy plist), and code signing (the failing host is an unsigned script).

[...]

PR #15253, opened 2026-09-06
gh-15112: Fix iCloud Passwords authorization popup on Apple silicon

### LINKED ISSUE #15112: Apple/iCloud Passwords verification code window no longer appears in Zen after updating to 1.21.15b
### Preliminary Checks

### What happened?

Description

Since updating Zen to 1.21.15b, the macOS system verification code window used to authorize access to iCloud/Apple Passwords (via the "Mot de passe Apple"/Apple Passwords extension) no longer appears at all in Zen. It's not that autofill fails or is slow — the verification code prompt itself simply never shows up anymore when it should.

I've reproduced this on 2 different Macs, both running Zen 1.21.15b. On both machines, the verification code window appears normally in another browser (Firefox & G Chrome), so this looks specific to Zen rather than to macOS, a specific machine, the Apple ID/iCloud account, or the beta OS version.

This looks distinct from the "Known" issue listed in the 1.21.15b release notes ("some users may experience slower performance when Zen accesses saved passwords... unexpected password prompts with Primary Password enabled") — I do not have a Primary Password enabled, and the issue here is the opposite: no prompt/code is triggered at all, rather than an unexpected extra prompt.

**Additional context**

Did not occur before updating to 1.21.15b.
Reproduced on 2 different Macs (one on macOS 27 "Golden Gate" Beta 27.0 — [OS de la 2e machine à confirmer si différent]) — since another browser works fine on both machines, this points more toward Zen than toward the beta OS or a machine-specific config.
Release notes for 1.21.15b mention changes to Zen's password storage, which may be related even though the listed "Known" issue describes a different symptom than the one reported here.

**Links**

Release notes: https://github.com/zen-browser/desktop/releases/tag/1.21.15b
Firefox 154.0 release notes: https://www.mozilla.org/en-US/firefox/154.0/releasenotes/

### Expected behavior

The macOS verification code window should appear as it did before the update.

### Actual behavior

The verification code window does not appear at all in Zen. Confirmed working as expected in another browser, on 2 different Macs.

### Steps to reproduce

Update Zen Browser to 1.21.15b (Firefox 154.0).
Go to a site with a saved credential in iCloud/Apple Passwords.
Trigger the password fill via the extension.
Expected: macOS shows the verification code window to authorize the action.
Actual: the verification code window never appears.

### Screenshots and videos

_No response_

### Version

1.21.15b.

macOS - aarch64

Extensions

```shell

```

lavindeep:
Opening iCloud Passwords in Zen on Apple silicon running macOS 27 can show six code-entry fields without launching the native verification-code window. This corrects the Unix subprocess `fcntl` binding so the pipe flags reach macOS correctly and the native helper can start.

Fixes #15112. Related to #15123.

### Cause and fix

The binding declares `fcntl(int, int, int)`, but the native function is `fcntl(int, int, ...)`. Apple silicon passes fixed and variadic arguments differently. Local runtime probes found that requesting `FD_CLOEXEC` (`1`) could instead set `0` or `FD_CLOFORK` (`2`); launches with the unintended close-on-fork flag failed. See [Apple's calling-convention documentation](https://developer.apple.com/documentation/xcode/writing-arm64-code-for-apple-platforms).

Firefox's current main, beta, release, ESR 153, and ESR 140 sources retain [the same incorrect declaration](https://github.com/mozilla-firefox/firefox/blob/156018d359c66937b49d1794a8414f11c2f28cfd/toolkit/modules/subprocess/subprocess_shared_unix.js#L57), so the ABI mismatch is present upstream too, even when a particular Firefox build does not exhibit the same visible symptom. A controlled comparison has not established which build change first exposed it in Zen.

The patch declares `fcntl` as variadic and passes `ctypes.int` flags at all six Unix subprocess call sites. A macOS xpcshell test checks the actual signaling and worker pipe flags, nonblocking mode, data transfer, and child exit. All four patches were generated with `npm run export`.

### Validation

Environment: Apple silicon, macOS 27.0 build 26A5425a, Zen 1.22b / Firefox 155.0.1 source, macOS 26.5 SDK. Runtime behavior on macOS 26 has not been tested.

- Followed the documented download, clean import, bootstrap, and English language-pack steps. All 262 Git patches imported successfully.
- `npm run build -- --jobs 8` passed as a full development build with tests enabled.
- Reversed only the three production patches in the generated source and ran `npm run build:ui`. The real xpcshell regression test failed because a worker pipe's `O_NONBLOCK` bit was `0`, expected `4`.
- Restored the fix, rebuilt the JavaScript, and ran the complete subprocess suite:

  ```sh
  cd engine
  ./mach xpcshell-test toolkit/modules/subprocess/test/xpcshell
  ```

  **8 tests passed, 0 failed, 0 retried.** This includes the new regression test.

The exporter-only follow-up produces source files byte-for-byte identical to the build used for those tests.

### Native popup verification

With official iCloud Passwords 3.3.0, the missing native window was reproduced in a separate, unmodified Zen runtime with a fresh profile. Loading the patched modules made the native **Verification Code** window appear in two fresh browser processes and profiles, without warm-up spawns or an extension-reload workaround. I also manually tested the extension and confirmed iCloud Passwords works.

The popup checks used a separate signed official Zen copy with temporary resource-module routing to load the patched JavaScript before installing the extension, preserving the executable's signing identity for Apple's helper. The freshly compiled development build was validated through xpcshell; its native iCloud popup was not retested.

### CI scope

The PR workflow downloads and bootstraps Firefox, imports the patches on Ubuntu, and runs `mach lint zen`. That lint target does not cover these subprocess files, and the workflow does not run xpcshell. The local macOS build, failing baseline test, passing subprocess suite, and native popup checks above provide the runtime evidence for this change.

### DIFF
--- a/src/toolkit/modules/subprocess/subprocess_shared_unix-js.patch
+++ b/src/toolkit/modules/subprocess/subprocess_shared_unix-js.patch
@@ -1,0 +1,18 @@
+diff --git a/toolkit/modules/subprocess/subprocess_shared_unix.js b/toolkit/modules/subprocess/subprocess_shared_unix.js
+--- a/toolkit/modules/subprocess/subprocess_shared_unix.js
++++ b/toolkit/modules/subprocess/subprocess_shared_unix.js
+@@ -54,12 +54,13 @@
+ 
+   dup: [ctypes.default_abi, ctypes.int, ctypes.int],
+ 
++  // Variadic arguments use a different calling convention on Apple silicon.
+   fcntl: [
+     ctypes.default_abi,
+     ctypes.int,
+     ctypes.int /* fildes */,
+     ctypes.int /* cmd */,
+-    ctypes.int /* ... */,
++    "...",
+   ],
+ 
+   getcwd: [
--- a/src/toolkit/modules/subprocess/subprocess_shared_unix-js.patch
+++ b/src/toolkit/modules/subprocess/subprocess_shared_unix-js.patch
@@ -1,7 +1,8 @@
 diff --git a/toolkit/modules/subprocess/subprocess_shared_unix.js b/toolkit/modules/subprocess/subprocess_shared_unix.js
+index afa2b92eee680d9e2deb44a805904dccd85fb526..2900bf133987728a73db75a0704f9e77fc6b3494 100644
 --- a/toolkit/modules/subprocess/subprocess_shared_unix.js
 +++ b/toolkit/modules/subprocess/subprocess_shared_unix.js
-@@ -54,12 +54,13 @@
+@@ -54,12 +54,13 @@ var libc = new Library("libc", LIBC_CHOICES, {
  
    dup: [ctypes.default_abi, ctypes.int, ctypes.int],
  
--- a/src/toolkit/modules/subprocess/subprocess_unix-sys-mjs.patch
+++ b/src/toolkit/modules/subprocess/subprocess_unix-sys-mjs.patch
@@ -1,7 +1,8 @@
 diff --git a/toolkit/modules/subprocess/subprocess_unix.sys.mjs b/toolkit/modules/subprocess/subprocess_unix.sys.mjs
+index f4c7f1bd3eb097c8261cdc65a2eb15d651389edd..1606d4d4fb352ef1e6e5234d1bc95b56734caec4 100644
 --- a/toolkit/modules/subprocess/subprocess_unix.sys.mjs
 +++ b/toolkit/modules/subprocess/subprocess_unix.sys.mjs
-@@ -36,9 +36,9 @@
+@@ -36,9 +36,9 @@ class UnixPromiseWorker extends PromiseWorker {
  
      this.signalFd = fds[1];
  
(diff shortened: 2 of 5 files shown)


mr-cheffy:
Wouldn't this better be submitted to Firefox?

mr-cheffy:
Please send this patch to Firefox and zen will pull it once they accept it. Thanks!

lavindeep:
Thanks @mr-cheffy, I've submitted the fix upstream to Mozilla:

- Bugzilla report with reproduction steps and test results: https://bugzilla.mozilla.org/show_bug.cgi?id=2069768
- Patch and regression test, with review requested from robwu: https://phabricator.services.mozilla.com/D323933

It is awaiting review. I'll follow up here once Mozilla accepts it.

lavindeep:
Following up as promised: the fix has landed upstream for Firefox 158 in [Mozilla bug 2069768](https://bugzilla.mozilla.org/show_bug.cgi?id=2069768), with [commit a02295e05498](https://github.com/mozilla-firefox/firefox/commit/a02295e05498).

Thanks @mr-cheffy for bringing it into Zen through #15328. Once Zen updates to Firefox 158, the separate D323933 patch can be removed.

PR #15262, opened 2026-09-06

iexavl:
Continuation of https://github.com/zen-browser/desktop/pull/15207

While looking for a solution I found a few more subtle issues that needed a fix (the first two commits here)

This works by making `ZenURLBarClosed` synchronous and moving the handler invocation inside of `UrlbarInput::pickResult` at the bottom instead of the top. This allows the handler to be invoked with specific arguments earlier inside the function, which is used to pass the specific tab being picked to the handler before that tab is shown, which allows us to set all necessary insets and whatever other styling beforehand. Then that invocation at the end of `pickResult` becomes more of a general fallback in case  `_zenHandleUrlbarClose`  never got invoked inside of the function.
It also solves the "math equation in the url bar" issue for free.

https://github.com/user-attachments/assets/578c99d9-0383-4583-9a85-07eb4c1d0f2e

### DIFF
--- a/src/browser/components/tabbrowser/content/tabbrowser-js.patch
+++ b/src/browser/components/tabbrowser/content/tabbrowser-js.patch
@@ -1,5 +1,5 @@
 diff --git a/browser/components/tabbrowser/content/tabbrowser.js b/browser/components/tabbrowser/content/tabbrowser.js
-index 43180487a2fac15b7a65fac94f233eeed9094990..ae1288974cae29da929f27b78ff8db542566eba2 100644
+index 43180487a2fac15b7a65fac94f233eeed9094990..3107b0d9fff57a9bde4a6c92c0fd61983a19e45d 100644
 --- a/browser/components/tabbrowser/content/tabbrowser.js
 +++ b/browser/components/tabbrowser/content/tabbrowser.js
 @@ -514,6 +514,7 @@
@@ -203,7 +203,7 @@ index 43180487a2fac15b7a65fac94f233eeed9094990..ae1288974cae29da929f27b78ff8db54
  
        let activeEl = document.activeElement;
 +      if (gURLBar._zenHandleUrlbarClose) {
-+        gURLBar._zenHandleUrlbarClose(true);
++        gURLBar._zenHandleUrlbarClose(true, false, newTab);
 +      }
        // If focus is on the old tab, move it to the new tab.
        if (activeEl == oldTab) {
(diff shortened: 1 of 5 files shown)


src/browser/components/urlbar/content/UrlbarInputBase-mjs.patch
```diff
@@ -1,362 +1,40 @@
...
-+    if (this.focusedViaMousedown || this.hasAttribute("zen-newtab")) {
-       this.view.autoOpen({ event });
-     } else {
-       if (this._untrimOnFocusAfterKeydown) {
-@@ -5605,9 +5735,16 @@ ${
-   }
- 
-   _on_mousedown(event) {
--    switch (event.currentTarget) {
-+    switch (event.zenOriginalTarget || event.currentTarget) {
```
mr-cheffy, line 321: Why remove these things?

mr-cheffy (changes requested):

src/browser/components/urlbar/content/UrlbarInputBase-mjs.patch
iexavl: > Why remove these things?

I haven't touched that... Or at least I didn't intend to if I did. Not sure if it came from the merge or it's something I messed up. There was a merge conflict here, and in tabbrowser, because tabbrowser was removed (or renamed, to Tabbrowser I guess?)  and I had to do a whole bunch of patch gymnastics so I can move the fix onto 156 firefox source. I'll recheck it tomorrow. Can you tell me how you deal with merge conflicts in these patch files? 
What I did was more or less: try to merge, get a conflict, which is in the .patch file (practically impossible to resolve) so instead of trying to resolve the conflict in the patch file I checked out my version of the patch file into the working directory, applied the patch, then stored the version of the patch coming from upstream in a temp patch file, applied that too to get the merge conflict markers, under engine/, resolved the conflict and re-exported the patch.
Any was to make this less of a pain?

iexavl: Now that I think about it because of how I had to move the changes into the merge commit to resolve the conflict here you probably can't determine what I even changed. I think I'll just create a new branch that's based on the current dev branch and commit the fix on top of that so it's clear what I actually did. Sounds good? (I could also just rebase on this branch, but you will hsve to also deal with that on your end)

iexavl: yeah.. this is getting a bit ridiculous but this branch is messed up after the merge. Not sure exactly what I did, but I made a new branch to clean this up and base it on dev.. Third time's the charm? Feel free to close this pr.

PR #15268, opened 2026-09-06
gh-15264: Add optional NVIDIA Video Super Resolution for Linux

alper-han:
## Summary

Adds optional NVIDIA Video Super Resolution (VSR) support for Linux.

This follows the Linux VSR proposal in #15264. It is separate from Zen's existing Windows DirectComposition/NVIDIA driver-overlay path: Linux uses NVIDIA's VFX VideoSuperRes API instead.

The feature is Linux/NVIDIA-only and disabled by default.

## Behavior

- `zen.video.super-resolution.enabled` defaults to `false`.
- VSR requires explicit user opt-in and runtime capability checks.
- The initial supported path is SDR, limited-range BT.709 NV12 external video.
- Existing decoder paths remain unchanged.
- The intended frame path remains GPU-resident: DMA-BUF -> GL -> CUDA/GL interop -> NVIDIA VFX -> GL -> WebRender.
- VFX execution uses a worker-owned CUDA stream.

Normal playback always takes priority. When VSR is unavailable or fails, Zen transparently presents the original video frame.

## NVIDIA Runtime

NVIDIA VFX is proprietary. This PR does not bundle or automatically download the VFX runtime or models.

The current implementation expects the NVIDIA VFX SDK Core and `nvvfxvideosuperres` feature/model to be installed externally. If unavailable, normal video playback continues.

Feedback on the preferred runtime installation and packaging approach would be welcome.

## Validation

Tested locally on Linux/Wayland with a GeForce RTX 4070 SUPER and Vulkan H.264 hardware decoding.

- Built successfully.
- Verified VSR playback and fallback behavior with 360p to 1080p content at 30 and 60 fps.
- Confirmed RDD sandbox observations remained `Seccomp=2` and `NoNewPrivs=1`.
- Verified exported VSR patches match the local engine changes.

### DIFF
--- a/prefs/zen/video-super-resolution.yaml
+++ b/prefs/zen/video-super-resolution.yaml
@@ -1,0 +1,28 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+- name: zen.video.super-resolution.enabled
+  value: false
+  condition: "defined(XP_LINUX) && defined(MOZ_WIDGET_GTK)"
+  mirror: always
+  cpptype: RelaxedAtomicBool
+  type: static
+- name: zen.video.super-resolution.scale-mode
+  value: 0
+  condition: "defined(XP_LINUX) && defined(MOZ_WIDGET_GTK)"
+  mirror: always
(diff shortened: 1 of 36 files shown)


mr-cheffy:
I feel like this would be way out of our scope to maintain. However, please do feel free to submit this to firefox and it'll get pulled into zen if they accept it.

Thanks!

PR #15285, opened 2026-09-08
gh-14755: fix duplicated owned tabs leaving folders

### LINKED ISSUE #14755: Duplicating tabs should create new tab in the same Folder and Inherrit older tab information
### Preliminary Checks

### What happened?

When duplicating tab it will not inherit previous tab's folder

### Expected behavior

Duplicating tap should create new tab in the same folder as older tab

### Actual behavior

duplicated tab will not get created on previous tab's folder

### Steps to reproduce

Just duplicate a tab that is under a folder

### Screenshots and videos

_No response_

### Version

1.21.9b

Linux (Tarball)

Tab Folders

```shell

```

bnquon:
updated the `SessionStore` patch (the real change is on line 184, the rest is just regenerated line number changes) so duplicated tabs stay in their Zen folder when `zen.folders.owned-tabs-in-folder` is enabled. previously, the duplicate function wouldn't check this config flag.

also added a regression test using `gBrowser.duplicateTab()` and renamed the existing tests to clarify that they cover adding tabs rather than duplicating it.

demo showing enabled config and expected duplicate behavior:

https://github.com/user-attachments/assets/d8f9726a-065a-4214-adc9-ade5e155f528

closes #14755

### DIFF
--- a/src/zen/tests/folders/browser_folder_owner_tabs.js
+++ b/src/zen/tests/folders/browser_folder_owner_tabs.js
@@ -73,7 +73,7 @@ add_task(async function test_Owned_Tab_Inside_Folder_Unpinned() {
   await SpecialPowers.popPrefEnv();
 });
 
-add_task(async function test_Duplicate_Tab_Inside_Folder() {
+add_task(async function test_Duplicate_Tab_Inside_Folder_Pref_Enabled() {
   await SpecialPowers.pushPrefEnv({
     set: [["zen.folders.owned-tabs-in-folder", true]],
   });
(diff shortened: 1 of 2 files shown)


src/zen/tests/folders/browser_folder_owner_tabs.js
```diff
@@ -72,3 +72,29 @@ add_task(async function test_Duplicate_Tab_Inside_Folder_Unpinned() {
   await removeFolder(folder);
   await SpecialPowers.popPrefEnv();
 });
+
+add_task(async function test_Duplicate_Tab_Inside_Folder() {
```
mr-cheffy: A test for when the pref is off would be great

mr-cheffy (changes requested):

src/zen/tests/folders/browser_folder_owner_tabs.js
bnquon: added! also changed my added test names so the enabled and disabled cases are clearer from a quick look

mr-cheffy (approved):
Thanks!

PR #15287, opened 2026-09-08
Float Picture-in-Picture over native-fullscreen Spaces on macOS

SMRehan:
Refs #6555 (discussion — 43 upvotes, no fix until now).

**TL;DR:** on macOS the PiP player is hidden behind any other app's native-fullscreen
Space. Reparenting `BaseWindow` to `NSPanel` (pref-gated, PiP-only) fixes it; every
other window is unchanged. Video attached in the first comment.

Makes the macOS Picture-in-Picture player float above other applications'
native-fullscreen Spaces. It already floats over normal desktops — this fixes the
fullscreen case, which is the long-standing gap discussed in #6555 (43 upvotes).

**Pref-gated on `zen.widget.macos.pip-over-fullscreen.enabled`, and every window
that is not the PiP player is byte-for-byte unchanged.**

### Why it works

Measured on macOS 26.6 against a real fullscreen Space, using
`SLSFindWindowByGeometry` as a hit-test oracle:

| Window configuration | Floats over another app's fullscreen Space? |
|---|---|
| NSWindow, any level, CanJoinAllSpaces + FullScreenAuxiliary | no |
| **Real NSPanel + NSWindowStyleMaskNonactivatingPanel + CanJoinAllSpaces, no FullScreenPrimary** | **yes** |
| same, but FullScreenPrimary still present | no |
| NSPanel without the nonactivating bit | no |

Exactly two things are load-bearing: the window must be a genuine `NSPanel` so the
nonactivating style bit sticks, and it must not carry `FullScreenPrimary`. Window
**level is irrelevant** — the existing `NSFloatingWindowLevel` is sufficient.

This is also why Mozilla's 2023 attempt at the same bug
([D170673](https://phabricator.services.mozilla.com/D170673)) had no visible effect:
it overrode the `-styleMask` *getter* on a window that was not an `NSPanel`, and
AppKit silently strips the bit from any non-panel class. The reviewer applied it,
saw no change, asked "what am I missing here?", and the patch was abandoned.

### What changed

- **`nsCocoaWindow.h`** — `BaseWindow` reparented from `NSWindow` to `NSPanel`
  (a strict subclass).
- **`nsCocoaWindow.mm`** — the three defaults `NSPanel` changes are restored so all
  other windows behave exactly as before (`-isReleasedWhenClosed`,
  `-hidesOnDeactivate`, `-canBecomeMainWindow`); `-cancel:` becomes a no-op because
  an `NSPanel` closes itself on Escape and Gecko owns Escape; `-accessibilitySubrole`
  is pinned because an `NSPanel` otherwise reports `AXDialog` to VoiceOver. The PiP
  flag flip lives in the existing `nsCocoaWindow::SetWindowClass` hook.
- **`player.xhtml`** — `macnativefullscreen="false"`, so `SyncAttributesToWidget()`
  doesn't put `FullScreenPrimary` back after the flip.
- **`prefs/zen/macos.yaml`** — the static pref.

No new patch files, and no new upstream anchor lines: the diff anchors on Zen's own
`SetWindowClass` block and on the stable `@interface BaseWindow` line.

### Testing

macOS 26.6. The behaviour only reproduces against a real fullscreen Space, so
this is manual; `browser_test_fullscreen_size.js` shows the same environmental
failures with and without the patch (native fullscreen can't complete unfocused).

Verified on a local build of the same change:

- PiP floats over another application's native-fullscreen Space, and over Zen's
  own fullscreen.
- Clicking the PiP controls pauses/resumes playback **without** switching Spaces
  and without taking the menu bar away from the fullscreen app.
- Escape does not close the player; ⌘W, "back to tab" and keyboard controls
  behave as before.
- With the pref off, the old behaviour returns exactly — PiP is hidden behind
  the fullscreen app again.
- Daily-driven a patched release build with a real profile (extensions, mods,
  workspaces): main-window focus and traffic lights, ⌘Tab, popovers and context
  menus, compact mode and session restore all unchanged.
- `leakcheck`: no window leaks at shutdown — the check that matters for a
  base-class change.
- `npm run lint` clean; patches apply cleanly on a fresh `dev` import at
  Firefox 155.0.1, and the tree builds (`./mach build`, opt).
- Screen recording attached below: the player floats over another app's
  fullscreen Space with the pref on, and is hidden with it off. The window
  server agrees — the PiP panel reports `onscreen=1` with the pref on and
  `onscreen=false` with it off, while the fullscreen app is frontmost.

### Upstream

This is Mozilla [bug 1688932](https://bugzilla.mozilla.org/show_bug.cgi?id=1688932)
(NEW, unassigned, open since 2021). I've written it up there too, including why the
2023 patch didn't work.

If you'd rather not carry this in `src/widget/cocoa/`, I'm happy to move it to
`src/external-patches/firefox/` and track it as an upstream-pending patch — the same
way the native-popovers fix is handled — so the delta disappears once it lands in
Firefox. Just say which you prefer.

### DIFF
--- a/prefs/zen/macos.yaml
+++ b/prefs/zen/macos.yaml
@@ -35,3 +35,13 @@
 - name: zen.widget.macos.override-system-swipe-gestures
   value: true
   condition: "defined(XP_MACOSX)"
+
+# Let the Picture-in-Picture player float above other applications'
+# native-fullscreen Spaces. Requires BaseWindow to be an NSPanel; see
+# nsCocoaWindow::SetWindowClass in nsCocoaWindow.mm (bug 1688932).
+- name: zen.widget.macos.pip-over-fullscreen.enabled
+  value: true
+  condition: "defined(XP_MACOSX)"
+  cpptype: bool
+  mirror: always
+  type: static
--- a/src/toolkit/components/pictureinpicture/content/player-xhtml.patch
+++ b/src/toolkit/components/pictureinpicture/content/player-xhtml.patch
@@ -1,7 +1,16 @@
 diff --git a/toolkit/components/pictureinpicture/content/player.xhtml b/toolkit/components/pictureinpicture/content/player.xhtml
-index 09c9c5000de96fcd4a9d26328408996540c2cbda..b6d93850bbfe63a4e21756c97415a6c1908dae25 100644
+index a723bc582857601d7fd048f3b4ac203bff28b616..c1d09bdb08555ba091116db316908b672172aaae 100644
 --- a/toolkit/components/pictureinpicture/content/player.xhtml
 +++ b/toolkit/components/pictureinpicture/content/player.xhtml
+@@ -6,7 +6,7 @@
+ <html xmlns="http://www.w3.org/1999/xhtml"
+       xmlns:xul="http://www.mozilla.org/keymaster/gatekeeper/there.is.only.xul"
+       windowtype="Toolkit:PictureInPicture"
+-      macnativefullscreen="true"
++      macnativefullscreen="false"
+       customtitlebar="true"
+       scrolling="false"
+       >
 @@ -18,6 +18,7 @@
      <link rel="localization" href="browser/browserSets.ftl"/>
      <script src="chrome://global/content/pictureinpicture/player.js"></script>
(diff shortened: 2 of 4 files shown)


SMRehan:
this is a video showing its working:

https://github.com/user-attachments/assets/7a361e4b-cbcb-46f6-9f13-2fc4d233feac

mr-cheffy:
I feel like this would be way out of our scope to maintain? However, please do feel free to submit this to firefox and it'll get pulled into zen if they accept it.

Thanks!

SMRehan:
Understood, thanks — it's up at https://phabricator.services.mozilla.com/D324239 and Mozilla are reviewing.

PR #15282, opened 2026-09-07
gh-15438: Implementing Zen Library

FlorianButz:
This PR aims to implement the Library from Arc into Zen.

- [x] Library button
  - [x] Animated icon
  - [x] Downloads should appear when hovering this button
- [x] Opening / Closing behaviors. The library should slide in, fade out the sidebar and push the web content away.
  - [x] Button animation
  - [x] Animation while swiping
- [x] Opening the History/Bookmarks should open the library instead

- [x] Tabs / Animated tab icons
- [x] History Page
  - [x] Search / Filter
- [x] Boost Page
  - [x] Search / Filter
- [x] Downloads Page
  - [x] Search / Filter
- [x] Spaces Page
- [ ] Media Page
- [ ] Space Routing page(?)

- [x] Tabs on right
- [x] Opening Library should close URL bar

---
Bugs
- [x] Traffic lights aren't aligned
- [x] URLBar snapping to left edge

PR #15309, opened 2026-09-09
Show system screenshot folder in the downloads panel

Lmears:
Adds a Screenshots section to the downloads popup that defaults to the operating system's screenshot folder. Users can drag recent image files directly from the popup or select a filename to reveal it in the file manager, with no initial folder picker when discovery succeeds.

The section shows the ten newest eligible images, refreshes on opening and every two seconds while open, and offers Change folder, Use system folder, and Disconnect controls. On macOS, discovery reads the screenshot preference with Gecko's plist reader (bounded to 1 MiB), expands `~/` paths, and uses Desktop if no location is configured. Other platforms use Gecko's screenshot-directory provider when available; otherwise the chooser remains available rather than guessing another folder. It reads only direct children; there is no clipboard integration or background polling while the popup is closed.

### Privacy and file handling

- Excluded from private windows. Automatic discovery does not persist a path; custom overrides and disconnect are local to the profile and are not opted into Sync. Disconnect persists across restarts; Use system folder removes the override. The default preference is `system` so an explicitly empty override is preserved by Firefox.
- Does not create download records, change download commands, upload files, decode previews, or launch files. Names are rendered as text and errors do not log paths.
- Rejects symbolic links in files and folder ancestors, directories, dotfiles, empty files, and files modified within the last second. Rechecks files at drag/reveal time. Native drag payloads contain only `application/x-moz-file` and permit copying only.
- Invalidates pending results on folder changes, disconnect, popup closure, and window unload. Scans cap directories at 2,000 entries and metadata concurrency at 32.
- A native path-based drag cannot prevent another local application from replacing the file after validation; this does not claim snapshot semantics or a security boundary against local programs.

### Validation

- Added browser-chrome tests registered under `zen/tests/downloads` for system-default/custom/disconnected precedence, filtering, ordering/limits, changed/deleted files, out-of-folder paths, cancellation, popup refresh, native drag payloads, disconnect, and private-window exclusion.
- Executed the file tests and additional integration checks by loading the source modules into an isolated temporary profile of installed Zen 1.21.16b (Firefox 154.0.1) on macOS. Passed live refresh, native copy-only drag payload, rejection of deleted files and nonempty file/folder symlinks, immediate disconnect clearing, and private-window exclusion.
- Repackaged a separate local test copy and verified automatic module loading, localization, actual macOS screenshot-location discovery with a `~/` setting, custom override, restoring the system folder, and disconnect persistence after a browser restart. The test installation and profiles are local artifacts, not part of this PR.
- JavaScript syntax, repository-provided Prettier, and `git diff --check` pass. No dependencies or lockfiles changed.
- **Not yet validated:** a full build of this `dev` revision (Firefox 155.0.1), the full `mach` browser-chrome harness, upstream lint, Windows/Linux, and end-to-end OS drags into web/native targets. Local `npm run lint` cannot run without the downloaded Firefox engine. This PR is a draft pending those checks and maintainer review.

Usage, behavior limits, and a manual release checklist are in `docs/screenshot-folder.md`.

### DIFF
--- a/docs/screenshot-folder.md
+++ b/docs/screenshot-folder.md
@@ -1,0 +1,50 @@
+# Screenshots in the downloads panel
+
+Open the downloads popup, select **Choose screenshots folder…**, and choose
+the folder where your operating system saves screenshots. The panel shows the
+ten most recently modified images in that folder. Drag a filename into a drop
+target, or select it to reveal the file in the system file manager.
+
+The folder is checked when the popup opens and every two seconds while it is
+open. Closed popups do no scanning. **Change folder…** selects another folder;
+**Disconnect** forgets the selection without touching the files.
+
+The selection is local to the browser profile and is not opted into Sync.
+This feature is disabled in private windows. It does not access the clipboard,
+upload images, decode previews, launch files, or add entries to download history.
+Existing downloads and their commands are unchanged.
+
+## File handling
+
+- Only direct children with PNG, JPEG, WebP, GIF, HEIC, TIFF, or BMP extensions
+  qualify. A selected folder may contain images other than screenshots; filenames
+  are not used to guess which application created them.
+- Hidden dotfiles, directories, symbolic links, empty files, and files modified
+  within the last second are excluded. The delay avoids common partial writes;
+  it is not a guarantee that another application has finished writing a file.
+- Scans reject folders with more than 2,000 entries and limit concurrent metadata
+  requests to 32. Use a dedicated, accessible screenshots folder.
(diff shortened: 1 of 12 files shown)


mr-cheffy:
Hi, sorry, but we are deprecating the downloads panel in favor of library (https://github.com/zen-browser/desktop/pull/15282)

Lmears:
> Hi, sorry, but we are deprecating the downloads panel in favor of library (#15282)

Awesome! I should have checked haha, I threw this together because I've been missing that feature from Zen! Love your work

PR #15322, opened 2026-09-09
thunderbird

hgarfer:
- **fix(mozconfig): drop MOZ_SOURCE_REPO export conflicting with comm-release**
- **fix(mozconfig): drop MOZ_SOURCE_CHANGESET export conflicting with comm-release**


hgarfer:
Very much a mistake, my bad

Issue #14938, opened 2026-08-12
Browser defaults to new tab instead of pinned tab on startup


SophieScott12:
### Preliminary Checks

- [x] I have read and understood the important section above.
- [x] I have searched existing issues and avoided creating duplicates.
- [x] I am not filing an enhancement request.
- [x] I have checked that this issue cannot be reproduced on Mozilla Firefox.
- [x] I have checked that this issue can be reproduced once I removed all my Mods and Custom CSS.

### What happened?

I have several pinned tabs, and the setting Continue Where You Left Off enabled.  If I close the zen browser with any two of the three pinned tabs open (or the other nonpinned tabs) and then reopen the browser, it opens on the last tab like normal.  But if I close with mail.zoho.com active, it always defaults to the new tab page instead of opening the mail.zoho.com tab.  This is extremely frustrating.

### Expected behavior

Open focused on mail tab

### Actual behavior

Opened focused on New Tab

### Steps to reproduce

Open Browser
Open several tabs
Pin mail.zoho.com
Close browser with mail tab focused
Reopen browser

### Version

1.21.13b

### What platform are you seeing the problem on?

Windows - x64

### What component is this issue related to?

Tabs

### Relevant log output if applicable

```shell

```

Propheticus:
@mr-cheffy I can reproduce this when closing the browser while on the first essential tab.
(only) then reopening will result in a newtab popup instead of returning to where I left off.

Retested by dragging other essentials to the 1st spot and the result was the same. (so not tied to any specific site)

bnquon:
ran into what may be a related case on the current version, 1.22b (Firefox 155.0.1) (aarch64) macOS.

i can consistently reproduce it when there are no normal tabs open across any workspace and the session only contains pinned or essential tabs. after fully closing and reopening Zen, it selects a new tab instead of the previously active pinned or essential tab.

i’ve opened a small fix for that specific case. if this issue also occurs while normal tabs are still open, it may be a separate problem 👍

ashvwinn:
Looks like it didnt close automatically. Fixed with https://github.com/zen-browser/desktop/issues/14938

PR #15324, opened 2026-09-10
gh-14938: fix continue where you left off for only pinned/essential tabs

bnquon:
fixes CWYLO selecting an empty new tab when the session only contains pinned or essential tabs. when CWYLO is enabled, these sessions now follow the normal restoration path so the previously active pinned/essential tab is selected.

BEFORE demo showing that CWYLO doesn't properly go back to a pinned or essentials tab with no other normal tab open:

https://github.com/user-attachments/assets/e815bdd3-cf9e-483f-a946-e19eac339860

AFTER: demo showing same setup properly goes back to the pinned or essentials tab with no other normal tab open:

https://github.com/user-attachments/assets/17f1b6ea-4a73-489c-955f-6cb4237db30b

related to #14938


bnquon:
realizing this issue is a little unclear on whether normal tabs were still open when the browser was closed. my PR specifically fixes CWYLO when the session only contains pinned or essential tabs. if this report is describing a different case, i’ll open a separate issue for the pinned/essential behavior if accepted.

mr-cheffy (approved):
Looks reasonable, thanks!

PR #15328, opened 2026-09-10
gh-15123: Fixed not spawining any native messaging host

### LINKED ISSUE #15123: 1.21.15b: runtime.connectNative() never spawns any native messaging host (regression from 1.21.14b)

mr-cheffy:


Bandit:
When is this being merged to main? Zen is unusable without 1Password's extension working.

mr-cheffy:
It's already on latest. Regardless, 1pass needs a bit more of work: https://docs.zen-browser.app/guides/1password

Bandit:
image

It's not working. It was working until ~4 updates ago (#15123), so this fix didn't fix it I guess

image

PR #15333, opened 2026-09-10
gh-13888: Add touch edge swipe to reveal the compact sidebar

### LINKED ISSUE #13888: No way to open the sidebar while in compact mode on touchscreen
### Preliminary Checks

### What happened?

i am using an tablet as a second screen and when zen is in compact mode, i cannot hover to the sidebar by touch like swiping or something unless i use a mouse and that is inconvenient

### Expected behavior

sidebar supposed to show up when i swipe from right side to left on my tablet

### Actual behavior

no sidebar shows up without mouse

### Steps to reproduce

use zen on a touch screen device like surface pro, tablet/mobile as second screen using apps like spacedesk

### Screenshots and videos

_No response_

### Version

1.20b

Windows - x64

Compact Mode

```shell

```

PapeThePope:
Closes #13888

Touchscreens cannot reveal the sidebar in compact mode, because Zen only listens for mouse hover at the screen edge. This adds a small activation zone at the sidebar edge that only reacts to touch pointers:

- swipe inwards from the screen edge to reveal the sidebar (uses the same `zen-user-show` state as "Toggle Floating Sidebar")
- swipe outwards on the sidebar to dismiss it, or let it hide automatically after a short delay
- touching/using the sidebar keeps it open

The zone is only created on devices with touch points (`navigator.maxTouchPoints > 0`) and is only active while compact mode is enabled and the sidebar is actually hidden, so it does not intercept input in other layouts. It sits in the edge strip (18px) and is disabled as soon as the sidebar is shown.

New prefs (in `prefs/zen/compact-mode.yaml`):
- `zen.view.compact.touch-swipe-reveal` (default `true`)
- `zen.view.compact.touch-swipe-auto-hide.duration` (default `4000`)

Testing: the gesture logic was validated on real touch hardware (Linux/Wayland) via an equivalent userChrome.js script - edge swipe reveal, auto-hide, swipe-out dismiss and keep-open interactions all work. I could not run a full Zen build in this environment, so this exact patch has only been lint/format checked; feedback on thresholds or hiding behaviour is welcome.

### DIFF
--- a/prefs/zen/compact-mode.yaml
+++ b/prefs/zen/compact-mode.yaml
@@ -40,6 +40,12 @@
 - name: zen.view.compact.debug
   value: false
 
+- name: zen.view.compact.touch-swipe-reveal
+  value: true
+
+- name: zen.view.compact.touch-swipe-auto-hide.duration
+  value: 4000
+
 # Do not edit manually
 - name: zen.view.compact.enable-at-startup
   hidden: true
--- a/src/zen/compact-mode/ZenCompactMode.mjs
+++ b/src/zen/compact-mode/ZenCompactMode.mjs
@@ -34,6 +34,20 @@ XPCOMUtils.defineLazyPreferenceGetter(
   true
 );
 
+XPCOMUtils.defineLazyPreferenceGetter(
+  lazy,
+  "COMPACT_MODE_TOUCH_SWIPE_REVEAL",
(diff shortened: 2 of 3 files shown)


mr-cheffy:
Hi, thanks for contributing! But this looks really messy / AI generated and I dont think we might even go through with this solution. Sorry

PR #15336, opened 2026-09-10
no-bug: Add long press activation method for Glance

octaviusz:


prefs/zen/glance.yaml
```diff
@@ -12,7 +12,10 @@
   value: true
 
 - name: zen.glance.activation-method
-  value: "alt" # ctrl, alt, shift
+  value: "ctrl" # ctrl, alt, shift
```
octaviusz: Everywhere else, `Ctrl` is used by default, but for some reason, here it's `Alt`.

PR #15365, opened 2026-09-11
gh-15363: Fix boost font shuffle off-by-one index

### LINKED ISSUE #15363: Boost font shuffle sometimes skips font update (off-by-one index)
### Preliminary Checks

### What happened?

In Zen Boosts editor, `shuffleBoost()` picks a random font with:

`commonFonts[Math.round(Math.random() * commonFonts.length)]`

When `Math.random()` is close to `1`, `Math.round` can return `commonFonts.length`, so the lookup is `undefined`. The `availFonts.includes(font)` check then fails and the font family is left unchanged — a silent no-op (~3% of shuffles).

Related closed PR: https://github.com/zen-browser/desktop/pull/15361

### Expected behavior

Every font shuffle should select a valid font index in `[0, length)` and update `fontFamily` when that font is available.

### Actual behavior

Roughly 3% of shuffles produce an out-of-range index, so the font does not change even though other shuffle fields still update.

### Steps to reproduce

1. Open the Zen Boosts editor on any page.
2. Trigger font/style shuffle repeatedly (dozens of times).
3. Observe that occasionally invert/brightness/etc. change but the font does not.

### Screenshots and videos

_No response_

### Version

1.22b

macOS - aarch64

Other

```shell

```

dodo-md:
## Summary
- Use `Math.floor` instead of `Math.round` when picking a random font in `shuffleBoost()`.
- Prevents an out-of-range index (`length`) that left `font` undefined and skipped the font update ~3% of the time.

fix: #15363

## Test plan
- [ ] Open Zen Boosts editor and trigger font shuffle repeatedly
- [ ] Confirm a font is applied on each shuffle (no silent no-ops)
- [ ] Confirm other shuffle fields (invert, brightness, contrast, saturation, gradient) still update as before

### DIFF
--- a/src/zen/boosts/ZenBoostsEditor.mjs
+++ b/src/zen/boosts/ZenBoostsEditor.mjs
@@ -1521,7 +1521,7 @@ ${cssSelector} {
   shuffleBoost() {
     const availFonts = this.fetchFontList();
     const commonFonts = this.commonFonts;
-    let font = commonFonts[Math.round(Math.random() * commonFonts.length)];
+    let font = commonFonts[Math.floor(Math.random() * commonFonts.length)];
     if (availFonts.includes(font)) {
       this.currentBoostData.fontFamily = font;
     }


mr-cheffy (approved):
Looks reasonable, thanks!

PR #15372, opened 2026-09-11
gh-15371: Match urlbar actions by word and by localized keywords

ast3150:
Idea / motivation: #15371

### Problem
The URL bar action search matched the query as one ordered character subsequence against the label. Natural queries failed: `pick theme` (word order), `find on page` (no `o` after "find "), `change theme` (synonym). The feature only worked when you knew the exact label wording.

### What
- **Word-level matching** in `ZenUBActionsProvider`: the query is split into words, each word is scored against the label words (any order) and the action's keywords. The result is the average per-word score, compared against the existing thresholds. The whole-label score is still computed and the higher one wins, so everything that matched before keeps its score and rank.
- **Optional localized keywords**: actions may declare `keywordsL10nId`; the comma-separated string lives in `zen-command-palette.ftl` next to the label. Keyword matches are scaled by 0.9 so label matches rank first. Seeded for theme picker, find in page, settings, copy URL, private window.
- **Precision guard**: every meaningful query word must match something. Only `a, an, the, to, in, on, of, for` may be skipped, and only when another word matched. No global synonym list, no additional typo tolerance.

### Behaviour (from a scoring simulation over the current action list, threshold 92)
| query | before | after |
|---|---|---|
| `pick theme` | nothing | Open Theme Picker |
| `change theme` | nothing | Open Theme Picker |
| `find on page` | nothing | Find in Page |
| `copy link` | nothing | Copy Current URL |
| `theme` | Open Theme Picker | unchanged |
| `open` | Open Theme Picker, Open Private Window | unchanged |
| `xyz theme` | nothing | nothing |

### Testing
- New `browser_ub_actions_keywords.js` covers word order and keyword matching. Existing `browser_ub_actions_search.js` (full label queries) is unaffected since exact matches still score 200.
- Scoring logic verified with a Node simulation of the two functions.
- Behaviour verified at runtime in Zen 1.22b (macOS, arm64) by loading this exact provider and actions module through a resource substitution and swapping the registered provider: `pick theme`, `change theme`, `find on page`, `copy link` resolve; unrelated queries still return nothing. A local build with the mochitests is in progress and results will be posted here.

### DIFF
--- a/locales/en-US/browser/browser/zen-command-palette.ftl
+++ b/locales/en-US/browser/browser/zen-command-palette.ftl
@@ -4,11 +4,15 @@
 
 zen-action-toggle-compact-mode = Toggle Compact Mode
 zen-action-open-theme-picker = Open Theme Picker
+zen-action-open-theme-picker-keywords = change theme, pick theme, appearance
 zen-action-new-split-view = New Split View
 zen-action-new-folder = New Folder
 zen-action-copy-current-url = Copy Current URL
+zen-action-copy-current-url-keywords = copy link, copy address
 zen-action-settings = Settings
+zen-action-settings-keywords = preferences, options
 zen-action-open-private-window = Open Private Window
+zen-action-open-private-window-keywords = incognito, private browsing
 zen-action-open-new-window = Open New Window
 zen-action-new-blank-window = New Blank Window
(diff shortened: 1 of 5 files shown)


mr-cheffy:

PR #15373, opened 2026-09-11
gh-10577: Add more urlbar actions for tabs, split view, spaces and history

ast3150:
Ideas / motivation: #10577 (duplicate tab, pinned tab actions, split view and spaces actions requested in the thread) and #14812 (history).

### What
Ten new URL bar actions, each mapping one to one onto an existing Zen or Firefox command. No new behaviour, only exposure:

| action | command | shown when |
|---|---|---|
| Duplicate Tab | `cmd_zenDuplicateTab` | not on an empty tab |
| Reset Pinned Tab | `cmd_zenPinnedTabReset` | selected tab is pinned and its URL changed |
| Replace Pinned URL with Current | `gZenPinnedTabManager.replacePinnedUrlWithCurrent(selectedTab)` | same |
| Split Tabs Vertically / Horizontally / as Grid | `cmd_zenSplitView{Vertical,Horizontal,Grid}` | more than one visible tab |
| Unsplit View | `cmd_zenSplitViewUnsplit` | a split view is active |
| New Space | `cmd_zenOpenWorkspaceCreation` | always |
| Unload Current Space | `cmd_zenUnloadWorkspace` | always |
| Show All History | `Browser:ShowAllHistory` | always |

Notes:
- "Replace Pinned URL" is a function command because `replacePinnedUrlWithCurrent()` without an argument falls back to `TabContextMenu.contextTab`, which is stale from the URL bar. Reset already uses the selected tab in `zen-sets.js`.
- Not included from the thread: move tab to folder / space (needs a target picker), layout switching (pref-based, no command), tab search (separate feature).

### Testing
- `browser_ub_actions_search.js` iterates every action and skips unavailable ones, so the new entries are covered for label search without extra fixtures.
- Every `l10nId` has an en-US string (checked by script).
- Behaviour verified at runtime in Zen 1.22b (macOS, arm64) by loading this exact actions module through a resource substitution: the entries appear and hide according to their `isAvailable` checks, and the commands act on the selected tab. A local build with the mochitests is in progress and results will be posted here.


mr-cheffy:

PR #15376, opened 2026-09-11
gh-10577: Add various requested url bar actions

ast3150:
## Why

Users requested more URL bar actions, for example reopen closed tab (gh-15369); duplicate tab, new space (gh-10577)

## What changed

- Added the commands to the command palette
- Every action runs an existing Zen or Firefox command. No new behavior.
- Actions are listed only when they apply:
  - Reopen closed tab: SessionStore reports a closed tab
  - Unsplit view: a split view is active

## Testing
Verified locally in Zen 1.22b via an autoconfig hook that loads this branch's actions module in place of the installed one. All actions appear in Cmd+L, availability checks run and the commands fire.


locales/en-US/browser/browser/zen-command-palette.ftl
```diff
@@ -18,6 +24,10 @@ zen-action-new-boost = New Boost
 zen-action-next-space = Next Space
 zen-action-previous-space = Previous Space
 zen-action-close-tab = Close Tab
+zen-action-reopen-closed-tab = Reopen Closed Tab
+zen-action-duplicate-tab = Duplicate Tab
```
mr-cheffy:
```suggestion
zen-action-duplicate-tab = Duplicate Current Tab
```

locales/en-US/browser/browser/zen-command-palette.ftl
```diff
@@ -5,6 +5,12 @@
 zen-action-toggle-compact-mode = Toggle Compact Mode
 zen-action-open-theme-picker = Open Theme Picker
 zen-action-new-split-view = New Split View
+zen-action-split-tabs-vertically = Split Tabs Vertically
+zen-action-split-tabs-horizontally = Split Tabs Horizontally
+zen-action-split-tabs-grid = Split Tabs as Grid
+zen-action-unsplit-view = Unsplit View
```
mr-cheffy:
```suggestion
zen-action-unsplit-view = Expand current view
```

mr-cheffy (changes requested):
I dont think we should include: `zen-action-replace-pinned-url`, `zen-action-show-all-history`, `zen-action-unload-space`, and `zen-action-split-tabs-*`. Please remove them.

ast3150:
I do think show history is genuinely valuable here, but ok, going with the removal...

mr-cheffy:
I agree, but we are trying to add commands on small batches as we are not sure how much they affect URL bar performance in the long run and how in-the-way it would get into the users.

mr-cheffy (approved):
Looks good, thanks!

PR #15406, opened 2026-09-15

iexavl:
Continuation of https://github.com/zen-browser/desktop/pull/15262#issue-5367815994


src/zen/split-view/ZenViewSplitter.mjs
```diff
@@ -2540,12 +2544,76 @@ class nsZenViewSplitter extends nsZenDOMOperatedFeature {
     try {
       return callback();
     } finally {
+      // Flush the new geometry now. Otherwise, because requestAnimationFrame
+      // runs before style computations, the zen-split-view-no-transition class
+      // is never even seen and we get a transition anyway.
+      this.tabBrowserPanel.getBoundingClientRect();
```
mr-cheffy: I don't like this.

PR #15442, opened 2026-09-17
Also search for name of pinned tabs in UrlBar

mduesterhoeft:
Also search for name of pinned tabs in UrlBar.

Currently the search in the UrlBar view (CMD+T) only searches in the open tabs by tab title. I do not find my pinned tabs by name, if I give them a custom name. This change adds support for also searching in the name of pinned tabs

image


mr-cheffy:
Hi! Thanks for the contribution! I was working on this apparently at the same time without noticing this PR... 

I think i'll go with my solution because it also supports searching for folders. But thanks for trying to contribute nonetheless!

mduesterhoeft:
@mr-cheffy really nice - thanks you. One thing that would also be really nice is to represent "recent tabs" in the UrlBar - e.g. when no search term is entered - then the user does not need to switch between Ctrl + Tab and Cmd+T

Issue #14876, opened 2026-08-07
Renaming silently fails and permanently disables all rename actions when the rename input's container has no layout box


Libera-sudo:
### Summary

If the element that hosts the rename input has no layout box, `input.focus()` in `gZenVerticalTabsManager.renameTabStart` fails **silently**. Because focus never lands, `blur` never fires, so `renameTabHalt` never runs and `this._tabEdited` is never reset. Since `renameTabStart` starts with `if (this._tabEdited) return;`, **every subsequent rename — tabs, folders and spaces alike — becomes a no-op until the browser is restarted.** The leftover `zen-renaming-tab="true"` on `:root` additionally prevents the sidebar from auto-hiding in compact mode.

No exception is thrown and nothing is logged, which makes this effectively undiagnosable from the user's side.

### Version

- Zen `1.21.10b` (BuildID `20260729085236`), macOS (aarch64)
- Also reproduces on `1.21.9b`

### Steps to reproduce

No mods required — a single userChrome.css rule is enough to trigger it:

1. Use vertical tabs with compact mode enabled.
2. Add to `chrome/userChrome.css` (and set `toolkit.legacyUserProfileCustomizations.stylesheets = true`):
   ```css
   .zen-current-workspace-indicator { display: none !important; }
   ```
3. Restart, then right-click the space and choose **Change name** (`cmd_zenChangeWorkspaceName`).
4. Nothing happens. No error in the Browser Console (parent process).
5. Now try to rename **any tab** (double-click) or **any folder** — those are dead too.
6. In compact mode, the sidebar no longer auto-hides.
7. Only restarting Zen restores any of it.

Step 2 is just a convenient trigger; anything that leaves the host container without a layout box reaches the same state.

### Root cause

`cmd_zenChangeWorkspaceName` routes the space's indicator label into the shared rename flow (`zen-sets.js`), and `renameTabStart` inserts the input as a child of `.zen-current-workspace-indicator`:

```js
// ZenUIManager.mjs — renameTabStart
document.documentElement.setAttribute("zen-renaming-tab", "true");
...
this._tabEdited.after(input);
input.focus();                 // <-- silently does nothing if there is no layout box
input.setSelectionRange(0, input.value.length, "backward");
input.addEventListener("blur", this._renameTabHalt);
```

`this._tabEdited` is only ever cleared on the very last line of the success path of `renameTabHalt`:

```js
// ZenUIManager.mjs — renameTabHalt
renameTabHalt(event) {
[...]

PR #15471, opened 2026-09-18
fix: clean up failed rename state

xWielo:
Related to #14876.

Fixes the shared tab, folder and space rename flow when the editor cannot receive focus.

The rename state is now cleaned up after focus failures and during completion, preventing a failed attempt from blocking later rename actions.

Adds a regression test for a tab inside a folder and verifies that renaming remains available afterwards.

Tests were not run locally because this checkout does not include engine/. Syntax and diff checks passed.


mr-cheffy (changes requested):
Please link an issue to this PR, like the rest of PRs.

xWielo:
> Please link an issue to this PR, like the rest of PRs.

#14876 was closed without a code fix. This PR implements the missing cleanup and regression test.

mr-cheffy:
Why does blur fail to be detected currently? What we should do is simply blur the input if one is already selected while trying to rename a new field

xWielo:
> Why does blur fail to be detected currently? What we should do is simply blur the input if one is already selected while trying to rename a new field

Blur does not fire when the input never received focus, for example when its parent has no layout box. I changed the flow to blur an existing rename input before starting another rename, with a direct cleanup fallback when that input was never focused.

mr-cheffy:
Not sure, this all seems a very hacky fix for this particular issue, and with the addition that you haven't tested it. I think i'll close this issue sorry.

xWielo:
> Not sure, this all seems a very hacky fix for this particular issue, and with the addition that you haven't tested it. I think i'll close this issue sorry.

Understood. Thanks for taking the time to review it.

